PR #4, opened 2018-11-15
build(npm): add script for cleaning

jleveugle:


package.json
```diff
@@ -9,6 +9,7 @@
   ],
   "scripts": {
     "build": "webpack --config ./packages/manager-layout-ovh/webpack.config.js --env=production",
+    "clean": "npx rimraf dist node_modules && lerna clean -y",
```
antleblanc: `lerna` won't be availabe if you launch `npx rimraf … node_module`

jleveugle: I think your comment is falsy

Cf:

```bash
➜  manager git:(clean) yarn clean
yarn run v1.10.1
$ npx rimraf dist node_modules && lerna clean -y
lerna notice cli v3.4.3
lerna info versioning independent
lerna info clean removing /Users/jleveugle/Documents/Work/github/ovh-ux/manager/packages/manager-core/node_modules
lerna info clean removing /Users/jleveugle/Documents/Work/github/ovh-ux/manager/packages/manager-layout-ovh/node_modules
lerna info clean removing /Users/jleveugle/Documents/Work/github/ovh-ux/manager/packages/manager-welcome/node_modules
lerna success clean finished
✨  Done in 12.08s.
```

frenautvh: isn't lerna installed globally in your test ?

jleveugle: It won't change something.

Lerna is in our dependencies, and scripts entries are scoped in ./node_modules/.bin folder, so if lerna is a dependency, it's available in scripts entries :)

Cf: https://docs.npmjs.com/misc/scripts

```
So, if your package.json has this:
	{ 
		"name" : "foo" ,
		"dependencies" : { "bar" : "0.1.x" },
		"scripts": { "start" : "bar ./test" }
	}
then you could run `npm start` to execute the `bar` script, which is exported into the `node_modules/.bin` directory on `npm install`.
```

antleblanc: Due to the fact that we invoke `rimraf` in order to delete `node_modules` folder we can probably be in a state where `lerna` binary is no more available.

PR #14, opened 2018-11-21
fix: add jquery in ProvidePlugin

cbourgois:
### Description of the Change

`angular.js` looks for `window.jQuery` (cf: https://github.com/angular/angular.js/blob/master/src/Angular.js#L1890 )

Before `angular.element` used `jqLite`, with this fix it's now `jQuery`.

_This solution is recommended in the [webpack documentation](https://webpack.js.org/plugins/provide-plugin/#usage-jquery-with-angular-1)_

7a0e47e - fix: add jquery in ProvidePlugin

### DIFF
--- a/packages/manager-layout-ovh/package.json
+++ b/packages/manager-layout-ovh/package.json
@@ -50,6 +50,7 @@
     "babel-eslint": "^8.2.6",
     "eslint": "^5.8.0",
     "eslint-config-airbnb-base": "^13.1.0",
+    "webpack": "^4.5.0",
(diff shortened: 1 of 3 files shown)


antleblanc:
> _This solution is recommended in the [webpack documentation](https://webpack.js.org/plugins/provide-plugin/#usage-jquery-with-angular-1)_

I'm a bit curious about why you've decided to close this pull request?

Could you be more precise of such a decision? Your feedback will be well appreciate:) :)

PR #22, opened 2018-11-26
build: add publish script for packages

jleveugle:


package.json
```diff
@@ -17,6 +17,7 @@
     "lint:html": "htmlhint 'packages/**/*.html'",
     "lint:js": "eslint --fix .",
     "lint:md": "remark -qf .",
+    "packages:publish": "npx lerna exec -- yarn publish --access=public",
```
antleblanc: As `lerna` is installed as a `devDependencies`, I didn't found any reason to prefer using `npx lerna` over `lerna` in that context (only if you are running `yarn install --production=false` I mean).

Otherwise I personally prefer `"publish:packages"` as script name. Later we can imagine things like `"publish:documentation"` for instance.

jleveugle: > We can run only `yarn publish` without re-installing the project :)

I prefer the first script name, because the first part (packages) is more generic than the second one (publish)

If i take your example with documentation, we can imagine:

```
packages:clean
packages:publish
documentation:clean
documentation:generate
documentation:publish
```

Imo, i prefer to keep this order: most generic at least generic

@cbourgois @frenautvh, what sounds good for you ?

frenautvh: i would say "publish" is the action and prefer 'publish:packages" to be consistent with "lint:js" for example

antleblanc: I will put some credits on @cbourgois answer:)

jleveugle: @cbourgois waiting your answer

I merge this PR atm, we should update the script name later :)

cbourgois: IMO `xxx:publish` sounds better (group functionnality by _context_)

I think it's better to have : 

```
packages:clean
packages:publish
documentation:clean
documentation:generate
documentation:publish
```

And if we want, add some 

`publish (run packages:publish && documentation:publish)`

PR #40, opened 2018-12-03
Refactor workspace : move packages in packages/manager

cbourgois:
### Description of the Change

Only changes in root `package.json` (see https://github.com/ovh-ux/manager/pull/40/files#diff-b9cfc7f2cdf78a7f4b91a753d10865a2R8) 
and in `.eslintrc.js`(see https://github.com/ovh-ux/manager/pull/40/files#diff-e4403a877d80de653400d88d85e4801aR18)

There is no changes in packages, there is only moving changes : 
* `/packages/manager-core` become `/packages/manager/core`
* `/packages/manager-welcome` become `/packages/manager/welcome`
* and so on.

7093e42 - refactor: move packages in packages/manager


jleveugle:
Good Job @cbourgois !!! :)

antleblanc:
What's about adding `manager/*` in the `workspaces` entry instead?

```diff
diff --git a/package.json b/package.json
index 5ddb48126d42..3adc57e1001d 100644
--- a/package.json
+++ b/package.json
@@ -5,7 +5,8 @@
   "description": "OVH Control Panel also known as Manager",
   "license": "BSD-3-Clause",
   "workspaces": [
-    "packages/*"
+    "packages/*",
+    "manager/*"
   ],
   "scripts": {
     "build": "lerna run build --stream --parallel",
```

ref: https://github.com/ovh-ux/manager/blob/develop/package.json#L8

cbourgois:
Why 2 entries ? Packages are now in `packages/manager`.

If we define workspaces to `packages/*`, no `package.json` are found.
If we define workspaces to `manager/*`, `manager` directory doesn't exist.

It's why I've changed it to `packages/manager/*`.

PR #49, opened 2018-12-10
perf(lodash): avoid importing global lodash

jleveugle:
Avoid importing global lodash and get these kind of errors: 

```
@ovh-ux/manager-freefax: [BABEL] Note: The code generator has deoptimised the styling of /Users/jleveugle/Documents/Work/github/ovh-ux/manager/node_modules/lodash/lodash.js as it exceeds the max of 500KB.
```

Smaller is better :D

### DIFF
--- a/packages/manager/freefax/src/notifications/freefax-notifications.factory.js
+++ b/packages/manager/freefax/src/notifications/freefax-notifications.factory.js
@@ -1,5 +1,10 @@
 import angular from 'angular';
-import _ from 'lodash';
+
+import assignIn from 'lodash/assignIn';
+import isBoolean from 'lodash/isBoolean';
+import isEmpty from 'lodash/isEmpty';
+import pick from 'lodash/pick';
+import snakeCase from 'lodash/snakeCase';
 
 export default /* @ngInject */ ($translate) => {
   const template = {
@@ -13,10 +18,10 @@ export default /* @ngInject */ ($translate) => {
      * @param {Object} data Data from AAPI
      */
   const FreefaxNotificationObject = function FreefaxNotificationObject(data) {
-    _.extend(
+    assignIn(
       this,
       template,
-      _.pick(
+      pick(
         data,
         Object.keys(template),
       ),
@@ -36,7 +41,7 @@ export default /* @ngInject */ ($translate) => {
      * Enter Edit Mode
      */
   FreefaxNotificationObject.prototype.edit = function edit() {
-    this.tempValue = _.pick(this, Object.keys(template));
+    this.tempValue = pick(this, Object.keys(template));
     this.toggleEdit(true);
   };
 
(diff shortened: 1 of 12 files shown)


packages/manager/overthebox/src/order/order-overTheBox.controller.js
```diff
@@ -1,4 +1,4 @@
-import _ from 'lodash';
+import first from 'lodash/first';
```
FredericEspiau: Please use `head` instead

https://github.com/lodash/lodash/wiki/Changelog#compatibility-warnings

yarn.lock
```diff
@@ -1866,11 +1866,6 @@ ajv@^6.5.3, ajv@^6.5.5:
     json-schema-traverse "^0.4.1"
     uri-js "^4.2.2"
 
-amdefine@>=0.0.4:
```
FredericEspiau: Lots of changes not related to the PR

packages/manager/overthebox/src/details/overTheBox-details.constant.js
```diff
@@ -1,4 +1,5 @@
-import _ from 'lodash';
+import first from 'lodash/first';
```
FredericEspiau: Please use `head` instead

https://github.com/lodash/lodash/wiki/Changelog#compatibility-warnings

packages/manager/freefax/src/notifications/freefax-notifications.factory.js
```diff
@@ -1,5 +1,10 @@
 import angular from 'angular';
-import _ from 'lodash';
+
+import extend from 'lodash/extend';
```
FredericEspiau: Please use `assignIn` instead

https://github.com/lodash/lodash/wiki/Changelog#compatibility-warnings

yarn.lock
jleveugle: due to yarn upgrade :)

antleblanc:
According to the `warning?` message you have from `manager-freefax` section, what's about applying the `perf` commit to this specific section instead?

It is a bit confusing here if I pay attention to the whole diff because you are targeting multiple `workspace` such as `core` and `overthebox` for instance.

I'm not convince about specify `lodash` as a `scope` to your commit message but it's probably a personal feeling.

What is your point of vue for example targeting a `section|workspace` and then applying the related `perf` commit to?

packages/manager/overthebox/src/details/overTheBox-details.constant.js
jleveugle: done :)

packages/manager/overthebox/src/order/order-overTheBox.controller.js
jleveugle: done

packages/manager/freefax/src/notifications/freefax-notifications.factory.js
jleveugle: done

@FredericEspiau thx for reviewing ;)

jleveugle:
@antleblanc The perf commit is cross workspaces, each workspace will get a trace in the history / changelog with specific files for the current workspace :)

We must see here a way to standardize the changes across all the workspaces (and the message commit too :)) 

For the scope, I tried. I can update it if needed, what is your purpose? :)

antleblanc:
Due to the fact that you keep only a single commit, the scope `lodash` sounds good.

The thing that I'm pointing out here is the description which is related to `@ovh-ux/manager-freefax`. 

This warning will be also displayed with the `core` and `overthebox` package too.

PR #52, opened 2018-12-11
fix(overthebox): add missing modify translation

cbourgois:
### Description of the Change

d9938bb - fix(overthebox): add missing modify translation

ref: OM-209

### DIFF
--- a/packages/manager/overthebox/src/details/overTheBox-details.html
+++ b/packages/manager/overthebox/src/details/overTheBox-details.html
@@ -136,9 +136,9 @@
                                           data-ng-bind="OverTheBoxDetails.device.deviceId">
                                     </span>
                                     <button class="btn btn-default no-transition"
-                                            data-translate="modify"
+                                            data-translate="overTheBox_deviceId_modify"
                                             type="button"
-                                            title="{{::'overTheBox_deviceId_button_title' | translate }} {{::'modify' | translate }}"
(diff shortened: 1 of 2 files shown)


packages/manager/overthebox/src/details/overTheBox-details.html
```diff
@@ -136,9 +136,9 @@
                                           data-ng-bind="OverTheBoxDetails.device.deviceId">
                                     </span>
                                     <button class="btn btn-default no-transition"
-                                            data-translate="modify"
+                                            data-translate="overTheBox_deviceId_modify"
                                             type="button"
-                                            title="{{::'overTheBox_deviceId_button_title' | translate }} {{::'modify' | translate }}"
+                                            title="{{::'overTheBox_deviceId_button_title' | translate }} {{::'overTheBox_deviceId_modify' | translate }}"
```
FredericEspiau: `Identifiant du périphérique : Modifier` ? 🤔

Anyway `title` is hardly accessible for screen-readers and mobile, you can remove it safely

antleblanc: @FredericEspiau Good catch.

As we are on the migration phase we definitely cannot put too much effort to address all that kind of things.

We have to tackle this migration in a small period that doesn't offer too much possibility to provide such a greate enhancement you pointed out.

But otherwise, feel free to open an Issue or a Pull request, it will we well appreciate:)

PR #61, opened 2018-12-12
fix: fix necessary double click on cancel button

JeremyDec:
MANAGER-1857


packages/manager/sms/src/users/change-password/telecom-sms-users-change-password.html
```diff
@@ -1,101 +1,71 @@
...
+      id="$ctrl.changePasswordUserForm"
+      data-ng-submit="$ctrl.changePassword()">
 
     <div class="modal-header">
         <button
             type="button"
             class="close float-right"
-            aria-label="{{ ::'sms_cancel' | translate }}"
-            data-ng-click="UsersChangePasswordCtrl.cancel()">
+            aria-label="{{ ::'cancel' | translate }}"
```
antleblanc: There is no reason to specify previous translation key I guess. You can stay with all news.

PR #63, opened 2018-12-13
refactor: update use of telecom-styles

cbourgois:
### Description of the Change

Build @ovh-ux/manager-telecom-styles
Use build in freefax, overthebox and sms packages
Remove unused less variables in telecom-styles

BREAKING CHANGE : @ovh-ux/manager-universes-styles is now @ovh-ux/manager-telecom-styles

    Before:
      import @ovh-ux/manager-telecom-styles/src/telecom.less in less

    After:
      import @ovh-ux/manager-telecom-styles in javascript

186f83e - refactor: update use of telecom-styles

ref:


packages/manager/telecom-styles/package.json
```diff
@@ -0,0 +1,21 @@
+{
+  "name": "@ovh-ux/manager-telecom-styles",
```
antleblanc: It would be great to add some meta (`description` and `keywords` for instance).

cbourgois: @antleblanc : Like this ?

```json 
  "description" : "Legacy styles from ovh-manager-telecom",
  "keywords": [
    "control",
    "ovh",
    "panel",
    "style",
    "telecom"
  ],
```

antleblanc: LGTM

PR #75, opened 2019-01-02
build: remove start entry in root package.json

jleveugle:


antleblanc:
We have to update the `README.md` file as well in order to provide some documentation.

PR #92, opened 2019-01-09
ci(cds): add workflow as code

jleveugle:


antleblanc:
@sguiheux Does the generated files could have the `.yaml` extensions instead? (ref: https://yaml.org/faq.html)

.cds/build-manager-dist.pip.yml
```diff
@@ -0,0 +1,18 @@
+version: v1.0
+name: build-manager-dist
+jobs:
+- job: Download
+  steps:
+  - script:
+    - '#!/bin/bash'
```
antleblanc: Does the literal style could be used from here? (eg: https://symfony.com/doc/current/components/yaml/yaml_format.html#strings)

.cds/build-manager-ng-app.pip.yml
```diff
@@ -0,0 +1,26 @@
+version: v1.0
+name: build-manager-ng-app
+parameters:
+  name:
+    type: string
+    description: npm module name
```
antleblanc: This description doesn't mean nothing right?

.cds/build-manager-dist.pip.yml
jleveugle: eg the documentation:

```
Strings containing any of the following characters must be quoted. Although you can use double quotes, for these characters it is more convenient to use single quotes, which avoids having to escape any backslash \:

:, {, }, [, ], ,, &, *, #, ?, |, -, <, >, =, !, %, @, `
```

.cds/build-manager-ng-app.pip.yml
jleveugle: Yes, we should update it ;)

jleveugle: The pipeline requires a parameter which is the name of the module you want to build.

What's your purpose ? :)

.cds/build-manager-dist.pip.yml
jleveugle: I think the best way is to have external scripts, and our job launch only this script.

We should avoid import our script directly in YAML files. But Step 2.

Here the proposal is to switch to "workflow as code", so it's only an cleaning extract of our remote workflow.

Next step: Make file / NPM script etc etc etc :)

antleblanc: Yes I mean/ask if we are able to write the script on multiline or not?

```yaml
|
  echo 'foo'
  cd/mv whatever
  …
```

.cds/build-manager-ng-app.pip.yml
antleblanc: It could simply reflect the `name`.

```txt
Build OVH Manager AngularJS applications
```

.cds/build-manager-dist.pip.yml
jleveugle: This code is autogenerated by CDS, but i think it's possible :)

.cds/build-manager-ng-app.pip.yml
jleveugle: It's not the name of the pipeline, it's the name of your npm package that you want to build like @ovh-ux/manager-sms or @ovh-ux/manager-core

-> `npm module name` is the description of the parameter "name"

antleblanc:
Is there any reason about closing this pull request?

PR #100, opened 2019-01-10
build: update rules to check if repository should be published

jleveugle:


scripts/common/repository.js
```diff
@@ -1,10 +1,13 @@
 /* eslint-disable global-require, import/no-dynamic-require, no-use-before-define */
+const concat = require('concat-stream');
 const bump = require('conventional-recommended-bump');
 const execa = require('execa');
 const fs = require('fs');
 const github = require('octonode');
 const path = require('path');
 const semver = require('semver');
+const gitRawCommits = require('git-raw-commits');
```
antleblanc: Both `require` could be sorted?

scripts/common/repository.js
```diff
@@ -209,9 +212,19 @@ class Repository {
...
+    return new Promise((resolve) => {
+      this.constructor.fetchTags().then((tag) => {
+        gitRawCommits({
+          format: '%B%n-hash-%n%H',
+          from: tag || '',
+          path: this.location,
+        })
+          .pipe(conventionalCommitsParser())
+          .pipe(concat((commits) => {
+            resolve(commits.filter(commit => commit.type === 'fix' || commit.type === 'feat' || commit.type === 'perf' || commit.notes.length > 0).length > 0);
```
antleblanc: Could be turned into an `include`?

```js
['fix', 'feat', 'perf'].include(commit.type) || commit.notes.length > 0
```

PR #101, opened 2019-01-11
feat(core): add sso config for ws servicetype

cbourgois:
### Description of the Change

3be50da — feat(core): add sso config for ws servicetype


antleblanc:
As we are migrating few sections, the `core` module (which is quite new) is evolving along the migration phase.

Giving the `ws` serviceType sounds more like adding a feature to this module rather than a fix.

Isn't it?

PR #103, opened 2019-01-15
fix(core): add missing ovh-ui-angular translations

frenautvh:
In order to translate ovh-oui-angular components in our standalone apps

### DIFF
--- a/packages/manager/modules/core/src/translate/ovh-ui-angular/index.js
+++ b/packages/manager/modules/core/src/translate/ovh-ui-angular/index.js
@@ -1,0 +1,110 @@
+import angular from 'angular';
+import 'angular-translate';
+import 'ovh-ui-angular';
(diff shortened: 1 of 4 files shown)


packages/manager/modules/core/src/translate/ovh-ui-angular/index.js
```diff
@@ -0,0 +1,110 @@
+import angular from 'angular';
+import 'angular-translate';
+import 'ovh-ui-angular';
+import '@uirouter/angularjs';
+import _ from 'lodash';
```
jleveugle: Can we avoid importing full lodash?

FredericEspiau: https://github.com/lodash/babel-plugin-lodash would be useful ;)

PR #109, opened 2019-01-16
feat(sms): add global sms order

cbourgois:
### Description of the Change

Update sms states to sms.service.*
Add sms.order state
BREAKING CHANGE:
    Sms service state is now 'sms.service' instead of 'sms'

e00e962 — feat(sms): add global sms order


cbourgois:
I should test back button in global order before merge.

PR #151, opened 2019-01-25
fix(overthebox): add missing warning section

frenautvh:


packages/manager/modules/overthebox/src/remote/index.js
```diff
@@ -6,6 +6,7 @@ import template from './overTheBox-remote.html';
 const moduleName = 'ovhManagerOtbRemote';
 
 angular.module(moduleName, [])
+  .run(/* @ngTranslationsInject:json ./translations */)
```
FredericEspiau: `json` and `./translations` are two parameters ? Then `:` states the start of the parameters and they are separated by spaces ? Whats if I want my translation folder to be called `./global translations` ? :D

FredericEspiau: Awesome !

PR #204, opened 2019-02-21
Upgrade ng-ovh-apiv7

antleblanc:


antleblanc:
It will be handled in a separate pull request.

PR #214, opened 2019-02-26
feat(cloud-universe-components): add currency service and filter

cbourgois:
### Description of the Change

ref: OM-319

ab4ed65 — feat(cloud-universe-components): add currency service and filter

### DIFF
--- a/packages/manager/modules/cloud-universe-components/src/currency/constants.js
+++ b/packages/manager/modules/cloud-universe-components/src/currency/constants.js
@@ -1,0 +1,38 @@
+export const CURRENCY_SUBSIDIARY = {
+  // america
+  CA: 'C$',
+  QC: 'C$',
+  WE: 'USD$',
+  WS: 'USD$',
+  // US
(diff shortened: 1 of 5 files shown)


packages/manager/modules/cloud-universe-components/src/currency/constants.js
```diff
@@ -0,0 +1,38 @@
+export const CURRENCY_SUBSIDIARY = {
```
FredericEspiau: I know this is not the point of this PR but https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/NumberFormat

packages/manager/modules/cloud-universe-components/src/currency/filter.js
```diff
@@ -0,0 +1,11 @@
+import _ from 'lodash';
```
jleveugle: We should use `lodash/import` :)

PR #251, opened 2019-03-12
Fix rems in cloud styles

frenautvh:
fix rems for cloud-style module
fix rems for cloud-universe-components


packages/manager/modules/cloud-styles/src/cloud.less
```diff
@@ -107,7 +122,10 @@
   @import './cloud/project/openstack/openstack';
   @import './cloud/project/openstack/users/openrc/openstack-users-openrc';
   @import './cloud/project/storage/storage-add/storage-add';
-  @import './cloud/project/storage/storage';
+
+  /* @TODO fix storage style */
+
+  /* @import './cloud/project/storage/storage'; */
```
antleblanc: Shouldn't be imported as well?

antleblanc: Shouldn't be imported?

packages/manager/modules/cloud-styles/.prerelease
```diff
@@ -0,0 +1 @@
+alpha
```
antleblanc: Is there any reason to no prefer using `yarn run split` command?

jleveugle: Splitting is the right way to test a module outside the monorepo, but it's complicated to split multiple branches and use splitted branch inside the monorepo. All of us are working on the same things, and we need to merge our work without publising a final release.

The .prerelease file let us manage prerelease per package. The main objective is to work on the same codebase, and avoid keeping a lot of big branches, with packages which need to be test by others packages insides monorepository. Let's merge this one on develop, if we have some issue with this package, we fix it and prerelease again.

When the package is stable, let's remove the .prerelease file and publish the final version.

PR #261, opened 2019-03-13
fix(manager-core): add missing peerDeps ngOvhSwimmingPoll

antleblanc:
# Add missing peerDeps

Since we upgraded to `@ovh-ux/ng-ovh-api-wrappers@^3.0.0`, we have to add "@ovh-ux/ng-ovh-swimming-poll" as a peerDependency

### DIFF
--- a/packages/manager/modules/core/package.json
+++ b/packages/manager/modules/core/package.json
@@ -28,6 +28,7 @@
     "@ovh-ux/ng-ovh-api-wrappers": "^3.0.0",
     "@ovh-ux/ng-ovh-http": "^4.0.1-beta.0",
     "@ovh-ux/ng-ovh-sso-auth": "^4.0.0-beta.0",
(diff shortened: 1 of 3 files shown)


antleblanc:
The commit message type has been set to a `fix` in order to be able to target a new patch for the given module `@ovh-ux/manager-core`.

Let's assume that you want to start the `@ovh-ux/manager-sms`.

Actually you get this following error from the console:

```js
Error: [$injector:modulerr] Failed to instantiate module smsApp due to:
[$injector:modulerr] Failed to instantiate module ovhManagerSms due to:
[$injector:modulerr] Failed to instantiate module oc.lazyLoad due to:
[$injector:nomod] Module 'ngOvhSwimmingPoll' is not available! You either misspelled the module name or forgot to load it. If registering a module ensure that you specify the dependencies as the second argument.

```

PR #273, opened 2019-03-15
fix vrack module / cloud styles

frenautvh:
- move vrack styles from cloud-style to vrack module
- fix vrack oui-pen icon size


packages/manager/modules/vrack/src/add/vrack-add.less
```diff
@@ -1,5 +1,4 @@
-@cloud_green_3: #65afa0;
-@cloud_green_5: #e1f0ed;
+@import '~@ovh-ux/manager-cloud-styles/src/cloud/cloud.variables.less';
```
antleblanc: Is `.less` needed?

frenautvh: yes because path name already has a dot

antleblanc: I'm a bit confused since you've removed the `.less` from here: https://github.com/ovh-ux/manager/pull/273/files#diff-1c6bdb7534e194962b4e2c770538a58bR1

PR #314, opened 2019-03-20
fix(pci): add cloud-legacy class to body

frenautvh:


frenautvh:
because ovh responsive popovers are appended to the body

antleblanc:
@frenautvh Could you explain why this PR is closed? If it's a duplicate or it doesn't fix the root cause, then could you please delete the branch? Thanks!

PR #331, opened 2019-03-25
feat: add uirouter-layout component

cbourgois:
### Description of the Change

* add `@ovh-ux/ng-ovh-uirouter-layout` package

ba23cf3a — feat: add uirouter-layout component

/cc @jleveugle @frenautvh @marie-j


packages/components/ng-ovh-uirouter-layout/src/index.js
```diff
@@ -0,0 +1,205 @@
...
+     *    - toChilds (default value: 'false'):
+     *      - can be a boolean: 'true' to declare the modal state to all direct parent childs.
+     *      - or an array of string that contains the childs of the direct parents where the modal
+     *        needs to be displayed.
+     *    - ignoreChilds: an array of string that contains states names where the modal state
+     *                    doesn't need to be displayed.
+     */
+    $stateProvider.decorator('layout', (state) => {
+      let modalLayout;
+      if (get(state, 'self.layout') === 'modal' || get(state, 'self.layout.name') === 'modal') {
```
jleveugle: We should check if `state.self.layout` is an `object` or a `string`

packages/components/ng-ovh-uirouter-layout/src/index.js
```diff
@@ -0,0 +1,205 @@
...
+        // close previous modal
+        if (modalInstance) {
+          modalInstance.close();
+        }
+
+        if (get(state, 'layout.name') === 'modal') {
+          const $state = transition.injector().get('$state');
+          const $uibModal = transition.injector().get('$uibModal');
+
+          const componentName = get(state, state.views.modal ? 'views.modal.component' : 'component');
```
jleveugle: I don't understand why we need to check `state.views.modal`. Why do we not use the default view ?

packages/components/ng-ovh-uirouter-layout/src/index.js
```diff
@@ -0,0 +1,205 @@
...
+          layout: { // don't know why full config must be defined???
+            name: 'modal',
+            toChilds: false,
+            ignoreChilds: [],
+          },
+        });
+      });
+    });
+
+    $urlService.listen(); // Start responding to URL changes
```
jleveugle: I'm not sure we need to stop and listen URL changes, because modules which are defining the routing will require this module :)

Cf: https://docs.angularjs.org/guide/module

> After that, the configuration blocks (config functions) are executed. This means the configuration blocks of the required modules execute before the configuration blocks of any requiring module.

We should remove this I think, because config for state will always be executed after config of this module (if dependency are proper) :)

packages/components/ng-ovh-uirouter-layout/rollup.config.js
```diff
@@ -0,0 +1,25 @@
...
+  },
+});
+
+export default [
+  config.cjs(),
+  config.umd({
+    output: {
+      globals: {
+        angular: 'angular',
+        d3: 'd3',
```
marie-j:
```suggestion
```

packages/components/ng-ovh-uirouter-layout/rollup.config.js
```diff
@@ -0,0 +1,25 @@
...
+    paths: [
+      path.resolve(__dirname, 'node_modules'),
+      path.resolve(__dirname, '../../../../node_modules'),
+    ],
+  },
+});
+
+export default [
+  config.cjs(),
+  config.umd({
```
marie-j: Don't you need to also  add 
```suggestion
  config.es()
```
?

packages/components/ng-ovh-uirouter-layout/src/index.js
cbourgois: You could define state like this : 

```js
$stateProvider
  .state('project.storage_blocks', {
    url: '/storages/blocks',
    layout: 'modal',
    views: {
        '@main': {
          component: 'myAwesomeComponent',
        },
        modal: {
          component: 'myModalComponent',
        },
    },
  );
```

the `views.modal` will open in modal.

packages/components/ng-ovh-uirouter-layout/rollup.config.js
cbourgois: 👍

packages/components/ng-ovh-uirouter-layout/src/index.js
cbourgois: Removed.

packages/components/ng-ovh-uirouter-layout/src/index.js
jleveugle: Strange no ? If you open a modal, why do we want to display another view in the background ?

packages/components/ng-ovh-uirouter-layout/src/index.js
jleveugle: nice ;)

PR #339, opened 2019-03-28
build(apps.vrack): remove repository entry from package.json file

antleblanc:
# Remove `repository` entry from `package.json` file

### ⚙️ Build

6b960da - build(apps.vrack): remove repository entry from package.json file

since this package is set to be private, there is no benefit of 
providing a `repository` entry.

### 🏠 Internal

- No QC required

### DIFF
--- a/packages/manager/apps/vrack/package.json
+++ b/packages/manager/apps/vrack/package.json
@@ -2,11 +2,6 @@
   "name": "@ovh-ux/manager-vrack-app",
   "version": "0.1.0",
   "private": true,
-  "repository": {
-    "type": "git",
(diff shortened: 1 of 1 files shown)


packages/manager/apps/vrack/package.json
```diff
@@ -2,11 +2,6 @@
   "name": "@ovh-ux/manager-vrack-app",
   "version": "0.1.0",
   "private": true,
-  "repository": {
-    "type": "git",
-    "url": "git+https://github.com/ovh-ux/manager.git",
-    "directory": "packages/manager/app/vrack"
```
antleblanc: Furthermore the path isn't correct (`app` instead of `apps`).

PR #343, opened 2019-03-29
feat(pci): add kubernetes to public-cloud app

marie-j:
# Add kubernetes 

cc \ @cbourgois @jleveugle @antleblanc

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/kubernetes/index.js
+++ b/packages/manager/modules/pci/src/projects/project/kubernetes/index.js
@@ -1,0 +1,15 @@
+import angular from 'angular';
+
+import '@ovh-ux/manager-kubernetes';
+
+import routing from './routing';
+
+const moduleName = 'ovhManagerPciProjectKubernetes';
+
+angular
+  .module(moduleName, [
+    'ovhManagerKubernetes',
+  ])
+  .config(routing);
+
+export default moduleName;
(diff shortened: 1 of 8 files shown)


packages/manager/modules/pci/src/projects/project/kubernetes/index.js
```diff
@@ -0,0 +1,15 @@
+import angular from 'angular';
+
+import kubernetes from '@ovh-ux/manager-kubernetes';
```
antleblanc:
```suggestion
import '@ovh-ux/manager-kubernetes';
```

packages/manager/modules/pci/src/projects/project/kubernetes/index.js
```diff
@@ -0,0 +1,15 @@
...
+
+import routing from './routing';
+
+const moduleName = 'ovhManagerPciProjectKubernetes';
+
+angular
+  .module(moduleName, [
+    kubernetes,
```
antleblanc:
```suggestion
    'ovhManagerKubernetes',
```

packages/manager/modules/pci/src/projects/project/kubernetes/index.js
jleveugle: Nop, it's in the same module

antleblanc: Due to the fact that you are importing a peerDependency, I encourage you to import the module as we already done across others modules.

jleveugle: You are right ;)

PR #344, opened 2019-03-29
build: bump ovh-api-services to v6.7.0

marie-j:
# Bump `ovh-api-services` to v6.7.0

cc \ @cbourgois @jleveugle @antleblanc

### DIFF
--- a/packages/manager/apps/freefax/package.json
+++ b/packages/manager/apps/freefax/package.json
@@ -34,7 +34,7 @@
     "angular-translate-loader-pluggable": "^1.3.1",
     "oclazyload": "^1.1.0",
     "ovh-angular-ui-confirm-modal": "^1.0.2",
-    "ovh-api-services": "^6.4.0",
+    "ovh-api-services": "^6.7.0",
     "ovh-ui-angular": "^2.21.4",
     "ovh-ui-kit": "^2.23.1",
     "ovh-ui-kit-bs": "^2.1.1"
--- a/packages/manager/apps/kubernetes/package.json
+++ b/packages/manager/apps/kubernetes/package.json
@@ -24,6 +24,7 @@
     "angular": "^1.7.5",
     "angular-ui-bootstrap": "^1.3.3",
     "font-awesome": "^4.7.0",
+    "ovh-api-services": "^6.7.0",
     "ovh-ui-angular": "~2.25.1",
     "ovh-ui-kit": "~2.25.0"
   },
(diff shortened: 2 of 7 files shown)


antleblanc:
Is the bump shouldn't be also applied to other modules?

jleveugle:
NPM Dependencies using angularjs should be bumped in one time and we should avoid keeping multiple versions of angularjs modules as external dependencies in the monorepository.

Will try to explain why with one example:

For example, we want to import `ng-moduleA@1.1.0` inside a first host app

Our host app `package.json`:

```json
{
	"name": "my-host-app",
	"dependencies": {
		"ng-moduleA": "^1.1.0",
		"ng-moduleB": "^6.4.0",
	}
}
```

Other fact, `ng-moduleB` is used in multiple packages inside monorepository and has a peerDependency to `ng-moduleA`:

```json
{
	"name": "ng-moduleB"
	"peerDependencies": {
		"ng-moduleA": "^1.1.0",
	}
}
```

... A long time ...

Oh! We want a new host app, named `my-host-app-2` and we need `ng-moduleA` with a new feature.

So, we have to declare a new feature in `ng-moduleA`. We are going to develop a `ng-moduleA@1.2.0`.

`package.json` from `my-host-app-2` looks like:

```json
{
	"name": "my-host-app",
	"dependencies": {
		"ng-moduleA": "^1.2.0",
		"ng-moduleB": "^6.4.0",
	}
}
```

At this step, all of our packages looks compatibles, and it's true!

BUT :)

Let's have a look inside our `node_modules` structure (due to yarn workspace):

```
/node_modules/ng-moduleA (@1.1.0)
/node_modules/ng-moduleB (@6.4.0)
/packages/my-host-app-2/node_modules/ng-moduleA (@1.2.0)
```

And here, we can have an issue, because `my-host-app-2` requires `ng-moduleB`, and `ng-moduleB` requires `ng-moduleA`. Due to path resolutions, `ng-moduleB` will search `ng-moduleA` like this:

```
./node_modules (its own node_modules)
../node_modules (root node_modules) -> find `ng_moduleA@1.1.0`
/packages/my-host-app-2/node_modules/ (due to webpack configuration, but it will never hit this path because it already found it before)
```

So, we will use `ng-moduleA@1.1.0` insides `my-host-app-2` and it will result by a crash, because we don't get our awesome feature developed in `ng-moduleA@1.2.0`

The first `./node_modules` is due to our configuration: https://github.com/ovh-ux/manager-webpack-config/blob/master/src/webpack.common.js#L68

In fact, we thought about removing it but it's a bad idea, because if we get the same case with a dependency or devDependency, we will get another wrong version because we will check inside absolute path before the module `node_modules` folder.

So, to resume, we can get this case only if peerDependencies hosted outside the monorepository (angularjs module here) are not up-to-date between all packages inside the monorepository. It's why we decided to bump all packages from another repositories at this moment.

Feel free to ask me some details if needed :D

See you!

PR #346, opened 2019-04-02
feat(project): user are now able to edit project and define default

jleveugle:


packages/manager/modules/pci/src/projects/project/edit/edit.html
```diff
@@ -0,0 +1,23 @@
+<div class="text-center" data-ng-if="$ctrl.loading.init"><oui-spinner size="l"></oui-spinner></div>
```
marie-j: Indentation
```suggestion
<div class="text-center" data-ng-if="$ctrl.loading.init">
  <oui-spinner size="l"></oui-spinner>
 </div>
```

packages/manager/modules/pci/src/projects/project/edit/edit.controller.js
```diff
@@ -0,0 +1,99 @@
...
+
+    this.loading = {
+      init: false,
+      submit: false,
+    };
+
+    this.$ngInit();
+  }
+
+  $ngInit() {
```
marie-j: Current convention is to use `$onInit` hook instead. We should probably talk if there is a difference and what's better especially with performances ( FYI : https://docs.angularjs.org/api/ng/service/$compile#life-cycle-hooks only mentions `$onInit`)

packages/manager/modules/pci/src/projects/project/edit/edit.controller.js
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  submit() {
+    this.loading.submit = true;
+
+    return this.OvhApiCloudProject
+      .v6()
+      .put(
+        {
+          serviceName: this.$stateParams.projectId,
```
marie-j: This should probably be stored in as an attribute (if ever we want to change the way of fetching this information it will be easier)

packages/manager/modules/pci/src/projects/project/edit/edit.html
```diff
@@ -0,0 +1,23 @@
...
+                description="{{ ::'pci_projects_project_edit_default_project_description' | translate }}"
+                model="$ctrl.isDefault">
+            </oui-checkbox>
+        </oui-field>
+        <oui-button variant="primary" type="submit" text="{{ ::'pci_projects_project_edit_update' | translate }}" disabled="$ctrl.loading.submit"></oui-button>
+    </form>
+    <div class="mt-4 pt-4 border-top">
+        <h2 data-translate="pci_projects_project_edit_remove"></h2>
+        <p>Lorem ipsum dolor sit amet, consectetur adipiscing elit. Nulla sit amet consequat turpis, at accumsan metus. Curabitur pellentesque condimentum mauris id iaculis. Sed id ex tristique, ornare tortor quis, bibendum lacus. Sed ornare elit eu malesuada auctor. Integer tempor nunc non tincidunt suscipit. Aenean faucibus, felis nec aliquet condimentum, est risus dapibus nibh, vel cursus massa mi sed nibh. Vivamus pulvinar arcu arcu, at consequat dui fermentum sed.</p>
+        <oui-button variant="secondary" text="{{ ::'pci_projects_project_edit_remove' | translate }}" disabled="$ctrl.loading.submit"></oui-button>
```
marie-j: `text` is deprecated. Prefer using a nested `span` or `{{ }}` notation https://ovh-ux.github.io/ovh-ui-kit-documentation/#!/oui-angular/button

packages/manager/modules/pci/src/projects/project/project.html
```diff
@@ -2,24 +2,34 @@
...
+                </oui-header-tabs-item>
             </oui-header-tabs>
         </oui-page-header>
-        <div class="p-5" data-ui-view>
+        <div class="p-5" data-ui-view="contentTab">
             <div class="container-fluid px-0">
                 <div class="row">
-                    <div class="col-sm-6 col-lg-4 col-xl mb-4 mb-xl-0" data-ng-repeat="action in $ctrl.actions track by $index">
+                    <div class="col-sm-6 col-lg-4 col-xl mb-4" data-ng-repeat="action in $ctrl.actions track by $index">
                         <a class="text-center w-100 d-block oui-tile py-4" href="#">
```
marie-j: Not yours but strange to have a `.oui-tile` on an anchor

packages/manager/modules/pci/src/projects/project/edit/edit.controller.js
jleveugle: Will switch to component instead of controller / view

packages/manager/modules/pci/src/projects/project/edit/edit.html
jleveugle: Documentation is outdated so :(

packages/manager/modules/pci/src/projects/project/project.html
jleveugle: Nop, because in this case, we want to get some clickable tiles :D

packages/manager/modules/pci/src/projects/project/edit/edit.html
jleveugle: done

PR #368, opened 2019-04-12
feat: add empty component

marie-j:
### Description of the change

Add component for empty pages 

### Related 
- [ ] https://github.com/ovh-ux/ovh-ui-kit/pull/386

### DIFF
--- a/packages/manager/modules/pci/src/components/project/empty/empty.component.js
+++ b/packages/manager/modules/pci/src/components/project/empty/empty.component.js
@@ -1,0 +1,13 @@
+import controller from './empty.controller';
+import template from './empty.html';
+
+export default {
+  controller,
+  template,
+  transclude: true,
+  bindings: {
+    imageAlt: '<',
+    imageSource: '<',
+    guides: '<?',
+  },
+};
--- a/packages/manager/modules/pci/src/components/project/empty/empty.html
+++ b/packages/manager/modules/pci/src/components/project/empty/empty.html
@@ -1,0 +1,36 @@
+<div class="container-fluid px-0 pci-project-empty">
+    <div class="row">
+        <div class="col-md-3">
+            <div class="p-5" data-ng-class="{ 'placeholder' : !$ctrl.imageSource }">
+                <img
+                    data-ng-src="{{:: $ctrl.imageSource || $ctrl.placeholder }}"
+                    data-ng-attr-alt="{{:: $ctrl.imageAlt }}"
+                    alt=""
+                    class="w-100">
+            </div>
+        </div>
+        <div class="col-md-9" data-ng-transclude></div>
+    </div>
+
+    <aside class="row pt-5">
+        <div class="col-sm-6 col-md-4 mb-4" data-ng-repeat="guide in $ctrl.guides track by $index">
+            <div class="py-2 oui-tile d-flex align-items-start">
+                <span class="d-block oui-icon oui-icon-guides" aria-hidden="true"></span>
+                <div class="ml-2">
+                    <h2 class="oui-tile__title oui-heading_4"  data-ng-bind=":: guide.title"></h2>
+                    <div class="oui-tile__body mt-3">
+                        <p class="oui-tile__description" data-ng-bind=":: guide.description"></p>
+                        <a class="oui-link_icon"
+                            data-ng-href="{{:: guide.link }}"
+                            target="_blank"
+                            rel="noopener">
+                            <span data-translate="pci_project_empty_more"></span>
+                            <span class="oui-icon oui-icon-external_link" aria-hidden="true"></span>
+                            <span class="sr-only" data-translate="pci_project_empty_new_tab"></span>
+                        </a>
+                    </div>
+                </div>
+            </div>
+        </div>
+    </aside>
+</div>
--- a/packages/manager/modules/pci/src/components/project/empty/index.js
+++ b/packages/manager/modules/pci/src/components/project/empty/index.js
@@ -1,0 +1,16 @@
+import angular from 'angular';
+import 'angular-translate';
+
+import component from './empty.component';
+
+import './index.less';
+
+const moduleName = 'ovhManagerPciComponentsProjectEmpty';
+
+angular
+  .module(moduleName, [
+    'pascalprecht.translate',
+  ])
+  .component('pciProjectEmpty', component);
+
+export default moduleName;
--- a/packages/manager/modules/pci/src/components/project/empty/translations/Messages_fr_FR.json
+++ b/packages/manager/modules/pci/src/components/project/empty/translations/Messages_fr_FR.json
@@ -1,0 +1,4 @@
+{
+  "pci_project_empty_more": "En savoir plus",
+  "pci_project_empty_new_tab": "Nouvel onglet"
+}
--- a/packages/manager/modules/pci/src/components/project/empty/empty.controller.js
+++ b/packages/manager/modules/pci/src/components/project/empty/empty.controller.js
@@ -1,0 +1,7 @@
+import placeholder from './placeholder.svg';
+
+export default class {
+  constructor() {
+    this.placeholder = placeholder;
+  }
+}
--- a/packages/manager/modules/pci/src/components/project/empty/index.less
+++ b/packages/manager/modules/pci/src/components/project/empty/index.less
@@ -1,0 +1,7 @@
+.pci-project-empty {
+  @import '~ovh-ui-kit/packages/oui-color/_variables.less';
+
+  .placeholder {
+    background-color: @oui-color-mischka;
+  }
+}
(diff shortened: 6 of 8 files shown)


packages/manager/modules/pci/src/components/project/empty/empty.html
```diff
@@ -0,0 +1,43 @@
...
+                        <span data-ng-bind=":: guide.title"></span>
+                    </div>
+                    <div class="oui-tile__body mt-3">
+                        <p class="oui-tile__description" data-ng-bind=":: guide.description"></p>
+                        <a class="oui-link_icon"
+                            data-ng-href="{{:: guide.link }}"
+                            target="_blank"
+                            rel="noopener">
+                            <span data-translate="pci_project_empty_more"></span>
+                            <span class="oui-icon oui-icon-external_link" aria-hidden="true"></span>
```
FredericEspiau: Could you add a `<span class="sr-only">Opens a new tab</span>` ? :)

packages/manager/modules/pci/src/components/project/empty/empty.html
```diff
@@ -0,0 +1,43 @@
...
+                    alt="">
+            </div>
+        </div>
+        <div class="col-md-9" data-ng-transclude></div>
+    </div>
+
+    <div class="row pt-5">
+        <div class="col-sm-6 col-md-4 mb-4" data-ng-repeat="guide in $ctrl.guides track by $index">
+            <div class="py-2 oui-tile d-flex align-items-start">
+                <div>
```
FredericEspiau:
```suggestion
                <span class="d-block" oui-icon oui-icon-guides" aria-hidden="true"></span>
```
?

packages/manager/modules/pci/src/components/project/empty/empty.html
```diff
@@ -0,0 +1,43 @@
...
+                    data-ng-src="{{:: $ctrl.placeholder }}"
+                    class="w-100"
+                    alt="">
+            </div>
+        </div>
+        <div class="col-md-9" data-ng-transclude></div>
+    </div>
+
+    <div class="row pt-5">
+        <div class="col-sm-6 col-md-4 mb-4" data-ng-repeat="guide in $ctrl.guides track by $index">
```
FredericEspiau:
```suggestion
        <div class="col-sm-6 col-md-4 mb-4" data-ng-repeat="guide in ::$ctrl.guides track by $index">
```

packages/manager/modules/pci/src/components/project/empty/empty.html
```diff
@@ -0,0 +1,43 @@
+<div class="container-fluid px-0 pci-project-empty">
+    <div class="row">
+        <div class="col-md-3">
+            <div class="p-5 placeholder">
+                <img data-ng-if="$ctrl.imageSource"
+                    data-ng-src="{{:: $ctrl.imageSource }}"
+                    data-ng-attr-alt="{{:: $ctrl.imageAlt }}"
+                    alt=""
```
FredericEspiau:
```suggestion
```
?

packages/manager/modules/pci/src/components/project/empty/empty.html
```diff
@@ -0,0 +1,43 @@
...
+                    class="w-100">
+               <img data-ng-if="!$ctrl.imageSource"
+                    data-ng-src="{{:: $ctrl.placeholder }}"
+                    class="w-100"
+                    alt="">
```
FredericEspiau: :+1:

packages/manager/modules/pci/src/components/project/empty/empty.component.js
```diff
@@ -0,0 +1,13 @@
+import controller from './empty.controller';
```
FredericEspiau: Wouldn't `controller.js` suffice as a file name ? I'd like to know as I use modules too :)

packages/manager/modules/pci/src/components/project/empty/index.js
```diff
@@ -0,0 +1,14 @@
+import angular from 'angular';
+
+import component from './empty.component';
+
+import './index.less';
+
+const moduleName = 'ovhManagerPciComponentsProjectEmpty';
+
+angular
+  .module(moduleName, [
```
antleblanc: Since you are using `translate` directive, I think you have to declare a dependency to `angular-translate` module. Isn't it?

packages/manager/modules/pci/src/components/project/empty/empty.component.js
antleblanc: Instead of have too many file named exactly the same across the entire repository, then the closest module/folder name is used as a prefix I think.

packages/manager/modules/pci/src/components/project/empty/empty.html
```diff
@@ -0,0 +1,43 @@
+<div class="container-fluid px-0 pci-project-empty">
+    <div class="row">
+        <div class="col-md-3">
+            <div class="p-5 placeholder">
+                <img data-ng-if="$ctrl.imageSource"
+                    data-ng-src="{{:: $ctrl.imageSource }}"
```
antleblanc: Feel free to tell me if you find this more or less readable but why not just using a single `` tag?

```suggestion
                    data-ng-src="{{:: $ctrl.imageSource || $ctrl.placeholder }}"
```

In that way you can get rid of the `ngIf`.

packages/manager/modules/pci/src/components/project/empty/empty.html
antleblanc: Due to [HTMLHint configuration](https://github.com/ovh-ux/manager/blob/master/.htmlhintrc#L2) and `ngAttr` usage, it will report an error.

packages/manager/modules/pci/src/components/project/empty/empty.html
```diff
@@ -0,0 +1,43 @@
...
+               <img data-ng-if="!$ctrl.imageSource"
+                    data-ng-src="{{:: $ctrl.placeholder }}"
+                    class="w-100"
+                    alt="">
+            </div>
+        </div>
+        <div class="col-md-9" data-ng-transclude></div>
+    </div>
+
+    <div class="row pt-5">
```
antleblanc: Looks like an `<aside>` element. Isn't it?

packages/manager/modules/pci/src/components/project/empty/empty.html
```diff
@@ -0,0 +1,43 @@
...
+        <div class="col-sm-6 col-md-4 mb-4" data-ng-repeat="guide in $ctrl.guides track by $index">
+            <div class="py-2 oui-tile d-flex align-items-start">
+                <div>
+                    <span class="oui-icon oui-icon-guides" aria-hidden="true"></span>
+                </div>
+                <div class="ml-2">
+                    <div class="oui-tile__title" >
```
antleblanc: Couldn't be an heading tag? (and merged with the single child element).

packages/manager/modules/pci/src/components/project/empty/index.js
```diff
@@ -0,0 +1,14 @@
+import angular from 'angular';
+
+import component from './empty.component';
+
+import './index.less';
+
+const moduleName = 'ovhManagerPciComponentsProjectEmpty';
```
antleblanc: Must be a dependency of the `ovhManagerPciComponentsProject` module. Right?

packages/manager/modules/pci/src/components/project/empty/empty.component.js
marie-j: Indeed, they are prefixed that way in order to be easier to differenciate. Imo, it gets quite hard and messy when you start having all your files named `controller`

packages/manager/modules/pci/src/components/project/empty/empty.html
marie-j: I don't know if it makes sense here. Let's suppose the `guides` binding changes, those changes won't be reflected here

packages/manager/modules/pci/src/components/project/empty/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,4 @@
+{
+  "pci_project_empty_more": "En savoir plus",
+  "pci_project_empty_new_tab": "S'ouvre dans un nouvel onglet"
```
jleveugle: "Nouvel onglet" is enougth > Shorter is better :D

packages/manager/modules/pci/src/components/project/empty/empty.html
FredericEspiau: 😅 

okay, it didn't make sense as there was a `ngAttrAlt` but I get it now

packages/manager/modules/pci/src/components/project/empty/empty.html
FredericEspiau: Is it the case right now ? Because if it's not, it would mean we should never use `::` or I don't understand your point :)

packages/manager/modules/pci/src/components/project/empty/empty.component.js
FredericEspiau: It's yours to see, I had the same point of you before and then I realized I could search by folder name in VSCode and in Chromium so I've never add any real issue :)

PR #371, opened 2019-04-16
feat(kubernetes): add kubernetes to pci module

marie-j:
# Migrate kubernetes to PCI module

Move kubernetes into pci module 

## Main change 

Kubernetes module routing


jleveugle:
> Meaning that we need to deprecated the current package `@ovh-ux/manager-kubernetes` on npm. Isn't it?

Yes! :)

PR #385, opened 2019-04-25
build: upgrade ovh-ui-kit to v2.30.0 and ovh-ui-angular to v2.30.0

marie-j:
# Upgrade dependencies

## Description of the change

* Bump `ovh-ui-kit` to v2.30.0
* Bump `ovh-ui-angular` to v2.30.0


jleveugle:
> Not necessary for both `kubernetes` Application and the Module since the module will no more be published (moved to pci).

Need it to current status of develop branch :)

> Merge of kubernetes section is an objective of public-cloud branch, so in these PR, we should bump kube too :)

PR #397, opened 2019-04-27
fix: update volume snapshots list

cbourgois:
### Description of the Change

- remove `CucCloudMessage` listener in modals
- add `goToSnapshots` resolve and use it in `goBack` sub resolve 
  - navigate to volume storages list and trigger a `CucCloudMessage`
  - if the message is a `success` message, `$state.go` is called with `reload: true` option
- move initial data loading in `resolve` and remove related code/translations
- Fix display of region (using `CucRegionService`)
- Fix missing dependencies (`PciProjectStorageBlockService` service and `ovhManagerPciStoragesBlocksBlockVolumeEdit`component are used in volume creation from snapshots)
- add lazy loading

e358dff6 — fix: update volume snapshots list

/cc @jleveugle @marie-j

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/storages/snapshots/snapshot/create-volume/create-volume.controller.js
+++ b/packages/manager/modules/pci/src/projects/project/storages/snapshots/snapshot/create-volume/create-volume.controller.js
@@ -20,35 +20,25 @@ export default class PciBlockStorageSnapshotsCreateVolumeController {
   }
 
   $onInit() {
-    this.loadings = {
-      init: false,
-      save: false,
-    };
+    this.isLoading = false;
+    this.loadMessages();
 
-    this.initLoaders();
-  }
-
-  initLoaders() {
-    this.loadings.init = true;
-
-    this.$translate.refresh()
-      .then(() => this.loadMessages())
-      .then(() => this.getSnapshot())
-      .catch((err) => {
-        this.CucCloudMessage.error(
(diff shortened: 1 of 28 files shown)


packages/manager/modules/pci/src/projects/project/storages/snapshots/snapshot/create-volume/create-volume.controller.js
```diff
@@ -61,42 +51,14 @@ export default class PciBlockStorageSnapshotsCreateVolumeController {
...
-          type: snapshot.volume.type,
-          snapshotId: snapshot.id,
-        });
-        return this.editStorage;
-      });
-  }
-
   save() {
-    this.loadings.save = true;
+    this.isLoading = true;
```
antleblanc: `isSaving` couldn't make more sense?

packages/manager/modules/pci/src/projects/project/storages/snapshots/snapshot/delete/delete.controller.js
```diff
@@ -1,90 +1,37 @@
...
-  }
-
-  refreshMessages() {
-    this.messages = this.messageHandler.getMessages();
+    this.isLoading = false;
   }
 
   deleteSnapshot() {
-    this.loadings.save = true;
+    this.isLoading = true;
```
antleblanc: `isDeleting` couldn't make more sense?

PR #406, opened 2019-04-30
fix(sidebar): add oui-clipboard to copy id of project

jleveugle:


packages/manager/apps/public-cloud/src/assets/theme/_sidebar.less
```diff
@@ -9,6 +9,33 @@ cloud-sidebar { /* stylelint-disable-line */
...
   box-shadow: 0 2pt 10pt rgba(40, 89, 192, 0.25);
 
+  oui-clipboard { /* stylelint-disable-line */
+    margin-bottom: 0;
+
+    .oui-clipboard__control {
+      border-bottom: 0;
+      padding: 0;
+      padding-left: 1.5rem;
+      font-weight: 400;
```
marie-j: There are variables available for this in `ui-kit` like `@oui-font-regular`

packages/manager/apps/public-cloud/src/assets/theme/_sidebar.less
```diff
@@ -22,6 +49,13 @@ cloud-sidebar { /* stylelint-disable-line */
     will-change: transform;
     transition: transform ease-in-out 500ms;
 
+    oui-clipboard { /* stylelint-disable-line */
```
FredericEspiau: Stylelint prevents creating styles for tags ? Maybe we should disable this rule then

jleveugle: Yep :)

PR #420, opened 2019-05-02
feat(public-cloud): store walkme value in local storage

marie-j:


packages/manager/apps/public-cloud/src/components/walkMe/walkme.service.js
```diff
@@ -28,6 +28,7 @@ export default class WalkMe {
     return new Tour({
       name: 'public-cloud-walkme',
       steps: this.getSteps(),
+      debug: true,
```
jleveugle: debug is always needed ? :)

marie-j: No, forgot to remove it my bad

PR #429, opened 2019-05-02
fix(pci.project.private.networks.vrack): update some translations

antleblanc:
# Update some translations

### :bug: Bug Fix

16646e9 - fix(pci.project.private.networks.vrack): update some translations

ref: MANAGER-2588


marie-j:
This is not the content in the mockups. Also it will break consistency across all the pages with same CTAs

antleblanc:
@marie-j Thanks for the feedback. I have update the pull request regarding the latest feedback that @fchabrolle gave me.

fchabrolle:
we can have both, however in this specific case i think it's more correct, if it was just "créer un vRack" it was ok but the cta has more text than that

PR #434, opened 2019-05-02
fix(pci.storage.container): apply margin between clipboard and tooltip

antleblanc:
# Apply margin between clipboard and tooltip

### :bug: Bug Fix

a946fff - fix(pci.storage.container): apply margin between clipboard and tooltip

ref: MANAGER-2516


cbourgois:
Also done in #435 (another tooltip added)

antleblanc:
@cbourgois Thanks for the notifications. Shouldn't be in a conflict after a merge.

PR #441, opened 2019-05-03
feat(pci.project.storage.objects): add empty page

cbourgois:
### Description of the Change

15751dc8 — feat(pci.project.storage.objects): add empty page

ref: MANAGER-2280


packages/manager/modules/pci/src/projects/project/storages/objects/onboarding/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,17 @@
+{
```
antleblanc: There are some empty translations. Is it needed?

cbourgois: Yes, to avoid to display translation key instead of nothing. We haven't all content for guides.

PR #458, opened 2019-05-07
fix(storages.blocks.block.volume.edit): name of volume is now optional

jleveugle:


packages/manager/modules/pci/src/projects/project/storages/blocks/block/edit/volume-edit/volume-edit.html
```diff
@@ -29,7 +29,7 @@ <h1 data-ng-bind="$ctrl.title"></h1>
             <input class="oui-input" type="text" name="name"
                 data-ng-model="$ctrl.storage.name"
                 data-ng-maxlength="255"
-                required>
+                >
```
FredericEspiau: Could you please move it one line above ? 😅

PR #472, opened 2019-05-08
fix(pci.project.instance): update instance status

cbourgois:
### Description of the Change

29c7d561 — fix(pci.project.instance): update instance status

ref: MANAGER-2619


packages/manager/modules/pci/src/projects/project/instances/instance/translations/Messages_fr_FR.json
```diff
@@ -15,15 +15,15 @@
   "pci_projects_project_instances_instance_status_REBOOT": "Redémarrage",
   "pci_projects_project_instances_instance_status_REBUILD": "Réinstallation",
   "pci_projects_project_instances_instance_status_RESCUED": "",
-  "pci_projects_project_instances_instance_status_RESIZED": "",
-  "pci_projects_project_instances_instance_status_REVERT_RESIZE": "",
+  "pci_projects_project_instances_instance_status_RESIZED": "Mise à jour",
```
antleblanc: I'm not sure about which status should be displayed in the UI but in that case the message lack some information. Isn't it?

Same for `REVERT_RESIZE`, `RESIZE`, and `VERIFY_RESIZE`.

PR #479, opened 2019-05-08
feat(us): configure us side and remove useless links

jleveugle:


packages/manager/modules/pci/src/projects/project/sidebar/sidebar.constant.js
```diff
@@ -62,17 +64,19 @@ export const MENU = [
...
-      {
-        options: {
-          state: 'pci.projects.project.kubernetes',
-        },
-        translation: 'cloud_sidebar_orchestration_kubernetes',
-      },
-    ],
-    translation: 'cloud_sidebar_orchestration',
-  },
+  Environment.getRegion() !== 'US'
```
antleblanc: An other approach would be to add for instance an exclude entry which can contains a list of regions.

Is that make sense to you?

```js
{
  …,
  exclude: ['US'] // or Environment.getRegion(),
  …,
}
```

PR #485, opened 2019-05-09
fix(public-cloud): set moment locale

cbourgois:
### Description of the Change

b931a6e2 — fix(public-cloud): set moment locale


packages/manager/modules/pci/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
antleblanc: Why not prefer the JSON format?

PR #522, opened 2019-05-13
Feat/kube create cluster

marie-j:
## Update API calls for Kubernetes and create cluster 

* Add cluster creation
* In order to be consistent with the new cluster creation, migrate api calls and refactor code to new standards (modal with states, lazyloading ...)


antleblanc:
A new version of `ovh-api-services` has been released and published :tada:!

ref: https://github.com/ovh-ux/ovh-api-services/releases/tag/v6.17.0

PR #523, opened 2019-05-13
build: bump ovh-api-services v6.17.0

marie-j:
# Bump ovh-api-services to v6.17.0


antleblanc:
Isn't reflected to the modules as well?

PR #540, opened 2019-05-13
feat(pci.projects.new): display the right min credit amount

Jisay:


packages/manager/modules/pci/src/projects/new/payment/credits/credits.html
```diff
@@ -8,9 +10,42 @@
     </p>
     <p data-translate="pci_projects_new_credits_explain2"></p>
 
-    <oui-field
+    <div class="oui-field"
```
Jisay: Can't use the component `oui-field` because of the bad encoding of the € symbol in the description

marie-j: Is it a bug due to `ui-kit`? Maybe we can fix it if that's the case

PR #549, opened 2019-05-13
fix(pci.sidebar): add filter on sidebar menu depending on region

marie-j:
## Filter on regions for responsive sidebar


packages/manager/modules/pci/src/projects/project/sidebar/sidebar.service.js
```diff
@@ -1,9 +1,13 @@
+import includes from 'lodash/includes';
+import isUndefined from 'lodash/isUndefined';
```
FredericEspiau: Just a thought, but now that we have `Lodash 4`, we should think about using `isNil` if we don't want to make a difference between `null` and `undefined` :)

PR #551, opened 2019-05-13
fix(pci.projects.new): fix creation error message

Jisay:


packages/manager/modules/pci/src/error/translations/Messages_fr_FR.json
```diff
@@ -7,9 +7,9 @@
...
-  "pci_error_paypal_account_not_verified": "Malheureusement vous ne pouvez pas créer de projet Public Cloud avec votre compte PayPal car celui-ci n'est pas vérifié. Nous vous invitons à faire les démarches nécessaires auprès de PayPal pour vérifier votre compte, puis de réessayer. Vous pouvez aussi modifier <a href=\"{{ href }}\">votre moyen de paiement par défaut</a>.",
-  "pci_error_unpaid_debts": "Malheureusement vous ne pouvez pas créer de projet Public Cloud car vous avez une dette en cours chez OVH. Nous vous invitons à régler le montant dû, puis de réessayer.",
-  "pci_error_invalid_payment_mean": "Malheureusement vous ne pouvez pas créer de projet Public Cloud car vous n'avez pas de moyen de paiement par défaut valide. Nous vous invitons à modifier <a href=\"{{ href }}\">votre moyen de paiement par défaut</a>."
+  "pci_error_account_not_eligible": "La création d’un nouveau projet Public Cloud n’est pas possible avec votre compte. Nous vous invitons à contacter notre support pour plus d’informations.",
```
jleveugle:
```suggestion
  "pci_error_account_not_eligible": "La création d’un nouveau projet Public Cloud est impossible avec votre compte. Nous vous invitons à contacter notre support pour plus d’informations.",
```

packages/manager/modules/pci/src/projects/new/new.html
```diff
@@ -1,7 +1,7 @@
 <div class="pci-projects-new h-100">
-    <div class="container-fluid h-100">
-        <div class="row h-100 justify-content-md-center">
-            <div class="col-md-7 p-5 bg-white h-100">
+    <div class="container-fluid overflow-auto h-100">
+        <div class="row minh-100 justify-content-md-center">
```
jleveugle: If you do that, i think the navbar isn't sticky anymore, right ?

Jisay: Nope, it's okay with that ;-)

packages/manager/modules/pci/src/error/translations/Messages_fr_FR.json
Jisay: This has been validated by content team

packages/manager/modules/pci/src/index.scss
```diff
@@ -4,4 +4,12 @@
   @import '~bootstrap4/scss/_mixins';
   @import '~bootstrap4/scss/_utilities';
   @import '~bootstrap4/scss/_grid';
+
+  .minh-100 {
```
FredericEspiau: Don't we have a utility CSS package like Bootstrap does ?

Jisay: Nope. There is `mh-100` but it's for `max-height` :-$
And for the class `overflow-auto`, it exists in bootstrap but in v4.3 ;-)

FredericEspiau: Maybe we should have one then, so we don't have to rely on Boostrap ;)

Jisay: But bootstrap is great! :heart: :smile:

packages/manager/modules/pci/src/projects/new/new.html
jleveugle: I think we don't need that because develop has been fixed in another way :)

Jisay: Nope, the white background is not extended if there is an overflow.
I have merged this branch into an other one, to keep everything at one plae. I close this PR.

PR #570, opened 2019-05-14
fix(pci): update responsive layout

cbourgois:
### Description of the Change

Fixes :
* datagrid filter button on hover
* datagrid header (top actions / search / filter button)
* creation forms layout on mobile

79bfbf74 — fix(pci): update responsive layout

### DIFF
--- a/packages/manager/apps/public-cloud/src/assets/theme/_stepper.less
+++ b/packages/manager/apps/public-cloud/src/assets/theme/_stepper.less
@@ -1,0 +1,8 @@
+@import "~ovh-ui-kit/packages/oui-responsive/_variables";
+
+@media screen and (max-width: @oui-responsive-breakpoint-tablet) {
+  oui-step-form { // stylelint-disable-line
+    padding-left: 0 !important;
+    padding-right: 0 !important;
+  }
+}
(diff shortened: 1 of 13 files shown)


packages/manager/apps/public-cloud/src/assets/theme/_stepper.less
```diff
@@ -0,0 +1,8 @@
+@import "~ovh-ui-kit/packages/oui-responsive/_variables";
+
+@media screen and (max-width: @oui-responsive-breakpoint-tablet) {
+  oui-step-form { // stylelint-disable-line
+    padding-left: 0 !important;
```
FredericEspiau: No way to increase the specificity instead of using `!important` ?

For example,
```css
oui-step oui-step-form { // stylelint-disable-line
    padding-left: 0;
    padding-right: 0;
}
```

This would make the specificity go from (0,1,0) to (0,2,0)

Or add an id to the stepper:
```css
#myStepper oui-step-form { // stylelint-disable-line
    padding-left: 0;
    padding-right: 0;
}
```

would make the specificity (1,1,0), it should be enough

packages/manager/modules/pci/src/components/project/images-list/images-list.html
```diff
@@ -18,7 +18,7 @@
 
                 <div class="container-fluid px-0">
                     <div class="row">
-                        <div data-ng-repeat="(distribution, images) in distributions" class="d-inline-block col-md-6 col-lg-4 my-3">
+                        <div data-ng-repeat="(distribution, images) in distributions" class="d-inline-block col-md-6 col-xl-4 my-3">
```
FredericEspiau: `track by` ?

PR #577, opened 2019-05-14
fix: update align on empty page

jleveugle:


packages/manager/modules/pci/src/components/project/empty/empty.html
```diff
@@ -1,4 +1,4 @@
-<div class="container px-0 pci-project-empty">
+<div class="container mx-0 px-0 pci-project-empty">
```
antleblanc: Isn't already defined?

ref: https://github.com/twbs/bootstrap/blob/master/scss/mixins/_grid.scss#L9-L10

jleveugle: mx-0 => Horizontal margin set to 0, instead of auto ;)

PR #582, opened 2019-05-14
refactor: remove unused ng-ovh-doc-url dependency

antleblanc:
# remove unused dependency

Since #521 remove the entire legacy part, this following dependency could
be removed.

It contains:

Component:
- ovhDocUrl

Provider
- ovhDocUrl

### :nail_care: Refactor

ad6f79e - refactor: remove unused ng-ovh-doc-url dependency

### :house: Internal

- No QC required.

ref: #521

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1607,6 +1607,24 @@
     lodash "^4.17.11"
     moment "^2.24.0"
 
+"@ovh-ux/manager-pci@^0.5.0-alpha.0":
+  version "0.5.0-alpha.3"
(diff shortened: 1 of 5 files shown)


yarn.lock
```diff
@@ -1607,6 +1607,24 @@
     lodash "^4.17.11"
     moment "^2.24.0"
 
+"@ovh-ux/manager-pci@^0.5.0-alpha.0":
```
antleblanc: `yarn install` from `96c54bf94b4f677ce080169aa6c64ec4633b1706` causes this following change.

FredericEspiau: Thanks for the explanation

PR #588, opened 2019-05-14
fix(pci.project.edit): associate label with input element

antleblanc:
# Associate label with input element

### :bug: Bug Fix

d8c6c21 - fix(pci.project.edit): associate label with input element

### :house: Internal

- No QC required.

ref: https://github.com/ovh-ux/ovh-ui-angular/blob/v3.0.2/packages/oui-field/src/field.controller.js#L66

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -1599,14 +1599,6 @@
     rollup-pluginutils "^2.3.3"
     slash "^2.0.0"
 
-"@ovh-ux/manager-navbar@^0.1.0-alpha.1":
-  version "0.1.0-alpha.1"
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -1599,14 +1599,6 @@
     rollup-pluginutils "^2.3.3"
     slash "^2.0.0"
 
-"@ovh-ux/manager-navbar@^0.1.0-alpha.1":
```
antleblanc: `yarn install` from `4976486a6b89967090fcecbc55e9d0bcd6bb30da` causes this following change.

PR #599, opened 2019-05-15
 feat(pci.projects.project.creating): display images into loader

Jisay:


packages/manager/apps/public-cloud/src/index.js
```diff
@@ -4,6 +4,7 @@ import { Environment } from '@ovh-ux/manager-config';
 /* eslint-disable import/no-webpack-loader-syntax, import/extensions */
 import 'script-loader!jquery';
 import angular from 'angular';
+import 'angular-animate';
```
antleblanc:
```suggestion
import ngAnimate from 'angular-animate';
```

packages/manager/apps/public-cloud/src/index.js
```diff
@@ -59,6 +60,7 @@ angular
     navbar,
     preload,
     walkMe,
+    'ngAnimate',
```
antleblanc: According to my previous comment (import dependencies in that way):

```suggestion
    ngAnimate,
```

packages/manager/modules/pci/src/index.scss
```diff
@@ -4,4 +4,12 @@
   @import '~bootstrap4/scss/_mixins';
   @import '~bootstrap4/scss/_utilities';
   @import '~bootstrap4/scss/_grid';
+
+  .minh-100 {
```
jleveugle: Do you still need this code with the fixs that have been applied on the branch `develop`?

Jisay: Yes I still need it because of white backgrounds that stop at 100% of the height. When scrolling the bg-white is cutted. Adding min-height 100% fix that

PR #609, opened 2019-05-16
fix(apps.public.cloud): set lang attribute

antleblanc:
# Set lang attribute

`lang` attribute is set based on the selected language from the 
localStorage.

### :bug: Bug Fix

5c6612f - fix(apps.public.cloud): set lang attribute

### :house: Internal

- No QC required.

### DIFF
--- a/packages/manager/apps/public-cloud/src/index.js
+++ b/packages/manager/apps/public-cloud/src/index.js
@@ -43,6 +43,11 @@ import routing from './index.routes';
 
 Environment.setRegion(__WEBPACK_REGION__);
 
+const setLangAttribute = /* @ngInject */ ($document) => {
+  const { 'univers-selected-language': lang } = localStorage;
+  $document.querySelectorAll('html')[0].setAttribute('lang', lang);
+};
+
 angular
   .module('ovhStack', [
     __NG_APP_INJECTIONS__,
@@ -80,4 +85,5 @@ angular
       }
     });
   })
+  .run(setLangAttribute)
   .run(/* @ngTranslationsInject:json ./translations */);


packages/manager/apps/public-cloud/src/index.js
```diff
@@ -43,6 +43,11 @@ import routing from './index.routes';
 
 Environment.setRegion(__WEBPACK_REGION__);
 
+const setLangAttribute = /* @ngInject */ ($document) => {
+  const { 'univers-selected-language': lang } = localStorage;
+  $document.querySelectorAll('html')[0].setAttribute('lang', lang);
```
jleveugle: lang attr should be in ISO 639-1, so we should split our lang and keep only the part before `_` :)

jleveugle: (it should be manage by core, not only for public-cloud too) :)

antleblanc: How to ensure the difference between `en` (US) and `en-GB` for instance?

A better option would be to turn the case into kebabCase from what `univers-selected-language` returns.

Is that make sense?

antleblanc: According to your last comment:

`ovhStack` module define the base layout by omitting the `lang` attribute.

Even if the `'univers-selected-language'` is setted from the `ovhManagerCore` module, changing the `lang` attribute must be delegated to the apps itself.

In that way we can imagine a different application that don't want to rely on that behavior.

jleveugle:
Hello @antleblanc,
Anna Bot for OVH UX <3

packages/manager/apps/public-cloud/src/index.js
antleblanc: ping @jleveugle

jleveugle:
Hello @antleblanc,
Anna Bot for OVH UX <3

jleveugle:
Hello @antleblanc,
Anna Bot for OVH UX <3

packages/manager/apps/public-cloud/src/index.js
jleveugle: I think it should be managed by core, like 'univers-selected-language'. In fact, you want to ha ve the same behavior when a module is used in standalone for instance.

If we want to manage that inside the host app, we should manage 'univers-selected-language' in the host app too.

antleblanc: @jleveugle  Thanks for you reply.

Delegate this behavior to the `@ovh-ux/manager-core` module isn't reliable and
here is my point of view (feel free to correct me if something is wrong or
incorrect).

Once again, if I'm in a case where I maintain for instance two separate
applications.

The first one which has a default layout that define `de` in the `lang`
attribute. This application is entirely translated in both `de` and `nl`.
So it means that we could safely change the lang attribute.

Whereas the second one which has a default layout that explicitly define
the `lang` attribute to `fr` and all the part of this application isn't
translated in any other languages. I definitely don't want to rely of this
kind of behavior.

That's why this behavior is delegated to the host application and not to the
module itself.

Furthermore you omitting to answer to the first question regarding the value
of the attribute. Splitting `univers-selected-language` value isn't reliable
because we cannot make the distinction between en (US vs GB) for instance. 

Thanks!

jleveugle: I think you forgot one thing, `@ovh-ux/manager-core` concerns our managers. We don't want to be compatible with all apps in the world, we only want to avoid get the same logic in 4 places :/

Today, managers are always in the same case, default language is french, and when this one is defined, we can set the lang attribute. When this language is updated, let's update the lang attribute too. This logic can be managed by core, because it's always the SAME LOGIC.

If I follow you, let's create some specific configuration for each project and forget tools like manager-config etc.

> Furthermore you omitting to answer to the first question regarding the value
of the attribute. Splitting univers-selected-language value isn't reliable
because we cannot make the distinction between en (US vs GB) for instance.

No, I don't forget that. I just give you the right answer, the lang should be in ISO 639-1 (https://www.w3schools.com/tags/ref_language_codes.asp). We don't want to distinguish en-GB and en-US, we only want to distinguish french vs english. In fact, lang attribute is used to avoid tools, like Voiceover, will not pronounce your french words in english.

PS: If you want to discuss it, I'm far enough away from you but I think we can cross each other :)

jleveugle:
Hello @antleblanc,
Anna Bot for OVH UX <3

packages/manager/apps/public-cloud/src/index.js
jleveugle: ping @antleblanc

antleblanc: Due to #1042,  there is no more distinction to do between `en_GB` and `en_US` meaning that spliting the actual `univers-selected-language` value would be enough.

jleveugle:
Hello @antleblanc,
Anna Bot for OVH UX <3

PR #613, opened 2019-05-16
Build improvements

jleveugle:


antleblanc:
Related to an issue with: `/proc/sys/fs/inotify/max_user_watches` on some GNU/Linux distributions.

Need to figure out if there is no downside effect regarding how file are watch. Is the watch is now slower than before for instance.

antleblanc:
FYI theses optimization seems to not work as fatal error are still reported.

```txt
 FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

jleveugle:
Hello @jleveugle,
Anna Bot for OVH UX <3

PR #616, opened 2019-05-16
fix(pci.projects.new): fix styles

Jisay:


antleblanc:
Your branch isn't correctly rebased on `develop` I guess.

Since #599 the commit `2ebf83d` is already applied.

PR #643, opened 2019-05-20
fix: redirect user and avoid block him on root page

jleveugle:


packages/manager/apps/public-cloud/src/index.html
```diff
@@ -16,7 +16,11 @@
             </div>
             <div class="d-flex w-100 h-100 position-relative">
                 <div class="w-100 h-100 position-absolute" style="top: 0; left: 0;">
-                    <div class="h-100 w-100" ui-view></div>
+                    <div class="h-100 w-100" data-ui-view>
+                        <div class="h-100 d-flex" data-ng-init="$ctrl.initRedirect()">
```
antleblanc: Couldn't be a part from the `$onInit` hook instead?

packages/manager/apps/public-cloud/src/index.js
```diff
@@ -13,7 +13,7 @@ import 'script-loader!jquery-ui/ui/minified/mouse.min';
 import 'script-loader!jquery-ui/ui/minified/draggable.min';
 import 'script-loader!moment/min/moment-with-locales.min.js';
 import 'script-loader!angular-ui-validate/dist/validate.js';
-import 'script-loader!bootstrap-tour/build/js/bootstrap-tour-standalone.min.js';
+import 'ovh-ui-angular';
```
antleblanc: No need to be encapsulated inside `eslint` block.

Issue #682, opened 2019-05-21
Application Public Cloud - $locationProvider - remove default value for the prefix


antleblanc:
**Is your feature request related to a problem? Please describe.**
If we pay attention to the URL prefix we still have `/!#/url`.

**Describe the solution you'd like**
- Remove existing prefix `'!'`.

```js
  .config(/* @ngInject */ $locationProvider => $locationProvider.hashPrefix(''))
```

**Describe alternatives you've considered**
- None.

**Additional context**
- All control panel (also known as Manager) doesn't have any prefix.

ref: https://docs.angularjs.org/api/ng/provider/$locationProvider

antleblanc:
It's now fixed. Will be available on the next release (cf:  #732)

PR #683, opened 2019-05-21
fix(apps.public.cloud): remove default location prefix

antleblanc:
# Remove default location prefix

### :bug: Bug Fix

6929364 - fix(apps.public.cloud): remove default location prefix

### :house: Internal

- No QC required.

ref: #682


antleblanc:
> If we merge this changes, **all external links** (…)

Where external links are supposed to be stored?

In addition, if you pay attention to some business logic inside the `/2api/universes` endpoint you will see that the prefix not present.

Hope it helps you,
Thanks!

antleblanc:
Ok got it. I will take a close look with Website team.

Thanks!

antleblanc:
Both Website and Email team confirm me that they do not rely on specific URL (only `/manager/public-cloud/`) is used.

PR #688, opened 2019-05-21
feat(pci): change link to legacy manager display

marie-j:


packages/manager/modules/pci/src/projects/project/project.html
```diff
@@ -4,6 +4,23 @@
...
+                    data-ng-click="$ctrl.goToLegacy()"
+                    data-translate="pci_projects_project_welcome_legacy">
+                </button>
+                <a class="oui-link pci-project-link_light"
+                    data-ng-if="$ctrl.user.ovhSubsidiary !== 'FR'"
+                    data-track-on="click"
+                    data-track-name="public-cloud_menu_back-to-cloud"
+                    data-track-type="action"
+                    data-ng-href="{{:: $ctrl.LEGACY_URL}}">
+                    <span data-translate="pci_projects_project_welcome_legacy"></span>
```
antleblanc: `translate` directive cannot be applied directly to the parent element?

packages/manager/modules/pci/src/projects/project/project.less
```diff
@@ -0,0 +1,9 @@
+@background-color: rgb(239, 249, 253);
+
+.pci-project-banner {
+  background-color: @background-color;
+
+  .pci-project-link_light {
+    color: rgb(77, 111, 146);
```
antleblanc: Parent rule is using color variable. Why it is not the same from here?

PR #727, opened 2019-05-27
fix(public-cloud): imort missing ovh-ui-angular peer deps

marie-j:
MANAGER-2694


marie-j:
Closing this as it will be handled directly in `ui-kit`

PR #738, opened 2019-05-27
fix(pci.project.private-network): prevent inputs from overlaping

marie-j:
MANAGER-2672

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/private-networks/add/add.html
+++ b/packages/manager/modules/pci/src/projects/project/private-networks/add/add.html
@@ -98,14 +98,14 @@
                                 data-label="{{:: 'pci_projects_project_network_private_create_localisation_range' | translate }}">
                                 <div class="d-flex flex-wrap flex-xl-nowrap justify-content-around align-items-center">
                                     <input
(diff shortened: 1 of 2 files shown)


antleblanc:
@marie-j There is a typo in  the scope of your commit message (`pic` => `pci`).

packages/manager/modules/pci/src/projects/project/private-networks/add/add.html
```diff
@@ -102,14 +102,16 @@
                                     type="text"
                                     name="{{:: region + '-range-start' }}"
                                     data-ng-model="subnet.start"
-                                    disabled>
+                                    disabled
+                                    style="min-width: 0">
```
antleblanc: Avoiding inline style would be better.

Putting this rule (with a comment) to one of these file would be great:

- manager/packages/manager/modules/pci/src/projects/project/project.less
- manager/packages/manager/modules/pci/src/index.scss

Is that make sense to you?

marie-j: This is something specific to the private network creation so it doesn't make sense to have it there

antleblanc:

Thanks!

PR #766, opened 2019-05-29
fix: project creation

Jisay:
ref: MANAGER-2730


antleblanc:

Thanks!

PR #794, opened 2019-06-04
fix(pci.projects.instances.new): missing input in oui-field

frenautvh:
See : MANAGER-2660


packages/manager/modules/pci/src/projects/project/instances/add/add.html
```diff
@@ -201,20 +201,20 @@ <h1 data-translate="pci_projects_project_instances_add_title"></h1>
                 data-label="{{ 'pci_projects_project_instances_add_script_label' | translate }}"
                 data-help-text="{{ 'pci_projects_project_instances_add_script_help' | translate }}"
                 data-size="xl">
+                <oui-textarea
+                    data-ng-show="$ctrl.showUserData"
```
FredericEspiau: Not an `ngIf` ?

frenautvh: yes oui-field will throw errors if it hasn't a input/text-area as a child element, so the element needs to exists in the DOM

PR #797, opened 2019-06-04
build: upgrade ovh-api-services to v6.23.0

marie-j:


FredericEspiau:
You should update the modules as well (this is what @antleblanc told me)

antleblanc:
@FredericEspiau It's not a mandatory but yes it keeps all `package.json` aligned across the entire repository.

PR #814, opened 2019-06-05
fix(pci.components.flavors-list): allow to translate bandwidth unit

marie-j:
MSB-55


FredericEspiau:
Unit conversion should probably be handled by us and not by the translators, don't you think ?

antleblanc:
@FredericEspiau Yes it could!

I assume that it will be also much more easier to do this across the entire monorepo as soon as all products from each control panel were migrated.

Because actually we still are dealing with both tools as you may know.

As for now I will merge this pull request in order to be able to create a new release.

Thanks!

antleblanc:
Even if the ticket is still in `in progress` I will merge this pull request in order to be able to provide a new release branch.

Thanks!

PR #821, opened 2019-06-07
build(deps): upgrade ovh-ui-angular to v3.2.2

frenautvh:
See : MANAGER-2687

**issue** : errors in js console due to oui-field component

**fix** : fix oui-field component, bump ovh-ui-angular to version 3.2.2


packages/manager/apps/freefax/package.json
```diff
@@ -35,7 +35,7 @@
     "oclazyload": "^1.1.0",
     "ovh-angular-ui-confirm-modal": "^1.0.2",
     "ovh-api-services": "^6.23.0",
-    "ovh-ui-angular": "^3.2.1",
+    "ovh-ui-angular": "^3.2.0",
```
marie-j: Should be `3.2.2` right ?

antleblanc:
@frenautvh Could you please resolve existing conflicts?

Thanks!

packages/manager/apps/freefax/package.json
frenautvh: i prefer ^3.2.0 since it includes 3.2.2 and further future patches 3.2.x, what do you think?

jleveugle: Depends on your needs!

Maybe you need a fix in ovh-ui-angular which means that you need at least the version 3.2.2. `^3.2.2` is a good compromise ;D

frenautvh:
@antleblanc done!

PR #825, opened 2019-06-10
build: bump ovh-ui-kit to v2.33.2

marie-j:
Bump `ovh-ui-kit` to v2.33.2

Fixes MSB-31


antleblanc:
Shouldn't be a `build` type commit message in that case?

marie-j:
> Shouldn't be a `build` type commit message in that case?

@antleblanc I've edited commit message

PR #832, opened 2019-06-11
fix(pci.components.project.images.list): update constants list

antleblanc:
# Update constants list

### :bug: Bug Fix

1468322 - fix(pci.components.project.images.list): update constants list

### :house: Internal

ref: MANAGER-2798


antleblanc:
Validated that we can keep the actual constant list. So I will close this pull request and cancel the ticket.

Thanks!

PR #846, opened 2019-06-13
fix: prevent onboarding to loop infinitely

marie-j:


packages/manager/modules/pci/src/projects/onboarding/onboarding.routing.js
```diff
@@ -16,6 +16,12 @@ export default /* @ngInject */ ($stateProvider) => {
         breadcrumb: () => null, // Hide breadcrumb
       },
       redirectTo: transition => transition.injector().get('publicCloud')
-        .handleDefaultProject(),
+        .getDefaultProject()
+        .then(projectId => (projectId ? ({
```
antleblanc: As discussed it could end up with an wrong behavior (which is already the case I think).

Assuming that we have two project A and B (the default).

If the user delete the B project, then we have to remove the preference.

marie-j: I agree that we should remove preference if default is removed but this is out of scope of this PR imo and will be taken care of in another one, if that's ok with you ?

PR #892, opened 2019-06-19
fix(pci.projet.user): update error translation keys

cbourgois:
### Description of the Change

a9aa9511 — fix(pci.projet.user): update error translation keys

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/users/download-openrc/download-openrc.controller.js
+++ b/packages/manager/modules/pci/src/projects/project/users/download-openrc/download-openrc.controller.js
@@ -58,7 +58,7 @@ export default class PciUsersDownloadOpenRcController {
           }));
       })
       .catch(err => this.goBack(this.$translate.instant(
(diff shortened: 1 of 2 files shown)


packages/manager/modules/pci/src/projects/project/users/download-openrc/download-openrc.controller.js
```diff
@@ -58,7 +58,7 @@ export default class PciUsersDownloadOpenRcController {
           }));
       })
       .catch(err => this.goBack(this.$translate.instant(
-        'pci_projects_project_users_download-openrc_error_save',
+        'pci_projects_project_users_download-openrc_error_openrc',
```
antleblanc: It would be nice if we can stick with `snakeCase` only for translations keys.

antleblanc:
It seems that there is a typo in your commit message `projet` > `project`.

Will be solved with a **Squash and merge** action.

PR #904, opened 2019-06-19
fix(pci): update flavors category title

cbourgois:
### Description of the Change

ced49e83 — fix(pci): update flavors category title

ref: MSB-54


packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.constants.js
```diff
@@ -3,22 +3,27 @@ export const DEFAULT_OS = 'linux';
 export const CATEGORIES = [
   {
     category: 'balanced',
+    title: 'General Purpose',
```
FredericEspiau: So the text will be `General Purpose` for all languages ?

cbourgois: Yes.

PR #906, opened 2019-06-20
fix: fix pci mobile sidebar

marie-j:
MANAGER-2971

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/sidebar/sidebar.service.js
+++ b/packages/manager/modules/pci/src/projects/project/sidebar/sidebar.service.js
@@ -15,9 +15,9 @@ export default class Sidebar {
     return [
       ...MENU.map(item => ({
         name: item.translation,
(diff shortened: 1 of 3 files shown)


packages/manager/modules/pci/src/projects/project/sidebar/sidebar.service.js
```diff
@@ -15,9 +15,9 @@ export default class Sidebar {
     return [
       ...MENU.map(item => ({
         name: item.translation,
-        title: this.$translate.instant(item.translation),
+        title: item.title,
```
FredericEspiau: What does this have to do with the schmilblick ?

marie-j: Long story short, sidebar not handled the same way on mobile, I've started a refactor following that issue  but in the meanwhile

PR #928, opened 2019-06-24
build: add both homepage and bugs entries in package.json files

antleblanc:
# Add both homepage and bugs entries in package.json files

### :gear: Build

1a044ef - build: add both homepage and bugs entries in package.json files

### DIFF
--- a/packages/manager/modules/navbar/package.json
+++ b/packages/manager/modules/navbar/package.json
@@ -1,7 +1,11 @@
 {
   "name": "@ovh-ux/manager-navbar",
-  "description": "OVH control panel navbar",
   "version": "1.0.1",
+  "description": "OVH control panel navbar",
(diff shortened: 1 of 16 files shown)


packages/manager/modules/navbar/package.json
```diff
@@ -23,6 +27,14 @@
     "start:dev": "lerna exec --stream --scope='@ovh-ux/manager-navbar' --include-filtered-dependencies -- yarn run dev",
     "start:watch": "lerna exec --stream --parallel --scope='@ovh-ux/manager-navbar' --include-filtered-dependencies -- yarn run dev:watch"
   },
+  "dependencies": {
```
antleblanc: Diff applied by running this following command:

```sh
npx sort-package-json modules/*/package.json
```

PR #608, opened 2019-05-16
Rename some ng-ui-router-* packages

antleblanc:
# Rename some ng-ui-router-* packages

Replace both package and module name for `ng-ui-router-breadcrumb`, `ng-ui-router-layout` and `ng-ui-router-line-progress`

### :nail_care: Refactor

a2455b4 - refactor(components.ng-ui-router-layout): renamed package name
5131f54 - refactor(components.ng-ui-router-breadcrumb): renamed package name
fadaaf2 - refactor(components.ng-ui-router-line-progress): renamed package name

### :house: Internal

- No QC required.

#### ng-ui-router-layout
- package is now named as `@ovh-ux/ng-ui-router-layout`

#### ng-ui-router-breadcrumb

- package and module are now respectively named as `@ovh-ux/ng-ui-router-breadcrumb` and `ngUiRouterBreadcrumb`

#### ng-ui-router-line-progress

- package and module are now respectively named as `@ovh-ux/ng-ui-router-line-progress` and `ngUiRouterLineProgress`

PR #931, opened 2019-06-25
feat(ng-uirouter-layout): add support for componentProvider

marie-j:
Ui Router allows to pass `componentProvider` for conditional component rendering. replicate that option for modal layouts


packages/components/ng-uirouter-layout/src/index.js
```diff
@@ -69,10 +70,19 @@ angular
         }
 
         if (get(state, 'layout.name') === 'modal') {
+          let componentName = get(state, state.views.modal ? 'views.modal.component' : 'component');
```
frenautvh: get(state, "views.modal.component", "component") ?

if you're calling state.views.modal explicitly you loose the benefits of using lodash "get"

marie-j: Indeed

marie-j:
> Hello,
> 
> Maybe we can provide a description of how to use it here:
> 
> https://github.com/ovh-ux/manager/tree/master/packages/components/ng-uirouter-layout
> 
> See ya

@jleveugle Sure, I will take care of this in another PR if that's ok with you ?

antleblanc:
It probably make more sense to review and merge a pending pull request that is actually renaming this module.

ref: ovh-ux/manager#608

marie-j:
> It probably make more sense to review and merge a pending pull request that is actually renaming this module.
> 
> ref: #608

@antleblanc The name of the module is not impacting this PR so I'm having trouble seeing why it makes more sense

antleblanc:
Both will ended up with a conflict if we use the merge button.

#608 was already rebased several times that's why I have left a comment in order to refer to this pull request.

PR #956, opened 2019-06-26
docs(ng-uirouter-layout): add usage examples

marie-j:


packages/components/ng-uirouter-layout/README.md
```diff
@@ -11,6 +11,8 @@ yarn add @ovh-ux/ng-uirouter-layout
 ```
 ## Usage
 
+### In `index.js`
```
antleblanc: Prefer specifying the current file by adding first commented line.

```js
// index.js
import ngUiRouterLayout from '@ovh-ux/ng-uirouter-layout';

angular
  .module('myApp', [
    ngUiRouterLayout,
  ]);
```
Same with the `routing.js` file example.

In addition it would great to provide real examples through the `examples` folder (recently introduced).

What do you think about it?

Thanks!

antleblanc:
Shouldn't target #931 instead?

marie-j:
> Shouldn't target #931 instead?

@antleblanc Could do, but this was in order so PR are not relying on each other

packages/components/ng-uirouter-layout/src/index.js
```diff
@@ -69,10 +70,19 @@ angular
         }
 
         if (get(state, 'layout.name') === 'modal') {
+          let componentName = get(state, 'state.views.modal.component', state.component);
           const $state = transition.injector().get('$state');
           const $uibModal = transition.injector().get('$uibModal');
 
-          const componentName = get(state, state.views.modal ? 'views.modal.component' : 'component');
+          const componentProvider = get(state, 'state.views.modal.componentProvider', state.componentProvider);
```
cbourgois: Theses changes are addressed in another PR https://github.com/ovh-ux/manager/pull/931 ?

marie-j: Oopsie ! Probably due to rebase, fixing it right away

packages/components/ng-uirouter-layout/README.md
```diff
@@ -21,10 +22,60 @@ angular
...
+```js
+// routing.js
+import controller from './controller';
+import template from './template.html';
+...
+
+  $stateProvider
+    .state('state.name', {
+      url,
+      views: {
```
cbourgois: In documentation you have a `views` property but not in example : 
```javascript
$stateProvider.state('foo', {
  url: '/foo',
  component: 'foo',
  layout: 'modal',
});
```

packages/components/ng-uirouter-layout/examples/component/index.js
```diff
@@ -0,0 +1,12 @@
+angular
+  .module('myApp', ['ui.router', 'ui.bootstrap', 'ngUiRouterLayout'])
+  .config(($stateProvider) => {
```
antleblanc: Previous examples was referring to a UMD version of components.

Here you are using some arrow functions. It won't work in some browsers.

In my opinions, all examples could fit into a single `index.html`

packages/components/ng-uirouter-layout/examples/.eslintrc.js
```diff
@@ -0,0 +1,5 @@
+module.exports = {
```
antleblanc: Related to my previous comment, this configuration could be dropped.

packages/components/ng-uirouter-layout/examples/component/index.js
marie-j: Imo it is easier to read if JS is extracted from HTML, especially as the examples can become more complex. I can switch to `function`

PR #961, opened 2019-06-27
fix(pci.projects.storages): prevent quota check on containers

marie-j:
MANAGER-2951

### DIFF
--- a/packages/manager/modules/pci/src/components/project/regions-list/index.js
+++ b/packages/manager/modules/pci/src/components/project/regions-list/index.js
@@ -4,6 +4,7 @@ import 'angular-translate';
 import 'ovh-ui-angular';
 
 import component from './regions-list.component';
+import service from './regions.service';
 
 const moduleName = 'ovhManagerPciRegionsList';
 
@@ -14,6 +15,7 @@ angular
     'pascalprecht.translate',
   ])
   .component('pciProjectRegionsList', component)
+  .service('PciProjectRegions', service)
   .run(/* @ngTranslationsInject:json ./translations */);
 
 export default moduleName;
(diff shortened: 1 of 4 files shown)


packages/manager/modules/pci/src/components/project/regions-list/index.js
```diff
@@ -14,6 +15,7 @@ angular
     'pascalprecht.translate',
   ])
   .component('pciProjectRegionsList', component)
+  .service('PciProjectRegions', service)
```
antleblanc: Is that make sense to you?
```suggestion
  .service('pciProjectRegions', service)
```

marie-j: How do we differentiate a `service` from a `component` in that case ? Because none of them is suffixed by their type according to our logic so far, the difference was made with that caps letter

antleblanc: LGTM ;)

FredericEspiau: @marie-j John Papa uses the `Service` suffix as `UpperCamelCase` is used by Javascript by data types and function constructors (`RegExp`, `TypeError`, `XMLHttpRequest` for example)

i.e.: Douglas Crockford's coding convention (https://www.crockford.com/code.html)

marie-j: As spoken with @antleblanc earlier, it makes sense to keep it with `CamelCase`, as it's a singleton there's only one possible instance

antleblanc: `TileCase` !== `camelCase` to be more precise.

Here the discussion has been marked as resolved because it could be ambiguous due to the fact the both component and service has the same name.

antleblanc:

Thanks!

PR #983, opened 2019-07-02
refactor(eslint): remove some global from root configuration

antleblanc:
# Remove some global from root configuration

## :nail_care: Refactor

6229133 - refactor(eslint): remove some global from root configuration

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -8,9 +8,6 @@ module.exports = {
     browser: true,
     jquery: true,
   },
-  globals: {
-    d3: true
-  },
   rules: {
     'no-bitwise': ['error', { allow: ['~'] }],
     'import/no-extraneous-dependencies': ['error', { devDependencies: true }],
--- a/packages/manager/modules/cloud-universe-components/.eslintrc.json
+++ b/packages/manager/modules/cloud-universe-components/.eslintrc.json
@@ -1,0 +1,6 @@
+{
+  "extends": "../../../../.eslintrc.js",
+  "globals": {
+    "d3": true
+  }
+}


FredericEspiau:
It's not a `bugfix` then 😅

antleblanc:
@FredericEspiau branch name is related to the [git flow](https://danielkummer.github.io/git-flow-cheatsheet/) branching naming convention.

jleveugle:
Why do we want to disable `import/no-unresolved` ?

And ... I think it's better to keep `.js` format instead of `.json`, we can add some comments in config file :)

antleblanc:
@jleveugle This pull request doesn't rely on the disabling the given rule: `import/no-unresolved`.

Please refer to this following commit to get a more-in-depth view: https://github.com/ovh-ux/manager/commit/f1a480ec6d10f524c47544bfd440f3de67656ec7#diff-e4403a877d80de653400d88d85e4801a

Concerning the choice to the `json` format for configuration files, it has been
discussed at the beginning when the monorepo was started nor configured.
It was the same for components.

The `json` was prefered over `*rc`, `yml`, `yaml` and `js`.

Even though it is still possible but adding comments isn't a blocking thing
I guess.

Here is a reference to a popular project for instance: https://github.com/microsoft/vscode/blob/master/.vscode/extensions.json.

Create a unified experience for configuration files is a nice to have and it
feel more natural when you have to deal with.

Do you share the same point of view as described?

Thanks!

jleveugle:
sry, I don't remember why I disable this rule, and the commit message doesn't help :/

> @jleveugle This pull request doesn't rely on the disabling the given rule: `import/no-unresolved`.

True, but, imo, we should enable it, instead of deleting that. I think it's more secure to keep this rule active.

> Concerning the choice to the `json` format for configuration files, it has been
> discussed at the beginning when the monorepo was started nor configured.
> It was the same for components.
> 
> The `json` was prefered over `*rc`, `yml`, `yaml` and `js`.

Rules can't evolve ? Again, I think JS format is better. In the future, if we need to require a package etc, all files will already be in JS, no need to switch again from JSON, that will result by a mix of js / json file again.

> Even though it is still possible but adding comments isn't a blocking thing
> I guess.
> 
> Here is a reference to a popular project for instance: [microsoft/vscode:.vscode/extensions.json@`master`](https://github.com/microsoft/vscode/blob/master/.vscode/extensions.json).
> 
> Create a unified experience for configuration files is a nice to have and it
> feel more natural when you have to deal with.

Completely agreed, so let's update all eslint configuration as JS file ? :)

antleblanc:
> True, but, imo, we should enable it, instead of deleting that. I think it's more secure to keep this rule active.

The rule has not been deleting. Once again the only difference was that I've tried to drop the `settings` part which depends on the `eslint-import-resolver-lerna` package.

Since this settings is still needed then I cannot convert the root ESLint configuration from `.js` to `.json`.

I have updated my commit.

> Rules can't evolve ? Again, I think JS format is better. In the future, if we need to require a package etc, all files will already be in JS, no need to switch again from JSON, that will result by a mix of js / json file again.

Rules can evolve. It means that a some configurations needs also to be converted to `.js` file. We have to take it in account.

> Completely agreed, so let's update all eslint configuration as JS file ? :)

Not ESLint configuration only I think but all configuration in general.

---

I have changed the scope of this pull request in order to remove some globals from the root configuration.

Switching configuration format could be addressed in a separate pull request.

Thanks!

PR #984, opened 2019-07-02
refactor(apps.public.cloud): rename application name

antleblanc:
# Rename application name

prevent using internal code name for the public cloud application

## :nail_care: Refactor 

f3e236b - refactor(apps.public.cloud): rename application name

## :house: Internal

- No QC required.

### DIFF
--- a/packages/manager/apps/public-cloud/src/index.html
+++ b/packages/manager/apps/public-cloud/src/index.html
@@ -5,8 +5,11 @@
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     </head>
-    <body class="ovhstack h-100" style="overflow: hidden;">
-        <div data-ng-app="ovhStack" data-ng-strict-di class="h-100 d-flex flex-column overflow-hidden" data-ng-controller="PublicCloudController as $ctrl">
+    <body class="ovh-public-cloud-app h-100" style="overflow: hidden;">
+        <div class="h-100 d-flex flex-column overflow-hidden"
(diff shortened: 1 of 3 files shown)


packages/manager/apps/public-cloud/src/index.html
```diff
@@ -5,8 +5,11 @@
         <meta charset="utf-8">
         <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
     </head>
-    <body class="ovhstack h-100" style="overflow: hidden;">
-        <div data-ng-app="ovhStack" data-ng-strict-di class="h-100 d-flex flex-column overflow-hidden" data-ng-controller="PublicCloudController as $ctrl">
+    <body class="ovh-public-cloud-app h-100" style="overflow: hidden;">
```
marie-j: Maybe this inline style could be removed 
```suggestion
    <body class="ovh-public-cloud-app h-100 overfow-hidden">
```

jleveugle:
> It's a `BREAKING CHANGE` right ? [ovh-ux/manager/pull/984/files#diff-a2bbac7801b0860a615e413187c5a71dR50](https://github.com/ovh-ux/manager/pull/984/files#diff-a2bbac7801b0860a615e413187c5a71dR50)

Not really, the host app is never published, so nobody can take this one and import it to use it inside another app ^.^

packages/manager/apps/public-cloud/src/index.html
antleblanc: @marie-j Thanks for the suggestions. I will take a closer look if `overflow: hidden` rule is still required.

PR #1000, opened 2019-07-03
fix(pci.project.instances): check if ip is undefined

marie-j:
MANAGER-3013

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/instances/instances.service.js
+++ b/packages/manager/modules/pci/src/projects/project/instances/instances.service.js
@@ -517,7 +517,7 @@ export default class PciProjectInstanceService {
   }
 
   getReverseIp({ ipAddresses }) {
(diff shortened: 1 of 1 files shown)


packages/manager/modules/pci/src/projects/project/instances/instances.service.js
```diff
@@ -517,7 +517,7 @@ export default class PciProjectInstanceService {
   }
 
   getReverseIp({ ipAddresses }) {
-    const { ip } = find(ipAddresses, { type: 'public', version: 4 });
+    const ip = get(find(ipAddresses, { type: 'public', version: 4 }), 'ip');
```
cbourgois: Or 
```suggestion
    const { ip } = find(ipAddresses, { type: 'public', version: 4 }) || { ip: null };
```

PR #1018, opened 2019-07-04
Fix/instance edit

marie-j:
MANAGER-3020

### DIFF
--- a/packages/manager/modules/pci/src/components/project/flavors-list/flavor-group.class.js
+++ b/packages/manager/modules/pci/src/components/project/flavors-list/flavor-group.class.js
@@ -9,9 +9,12 @@ import { DEFAULT_OS } from './flavors-list.constants';
 
 export default class FlavorGroup {
   constructor(flavors) {
-    Object.assign(this, omit(find(flavors, { osType: DEFAULT_OS }), [
-      'regions', 'id', 'osType', 'planCodes',
-    ]));
+    Object.assign(
+      this,
+      omit(find(flavors, flavor => flavor.osType === DEFAULT_OS && !flavor.isFlex()), [
+        'regions', 'id', 'osType', 'planCodes',
+      ]),
+    );
 
     this.availableRegions = uniq(
       map(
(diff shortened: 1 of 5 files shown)


packages/manager/modules/pci/src/projects/project/instances/instance/edit/edit.html
```diff
@@ -147,7 +147,7 @@ <h1>{{:: 'pci_projects_project_instances_instance_edit_title' | translate }}</h1
         </form>
     </div>
 
-    <!-- <div class="mt-4 pt-4 border-top">
+    <div class="mt-4 pt-4 border-top">
```
FredericEspiau: `border-top` :(

Looks like we really need a true CSS utility framework

marie-j: No bootstrap4 utils https://getbootstrap.com/docs/4.0/utilities/borders/

packages/manager/modules/pci/src/components/project/flavors-list/flavor-group.class.js
```diff
@@ -9,9 +9,12 @@ import { DEFAULT_OS } from './flavors-list.constants';
 
 export default class FlavorGroup {
   constructor(flavors) {
-    Object.assign(this, omit(find(flavors, { osType: DEFAULT_OS }), [
-      'regions', 'id', 'osType', 'planCodes',
-    ]));
+    Object.assign(
```
FredericEspiau:
```suggestion
    assign(
```

(my understanding is that we should be using `Lodash` most of the time)

marie-j: @FredericEspiau `Lodash` is useful for defensive pattern (e.g: if I'm not sure the value I'll be giving will be defined) but here `this` is defined so no need. 
To be honest I don't know what's best but with using `Object.assign` it avoids importing more things to the class

antleblanc:
@jleveugle Please pay attention! This pull request shouldn't have been "Squash and merge".

We were lucky because some `fix`'s commits are already in the `develop` branch which relies on the `@ovh-ux/manager-pci` module.

Without theses commits we weren't be able to provide a new patch for the expected module.

PR #1019, opened 2019-07-05
fix(analytics-data-platform): reuse pci-project-regions-list component

ganeshkumar1989:
* The pci-project-regions-list component has been reused for regions
* The fields in the deploy page have been resized to xl size


ganeshkumar1989:
> Imo, it is more a refactor commit than a fix as it was working before

Yeah... I did not find that label earlier. Just added it :)

cbourgois:
> > Imo, it is more a refactor commit than a fix as it was working before
> 
> Yeah... I did not find that label earlier. Just added it :)

I think you should update your commit message too.

PR #1031, opened 2019-07-08
refactor(i18n): switch from xml to json

jleveugle:


packages/manager/modules/telecom-dashboard/src/translations/Messages_en_AU.json
```diff
@@ -0,0 +1,40 @@
+{
+  "telecom_dashboard_title": "Summary of my services ",
```
cbourgois: Extra space but iso with XML.

```suggestion
  "telecom_dashboard_title": "Summary of my services",
```

jleveugle: Should be managed in another PR ;)

Here, we only want to switch from XML to JSON :)

PR #1041, opened 2019-07-09
feat(onboarding): display onboarding only if required

jleveugle:
MANAGER-2989


antleblanc:
@jleveugle Is there any internal ticket reference related to that feature? Thanks!

PR #1042, opened 2019-07-09
refactor(navbar): remove non translated english locales

Jisay:


Jisay:
Yes it makes sense to remove it. But it should also be good to have that config at only one place?

antleblanc:
@Jisay Is there any internal ticket reference related to that change?

marie-j:
> Yes it makes sense to remove it. But it should also be good to have that config at only one place?

@Jisay I agree, we should have this list at one place and reuse it everywhere its needed

Jisay:
> > Yes it makes sense to remove it. But it should also be good to have that config at only one place?
> 
> @Jisay I agree, we should have this list at one place and reuse it everywhere its needed

@marie-j I have centralized the available languages in ```@ovh-ux/manager-core``` as ```@ovh-ux/manager-navbar``` has this module as peerDependency.

PR #1051, opened 2019-07-10
feat: add less-plugin-inline-svg

euhmeuh:


euhmeuh:
As seen with @antleblanc, we must use a not yet released version of the plugin, so we decided to point to the latest commit.

packages/manager/tools/component-rollup-config/package.json
```diff
@@ -42,6 +42,7 @@
     "camelcase": "^5.0.0",
     "estree-walker": "^0.6.0",
     "less": "^3.9.0",
+    "less-plugin-inline-svg": "https://github.com/atlassian/less-plugin-inline-svg#57ef58a5b2839571220f2bd651237ad8a56a2d51",
```
jleveugle: I'm not for using this in the common configuration, for multiples reasons:

* We are targeting a specific commit :/
* I think that we should avoid adding specific function in our less (our less today is standard and can be used with webpack / rollupjs, we don't have plugin for specific one)
* I think that we have better solutions to handle this case (like https://github.com/webpack-contrib/url-loader for webpack, which will use file-loader if the file size is over the limit). Here, if the svg is big, we are going to increase size of our css, and duplicate the SVG base64 if it's use multiple times.

You can use it, but I suggest you to add it into your specific rollup.config.js inside your project for this time. Let's try, and if the plugin is OK for all of us, we will add it in `component-rollup-config`. `component-rollup-config` is a template, you can override it for specific case.

See ya!

euhmeuh: Good point! I'll try that.

PR #1059, opened 2019-07-11
feat(office): add office module

frenautvh:
⚠️ **Not tested !** ⚠️ 

Moving ovh-module-office to the monorepo (add small cleanup + es6 imports)


frenautvh:
@jleveugle i think we need manager web in the monorepo to check if everything is working correctly

packages/manager/modules/office/README.md
```diff
@@ -0,0 +1,29 @@
+# @ovh-ux/office
```
antleblanc: To stay aligned with actual modules I think we should name it like:

- **@ovh-ux/manager-office**

packages/manager/modules/office/package.json
```diff
@@ -0,0 +1,35 @@
+{
+  "name": "@ovh-ux/office",
+  "version": "7.0.8",
```
antleblanc: Since you are changing the package name I think we can init the semver to v1.0.0.

packages/manager/modules/office/package.json
```diff
@@ -0,0 +1,35 @@
...
+  "description": "OVH Microsoft Office module",
+  "keywords": [
+    "angular",
+    "microsoft"
+  ],
+  "repository": "@ovh-ux/office",
+  "license": "BSD-3-Clause",
+  "author": "OVH SAS",
+  "main": "./src/index.js",
```
antleblanc: Missing both **browser** and **module** entries.

frenautvh: I used rollup to check if the build was successful but i think we can remove it now and only use src/index.js as a main entry point. Are we ok with that @jleveugle @cbourgois @marie-j ?

jleveugle: Nop, this module will only provide a main entry targeting the src/index.js at this time.

antleblanc: Must be also reflected to the rollup.config.js file then.

jleveugle: `rollup.config.js` is now removed :)

jleveugle:
> Is XML translation files have to be converted to JSON as well for the migration?

done

packages/manager/modules/office/package.json
antleblanc: Ok got it! Thanks!

PR #1067, opened 2019-07-12
Fix/ca build private registry

varun257:
## Remove private registry for CA region

resolves MANAGER-3018


varun257:
@antleblanc can this go as hotfix ?

antleblanc:
@varun257 For sure.

/cc @lizardK @jleveugle @cbourgois @marie-j

marie-j:
@varun257 Could you please rebase on `master`? Thanks

jleveugle:
Do you test the `US` build ? We should check if the menu (orchestration) appears empty or if we remove it because it is empty :)

PR #1068, opened 2019-07-12
module sharepoint

frenautvh:


cbourgois:
We have to transform translations `xml` to `json`.

packages/manager/modules/sharepoint/.htmlhintrc
```diff
@@ -0,0 +1,25 @@
...
+  "inline-style-disabled": false,
+  "space-tab-mixed-disabled": "space",
+  "spec-char-escape": true,
+  "src-not-empty": true,
+  "style-disabled": true,
+  "tag-pair": true,
+  "tag-self-close": false,
+  "tagname-lowercase": true,
+  "title-require": true
+}
```
marie-j: Isn't there a shared configuration for this one ?

jleveugle:
> We have to transform translations `xml` to `json`.

done

packages/manager/modules/sharepoint/.htmlhintrc
jleveugle: Not in this case, it's a specific one for that module (options are different from the root file :/)

PR #1077, opened 2019-07-16
fix(pci-components-region-list): region update

ganeshkumar1989:
MANAGER-3036

The region is not being set when only one region is present in a continent.

The region variable is not set correctly so that the select picker is properly displayed when displaySelectedRegion is true.

### DIFF
--- a/packages/manager/modules/pci/src/components/project/regions-list/regions-list.controller.js
+++ b/packages/manager/modules/pci/src/components/project/regions-list/regions-list.controller.js
@@ -78,7 +78,8 @@ export default class RegionsListController {
 
   onMacroChange(macro, regions) {
     if (regions.length === 1) {
-      this.onRegionChange(regions[0]);
+      [this.region] = regions;
+      this.onRegionChange(this.region);
     }
   }
 


packages/manager/modules/pci/src/components/project/regions-list/regions-list.controller.js
```diff
@@ -78,6 +78,7 @@ export default class RegionsListController {
 
   onMacroChange(macro, regions) {
     if (regions.length === 1) {
+      [this.region] = regions;
       this.onRegionChange(regions[0]);
```
antleblanc: In that way you can refer to your previous assignation:

```suggestion
      this.onRegionChange(this.region);
```

antleblanc:
@ganeshkumar1989 Is there any internal ticket reference related to this pull request?

packages/manager/modules/pci/src/components/project/regions-list/regions-list.controller.js
ganeshkumar1989: This is done. Thanks :)

ganeshkumar1989:
I just updated the ticket number in the description

PR #1090, opened 2019-07-18
fix(private-registry): fix js error

varun257:
remove oui-field to fix js error

resolves MANAGER-2940

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/private-registry/create/create.html
+++ b/packages/manager/modules/pci/src/projects/project/private-registry/create/create.html
@@ -19,18 +19,16 @@
         </oui-field>
         <div data-ng-if="$ctrl.acceptTermsAndConditions">
             <p data-translate="private_registry_beta_warning"></p>
-            <oui-field>
-                <oui-checkbox
-                    data-model="$ctrl.agreementsAccepted"
-                    data-name="agreements"
-                    data-required>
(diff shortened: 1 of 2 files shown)


packages/manager/modules/pci/src/projects/project/private-registry/create/create.html
```diff
@@ -19,18 +19,16 @@
...
-                    data-name="agreements"
-                    data-required>
-                    <span data-translate="private_registry_new_project_contracts"
-                        data-translate-values="{
-                            contractsLinks: $ctrl.getCompiledLinks('<a class=\'oui-link_icon\' href=\'<%= pdf %>\' target=\'_blank\' rel=\'noopener\'><%= name %><span class=\'oui-icon oui-icon-external_link\' aria-hidden=\'true\'></span></a>')
-                        }">
-                    </span>
-                </oui-checkbox>
-            </oui-field>
+            <oui-checkbox
```
cbourgois: Or maybe 
```html 
<oui-field label="{{ :: 'private_registry_beta_warning' | translate }}">
  <oui-checkbox>...</oui-checkbox>
</oui-field>
```

varun257: There will be an error unless we add a `form`. But `form` is not used here because we did not want form `oui-field` error message on validation.

PR #1096, opened 2019-07-19
fix(apps.support): add missing @ovh-ux/manager-config dependency

antleblanc:
# Add missing @ovh-ux/manager-config dependency

## :bug: Bug Fix

fix(apps.support): add missing @ovh-ux/manager-config dependency

## :house: Internal

- No QC required.

/cc @FredericEspiau

### DIFF
--- a/packages/manager/modules/support/package.json
+++ b/packages/manager/modules/support/package.json
@@ -21,7 +21,7 @@
     "@ovh-ux/component-rollup-config": "^5.1.0"
   },
   "peerDependencies": {
-    "@ovh-ux/manager-core": "^6.0.1",
+    "@ovh-ux/manager-core": "^6.1.1",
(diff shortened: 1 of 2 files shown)


packages/manager/modules/support/package.json
```diff
@@ -21,7 +21,8 @@
     "@ovh-ux/component-rollup-config": "^5.1.0"
   },
   "peerDependencies": {
-    "@ovh-ux/manager-core": "^6.0.1",
+    "@ovh-ux/manager-config": "^0.3.0",
```
cbourgois: Is `@ovh-ux/manager-config` directly used in support module ?

antleblanc: nope! removed! thanks!

PR #1118, opened 2019-07-25
fix(exchange): replace upperCase with toUpper

frenautvh:
issue : `_.upperCase('FOO_BAR') = 'FOO BAR';`
fix : `_.toUpper('FOO_BAR') = 'FOO_BAR';`

one of the multiple reasons i don't like lodash ...


FredericEspiau:
`one of the multiple reasons i don't like lodash ...`

Did you know that `_.find` will disappear with Lodash 5 ? But somehow, we still think that Lodash is the same as polyfills 🤔

PR #1121, opened 2019-07-26
build(deps): upgrade ovh-ui-angular to v3.5.1

radireddy:
not able to change storage size using keyboard

This issue was in oui-numeric component. I have fixed the component and the fix is available in 3.5.1 release.  I have updated PCI with this release.

MBP-505

### DIFF
--- a/packages/manager/apps/pci/package.json
+++ b/packages/manager/apps/pci/package.json
@@ -45,7 +45,7 @@
     "ovh-api-services": "^7.0.0",
     "ovh-common-style": "^5.0.0",
     "ovh-manager-webfont": "^1.2.0",
-    "ovh-ui-angular": "^3.3.3",
+    "ovh-ui-angular": "^3.5.1",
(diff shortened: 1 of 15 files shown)


packages/manager/apps/pci/package.json
```diff
@@ -45,12 +45,12 @@
     "ovh-api-services": "^7.0.0",
     "ovh-common-style": "^5.0.0",
     "ovh-manager-webfont": "^1.2.0",
-    "ovh-ui-angular": "^3.3.3",
+    "ovh-ui-angular": "^3.5.1",
```
marie-j: Should be upgraded in all `apps` 
`yarn upgrade-interactive --latest ovh-ui-angular`

marie-j:
Even if this fixes the issue, it would be easier if the commit mentions the version bump instead imho

radireddy:
updated commit message

PR #1123, opened 2019-07-26
build: normalize json with webpack loader

frenautvh:
issue : {0} was not replaced with {{ t0 }} when importing translation json files
fix : add webpack loader

### DIFF
--- a/packages/manager/tools/component-rollup-config/src/plugins/common/index.ts
+++ b/packages/manager/tools/component-rollup-config/src/plugins/common/index.ts
@@ -1,0 +1,15 @@
+const translationNormalize = (text) => {
+  if (text) {
+    const filtered = text
+      .toString()
+      .replace(/&#13;\n/g, ' ') // carriage returns
+      .replace(/&#160;/g, ' ') // spaces
+      .replace(/\{(\s?\d\s?)\}/g, '{{t$1}}'); // {0} => {{t0}}
+    return filtered;
+  }
+  return text;
+};
+
+export = {
+  translationNormalize,
+}
--- a/packages/manager/tools/component-rollup-config/src/plugins/common/index.ts
+++ b/packages/manager/tools/component-rollup-config/src/plugins/common/index.ts
@@ -4,7 +4,7 @@ const translationNormalize = (text) => {
       .toString()
       .replace(/&#13;\n/g, ' ') // carriage returns
       .replace(/&#160;/g, ' ') // spaces
-      .replace(/\{(\s?\d\s?)\}/g, '{{t$1}}'); // {0} => {{t0}}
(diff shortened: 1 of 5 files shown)


packages/manager/tools/component-rollup-config/src/plugins/common/index.ts
```diff
@@ -0,0 +1,15 @@
+const translationNormalize = (text) => {
+  if (text) {
+    const filtered = text
+      .toString()
+      .replace(/&#13;\n/g, ' ') // carriage returns
+      .replace(/&#160;/g, ' ') // spaces
+      .replace(/\{(\s?\d\s?)\}/g, '{{t$1}}'); // {0} => {{t0}}
```
FredericEspiau:
```suggestion
      .replace(/\{(\s*\d\s*)\}/g, '{{t$1}}'); // {0} => {{t0}}
```

packages/manager/tools/component-rollup-config/src/plugins/common/index.ts
```diff
@@ -0,0 +1,15 @@
+const translationNormalize = (text) => {
+  if (text) {
+    const filtered = text
+      .toString()
+      .replace(/&#13;\n/g, ' ') // carriage returns
+      .replace(/&#160;/g, ' ') // spaces
```
FredericEspiau: https://stackoverflow.com/questions/31790440/regex-to-replace-no-break-space < may be more appropriate

packages/manager/tools/component-rollup-config/src/plugins/common/index.ts
```diff
@@ -0,0 +1,15 @@
+const translationNormalize = (text) => {
+  if (text) {
+    const filtered = text
+      .toString()
+      .replace(/&#13;\n/g, ' ') // carriage returns
```
FredericEspiau:
```suggestion
      .replace(/\r\n/g, ' ') // carriage returns
```

packages/manager/tools/component-rollup-config/src/plugins/common/index.ts
frenautvh: it's a regex that replace the html code `&#160;` by a whitespace in our json, it's different from replacing a single no-break-space character

packages/manager/tools/component-rollup-config/src/plugins/common/index.ts
frenautvh: same here the regex is about replacing the `&#13;` html code

FredericEspiau: Why not replacing it why `\r` then ?

PR #1129, opened 2019-07-29
fix(users): avoid scroll roles matrix

ganeshkumar1989:
Enable word wrap for column headers and remove extra padding

MBP-521

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/users/roles-matrix/roles-matrix.less
+++ b/packages/manager/modules/pci/src/projects/project/users/roles-matrix/roles-matrix.less
@@ -26,5 +26,10 @@
       border-left: none;
       box-shadow: none;
     }
+
+    .oui-datagrid__header {
+      padding: 0.125rem 0.25rem;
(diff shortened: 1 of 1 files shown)


packages/manager/modules/pci/src/projects/project/users/roles-matrix/roles-matrix.less
```diff
@@ -26,5 +26,10 @@
       border-left: none;
       box-shadow: none;
     }
+
+    .oui-datagrid__header {
```
FredericEspiau: Maybe this should be done in `ui-kit`, it looks like the best behaviour for most datagrids

AxelPeter: No, Not in this form. 
It's too specific for this case since the datagrid is designed to have a responsive form with the scrollbar. For example, this will have collateral effect with the sticky cells, that will not adapt their heights with the non-sticky cells.

PR #1133, opened 2019-07-30
fix(components-flavors-list): available regions

ganeshkumar1989:
Filter the available regions based on individual flavor availability and also set available flag for flavorGroup correctly

MBP-507 & MBP-506

### DIFF
--- a/packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.service.js
+++ b/packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.service.js
@@ -5,6 +5,7 @@ import isNil from 'lodash/isNil';
 import map from 'lodash/map';
 import mapValues from 'lodash/mapValues';
 import omit from 'lodash/omit';
+import some from 'lodash/some';
 
 import Flavor from './flavor.class';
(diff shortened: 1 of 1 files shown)


packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.service.js
```diff
@@ -34,9 +35,11 @@ export default class FlavorsList {
       .then(({ flavors, prices }) => map(
         groupBy(flavors.filter(({ planCodes }) => !isNil(planCodes.hourly)), 'name'),
         groupedFlavors => new Flavor({
-          ...omit(groupedFlavors[0], ['id', 'region']),
+          ...omit(groupedFlavors[0], ['available', 'id', 'region']),
+          available: reduce(groupedFlavors,
```
frenautvh: maybe use https://lodash.com/docs/4.17.15#some ?

`available: some(groupedFlavors, 'available')`

ganeshkumar1989: Thanks a lot @frenautvh . This change is done :)

PR #1152, opened 2019-08-02
build(office): add missing peer dependencies

frenautvh:


packages/manager/modules/office/package.json
```diff
@@ -20,8 +20,12 @@
     "@ovh-ux/component-rollup-config": "^6.0.0"
   },
   "peerDependencies": {
-    "@ovh-ux/ovh-utils-angular": "^14.0.2",
+    "@ovh-ux/ng-ovh-swimming-poll": "^4.0.0",
```
antleblanc: I cannot find any reference related to this peer dependency inside the Office module.

Could you please confirm me if that dependency is needed?

Thanks!

packages/manager/modules/office/package.json
```diff
@@ -20,8 +20,12 @@
...
+    "@ovh-ux/ng-ovh-utils": "^14.0.3",
+    "@ovh-ux/ng-ovh-web-universe-components": "^5.0.0",
     "angular": "^1.7.5",
-    "angular-ui-bootstrap": "1.3.3"
+    "angular-translate": "^2.18.1",
+    "angular-ui-bootstrap": "1.3.3",
+    "ovh-ui-angular": "^3.5.1"
```
antleblanc: Following service is used `OvhHttp`. It means that `@ovh-ux/ng-ovh-http` should also be added as a peer dependency.

PR #1156, opened 2019-08-05
feat(apps.public.cloud): manage both feedback's url for FR and EN

antleblanc:
# Manage both feedback's url for FR and EN

## :sparkles: Feature

e9b8281 - feat(apps.public.cloud): manage both feedback's url for FR and EN

## :house: Internal

- MANAGER-3000

## :building_construction: Continuous Integration

- [x] Update `build-manager-ng-app` pipeline (export both `FEEDBACK_URL_EN` and `FEEDBACK_URL_FR`)


antleblanc:
Wrong branch naming but can't update it since the Pull request has been created.

packages/manager/modules/pci/src/projects/project/project.controller.js
```diff
@@ -63,6 +63,12 @@ export default class ProjectController {
       });
   }
 
+  get feedbackUrl() {
+    return this.user.ovhSubsidiary.toUpperCase() === 'FR'
```
marie-j: Is it possible to have something like 
```suggestion
return [`__FEEDBACK_URL_${this.user.ovhSubsidiary.toUpperCase()}__`] || __FEEDBACK_URL_EN__
```
and does it make sense

packages/manager/modules/pci/src/projects/project/project.html
```diff
@@ -10,22 +10,38 @@
...
+                        data-ng-href="{{ $ctrl.feedbackUrl }}"
+                        data-translate="pci_projects_project_welcome_feedack">
+                    </a>
+                </div>
+                <div class="ml-auto">
+                    <button
+                        type="button"
+                        class="oui-link pci-project-link_light"
+                        data-ng-if="$ctrl.user.ovhSubsidiary === 'FR'"
+                        data-ng-click="$ctrl.goToLegacy()"
```
marie-j: Does it make sense to keep this special behaviour as feedback url is now given above ?

packages/manager/modules/pci/src/projects/project/translations/Messages_fr_FR.json
```diff
@@ -14,6 +14,7 @@
   "pci_projects_project_documentation_see_all_public_cloud_guides": "See all Public Cloud guides",
 
   "pci_projects_project_welcome": "Bienvenue sur la nouvelle interface Public Cloud",
+  "pci_projects_project_welcome_feedack": "Feedback",
```
marie-j:
```suggestion
  "pci_projects_project_welcome_feedback": "Feedback",
```

packages/manager/modules/pci/src/projects/project/project.controller.js
antleblanc: Sorry it's not possible. That doesn't return the same value.

packages/manager/modules/pci/src/projects/project/project.html
antleblanc: I will ask it directly into the associated ticket. If not then I will create a new to be able to address any upcoming changes.

packages/manager/modules/pci/src/projects/project/translations/Messages_fr_FR.json
antleblanc: Good catch!

frenautvh:
instead of "do not merge" you could create a draft :)

antleblanc:
IMO A draft would be nice if the feature is not fully finished (which is not the case here).

Plus an approval is required in order to be able to provide a staging.

PR #1165, opened 2019-08-08
fix(theme): navbar overlap

ganeshkumar1989:
In mobile view, the clipboard icon (if it is available) overlaps with the navbar. This is because the z-index of both the clipboard icon and the side bar is 2.
To fix this issue, the z-index for the navbar is increased to 3

MBP-541

### DIFF
--- a/packages/manager/apps/public-cloud/src/assets/theme/_navbar.less
+++ b/packages/manager/apps/public-cloud/src/assets/theme/_navbar.less
@@ -8,7 +8,7 @@
 .oui-navbar {
   background-color: #fff;
   box-shadow: 0 2pt 4pt rgba(17, 63, 109, 0.15);
-  z-index: 2;
+  z-index: 3;
 
   .oui-navbar__brand {
     color: rgb(40, 89, 192);


packages/manager/apps/public-cloud/src/assets/theme/_navbar.less
```diff
@@ -8,7 +8,7 @@
 .oui-navbar {
   background-color: #fff;
   box-shadow: 0 2pt 4pt rgba(17, 63, 109, 0.15);
-  z-index: 2;
+  z-index: 3;
```
marie-j: I think you'll need to increase the `sidebar` z-index as it is not related the navbar anymore 
https://github.com/ovh-ux/manager/blob/91336be8f23e6df3468a9dfc420b4dce1ca97ff7/packages/manager/apps/public-cloud/src/assets/theme/_sidebar.less#L16

ganeshkumar1989: Sorry. My mistake. Its navbar and not sidebar. Navbar is the one that has links to Manager Web, dedicated etc

PR #1170, opened 2019-08-09
fix(pci.project.kubernetes.details.service): adjust popover placement

ganeshkumar1989:
Set popover placement to auto so that it auto-adjusts for mobile view and does not get clipped

MBP-540

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/kubernetes/details/service/service.html
+++ b/packages/manager/modules/pci/src/projects/project/kubernetes/details/service/service.html
@@ -84,8 +84,8 @@
                 <dl class="oui-tile__definition">
                     <dt class="oui-tile__term">
                         <span data-translate="kube_service_file"></span>
-                        <oui-popover>
-                            <button type="button" class="oui-popover-button" oui-popover-trigger></button>
+                        <oui-popover data-oui-popover-placement="auto">
+                            <button type="button" class="oui-popover-button" data-oui-popover-trigger></button>
                             <oui-popover-content>
                                 <span data-translate="kube_service_file_help"></span>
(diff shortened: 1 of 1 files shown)


ovh-cds:
CDS Report web#4506.0 ✘  
* 
  * Build CA ✘  
  * Build EU ✘  
  * Build US ✘

ovh-cds:
CDS Report web#4507.0 ✘  
* 
  * Build CA ✘  
  * Build EU ✘  
  * Build US ✘

packages/manager/modules/pci/src/projects/project/kubernetes/details/service/service.html
```diff
@@ -84,8 +84,8 @@
                 <dl class="oui-tile__definition">
                     <dt class="oui-tile__term">
                         <span data-translate="kube_service_file"></span>
-                        <oui-popover>
-                            <button type="button" class="oui-popover-button" oui-popover-trigger></button>
+                        <oui-popover data-oui-popover-placement="auto">
```
jleveugle: Why is not the default behavior ? :(

ganeshkumar1989: Right now, the default behavior is right. Will discuss with Axel on this :)

AxelPeter: The default behaviour was asked `top-center` in the guidelines. Need to see how reacts Popper.js, but we could set it to default `auto` for the next major update (v4).

But here, the preoccupation should be that we should not use `oui-popover` component anymore (deprecated) :p
It should be used as a directive, to help for the migration.

PR #1178, opened 2019-08-12
refactor: remove useless images / styles

jleveugle:
Hello,

Grepping classes / images each time, looks like we don't use them :)

So, small clean! :)


cbourgois:
🎉

PR #1186, opened 2019-08-13
build: avoid duplicating dependencies

jleveugle:
Hello:

Harmonize dependencies to avoid multiple jquery-ui, lodash and more on @ovh-ux/manager-web :)


antleblanc:
I'm not sure to understand why a `fix` commit message is associated with a `perfomance` label.

Is there any reason?

Existing `pref` as commit message type (if needed) but in this case it's related to some bump so we used to specifiy it with the `build` type as commit message.

Is that make more sense?

Thanks!

cbourgois:
`fix` commit for `manager-web` with modifications on other packages (`freefax`, `overthebox` and so on) in the same commit : it will release all theses packages as patch. Is this what we want ?

PR #1187, opened 2019-08-13
feat(webpack-config): add javascriptEnabled less-loader option

frenautvh:


antleblanc:
Would be nice to also reflect it to the documentation.

PR #1196, opened 2019-08-13
docs(webpack-config): update readme with lessJavascriptEnabled option

frenautvh:


packages/manager/tools/webpack-config/README.md
```diff
@@ -34,6 +34,7 @@ The following configuration parameters are _optionals_:
    relatively to the current resource path. You can override this behavior by providing
    an absolute translationsRoot path. This path will then be used to resolve translations
    instead of current resource path.
+- _lessJavascriptEnabled_: adds _javascriptEnabled_ option to the less compiler
```
antleblanc: It could be right next the `_lessPath_`  parameter since both are related to **Less** usage.

Isn't it?

frenautvh: no because this parameter is optional, you can see there are two section : mandatory & optional params :)

antleblanc: You're right. Didn't notice.

Thanks!

PR #1197, opened 2019-08-13
feat: manage incomplete nic

Jisay:
Require:
- [x] ovh-ux/ng-ovh-sso-auth#21 to be released


yarn.lock
```diff
@@ -1725,6 +1725,14 @@
     shuffle-seed "^1.1.6"
     titleize "^1.0.1"
 
+"@ovh-ux/manager-server-sidebar@^0.4.2":
```
cbourgois: The version in the repository is `0.5.0`. Should not be in yarn.lock.

cbourgois: Please rebase on `develop` (changes done in https://github.com/ovh-ux/manager/commit/8d5d59e25b5119d8d53d19a9b5907f746699d161)

packages/manager/apps/sign-up/src/index.js
```diff
@@ -62,6 +62,9 @@ angular
     TranslateServiceProvider.setUserLocale(userLocale);
     $translateProvider.use(userLocale);
   })
+  .config(/* @ngInject */ (ssoAuthenticationProvider) => {
+    ssoAuthenticationProvider.allowIncompleteNic(true);
```
antleblanc: Is the `Public Cloud` application should also enable incomplete nic?

Jisay: As nobody knows exactly what can happen with incomplete nic, all applications do not allow incomplete nic before somebody detect that there is no risk to allow incomplete nics :-)

yarn.lock
Jisay: done, thx

packages/manager/apps/sign-up/src/index.js
marie-j: It should be the other way round therefore `allowIncompleteNic(false)`

Jisay: Not for sign-up. The /auth page will create incomplete nics, then the creation script will redirect the incomplete nic to the sign-up app page. The app have to allow `incomplete` nics in order to complete them.

marie-j: Oopsie I misread, my bad

PR #1204, opened 2019-08-14
Feat/exchange subtree

marie-j:


antleblanc:
@marie-j Couldn't be a single commit `fix(exchange): git subtree pull ovh-module-exchange`?

I'm asking this because I noticed the same commit message has been repeated twice.

Thanks!

PR #1205, opened 2019-08-14
build(webpack): get access to the region environment variable

antleblanc:
# Get access to the `REGION` environment variable

a28c927 - build(webpack): get access to the region environment variable

## :house: Internal

- No QC required.

### DIFF
--- a/packages/manager/apps/pci/webpack.config.js
+++ b/packages/manager/apps/pci/webpack.config.js
@@ -7,7 +7,7 @@ module.exports = (env = {}) => {
     template: './index.html',
     basePath: '.',
     root: path.resolve(process.cwd()),
-  }, env);
+  }, process.env.REGION ? Object.assign(env, { region: process.env.REGION }) : env);
 
   return merge(config, {
     entry: path.resolve('./index.js'),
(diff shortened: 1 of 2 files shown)


antleblanc:
All apps that refer to a Telecom product hasn't been changed.

Let me know if you think that all theses configurations must be updated as well or not.

Thanks!

FredericEspiau:
I would have added it everywhere because right now we only have 1 difference but maybe in 1 year we will have 50 differences and it would be hard to keep track of why and where specific code was done :)

marie-j:
I agree with @FredericEspiau this should probably be standard for all apps. Would it make sense to have it directly in our common webpack config ?

antleblanc:
Yes it is!

You can refer to the actual configuration from [`@ovh-ux/manager-webpack-dev-server`](https://github.com/ovh-ux/manager/blob/master/packages/manager/tools/webpack-dev-server/src/config.ts#L8).

Previous controls panels used to pass the environment via a `yarn run start|build:REGION` command.

The way that the Public Cloud application is passing the environment variable to the development server configuration is the right way so that's why for all applications who can refer to multiple regions I have submitted this change.

In addition if you refer to the APIs, it doesn't make sense to provide this environment variable for all applications (meaning the telecom one for instance) because all API routes won't exist.

PR #1208, opened 2019-08-16
feat(pci.projects): allow to delete currently creating project

marie-j:
## Allow to delete currently creating project

Add option to delete project being in creation 

MBE-374

### DIFF
--- a/packages/manager/modules/pci/src/projects/projects.routing.js
+++ b/packages/manager/modules/pci/src/projects/projects.routing.js
@@ -1,15 +1,12 @@
 import map from 'lodash/map';
 
-import controller from './projects.controller';
-import template from './projects.html';
+import Project from './Project.class';
 
 export default /* @ngInject */ ($stateProvider) => {
   $stateProvider
     .state('pci.projects', {
       url: '/projects',
-      controller,
-      controllerAs: '$ctrl',
-      template,
+      component: 'pciProjects',
       redirectTo: (transition) => {
         const projectsPromise = transition.injector().getAsync('projects');
         return projectsPromise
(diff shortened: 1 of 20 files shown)


packages/manager/modules/pci/src/projects/projects.routing.js
```diff
@@ -23,11 +20,38 @@ export default /* @ngInject */ ($stateProvider) => {
...
       resolve: {
         breadcrumb: /* @ngInject */ () => null,
+        confirmDeletion: /* @ngInject */ $state => project => $state.go('pci.projects.remove', { projectId: project.project_id }),
+        defaultProject: /* @ngInject */
+          PciProjectsService => PciProjectsService.getDefaultProject(),
+        getProject: /* @ngInject */ OvhApiCloudProject => project => OvhApiCloudProject
+          .v6()
+          .get({ serviceName: project.serviceName })
+          .$promise
+          .then(projectDetails => new Project(projectDetails)),
```
antleblanc: I don't know if you share the same feeling as mine but it means that all
properties returned by the API `cloud.Project` will be accessible on the Front.

Because in your class constructor you are using ` Object.assign(this, project);`.

But in some cases, some of these properties aren't relevant (or they act more
as a private one).

That's why the getters can provide a more granular control layer on what can
or cannot be accessible on the Front.

Any feedback is well appreciate.

Thanks!

packages/manager/modules/pci/src/projects/translations/Messages_fr_FR.json
```diff
@@ -5,5 +5,7 @@
   "pci_projects_status_ok": "Actif",
   "pci_projects_status_deleting": "Suppression en cours",
   "pci_projects_status_creating": "Création en cours",
-  "pci_projects_project_show": "Visualiser"
+  "pci_projects_project_show": "Visualiser",
+  "pci_projects_project_delete": "Supprimer",
+  "pci_projects_project_delete_success": "Une confirmation de suppression vient de vous être envoyé par email."
```
antleblanc:
```suggestion
  "pci_projects_project_delete_success": "Une confirmation de suppression vient de vous être envoyée par e-mail."
```

packages/manager/modules/pci/src/projects/projects.routing.js
marie-j: I don't understand your point, because using the class or directly `projectDetails` will expose the same API payload

antleblanc: Oh! nevermind then!

I asked myself if there was a way to make a distinction between public and private properties which is not the case as of now.

PR #1210, opened 2019-08-19
build: remove pci rollup config

marie-j:
Package is now private so there's no need to provide dist


antleblanc:
Since it change how the Public Cloud application is launched I think we should replicate this into a `README.md` file.

I noticed that `packages/manager/apps/public-cloud` doesn't have one yet.

Maybe it could be a good starting point to provide a `README.md` for this application, isn't it?

packages/manager/apps/public-cloud/package.json
```diff
@@ -8,9 +8,9 @@
     "build": "webpack --env.production",
     "dev": "webpack-dev-server",
     "dev:watch": "yarn run dev",
-    "start": "lerna exec --stream --scope='@ovh-ux/manager-public-cloud' --include-filtered-dependencies -- yarn run build",
-    "start:dev": "lerna exec --stream --scope='@ovh-ux/manager-public-cloud' --include-filtered-dependencies -- yarn run dev",
-    "start:watch": "lerna exec --stream --parallel --scope='@ovh-ux/manager-public-cloud' --include-filtered-dependencies -- yarn run dev:watch"
+    "start": "lerna exec --stream --scope='@ovh-ux/manager-public-cloud' --include-filtered-dependencies -- npm run build --if-present",
```
antleblanc: Is there any benefit to make a distinction between `yarn` and `npm` in that case?

marie-j: `yarn` doesn't have a `--if-present` option

antleblanc: Noticed!

But in my point of view I think we should keep all theses scripts. In that way it will ensure, as a module level, that we can keep building it and figured out what's going otherwise it will be delegate to the Public Cloud webpack config which can increase the building time for this given module.

PR #1234, opened 2019-08-27
fix(web): allow to rename private database service

cbourgois:
### Description of the Change

9ce911421 — fix(web): allow to rename private database service


antleblanc:
@cbourgois Is there any internal ticket reference related to this fix?

I also noticed that the actual Web Control Panel doesn't have this fix.

Is it normal?

jleveugle:
The actual web has not the same dependencies (angularjs...)

So yes, we fix some new behavior :/

PR #1271, opened 2019-09-04
refactor: fix all eslint warnings

cbourgois:
### Description of the Change

Before :
```bash
yarn lint:js
$ eslint --fix .
✖ 3396 problems (344 errors, 3052 warnings)
```

After : 
```bash
yarn lint:js
$ eslint --fix .
✨  Done in 82.25s.
````

ca95d4b16 — refactor: fix all eslint warnings

### CODE REVIEW

File: packages/manager/apps/dedicated/src/app/account/otrs/filter/html-string-linky.filter.js [hunk 1]
```diff
@@ -1,13 +1,14 @@
 import escape from 'lodash/escape';
 import forEach from 'lodash/forEach';
 import unescape from 'lodash/unescape';
+import head from 'lodash/head';
```
FredericEspiau: Imo it's not a good idea to copy/paste Angular code

antleblanc:
It sounds quiet confusing because your commit message refer to ESLint's warnings but in the same time you also reported:

> ```shell
> yarn lint:js
> $ eslint --fix .
> ✖ 3396 problems (344 errors, 3052 warnings)
> ```

I'm curious about where theses 344 errors came from.

Is the `eslint --fix .` command that's keeps reporting some warning as an issue?

For a **Continuous Integration** concern (meaning running `yarn run lint:js`) I think we should add the `--quiet` flag to prevents reporting _errors_ on warnings.

see: https://eslint.org/docs/user-guide/command-line-interface#options

Thanks!

FredericEspiau:
@antleblanc sounds quite confusing cause you should be doing something else than watching Github right now :sun: 🚲 📖 📺 🌲 🍪

cbourgois:
For example, a lot of errors came from this file https://github.com/ovh-ux/manager/blob/develop/packages/manager/apps/dedicated/src/app/dedicated/server/installation/ovh/dedicated-server-installation-ovh.controller.js#L862.
I think theses files were imported with `git subtree` and automatically committed without linting.

Concerning `--quiet` it is a workaround, but we have to fix :x: `errors` and considers ⚠️  `warnings`, if not what are the purpose of `warnings` :confused: ?

antleblanc:
Seems linked to the previous [`.eslintignore`](https://github.com/ovh-ux/ovh-manager-dedicated/blob/master/.eslintignore#L4) file that wasn't kept as is during the migration.

Goals of a warning is to warns us I guess;)

But in the same time warnings shouldn't rely on the test phase made by the Continuous Integrations.

That's why for all the parts of the Control Panel we specified the `--quiet` flag:

- https://github.com/ovh-ux/ovh-manager-web/blob/f26aabde028c7774928b9ec5f93141172df030bd/package.json#L28
- https://github.com/ovh-ux/ovh-manager-dedicated/blob/39c749b8a035755fc7c88b634c5d7bab28300151/package.json#L28
- https://github.com/ovh-ux/ovh-manager-cloud/blob/31926801e1b89cea047b8404e5b28e2baa3aa7e0/package.json#L31
- https://github.com/ovh-ux/ovh-manager-telecom/blob/a819408fc62cfb1a0579a28e2ef7849f9e281748/package.json#L30

Maybe we can add this flag on the `lint:js` script entry.

PR #1295, opened 2019-09-11
feat(support): add manager-navbar to select lang in support app

frenautvh:
for testing purposes, adding ovh-manager-navbar to support-app in order to be able to switch selected language when testing


FredericEspiau:
Good idea !

PR #1296, opened 2019-09-11
feat(support): add urgency selector

frenautvh:
add urgency selector for 'business' & 'enterprise' support level

### DIFF
--- a/packages/manager/modules/support/src/new-ticket/creation-form/creation-form.controller.js
+++ b/packages/manager/modules/support/src/new-ticket/creation-form/creation-form.controller.js
@@ -1,9 +1,30 @@
 import findLast from 'lodash/findLast';
 
 export default class SupportNewCreationFormController {
+  /* @ngInject */
+  constructor(SupportNewTicketService) {
+    this.SupportNewTicketService = SupportNewTicketService;
+  }
+
+  $onInit() {
+    return this.SupportNewTicketService.getSupportLevel().then((level) => {
+      this.supportLevel = level;
+      if (this.shouldSelectUrgency()) {
+        return this.SupportNewTicketService.getUrgencies().then((urgencies) => {
+          this.urgencies = urgencies;
+        });
+      }
+      return level;
+    });
+  }
+
   get subject() {
     const issue = findLast(this.issues, 'subject');
     return issue ? issue.subject : undefined;
+  }
+
+  shouldSelectUrgency() {
+    return ['business', 'enterprise'].indexOf(this.supportLevel) >= 0;
   }
 
   submitForm(isSuccess) {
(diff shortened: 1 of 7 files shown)


packages/manager/modules/support/src/new-ticket/creation-form/creation-form.controller.js
```diff
@@ -1,17 +1,39 @@
...
+    });
+  }
+
   get subject() {
     const issue = findLast(this.issues, 'subject');
     return issue ? issue.subject : undefined;
   }
 
+  shouldSelectUrgency() {
+    return ['business', 'enterprise'].indexOf(this.supportLevel) >= 0;
```
FredericEspiau: Wouldn't it be better to have capabilities for each support level ?

````js
class SupportLevel {
   shouldSelectUrgency() {
        return includes(
            ['business', 'enterprise'],
            this.supportLevel,
        );
    }
}
````

this way capabilities and duties of each support level would be at the same place :)

File: packages/manager/modules/support/src/new-ticket/creation-form/creation-form.html [hunk 2]
```diff
@@ -1,6 +1,16 @@
 <form novalidate
       name="supportNewCreationForm">
 
+  <oui-field data-label="{{ 'ovhManagerSupport_new_urgency' | translate }}"
```
FredericEspiau:
```suggestion
  <oui-field data-label="{{:: 'ovhManagerSupport_new_urgency' | translate }}"
```

packages/manager/modules/support/src/new-ticket/issues-form/issues-form.controller.js
```diff
@@ -1,14 +1,16 @@
-import constants from './issues-form.constant.js';
+import constants from './issues-form.constant';
 
 export default class SupportNewIssuesFormController {
   /* @ngInject */
-  constructor($q, $state, $translate, OvhApiService, OvhApiSupport, OvhApiMe) {
+  constructor($q, $state, $translate, OvhApiService,
```
FredericEspiau:
```suggestion
  constructor(
    $q,
    $state,
    $translate,
    OvhApiService,
    OvhApiSupport,
    OvhApiMe,
    SupportNewTicketService,
  )
```
makes it easier to add injections imo :)

packages/manager/modules/support/src/new-ticket/new-ticket.service.js
```diff
@@ -0,0 +1,34 @@
+export default class SupportNewTicketService {
+  /* @ngInject */
+  constructor($translate, OvhApiMe, OvhApiSupport) {
+    this.$translate = $translate;
+    this.OvhApiMe = OvhApiMe;
+    this.OvhApiSupport = OvhApiSupport;
+  }
+
+  static get name() {
+    return 'SupportNewTicketService';
```
FredericEspiau: This is weird that the service knows it's own name, doesn't it ?

My way to tackle this is to 

````js
export const name = 'SupportNewTicketService';

export const definition = class SupportNewTicketService {}

export default {
  definition,
  name,
}
  
````

PR #1315, opened 2019-09-13
refactor(telecom): remove unused link

FredericEspiau:
MANAGER-2803


cbourgois:
https://github.com/ovh-ux/ovh-manager-telecom is not yet deprecated, so changes should by made in specific repository.
When merged, we should update the subtree of telecom manager in this repository.

PR #1327, opened 2019-09-18
fix: avoid displaying loader and data at the same time

jleveugle:
Internal ref: MBP-595


antleblanc:
What's about specifying the scope (for the Web application) such as:

```txt
fix(email.domain.account): improve loading display

avoid displaying loader and data at the same time.

Internal ref: MBP-595
```

Is that look like correct to you?

PR #1344, opened 2019-09-23
Update ESLint configurations

antleblanc:
# Update ESLint configurations

- Remove unused globals.
- Add both plugins `eslint-plugin-angular@^4.0.1` and 
`eslint-plugin-markdown@^1.0.0`.

## 👷 Build

4175fe8 - build(eslint): remove extra .eslintignore files
20c3761 - build(eslint): improve all configurations

## 🏠 Internal

- No quality check required.


.eslintrc.js
```diff
@@ -5,9 +5,14 @@ module.exports = {
   parser: 'babel-eslint',
   extends: 'airbnb-base',
   env: {
+    'angular/mocks': true,
     browser: true,
     jquery: true,
   },
+  plugins: [
+    'angular',
```
FredericEspiau: Maybe we could enable some rules by the way 😄 

https://www.npmjs.com/package/eslint-plugin-angular#rules

antleblanc: @FredericEspiau Sure!

Is there any rules that you want to be turned on?

If you have some, let's discuss about it;)

PR #1345, opened 2019-09-24
fix(group): remove debounce

antleblanc:
# Remove debounce

## :bug: Bug Fix

389ea44 - fix(group): remove debounce

## :link: Related

- ovh-ux/ovh-module-exchange#330

## :house: Internal

- No quality check required.


FredericEspiau:
Why though ? It was added at the time to ease the API and enhance the user experience

antleblanc:
@FredericEspiau It brings a infinite loop when using the search field.

The model is already debounce in the view (modelOptions).

PR #1346, opened 2019-09-24
feat(access.exchange.modem): add exchange for modem

Steffy29:
## ✨ Feature
feat(xdsl.access.exchange) : Add exchange button to exchange modem with last version

## 🏠 Internal

Ref: UXCT-187


packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.controller.js
```diff
@@ -0,0 +1,88 @@
...
+          rmas.forEach(id => this.OvhApiXdsl.RMA()
+            .v6()
+            .get(
+              {
+                xdslId: this.xdslId,
+              },
+              { id },
+            )
+            .$promise.then((rma) => {
+              const addRma = {
```
JeremyDec: I guess `rma` have more properties than you need to return in a new variable ? If no you could push directly `rma`

packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.html
```diff
@@ -0,0 +1,85 @@
+<div class="widget-presentation">
+
+    <h2 class="widget-presentation-title" data-translate="xdsl_access_comfort_exchange_title"></h2>
+
+    <tuc-toast-message></tuc-toast-message>
+
+    <div data-ng-if="$ctrl.isRMAOpened" class="mb-3">
+        <span class="row" data-translate="xdsl_access_comfort_exchange_already_opened_rma"></span>
```
JeremyDec: You do need `row`class for that `<span>` ? Maybe an helper could be useful (like ml-x)

packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.controller.js
Steffy29: yes it is. But I don't need to display all properties.

packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.html
Steffy29: yes, you're right.

packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.constant.js
```diff
@@ -0,0 +1,18 @@
+export const COMFORT_EXCHANGE_TYPE_ERROR = {
```
jleveugle: Why do we need this ?

packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.controller.js
```diff
@@ -0,0 +1,88 @@
...
+        priceWithTax: null,
+        tax: null,
+      },
+      isSuccess: false,
+    };
+    this.isAvailable = false;
+    this.getListOpenedRMA();
+  }
+
+  getListOpenedRMA() {
```
cbourgois: IMO, it should be in components bindings and injected via state resolves.

packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.constant.js
Steffy29: To map with error code returned by the API

jleveugle: Yes, I understood that, but why do you need an array to do that?

Maybe we can concat the error code with the translation directly, and if the translation doesn’t exist for an error, we can use a generic translation.

As example, you can compare the translated key and the result, if result === key, the translation doesn’t exist and you can use a fallback message.

If one day, we have a new kind of errors, we will avoid patching this array and the translation file. We will patch only one file.

What do you think about it?

Steffy29: as you want

Steffy29: sorry, I just use errBase and errContactShipping. Others error numbers are mapping with translations.

packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.html
```diff
@@ -0,0 +1,85 @@
...
+                <span class="col-sm-3 oui-tile__term" data-translate="xdsl_access_comfort_exchange_tax"></span>
+                <span class="col-sm-4 oui-tile__description" data-ng-bind="$ctrl.exchange.order.tax.text"></span>
+            </div>
+            <div class="row">
+                <span class="col-sm-3 oui-tile__term"
+                    data-translate="xdsl_access_comfort_exchange_price_with_tax"></span>
+                <span class="col-sm-4 oui-tile__description" data-ng-bind="$ctrl.exchange.order.priceWithTax.text"></span>
+            </div>
+
+            <div class="row">
```
antleblanc: Is this `row` really needed?

Because a `row` always have a `col-x-x` as a child which is not the case here.

ref: https://github.com/twbs/bootlint/wiki/E013

packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.service.js
```diff
@@ -0,0 +1,20 @@
...
+    return this.OvhApiXdsl.RMA()
+      .v6()
+      .query({
+        xdslId,
+      })
+      .$promise.then((rmas) => {
+        if (rmas.length > 0) {
+          return rmas;
+        }
+        return null;
```
frenautvh: i think it's a bad practice to return null if the length is zero. You expect the method to return an empty array if there are no rmas.

packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.controller.js
```diff
@@ -0,0 +1,84 @@
...
+      this.openedRMAs.forEach(id => this.OvhApiXdsl.RMA()
+        .v6()
+        .get(
+          {
+            xdslId: this.xdslId,
+          },
+          { id },
+        )
+        .$promise.then((rma) => {
+          const addRma = {
```
frenautvh: RmaToAdd

frenautvh: or simpler : this.rmas.push({ ... })

jleveugle:
Hello @Steffy29,
Anna Bot for OVH UX <3

packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.service.js
```diff
@@ -0,0 +1,15 @@
...
+    this.OvhApiXdsl = OvhApiXdsl;
+  }
+
+  getOpenedRMAs(xdslId) {
+    return this.OvhApiXdsl.RMA()
+      .v6()
+      .query({
+        xdslId,
+      })
+      .$promise.then(rmas => rmas);
```
FredericEspiau:
```suggestion
      .$promise;
```

packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.html
FredericEspiau: If you only have a `col-xx-12` under a `row`, you can safely remove both, this is equivalent to `class="d-block"` :)

packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.controller.js
FredericEspiau: Why was this resolved ? It's not resolved

Steffy29: I'm not agree with this remark. This method is called only if there is opened RMA, not every time we're going on the RMA page. So I didn't do in the resolve.

packages/manager/apps/telecom/src/app/telecom/pack/xdsl/access/comfortExchange/pack-xdsl-access-comfort-exchange.controller.js
```diff
@@ -0,0 +1,82 @@
...
+
+    if (this.openedRMAs) {
+      this.getListOpenedRMA();
+    }
+  }
+
+  getListOpenedRMA() {
+    if (this.openedRMAs.length > 0) {
+      this.rmas = [];
+      this.openedRMAs.forEach(id => this.OvhApiXdsl.RMA()
```
antleblanc: Could be turned into a `lodash/map` I think, instead of using a forEach + push.

PR #1351, opened 2019-09-24
Fix/dedicated import

cbourgois:
### Description of the Change

* bump `@ovh-ux/ng-ovh-web-universe-components` from `^4.0.3` to `^7.1.0`
* replace `@ovh-ux/ovh-utils-angular` by `@ovh-ux/ng-ovh-utils`
* remove `bootstrap-datetimepicker` dependency

dbf120a074 — build(dedicated): bump dependencies
0c22a131eb — fix(dedicated): update less rem
221dc526d9 — build(dedicated): remove unused dependency


packages/manager/apps/dedicated/client/app/css/less/common/buttonIcons.less
```diff
@@ -100,7 +100,7 @@
 
   &.icon-order-block,
   &.icon-order-bloc,
-  &.icon-purchase-upgrade, { .icon_position(8, 3); }
```
antleblanc: Good catch!

packages/manager/apps/dedicated/client/app/css/source.less
```diff
@@ -20,7 +17,7 @@
 @import "less/common/buttonIcons";
 @import "less/common/button";
 @import "less/common/countryIcon";
-@import "less/common/datetimepicker";
+// @import "less/common/datetimepicker";
```
antleblanc: Could be removed then?

cbourgois: I hope, I'm searching if it is used or not. If no, we could remove `bootstrap-datetimepicker` too https://github.com/ovh-ux/manager/blob/master/packages/manager/apps/dedicated/package.json#L87

cbourgois: Not used, removed.

PR #1356, opened 2019-09-25
build: refactoring build process

jleveugle:
Add internal build process instead of using lerna.

The main difference betweeen the 2 methods is that the new build script is using Worker Thread to split intensive build tasks in parallel.

### DIFF
--- a/packages/manager/apps/dedicated/webpack.config.js
+++ b/packages/manager/apps/dedicated/webpack.config.js
@@ -52,6 +52,8 @@ module.exports = (env = {}) => {
     },
   }, env);
 
+  const WEBPACK_REGION = `'${_.upperCase(env.region || process.env.REGION || 'EU')}'`;
+
   config.plugins.push(new webpack.DefinePlugin({
     WEBPACK_ENV: {
       region: JSON.stringify(env.region),
@@ -60,6 +62,7 @@ module.exports = (env = {}) => {
   }));
 
   // Extra config files
+  const extrasRegion = glob.sync(`./.extras-${WEBPACK_REGION}/**/*.js`);
   const extras = glob.sync('./.extras/**/*.js');
 
   return merge(config, {
@@ -71,7 +74,10 @@ module.exports = (env = {}) => {
       ]
         .concat(glob.sync('./client/app/**/*.module.js'))
         .concat(glob.sync('./client/app/components/**/!(*.module).js')),
-    }, bundles, extras.length > 0 ? { extras } : {}),
+    },
+    bundles,
+    extras.length > 0 ? { extras } : {},
+    extrasRegion.length > 0 ? { extrasRegion } : {}),
     output: {
       path: path.resolve(__dirname, 'dist'),
       filename: '[name].[chunkhash].bundle.js',
@@ -82,7 +88,7 @@ module.exports = (env = {}) => {
     plugins: [
       new webpack.DefinePlugin({
         __NG_APP_INJECTIONS__: process.env.NG_APP_INJECTIONS ? `'${process.env.NG_APP_INJECTIONS}'` : 'null',
-        __WEBPACK_REGION__: process.env.REGION ? `'${process.env.REGION.toUpperCase()}'` : '"EU"',
+        __WEBPACK_REGION__: WEBPACK_REGION,
       }),
     ],
   });
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -1,0 +1,105 @@
+#!/usr/bin/env node
+const execa = require('execa');
+const program = require('commander');
+const path = require('path');
+const remove = require('lodash/remove');
+const EventEmitter = require('events');
+const { Worker } = require('worker_threads');
+
+const myEmitter = new EventEmitter();
+
+let modules;
+let i = 0;
+
+async function retrieveDependencies(_modules) {
+  const packages = await Promise.all(
+    _modules.map(pck => execa
+      .shell(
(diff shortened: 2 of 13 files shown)


scripts/build.js
```diff
@@ -0,0 +1,107 @@
...
+
+const myEmitter = new EventEmitter();
+
+let modules;
+let i = 0;
+
+async function retrieveDependencies(_modules) {
+  const rtn = _modules;
+  const packages = await Promise.all(
+    rtn.map(pck => execa
```
marie-j: As `map` creates a copy, why don't you directly use `_.modules`

scripts/build.js
```diff
@@ -0,0 +1,107 @@
...
+    const deps = pck.dependents.filter(
+      dep => !modules.done.find(el => el.name === dep.name),
+    );
+
+    if (deps.length === 0) {
+      const workerData = remove(modules.todo, p => p.name === pck.name)[0];
+      modules.doing.push(workerData);
+
+      const promise = program.dryRun
+        ? new Promise(resolve => resolve())
```
marie-j:
```suggestion
        ? Promise.resolve()
```
?

scripts/build.js
```diff
@@ -0,0 +1,107 @@
...
+            console.log(
+              `Launching intensive build task for package ${workerData.name}`,
+            );
+          });
+          worker.on('message', (messageFromWorker) => {
+            console.log(messageFromWorker);
+          });
+          worker.on('error', reject);
+          worker.on('exit', (code) => {
+            if (code !== 0) {
```
marie-j:
```suggestion
            if (code) {
```
?

scripts/build.js
cbourgois: You mean `_modules`

scripts/build.js
```diff
@@ -0,0 +1,107 @@
...
+          p => p.name !== pck.name,
+        );
+        return Object.assign(pck, { dependents });
+      })),
+  );
+  return packages;
+}
+
+function unstack() {
+  console.log(`-=-=-=- #${i} -=-=-=-`);
```
cbourgois: `♪┏(・o･)┛♪┗ ( ･o･) ┓♪`

scripts/build.js
```diff
@@ -0,0 +1,107 @@
...
+  });
+  i += 1;
+}
+
+program
+  .option('--dry-run', 'Launch the build script without creating dist')
+  .option('-p, --package [package]', 'Scope build to a specific package and its dependencies')
+  .action(() => {
+    execa
+      .shell(`lerna list -alp --json --top-o-sort ${program.package ? `--scope=${program.package} --include-filtered-dependencies` : ''}`)
```
antleblanc: Regarding the package documentation, it's recommended to prefer using `execa()` instead of `execa.shell()`.

ref: https://github.com/sindresorhus/execa/tree/v1.0.0#execashellcommand-options

antleblanc: I cannot find the flag `--top-o-sort` for the `lerna list` command.

Shouldn't be [`toposort`](https://github.com/lerna/lerna/tree/master/commands/list#--toposort) instead?

scripts/build.js
```diff
@@ -0,0 +1,107 @@
...
+    );
+
+    if (deps.length === 0) {
+      const workerData = remove(modules.todo, p => p.name === pck.name)[0];
+      modules.doing.push(workerData);
+
+      const promise = program.dryRun
+        ? new Promise(resolve => resolve())
+        : new Promise((resolve, reject) => {
+          const worker = new Worker(path.join(__dirname, '/build_module.js'), {
```
antleblanc: Could the script be moved into a new directory (eg: scripts/worker)?

Because I find this quite confusing to have this script at the root of the `scripts` folder.

scripts/build_module.js
```diff
@@ -0,0 +1,6 @@
+const { parentPort, workerData } = require('worker_threads');
+const execa = require('execa');
+
+execa
+  .shell(`cd ${workerData.location} && npm run build --if-present`)
```
antleblanc: Recommended to prefer using `execa()`.

scripts/build.js
jleveugle: Except that here, I want to use shell command, not execute a file :/

`execa()` is used to execute a file :)

jleveugle: For toposort, this option is now useless in this, I removed it ;)

scripts/build.js
jleveugle: Ok, if you prefer, but we should keep in mind that this script can be executed in standalone, without any worker ;)

scripts/build_module.js
jleveugle: command here, i don’t use a file (and I can’t ? ^^)

packages/manager/apps/dedicated/webpack.config.js
```diff
@@ -52,6 +52,13 @@ module.exports = (env = {}) => {
     },
   }, env);
 
+  let WEBPACK_REGION;
```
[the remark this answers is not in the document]
jleveugle: done ;)

scripts/build.js
antleblanc: Nothing blocker but we might consider bump to the latest version of execa which drop the `.shell` method in flavor of `.command` instead.

I will open a pull request to address it later on.

ref: https://github.com/sindresorhus/execa/releases/tag/v2.0.0

PR #1357, opened 2019-09-25
fix(support): refactoring, fix some ux synchonization glitches

frenautvh:
Some refactoring & bugfix

refactoring : 
  - since a single issue is used to generate the creation form instead of a list of issues, replace the list of issues with a single one
  - refactor issue-selector by removing unused bindings and simplify controller
bugfix : 
  - use issueParams in order to prevent the issue-selector to be refreshed multiple times when category changed

### DIFF
--- a/packages/manager/modules/support/src/tickets/new-ticket/creation-form/creation-form.controller.js
+++ b/packages/manager/modules/support/src/tickets/new-ticket/creation-form/creation-form.controller.js
@@ -1,5 +1,4 @@
-import findLast from 'lodash/findLast';
-import last from 'lodash/last';
+import get from 'lodash/get';
 
 export default class SupportNewCreationFormController {
   /* @ngInject */
@@ -20,12 +19,7 @@ export default class SupportNewCreationFormController {
   }
 
   get subject() {
-    const issue = findLast(this.issues, 'subject');
-    return issue ? issue.subject : undefined;
-  }
-
-  getLastIssue() {
-    return last(this.issues);
+    return get(this.issue, 'subject');
   }
 
   shouldSelectUrgency() {
@@ -36,7 +30,7 @@ export default class SupportNewCreationFormController {
     this.onSubmit({
       result: {
         isSuccess,
-        issues: this.issues,
+        issue: this.issue,
         subject: this.subject || this.customSubject,
         urgency: this.urgency,
       },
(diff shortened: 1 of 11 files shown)


packages/manager/modules/support/src/tickets/new-ticket/creation-form/creation-form.controller.js
```diff
@@ -20,12 +19,7 @@ export default class SupportNewCreationFormController {
   }
 
   get subject() {
-    const issue = findLast(this.issues, 'subject');
-    return issue ? issue.subject : undefined;
-  }
-
-  getLastIssue() {
-    return last(this.issues);
+    return get(this.issue, 'subject');
```
JeremyDec: Since the binding shows that `issue: '<'` (issue can't be null), you can directly `return this.issue.subject` no ?

packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
```diff
@@ -41,26 +43,41 @@ export default class SupportNewIssuesFormController {
...
   }
 
   onCategoryChange() {
-    this.issues = null;
     this.serviceType = null;
-    this.fetchServices();
+    this.service = null;
+    this.isUnknownService = false;
+    this.issue = null;
+    if (this.category.id === 'account') {
```
JeremyDec: const for `account` ? (Also used in `html`)

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.service.js
```diff
@@ -29,6 +31,22 @@ export const definition = class SupportNewTicketService {
   getSupportLevel() {
     return this.OvhApiMe.v6().supportLevel().$promise;
   }
+
+  createTicket(issue, subject, serviceName, urgency) {
+    let body = '';
```
FredericEspiau:
```suggestion
    const body = `${issue.subject}\n
    
    ${join(
      map(
        issue.fields,
        (field) => `${field.label}\n${field.default}\n`
      ),
      '\n',
    )}
    \n
    `;
```

don't think this is necessarely better though

ovh-cds:
CDS Report create-unified-dist#5496.0 ■   
* Stage 1
  * Download ■

packages/manager/modules/support/src/tickets/new-ticket/creation-form/creation-form.controller.js
frenautvh: i don't think '<' garantes that the binding will be defined

PR #1362, opened 2019-09-25
fix(telecom.pack.xdsl.modem.port): send correct value to API

FredericEspiau:
In Lodash 3, you could do `pick(xxx, identity)` to remove empty values but with Lodash 4 you have to do `pickBy(xxx, identity)`

No other usages for the rest of the monorepo (except Dedicated but the prod version is not in the monorepo yet)

MANAGER-3355

### DIFF
--- a/packages/manager/apps/telecom/src/app/telecom/pack/xdsl/modem/router/ports/pack-xdsl-modem-port.factory.js
+++ b/packages/manager/apps/telecom/src/app/telecom/pack/xdsl/modem/router/ports/pack-xdsl-modem-port.factory.js
@@ -2,6 +2,7 @@ import assignIn from 'lodash/assignIn';
 import identity from 'lodash/identity';
 import isBoolean from 'lodash/isBoolean';
 import pick from 'lodash/pick';
+import pickBy from 'lodash/pickBy';
 import without from 'lodash/without';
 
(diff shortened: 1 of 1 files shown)


packages/manager/apps/telecom/src/app/telecom/pack/xdsl/modem/router/ports/pack-xdsl-modem-port.factory.js
```diff
@@ -1,7 +1,8 @@
 import assignIn from 'lodash/assignIn';
-import identity from 'lodash/identity';
 import isBoolean from 'lodash/isBoolean';
 import pick from 'lodash/pick';
+import pickBy from 'lodash/pickBy';
+import identity from 'lodash/identity';
```
cbourgois: a-z

frenautvh:
it seems that the lodash "pick" on line 30 should be replaced by pickBy aswell

cbourgois:
@frenautvh nope. When we uses `_.pick` with a function in `lodash@3`, it should be replace by `_.pickBy` in `lodash@4`.

cf : 
* https://github.com/lodash/lodash/wiki/Changelog#jan-12-2016--diff--docs 
* https://lodash.com/docs/3.10.1#pick 
* https://lodash.com/docs/4.17.15#pick

PR #1368, opened 2019-09-26
fix(hosting): add decimal for quota size calculation

JeremyDec:
MBE-412

### DIFF
--- a/packages/manager/apps/web/client/app/hosting/general-informations/GENERAL_INFORMATIONS.html
+++ b/packages/manager/apps/web/client/app/hosting/general-informations/GENERAL_INFORMATIONS.html
@@ -76,7 +76,7 @@
                                     aria-valuemax="100">
                                     <span
                                         class="oui-progress__label float-left"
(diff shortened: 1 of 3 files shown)


packages/manager/apps/web/client/app/hosting/general-informations/GENERAL_INFORMATIONS.html
```diff
@@ -76,15 +76,15 @@
                                     aria-valuemax="100">
                                     <span
                                         class="oui-progress__label float-left"
-                                        data-ng-bind="::convertBytesSize(hosting.quotaUsed.value, hosting.quotaUsed.unit, 2) + ' / ' + convertBytesSize(hosting.quotaSize.value, hosting.quotaSize.unit)"
+                                        data-ng-bind="::convertBytesSize(hosting.quotaUsed.value, hosting.quotaUsed.unit, 2) + ' / ' + convertBytesSize(hosting.quotaSize.value, hosting.quotaSize.unit, 2)"
```
jleveugle: It should be a variable instead of doing the same thing two times ^^

JeremyDec: done :)

PR #1373, opened 2019-09-26
fix(server-sidebar): add serviceName as item id

marie-j:
### Related 
DTRUN-3377


antleblanc:
@marie-j Shouldn't target `master` branch instead?

PR #1375, opened 2019-09-30
fix(core): remove extra angular-i18n imports

antleblanc:
# Remove extra i18n files

Using following command:

```sh
$ find . -type f -name '*Messages_nl_NL.json' -delete
```

## :bug: Bug Fix

32fd024 - fix(core): remove extra angular-i18n imports

## :recycle: Recycle

86fd5c3 - refactor: remove extra i18n files

## :heavy_check_mark: TODOs

- [ ] Edit some pipelines' configurations.

## :link: Related

- ovh-ux/manager#1253

## :question: Question

- Should we have to keep this (https://github.com/ovh-ux/manager/pull/1260/files) from the`manager-core` module?

## :house: Internal

- No quality check required.

cc @blaryjp


jleveugle:
> Should we have to keep this (ovh-ux/manager/pull/1260/files) from themanager-core module?

Yes! If a user has an old cookie with NL language, we need to set the default language to EN instead of FR, which is our default language.

To remove this code, we should switch our fallback language to EN :)

PR #1391, opened 2019-10-01
fix(emailpro): ngInject

frenautvh:
See : DTRUN-3614


antleblanc:
Commit message could be a bit more explicit.

Will be updated when using "**Squash and merge**" from GitHub UI.

antleblanc:
@frenautvh Is there any internal ticket reference related to this patch?

Thanks!

frenautvh:
@antleblanc DTRUN-3614, no internal JIRA for QC since the fix is really obvious here

PR #1402, opened 2019-10-02
fix(emailpro): remove duplicate entry into translation

JeremyDec:
DTRUN-3581 / MANAGER-3443


frenautvh:
if i'm not mistaken you cannot modify non french translations directly, you have to use the translator tool

JeremyDec:
@frenautvh ok even for a mistaken value ? Not for a text modification ?

frenautvh:
@JeremyDec i think your modifications will be overrided by the next translation "retrieve", you have fix the translation manually in the translator tool

JeremyDec:
Ok, seen with @antleblanc, translation process launched in order to fix this issue. Closing PR

PR #1405, opened 2019-10-03
feat(support): add atInternet tracking

frenautvh:


packages/manager/modules/support/src/index.js
```diff
@@ -1,5 +1,6 @@
 import angular from 'angular';
 
+import ngAtInternet from '@ovh-ux/ng-at-internet';
```
antleblanc: Importing peerDependencies in your modules should be done in that way:

```js
import '@ovh-ux/ng-at-internet';
```

It could also be applied for the rest.

ovh-cds:
CDS Report deploy_uxci_us#5733.0 ✘  
* deploy_uxci
  * deploy_uxci ✘

packages/manager/modules/support/src/index.js
FredericEspiau: It's my fault, don't do as I do @frenautvh  😄

frenautvh: Hello guys, i just added a commit to remove rollup and makes the main point to src/index.js. It's better that way because the dev building time is reduced a lot.

Now i don't think it's an issue to import peer dep like that since it ony was a problem when the build was targeting a commonJS with rollup. Using webpack it's safe.

What do you think?

antleblanc: `rollup.config.js` can also be removed then.

antleblanc: If this package turned as a private one it means that you cannot rely on the actual version of the Dedicated Control Panel (ovh-ux/ovh-manager-dedicated I mean).

It clearly depends on the time frame to be able to put the Dedicated Control Panel in production through the monorepo (just to keeping it in mind).

packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
```diff
@@ -113,6 +115,10 @@ export default class SupportNewIssuesFormController {
   }
 
   submitForm(isSuccess) {
+    this.atInternet.trackClick({
```
FredericEspiau: There is also a tracking when you arrive on the `/new` page, isn't it ?

packages/manager/modules/support/src/index.js
frenautvh: > `rollup.config.js` can also be removed then.

done thanks

PR #1420, opened 2019-10-07
feat(support): import translations

frenautvh:
import translations files of future support module


antleblanc:
@blaryjp Could you please confirm that translations can/should be submitted directly via the `feat/support` branch instead?

See: https://github.com/ovh-ux/manager/compare/develop...feat/support?expand=1

Thanks!

PR #1437, opened 2019-10-09
fix(pci.new): redirect to custom error page

marie-j:
DTRUN-3863


antleblanc:
**Internal ticket reference**: MANAGER-3731

PR #1440, opened 2019-10-10
fix(hosting.website-coach): update order funnel redirection

marie-j:
MANAGER-3497


cbourgois:
Should target `master` ?

antleblanc:
@cbourgois Since this ticket doesn't rely on the DTRUN project I think we can keep targeting `develop` branch instead.

PR #1459, opened 2019-10-14
feat(support): add translation for Carrier SIP

FredericEspiau:
I've also removed duplicated translations

Signed-off-by: Frédéric Espiau <frederic.espiau@corp.ovh.com>


jleveugle:
> I've also removed duplicated translations

It should be an another commit, like: 

refactor(support): remove unused translations
feat(support): add translation for Carrier SIP …

FredericEspiau:
This is done @jleveugle 😉

PR #1465, opened 2019-10-15
build: refer to the latest version of manager-carrier-sip module

antleblanc:
# Refer to the latest version of manager-carrier-sip module

0679d26 - build: refer to the latest version of manager-carrier-sip module

### DIFF
--- a/yarn.lock
+++ b/yarn.lock
@@ -43,6 +43,26 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
+"@babel/core@^7.1.2":
+  version "7.6.4"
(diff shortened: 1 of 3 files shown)


yarn.lock
```diff
@@ -43,6 +43,26 @@
     semver "^5.4.1"
     source-map "^0.5.0"
 
+"@babel/core@^7.1.2":
```
cbourgois: Should not be here.

antleblanc: Seems related to "multiple" version of `@ovh-ux/component-rollup-configuration`.

I will fix this in a separate Pull request.

Thanks for pointing out this.

antleblanc: Fixed by #1469

PR #1467, opened 2019-10-15
Feat/support feedback

FredericEspiau:


packages/manager/modules/support/src/tickets/new-ticket/new-ticket.html
```diff
@@ -69,3 +69,6 @@
                 data-on-click="$ctrl.goToTickets()">
     </oui-button>
 </oui-message>
+
+<support-tickets-new-feedback
```
frenautvh: why creating a component here? it looks super overkill, it's just a link

FredericEspiau: I was expecting to have the link at more than one place but everything is in one page 😅 

Then I thought it could be migrated to a full component if it's ever exported to its own

FredericEspiau: The component is now used at multiple places as it was really ugly to have the links below loading screen

packages/manager/modules/support/src/tickets/new-ticket/feedback/feedback.html
```diff
@@ -0,0 +1,5 @@
+<a
+    data-ng-href="{{ $ctrl.url }}"
+    rel="noopener"
+    target="_blank"
+    data-translate="support_tickets_new_feedback"></a>
```
marie-j: No icon or aria description to indicate it opens in a new window ?

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.html
```diff
@@ -52,6 +52,9 @@
        data-translate-values="{ state: 'support.tickets.ticket({ id: $ctrl.ticketId })' }"
        data-translate-compile>
     </p>
+
+    <support-tickets-new-feedback
+        data-language="$ctrl.TranslateService.getUserLocale(true)"></support-tickets-new-feedback>
```
marie-j: As you always get the `language` the same way, any reason why you don't do it in the component itself ?

packages/manager/modules/support/src/tickets/new-ticket/feedback/feedback.html
FredericEspiau: Shameful indeed !

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.html
FredericEspiau: It don't want it linked to the API, components should be as dumb as a rock 😄

PR #1468, opened 2019-10-15
feat(core): add locale header

marie-j:


packages/manager/modules/core/src/index.js
```diff
@@ -185,6 +187,9 @@ angular
     $httpProvider.interceptors.push('OvhSsoAuthInterceptor');
     $httpProvider.interceptors.push('OvhNgRequestTaggerInterceptor');
   })
+  .config(($httpProvider) => {
```
antleblanc: Explicit annotation could be added:

```suggestion
  .config(/* @ngInject */($httpProvider) => {
```

FredericEspiau:
Why ?

marie-j:
> Why ?

@FredericEspiau Locale is needed in a 2API but instead of passing a query param each time we will need it, we just need to get the header. I don't know if that makes more sense. 
A benefit from this we can imagine is we want API (or 2API) to sort something depending on user's language, we don't need to get and pass it each time, it is only done at one place and avoids duplication

antleblanc:
@marie-j Why not considering using this information by doing:

```txt
GET /me
{
  "language": "..."
}
```

FredericEspiau:
£is it not possible to use `Content-Language` instead ? It's the header intended for this usage

packages/manager/modules/core/src/translate/translate.constants.js
```diff
@@ -0,0 +1,5 @@
+export const HEADER_LOCALE = 'Content-Language';
```
FredericEspiau: Shouldn't we have a custom header instead ? It would give us more control !

FredericEspiau: (I'm kidding)

marie-j: 😭

PR #1470, opened 2019-10-15
Feat/support optional

FredericEspiau:


packages/manager/modules/support/src/tickets/new-ticket/creation-form/creation-form.controller.js
```diff
@@ -36,4 +40,16 @@ export default class SupportNewCreationFormController {
...
+
+  computeFieldLabel(field) {
+    return field.mandatory
+      ? field.label
+      : this.$translate.instant(
+        'ovhManagerSupport_new_creation_optional',
+        { label: field.label },
+        undefined,
+        false,
+        'sceParameters',
```
frenautvh: why sceParameters?

FredericEspiau: Fair question, I've added a comment to make it clearer, what do you think ? :)

PR #1474, opened 2019-10-15
fix(dedicated): update ip range

cbourgois:
### Description of the Change

ref: MANAGER-3753

1376c06164 — fix(dedicated): update ip range

### DIFF
--- a/packages/manager/apps/dedicated/client/app/ip/ip-range.service.js
+++ b/packages/manager/apps/dedicated/client/app/ip/ip-range.service.js
@@ -35,11 +35,10 @@ angular.module('Module.ip.services').service('IpRange', function IpRange() {
   function getStrCopy(str, _copies) {
     const copies = _copies > 0 ? _copies : 1;
 
-    return times(copies + 1, str).join('');
+    return times(copies, () => str).join('');
   }
 
-  function convertDecToBase(dec, base, _length, _padding) {
-    const padding = _padding || '0';
+  function convertDecToBase(dec, base, _length, padding = '0') {
     let num = dec.toString(base);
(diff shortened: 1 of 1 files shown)


packages/manager/apps/dedicated/client/app/ip/ip-range.service.js
```diff
@@ -35,11 +35,10 @@ angular.module('Module.ip.services').service('IpRange', function IpRange() {
   function getStrCopy(str, _copies) {
     const copies = _copies > 0 ? _copies : 1;
 
-    return times(copies + 1, str).join('');
+    return times(copies, () => str).join('');
```
FredericEspiau: What was the issue there ? Lodash conversion ?

packages/manager/apps/dedicated/client/app/ip/ip-range.service.js
```diff
@@ -35,11 +35,10 @@ angular.module('Module.ip.services').service('IpRange', function IpRange() {
...
   }
 
-  function convertDecToBase(dec, base, _length, _padding) {
-    const padding = _padding || '0';
+  function convertDecToBase(dec, base, _length, padding = 0) {
```
FredericEspiau:
```suggestion
  function convertDecToBase(dec, base, _length, padding = '0') {
```
according to the previous code 😄

packages/manager/apps/dedicated/client/app/ip/ip-range.service.js
cbourgois: Yes, related to Lodash conversion when importing the `ovh-manager-dedicated` repository into `manager` repository.

PR #1477, opened 2019-10-16
fix(web): fix hosting price display

varun257:
fix display of price while ordering new web hosting package

resolves MANAGER-3615

### DIFF
--- a/packages/manager/apps/web/client/app/domain/webhosting-enable/domain-enable-web-hosting.controller.js
+++ b/packages/manager/apps/web/client/app/domain/webhosting-enable/domain-enable-web-hosting.controller.js
@@ -188,7 +188,10 @@ angular.module('App').controller(
     }
 
     getResumePrice(price) {
-      return price.value === 0 ? this.$translate.instant('price_free') : this.$translate.instant('price_ht_label', { price: price.text });
+      // If price value is 0, the price is included, or else we display price
+      // Adding sceParameters will tell translator not to sanitize the value and trust as HTML
+      return price.value === 0 ? this.$translate.instant('price_free')
+        : this.$translate.instant('domain_hosting_price_ht_label', { price: price.text }, undefined, false, 'sceParameters');
     }
 
     getSelectedOfferOrderInfos() {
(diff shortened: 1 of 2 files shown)


packages/manager/apps/web/client/app/domain/webhosting-enable/domain-enable-web-hosting.controller.js
```diff
@@ -188,7 +188,7 @@ angular.module('App').controller(
     }
 
     getResumePrice(price) {
-      return price.value === 0 ? this.$translate.instant('price_free') : this.$translate.instant('price_ht_label', { price: price.text });
+      return price.value === 0 ? this.$translate.instant('price_free') : `${price.text} ${this.$translate.instant('price_ht_label')}`;
```
FredericEspiau: Could you replace the translation by `{{price}} HT` instead ?

varun257: Okay sure will do. I thought of using the translation which was already there :)

FredericEspiau: Just make sure it's not used elsewhere and that you are not breaking anything 😄

varun257: I tried it, the problem is the `"€"` that has to come between price` `and `HT`, even if i add a translation which will be only for `"HT"`, i will have to concatenate `price.text` and translation in the end

FredericEspiau: I don't get it, please tell me:
- what happens when you have a `{{price}} HT` translation ? Isn't the `€` already in the `price`text ?

varun257: Yes `€` is in the text but the problem is, its getting displayed as `&#8364;`. Its not getting sanitized.

varun257: This works! thanks

packages/manager/apps/web/client/app/domain/webhosting-enable/domain-enable-web-hosting.controller.js
```diff
@@ -188,7 +188,8 @@ angular.module('App').controller(
     }
 
     getResumePrice(price) {
-      return price.value === 0 ? this.$translate.instant('price_free') : `${price.text} ${this.$translate.instant('domain_hosting_price_ht_label')}`;
+      return price.value === 0 ? this.$translate.instant('price_free')
+        : this.$translate.instant('domain_hosting_price_ht_label', { price: price.text }, undefined, false, 'sceParameters');
```
FredericEspiau: Maybe you could add a comment to explain what you are doing here :)

PR #1125, opened 2019-07-26
feat(server-sidebar): enable vps section for US accounts

JeremyDec:
ref: MANAGER-3090

## :rocket: Feature
Enable VPS section for US accounts

PR #1481, opened 2019-10-16
fix(server.sidebar): hide vps section for US until prod of feature

JeremyDec:
ref : MANAGER-3074

Revert enabling vps section

Disable section will have to be reverted once the prod of vps feature will be possible (See : https://github.com/ovh-ux/ovh-manager-cloud/pull/1539)


antleblanc:
Revert: #1125

PR #1489, opened 2019-10-17
fix(dedicated): override width for large modals

marie-j:
ref: MANAGER-3751


marie-j:
Discussed altogether, will refactor and reopen 😉

PR #1503, opened 2019-10-18
fix: replace legacy autocomplete by oui-autocomplete

jleveugle:


packages/manager/apps/dedicated/client/app/dedicated/server/dns/add/dedicated-server-dns-add.html
```diff
@@ -21,25 +21,10 @@
...
-                                <ul>
-                                    <li class="type-off-item"
-                                        data-type-off-item="ip"
-                                        data-ng-repeat="ip in ipdata.t"
-                                        data-ng-bind="ip.name">
-                                    </li>
-                                </ul>
-                            </div>
-                        </div>
+                        <input id="ip" type="text" class="oui-input" ng-model="entry.ip" oui-autocomplete="ipdata.t" oui-autocomplete-property="name">
```
antleblanc:
```suggestion
                        <input id="ip" type="text" class="oui-input" data-ng-model="entry.ip" oui-autocomplete="ipdata.t" oui-autocomplete-property="name">
```

Because we tend to use `data-` prefix for all `ng-*` attributes.

PR #1515, opened 2019-10-21
fix: instance capabilities

ganeshkumar1989:
Selective features have to be enabled or disabled based on the instance capabilities, which is got from the API.

**No resize (capability name: "resize")**
- Instance Edit panel > Template (disable Modify Template)
**No snapshots (capability name: "snapshot")**
- Instance list view > ... > "Create a backup" (either disable link, or allow popup with button disabled)
- "Create a backup" link on instance detail view (either disable link, or allow popup with button disabled)
- "Create a backup" popup: disable button
- Instance list view > ... > "Create an automatic backup" (either disable link, or allow link to workflow management that will bi disabled)
- "Create an automatic backup" link on Instance detail view (either disable link, or allow link to workflow management that will bi disabled)
- Create a schedule page in Workflow management: when selecting a target resource: disable select for not allowed instances
**No volumes (capability name: "volume")**
- Instance detail view > Volumes > ... > Attach a volume (disable link or allow popup with disable message/button)
**No IPFO (capability name: "failoverip")**
- Failover IP > Modify instance > Do not allow to select ironic instances
- Failover IP > Actions > Buy Cloud failover IPs > Do not allow to select ironic instances

MANAGER-3754


packages/manager/modules/pci/src/projects/project/failover-ips/failover-ips.module.js
```diff
@@ -23,6 +24,7 @@ angular
   ])
   .config(routing)
   .component('pciProjectFailoverIps', component)
+  .service('PciProjectsProjectInstanceService', instanceService)
```
cbourgois: Already defined https://github.com/ovh/manager/blob/master/packages/manager/modules/pci/src/projects/project/instances/instances.module.js#L55

packages/manager/modules/pci/src/projects/project/failover-ips/order/order/order.controller.js
```diff
@@ -59,4 +60,8 @@ export default class FailoverIpController {
     this.$window.open(`${ORDER_URL}${JSURL.stringify([order])}`, '_blank');
     this.goBack();
   }
+
+  static isDisabledInstance(instance) {
```
FredericEspiau: You are repeating this function over and over, maybe you should make a service :)

marie-j: Even better, you should maybe add it to the `Instance` model :)

FredericEspiau: Wouldn't the instance do too much though ? 🤔

cbourgois: The `instance` may contain its capabilities or have an `InstanceCapabilities` service with a `getForInstance(instance)` method that returns an `InstanceCapabilities` object.

packages/manager/modules/pci/src/projects/project/failover-ips/failover-ips.module.js
ganeshkumar1989: This is fixed by adding the module as a dependency. Thanks :)

packages/manager/modules/pci/src/projects/project/failover-ips/order/order/order.controller.js
ganeshkumar1989: A new method `isCapabilityEnabled` has been added to the `Instance` class, since it is very closely related to an `Instance`. This also avoids the code complication of injecting a new service every required and assigning it to scope, if the `capabilities` is a part of a service.
Please let me know your thoughts if you think there is a better approach :)

packages/manager/modules/pci/src/projects/project/instances/instances.service.js
```diff
@@ -129,6 +135,12 @@ export default class PciProjectInstanceService {
       }));
   }
 
+  static transformCapabilities(capabilities) {
+    return reduce(capabilities, (flavorCapabilities, currentCapability) => set(
```
marie-j: I think we should avoid using `set` on an object when it's possible and instead return an updated copy of the original object. 
In that particular case it shouldn't have much side effects but in others it can have

packages/manager/modules/pci/src/projects/project/failover-ips/failover-ips.routing.js
```diff
@@ -64,15 +64,20 @@ export default /* @ngInject */ ($stateProvider) => {
...
           projectId,
         ) => OvhApiCloudProject
           .Instance()
           .v6()
           .query({
             serviceName: projectId,
           })
-          .$promise
+          .$promise.then(instances => $q.all(
+            map(instances, instance => PciProjectsProjectInstanceService
```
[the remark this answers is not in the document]
ganeshkumar1989: This is moved to service. Thanks :)

packages/manager/modules/pci/src/projects/project/instances/instance/edit/edit.controller.js
```diff
@@ -16,6 +17,7 @@ export default class PciInstanceEditController {
     this.CucCloudMessage = CucCloudMessage;
     this.CucRegionService = CucRegionService;
     this.PciProjectsProjectInstanceService = PciProjectsProjectInstanceService;
+    this.CAPABILITIES = CAPABILITIES;
```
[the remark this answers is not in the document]
ganeshkumar1989: This is done. That makes the code cleaner. Thanks!

packages/manager/modules/pci/src/projects/project/instances/instances.service.js
```diff
@@ -79,6 +80,7 @@ export default class PciProjectInstanceService {
             flavorId: instance.flavorId,
           })
           .$promise
+          .then(flavor => set(flavor, 'capabilities', this.constructor.transformCapabilities(get(flavor, 'capabilities', []))))
```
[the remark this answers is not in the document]
ganeshkumar1989: This is fixed

packages/manager/modules/pci/src/projects/project/instances/instances.service.js
ganeshkumar1989: The `set` call is removed and replaced by a new object return

PR #1525, opened 2019-10-22
feat(*): version number

radireddy:
send version number header in all API calls
#MANAGER-3757


packages/manager/apps/dedicated/client/app/app.js
```diff
@@ -240,4 +241,10 @@ angular
         });
       });
   })
-  .constant('UNIVERSE', 'DEDICATED');
+  .constant('UNIVERSE', 'DEDICATED')
+  .config(/* @ngInject */($httpProvider) => {
+    $httpProvider.interceptors.push('OvhNgRequestTaggerInterceptor');
```
marie-j: This is already done by `@ovh-ux/manager-core`

packages/manager/apps/dedicated/client/app/app.js
```diff
@@ -240,4 +241,10 @@ angular
...
+  })
+  .run(/* @ngInject */ (OvhNgRequestTaggerInterceptor) => {
+    OvhNgRequestTaggerInterceptor.setHeaderVersion(Environment.getVersion());
```
marie-j: Can this be handled by `CoreConfig` like the region ?

packages/manager/modules/request-tagger/src/factory.js
```diff
@@ -21,14 +27,18 @@ export default /* @ngInject */ () => {
...
+        const headerConfig = {
           ...config,
           headers: {
             ...config.headers,
             [HEADER_REQUEST_ID]: `${Date.now()}-${requestIndex}`,
             ...overridenHeaders,
           },
         };
+        if (!isEmpty(headerVersion)) {
+          headerConfig.headers[HEADER_VERSION] = `${headerVersion}`;
```
marie-j:
```suggestion
          headerConfig.headers[HEADER_VERSION] = headerVersion;
```

FredericEspiau: Hum I think this is useful to have `2.0` written as `2.0` (as a string) instead of just `2` (as a number)

packages/manager/apps/dedicated/client/app/app.js
cbourgois: https://github.com/ovh/manager/blob/master/packages/manager/modules/core/src/index.js#L186

packages/manager/apps/dedicated/client/app/app.js
cbourgois: the header version should be configured in `@ovh-ux/manager-core`

packages/manager/modules/core/src/index.js
```diff
@@ -32,6 +34,8 @@ import {
 
 const moduleName = 'ovhManagerCore';
 
+Environment.setVersion(__VERSION__);
```
cbourgois: Should be done in manager applications.

PR #1529, opened 2019-10-22
feat(support): access to ticket creation from url

FredericEspiau:
Makes it possible to access the form using a URL of this kind:
```
support/tickets/new?serviceTypeName=nameOfTheType&serviceName=nameOfMyService
```

### DIFF
--- a/packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
+++ b/packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
@@ -1,5 +1,6 @@
 import filter from 'lodash/filter';
 import get from 'lodash/get';
+import isString from 'lodash/isString';
 import orderBy from 'lodash/orderBy';
 
 const CATEGORY_ACCOUNT = 'account';
@@ -28,13 +29,15 @@ export default class SupportNewIssuesFormController {
   }
 
   $onInit() {
+    this.defaultServiceTypeName = isString(this.serviceTypeName)
+      ? this.serviceTypeName.toLowerCase()
+      : null;
+
+    this.currentServiceName = null;
     this.categories = null;
-    this.serviceTypes = null;
-    this.services = [];
     this.category = null;
-    this.serviceType = null;
     this.issue = null;
-    this.issueParams = null;
+    this.issueParams = {};
     this.isLoading = true;
     return this.$q.all({
       categories: this.SupportNewTicketService.getCategories(),
@@ -47,9 +50,23 @@ export default class SupportNewIssuesFormController {
         label: this.$translate.instant(`ovhManagerSupport_new_serviceType_${name}`),
       }));
       this.serviceTypes = orderBy(this.serviceTypes, [type => (type.label || '').toLowerCase()]);
-    }).finally(() => {
-      this.isLoading = false;
-    });
+
+      if (this.serviceTypes && this.defaultServiceTypeName) {
+        this.defaultServiceType = this.findServiceByName(this.defaultServiceTypeName);
+        this.serviceType = this.defaultServiceType;
+
+        this.updateIssueParamsPartially({
+          serviceType: this.serviceType,
+        });
+
+        return this.fetchServices();
+      }
+
+      return null;
+    })
+      .finally(() => {
+        this.isLoading = false;
+      });
   }
 
   get guides() {
@@ -60,34 +77,51 @@ export default class SupportNewIssuesFormController {
     return filter(get(this.issue, 'selfCareResources'), { type: 'tip' });
   }
 
+  findServiceByName(name) {
+    return this.serviceTypes
+      .find(type => type.name.toLowerCase() === name);
+  }
+
   onCategoryChange() {
-    this.serviceType = null;
-    this.service = null;
-    this.isUnknownService = false;
-    this.issue = null;
-    if (this.category.id === CATEGORY_ACCOUNT) {
-      this.issueParams = {
-        category: this.category,
-        serviceType: undefined,
-      };
-    } else {
-      this.issueParams = null;
+    if (!this.defaultServiceType && !this.serviceName) {
+      this.serviceType = null;
+      this.currentServiceName = null;
+      this.isUnknownService = false;
+      this.issue = null;
     }
+
+    this.updateIssueParamsPartially({
+      category: this.category,
+    });
   }
 
   onServiceTypeChange() {
-    this.service = null;
+    this.defaultServiceTypeName = null;
+    this.defaultServiceType = null;
+    this.serviceName = null;
+    this.currentServiceName = null;
     this.isUnknownService = false;
     this.issue = null;
-    this.issueParams = {
-      category: this.category,
+    this.updateIssueParamsPartially({
       serviceType: this.serviceType,
+    });
+
+    return this.fetchServices();
+  }
+
+  updateIssueParamsPartially({ category, serviceType }) {
+    this.issueParams = {
+      category: category
+        || this.issueParams.category,
+      serviceType: serviceType
+        || this.issueParams.serviceType
+        || this.defaultServiceType,
     };
-    this.fetchServices();
   }
 
   onServiceChange() {
-    if (this.service) {
+    this.serviceName = null;
+    if (this.currentServiceName) {
       this.isUnknownService = false;
     }
   }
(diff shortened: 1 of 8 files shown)


packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
```diff
@@ -111,9 +146,32 @@ export default class SupportNewIssuesFormController {
       external: false,
     }).$promise.then((items) => {
       this.services = items;
+
+      if (this.defaultServiceType != null) {
+        const inputService = this.services
+          .find(service => service.serviceName === this.defaultServiceName);
+
+        if (inputService == null) {
```
frenautvh: not sure it matters here but conventionally with use strict comparisons : ===

packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
```diff
@@ -28,13 +29,17 @@ export default class SupportNewIssuesFormController {
   }
 
   $onInit() {
+    this.defaultServiceName = this.serviceName;
```
frenautvh: what is the use of defaultServiceName? can't we just use a single serviceName var? if not please add a little comment for clarity because i find it a little bit confusing that we have two vars that describe the same value

packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
FredericEspiau: `x == null` is the same as `x === null && x === undefined` -> this is a useful way to make sure a variable is defined :)

frenautvh: i commonly use ```if (!inputService) { ... }``` but i guess either way is fine

packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
```diff
@@ -28,13 +29,17 @@ export default class SupportNewIssuesFormController {
...
+    this.serviceName = null;
```
marie-j: I think you shouldn't reset a bindings, I don't expect the binding I pass to have a different value inside and outside my component, maybe you should use another variable or add an alias in your bindings declaration to state that your internal name will be `this.defaultServiceName`
```suggestion
    serviceName: '<defaultServiceName'
```

packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
```diff
@@ -60,49 +79,65 @@ export default class SupportNewIssuesFormController {
     return filter(get(this.issue, 'selfCareResources'), { type: 'tip' });
   }
 
+  findServiceByName(name) {
+    return this.serviceTypes
+      .find(type => type.name.toLowerCase() === name);
```
marie-j: `type.name` is always defined ?

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.controller.js
```diff
@@ -68,4 +71,13 @@ export default class SupportNewController {
       this.issue = null;
     }
   }
+
+  handleBackButton() {
+    if (this.step !== 'creation') {
+      return this.$state.go('support.tickets');
+    }
+
+    this.step = 'issues';
```
marie-j: Could it be possible and would it make sense to handle it as states instead ?

packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
FredericEspiau: `this.serviceName` is the name of the service to get tickets fort
`this.defaultServiceName` the name that was given to the component (by the URL for example). This way, I know if the current name of the service has changed compared to what the user gave to the component

The component acts differently if you give it a  `serviceName` and `serviceTypeName`: if you select a `category` in the `category` select box, then the `serviceName` and `serviceTypeName` won't be reset ! If no `serviceName` and `serviceTypeName` are given, then they are both reset

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.html
```diff
@@ -1,10 +1,12 @@
-<oui-back-button data-state="support.tickets">
+<oui-back-button data-on-click="$ctrl.handleBackButton()">
```
marie-j:
```suggestion
<oui-back-button data-on-click="$ctrl.goBack()">
```
? The binding doesn't have to be aware if it's a back button or something else I think

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.html
```diff
@@ -19,38 +21,37 @@
...
+        </span>
+    </oui-message>
+
     <p data-translate="ovhManagerSupport_new_creation_tip_1"></p>
     <p data-ng-if="$ctrl.guideURL">
         <span data-translate="ovhManagerSupport_new_creation_tip_guides"></span>
         <a data-ng-href="{{ $ctrl.guideURL }}"
-           target="_blank"
-           data-ng-bind="$ctrl.guideURL">
+            target="_blank"
```
marie-j: No external icon nor `rel=noopener` ?

packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
FredericEspiau: I don't do it because `!value` matches `false`, `''`, '0', and so on

packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
FredericEspiau: Yes good point

packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
FredericEspiau: Yes, and if it's not , it crashes, both are wanted behaviors

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.controller.js
FredericEspiau: This wasn't done by me, and I don't have time redo entirely the component unfortunately, but yes, this should be done and it's actually in my todo list for v2 (which is already on the way 😉 )

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.html
FredericEspiau: I haven't touched this,  I think you are looking at whitespace changes 😄 

I'll fix this though !

ovh-cds:
CDS Report deploy_uxci_us#6399.0 ✘  
* deploy_uxci
  * deploy_uxci ✘

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.html
FredericEspiau: Done

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.html
```diff
@@ -19,38 +21,47 @@
...
+            <span class="sr-only" data-translate="ovhManagerSupport_new_open_new_tab"></span>
         </a>
     </p>
     <p data-ng-if="$ctrl.forumURL">
         <span data-translate="ovhManagerSupport_new_creation_tip_forum"></span>
-        <a data-ng-href="{{ $ctrl.forumURL }}"
-           target="_blank"
-           data-ng-bind="$ctrl.forumURL">
+        <a
+            class="oui-link oui-link_icon"
```
FredericEspiau: Done here too @marie-j 😅

packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
FredericEspiau: Done @marie-j

ovh-cds:
CDS Report deploy_uxci_eu#6522.0 ✘  
* deploy_uxci
  * deploy_uxci ✘

ovh-cds:
CDS Report deploy_uxci_ca#6522.0 ✘  
* deploy_uxci
  * deploy_uxci ✘

ovh-cds:
CDS Report deploy_uxci_us#6522.0 ✘  
* deploy_uxci
  * deploy_uxci ✘

PR #1536, opened 2019-10-23
Feat/prepare vps sidebar

JeremyDec:
ref: MANAGER_3074
## :rocket: Add vps order link
Add link to order vps for US accounts

### DIFF
--- a/packages/manager/modules/server-sidebar/src/order.constants.js
+++ b/packages/manager/modules/server-sidebar/src/order.constants.js
@@ -367,6 +367,9 @@ export const ORDER_URLS = {
     express_review_base: {
       US: 'https://us.ovhcloud.com/order/express/#/express/review',
     },
+    vps: {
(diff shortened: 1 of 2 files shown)


packages/manager/modules/server-sidebar/src/order.constants.js
```diff
@@ -439,7 +442,7 @@ export const SIDEBAR_ORDER_CONFIG = [
     target: '_blank',
     external: true,
     app: [CLOUD],
-    regions: ['EU', 'CA'],
+    regions: ['EU', 'CA', 'US'],
```
marie-j: Removing `regions` won't do ?

cbourgois: 👍 https://github.com/ovh/manager/blob/master/packages/manager/modules/server-sidebar/src/controller.js#L73-L80

JeremyDec: done

PR #1493, opened 2019-10-18
New release

antleblanc:
# 📦 New release

## ✨ Feature

- [x] UXCT-187 - feat(pack.xdsl): add access exchange for modem (#1346)

### Other feature

6e145fa - feat(config): use config settings to set some environment variables (#1457)
ae7ffff - feat(core): add translations for oui-criteria-adder filter (#1486)
ec808d3 - feat(telecom): use @ovh-ux/ng-ui-router-layout (#1490)
35aa33f - feat(support): submit translations (#1510)

## 🐛 Bug Fix

- [x] MANAGER-3233 - fix(dbaas.logs.detail.roles): add row size and pagination to roles (#1482)
- [x] MANAGER-3550 - fix(domain): add expiration date for dns anycast option (#1458)
- [x] MANAGER-3749 - fix(dedicated): sort styles import (#1519)
- [x] MANAGER-3751 - fix(dedicated): override width for large modals (#1494)
- [x] MANAGER-3782, MANAGER-3840, MANAGER-3838, MANAGER-3839 - fix(dedicated): add bootstrap4 grid and update rem (#1519)
- [x] MANAGER-3785 - fix: replace legacy autocomplete by oui-autocomplete (#1503)
- [x] MANAGER-3786 - fix(dedicated): update spacing (#1520)
- [x] MANAGER-3823 - fix: fast polling now works correctly (#1518)

### Other bugfix

feb2e36 - fix(sign-up): set OVHcloud logo (#1394)
ff34a6a - fix(dedicated): remove max-width for oui-message (#1498)
af0b5ce - fix: replace lodash imports (#1511)
7e23001 - fix(private.database): fix typo (#1522)

## ♻️ Refactor

- [x] MANAGER-3796 - refactor: remove unused calls & variables (csid, XSRFTOKEN) (#1504)

### Other refactor

406d32a - refactor: remove ovh-ui-angular duplicated translation configuration (#1486)

## ⬆️ Upgrade

3ea11fd - build: upgrade dependencies
17a0080 - build: upgrade ovh-ui-kit-bs
cc7b7f7 - build: bump ovh-ui-angular to 3.9.5

## 👷 Build

73eb9df - build(yarn): fix yarn.lock (conflict on bootstrap4 dependency)
774b2d4 - build(less): fix bad path
144c43f - build(dedicated): update region extras path (#1506)
1bd2121 - build: remove lodash from eslint globals (#1511)

## 📝 Documentation

[...]

PR #1537, opened 2019-10-23
fix managers on ie11 

jleveugle:
Downgrade punycode which is no more compatible with browser :)

So downgrade it to version 1.4.1


antleblanc:
It will fix the previous release (#1493) especially for Web and Telecom control panel.

Thanks!

PR #1587, opened 2019-10-25
fix(exchange): fix help button on modal

varun257:
close wizard help on close of modal

resolves MANAGER-3110

### DIFF
--- a/packages/manager/apps/web/client/app/css/less/common/wizard.less
+++ b/packages/manager/apps/web/client/app/css/less/common/wizard.less
@@ -21,6 +21,7 @@
   border-top-right-radius: 6px;
   font-size: 90%;
   left: 100%;
+  max-height: 0;
   margin-top: 1px;
   overflow-x: hidden;
   overflow-y: auto;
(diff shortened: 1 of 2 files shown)


packages/manager/apps/web/client/app/css/less/common/wizard.less
```diff
@@ -21,6 +21,7 @@
   border-top-right-radius: 6px;
   font-size: 90%;
   left: 100%;
+  max-height: 0;
```
marie-j: Maybe instead we should think of uniformizing with other modals/pages ? What do you think ?

varun257: This is used in a lot of places in `web`. I think we need to also add this component to `ui-kit` as well.

antleblanc: Uniformizing part that came from `ng-ovh-utils`, since it's has been deprecated, is not necessary.

We should probably found an other solution regarding how wizard could be turned maybe in a stepper form for instance. It will facilitate how help message could be displayed.

PR #1590, opened 2019-10-25
fix(deps): downgrade ovh-ui-angular to v3.9.0

antleblanc:
# Downgrade ovh-ui-angular to v3.9.0

## :ambulance: Hotfix

user can now scroll across header tabs items on small devices

c71751f - fix(deps): downgrade ovh-ui-angular to v3.9.0

## :link: Related

- ovh-ux/ovh-ui-angular#438

## :house: Internal

ref: MANAGER-3889

### DIFF
--- a/packages/manager/apps/telecom/package.json
+++ b/packages/manager/apps/telecom/package.json
@@ -135,7 +135,7 @@
     "ovh-manager-webfont": "^1.0.1",
     "ovh-ng-input-password": "^1.2.5",
     "ovh-ngstrap": "^4.0.2",
-    "ovh-ui-angular": "^3.9.5",
+    "ovh-ui-angular": "3.9.0",
     "ovh-ui-kit": "^2.35.2",
     "ovh-ui-kit-bs": "^2.2.0",
     "popper.js": "^1.16.0",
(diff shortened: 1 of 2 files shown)


ovh-cds:
CDS Report build-monorepository#6493.0 ✘  
* 
  * Build ✘

packages/manager/apps/telecom/package.json
```diff
@@ -135,7 +135,7 @@
     "ovh-manager-webfont": "^1.0.1",
     "ovh-ng-input-password": "^1.2.5",
     "ovh-ngstrap": "^4.0.2",
-    "ovh-ui-angular": "^3.9.5",
+    "ovh-ui-angular": "3.9.0",
```
marie-j: I'm not sure this is a good idea given we're in a angualr monorepository context

FredericEspiau: This requires testing yes :)

FredericEspiau: Yes I understand, hence testing is required :)

antleblanc: @marie-j I agree that it is not a ideal solution but I'm relaying this as a hotfix.

In fact both application "Public Cloud" and "Telecom" are impacted.

As a quick fix I prefer downgrade the version of `ovh-ui-angular` in order to be able to apply the right patch.

Why? Because version [`v3.9.1`](https://github.com/ovh-ux/ovh-ui-angular/blob/master/CHANGELOG.md#391-2019-10-10) introduce this regression.

antleblanc: Seems to be fixed by [`v3.9.8`](https://github.com/ovh-ux/ovh-ui-angular/blob/master/CHANGELOG.md#398-2019-10-24). So I close this Pull request.

PR #1635, opened 2019-10-29
Feat/miaw cdn order (2/2 API)

AxelPeter:
ref: MANAGER-3459

This is the second part that followed https://github.com/ovh/manager/pull/1488

- Update `hosting.cdn.order` to use only calls from `OrderService`.
- Update page layout to a stepper with 2 steps.
- Update translations to use `translate namespace`
- If the option is free, `pay` is replaced by `activate`. On the Dashboard, to avoid adding an umpteenth API call on the existing stack, I prefer using the existing informations from `Hosting`.


ovh-cds:
CDS Report build-monorepository#6715.0 ✘  
* 
  * Build ✘

PR #1650, opened 2019-10-30
fix(dedicated): fix purchase order for ca

varun257:
add condition to show generate purchase order button

resolves MANAGER-3904


marie-j:
Seems to be a duplicate of https://github.com/ovh/manager/pull/1651

PR #1651, opened 2019-10-30
fix(license.order): allow to order license if there is no contracts

marie-j:
ref: DTRSD-4714


marie-j:
Need to check with API team if this behaviour is legit

PR #1682, opened 2019-10-31
Feat/miaw cdn order (2/2 API)

AxelPeter:
ref: MANAGER-3459

This is the second part that follows https://github.com/ovh/manager/pull/1488

- Update `hosting.cdn.order` to use only calls from `OrderService`.
- Update page layout to a stepper with 2 steps.
- Update translations to use `translate namespace`
- If the option is free, `pay` is replaced by `activate`. On the Dashboard, to avoid adding an umpteenth API call on the existing stack, I prefer using the existing informations from `Hosting`.

### DIFF
--- a/packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
+++ b/packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
@@ -1,77 +1,49 @@
-import get from 'lodash/get';
-
 export default class {
   /* @ngInject */
-  constructor($q, $translate, $window, HostingOptionOrder, Alerter) {
-    this.$q = $q;
-    this.$translate = $translate;
-    this.$window = $window;
-    this.HostingOptionOrder = HostingOptionOrder;
-    this.Alerter = Alerter;
+  constructor($timeout) {
+    this.$timeout = $timeout;
   }
 
   $onInit() {
-    this.loading = true;
+    // Auto-select duration
+    [this.price] = this.catalogAddon.pricings;
+    this.interval = this.price.interval;
+    this.isEditable = true;
 
-    this.model = {};
-    this.durations = [];
-    this.details = {};
-
-    this.submitText = this.isOptionFree
-      ? this.$translate.instant('hosting_cdn_order_submit_activate')
-      : this.$translate.instant('hosting_cdn_order_submit_pay');
-
-    [this.model.offer] = this.availableOffers;
-    this.getDuration();
+    if (this.catalogAddon.pricings.length === 1) {
+      // Go directly to the next step
+      this.currentIndex = 1;
+      this.isEditable = false;
+    }
   }
 
-  getDuration() {
-    const queue = [];
-    this.loading = true;
+  resetCart() {
+    if (this.cart) {
+      this.cart = undefined;
+      this.cartId = undefined;
+    }
+  }
 
-    this.HostingOptionOrder.getOrderDurations('cdn', {
-      offer: get(this.model, 'offer.value'),
-    })
-      .then((durations) => {
-        this.durations = durations;
+  async prepareCheckout() {
+    if (!this.cart && !this.checkoutLoading) {
+      this.checkoutLoading = true;
+      const { cart, cartId } = await this.prepareOrderCart();
 
-        if (this.durations.length === 1) {
-          [this.model.duration] = this.durations;
-        }
-
-        angular.forEach(this.durations, (duration) => {
-          queue.push(this.HostingOptionOrder.getOrderDetailsForDuration('cdn', duration, {
-            offer: get(this.model, 'offer.value'),
-          }).then((details) => {
-            this.details[duration] = details;
-          }));
-        });
-
-        this.$q.all(queue);
-      })
-      .catch(error => this.goBack(
-        this.$translate.instant('hosting_dashboard_cdn_order_error', { message: get(error, 'data.message') }),
-        'danger',
-      ))
-      .finally(() => {
-        this.loading = false;
+      this.$timeout(() => {
+        this.cart = cart;
+        this.cartId = cartId;
+        this.checkoutLoading = false;
       });
+    }
   }
 
-  makeOrder() {
-    this.loading = true;
-    this.HostingOptionOrder.makeOrder('cdn', this.model.duration, {
-      offer: get(this.model, 'offer.value'),
-    })
-      .then((order) => {
-        this.$window.open(order.url, '_blank');
-        this.goBack(
-          this.$translate.instant('hosting_dashboard_cdn_order_success', { t0: order.url }),
-        );
-      })
-      .catch(error => this.goBack(
-        this.$translate.instant('hosting_dashboard_cdn_order_error', { message: get(error, 'data.message') }),
-        'danger',
-      ));
+  checkout() {
+    this.checkoutLoading = true;
+
+    this.checkoutOrderCart(
+      // Autovalidate if the order is free
+      this.isOptionFree ? this.isOptionFree : this.autoPayWithPreferredPaymentMethod,
+      this.cartId,
+    );
   }
 }
--- a/packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
+++ b/packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
@@ -1,6 +1,7 @@
 export default class {
   /* @ngInject */
-  constructor($timeout) {
+  constructor($filter, $timeout) {
+    this.$filter = $filter;
     this.$timeout = $timeout;
   }
 
@@ -45,5 +46,12 @@ export default class {
       this.isOptionFree ? this.isOptionFree : this.autoPayWithPreferredPaymentMethod,
       this.cartId,
     );
+  }
+
+  getDuration(_interval_) {
+    const interval = _interval_.toString();
+    const duration = this.$filter('wucDuration')(interval, 'longDate');
+
+    return duration;
   }
 }
--- a/packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.html
+++ b/packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.html
@@ -8,58 +8,61 @@
 </div>
 
 <div data-ng-if="!$ctrl.loading">
-    <oui-message
-        data-ng-if="!$ctrl.isOrderable"
-        data-type="error">
-        <p class="m-0" data-translate="hosting_dashboard_cdn_order_not_available"></p>
-    </oui-message>
+    <p data-translate="hosting_cdn_order_description"></p>
 
-    <form novalidate
-        name="cdnOrder"
-        data-ng-if="$ctrl.isOrderable"
-        data-ng-submit="$ctrl.makeOrder()">
-        <p data-translate="hosting_cdn_order_text"></p>
+    <oui-stepper
+        data-current-index="$ctrl.currentIndex"
+        data-on-finish="$ctrl.checkout()">
+        <oui-step-form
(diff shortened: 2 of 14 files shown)


ovh-cds:
CDS Report build-monorepository#6821.0 ✘  
* 
  * Build ✘

packages/manager/apps/web/client/app/hosting/cdn/order/translations/Messages_fr_FR.json
```diff
@@ -1,12 +1,14 @@
...
-  "hosting_cdn_order_duration": "Votre durée",
-  "hosting_cdn_order_duration_tooltip": "Regardless of where your new hosting package is set up, your domain name will continue to point to the current records. You will need to make the changes in your DNS zone when your new website is ready. This is an ideal choice if you do not want your current website to go down.",
-  "hosting_cdn_order_contracts": "Je reconnais avoir bien pris connaissance",
-  "hosting_cdn_order_contracts_of": "des",
-  "hosting_cdn_order_contracts_and": "et",
-  "hosting_cdn_order_submit_activate": "Activer",
-  "hosting_cdn_order_submit_pay": "Payer",
-  "hosting_cdn_order_cancel": "Annuler"
+  "hosting_cdn_order": {
+    "title": "Commander un CDN",
```
antleblanc: :warning: Caution with this namepace feature because I'm not :100:% sure that our actual memsource configuration is compatible with. 

@blaryjp What's your opinion regarding our actual `.memsourcerc.yaml` configuration?

Thanks!

packages/manager/apps/web/client/app/components/order-catalog-price/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,8 @@
+{
+  "price_free": "Inclus",
+  "price_tax_excl_label": "{{ price }} HT",
```
FredericEspiau: You forgot about the difference between `GST` and `VAT`

packages/manager/apps/web/client/app/components/order-contracts/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,7 @@
+{
+  "order_contracts": {
+    "label": "Je reconnais avoir bien pris connaissance",
+    "separator_and": "et",
+    "separator_of": "des"
```
FredericEspiau: Both are not translatable, please do otherwise :)

packages/manager/apps/web/client/app/components/order-payment-means/order-payment-means.html
```diff
@@ -0,0 +1,30 @@
+<div data-translate-namespace="order_payment_means">
+    <oui-field data-label="{{:: $ctrl.label }}">
+        <oui-radio
+            name="{{:: ctrl.name }}"
```
FredericEspiau:
```suggestion
            name="{{::$ ctrl.name }}"
```

packages/manager/apps/web/client/app/hosting/hosting.routes.js
```diff
@@ -1,24 +1,33 @@
-angular.module('App').config(($stateProvider) => {
+import controller from './hosting.controller';
+import template from './hosting.html';
+
+export default /* @ngInject */ ($stateProvider) => {
   $stateProvider.state('app.hosting', {
     url: '/configuration/hosting/:productId?tab',
-    templateUrl: 'hosting/hosting.html',
-    controller: 'HostingCtrl',
+    template,
```
FredericEspiau: Why didn't you use a component ?

packages/manager/apps/web/client/app/hosting/cdn/order/translations/Messages_fr_FR.json
blaryjp: You MUST NOT do that actually.

packages/manager/apps/web/client/app/hosting/hosting.routes.js
AxelPeter: Not refactored

packages/manager/apps/web/client/app/hosting/cdn/order/translations/Messages_fr_FR.json
AxelPeter: I give it a try, I'll go back to the old way ;)

packages/manager/apps/web/client/app/components/order-payment-means/order-payment-means.html
AxelPeter: Good catch, fixed

packages/manager/apps/web/client/app/components/order-contracts/translations/Messages_fr_FR.json
AxelPeter: Was done like that on another manager. I'll check another solution.
Have you an idea on your side ?

packages/manager/apps/web/client/app/components/order-catalog-price/translations/Messages_fr_FR.json
AxelPeter: GST / VAT on french translations ?

packages/manager/apps/web/client/app/components/order-contracts/translations/Messages_fr_FR.json
AxelPeter: Splitted in 4 more friendly translations

packages/manager/apps/web/client/app/hosting/cdn/order/translations/Messages_fr_FR.json
antleblanc: @AxelPeter @blaryjp I think we should give a try as soon as the configuration can handle it properly. I will create a task for this.

Thanks again for this kind of enhancements/suggestions!

packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.html
```diff
@@ -8,58 +8,61 @@
...
-        </oui-field>
+            <!-- Durations -->
+            <oui-field>
+                <oui-radio
+                    name="interval"
+                    data-ng-repeat="pricing in ::$ctrl.catalogAddon.pricings track by $index"
+                    data-model="$ctrl.interval"
+                    data-value="pricing.interval"
+                    data-required>
+                    <span data-ng-bind=":: '' + pricing.interval | wucDuration:'longDate'"></span>&nbsp;:&nbsp;
```
antleblanc: Why using concatenation? `'' + pricing.internal` I mean.

And using `&nbsp;:` is fine in French but it's not true in English for instance.
Maybe we can move it into a translation file, Isn't it?

antleblanc:
Hi @AxelPeter!

I didn't notice but since my previous approval, some files are now in conflicts.
Could you please resolve them?

Thanks!

packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.html
AxelPeter: `wucDuration` take a string as input, it's a quick way to convert this number value to a string.

For `:`, I don't think having a translation `{{duration}} :` is a good idea. 
The catalog price that follows is in a component. I don't see any other option than that. Have you a recommandation ?

packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
```diff
@@ -1,77 +1,49 @@
...
 
-        this.$q.all(queue);
-      })
-      .catch(error => this.goBack(
-        this.$translate.instant('hosting_dashboard_cdn_order_error', { message: get(error, 'data.message') }),
-        'danger',
-      ))
-      .finally(() => {
-        this.loading = false;
+      this.$timeout(() => {
```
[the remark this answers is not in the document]
AxelPeter: Both works, but I prefer this way to avoid using `$q` in my services. All that is needed here, is a $digest :)

AxelPeter:
@antleblanc A rebase was needed, fixed the conflicts ;)

AxelPeter:
Fix DCO check

packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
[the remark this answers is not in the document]
marie-j: I'm curious why :) Instead of adding a dependency to `$q` you add it to `$timeout`

packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.routing.js
```diff
@@ -1,43 +1,103 @@
+import get from 'lodash/get';
+
 export default /* @ngInject */ ($stateProvider) => {
   $stateProvider.state('app.hosting.cdn.order', {
     url: '/order',
     component: 'hostingCdnOrder',
     resolve: {
-      goBack: /* @ngInject */ goToHosting => goToHosting,
+      autoPayWithPreferredPaymentMethod: /* @ngInject */
```
marie-j: This sounds more like a function than a boolean

packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.routing.js
```diff
@@ -1,43 +1,103 @@
...
+            .checkoutOrderCart(autoPayWithPreferredPaymentMethod, cartId);
+
+          if (isOptionFree) {
+            goBack(
+              $translate.instant('hosting_dashboard_cdn_order_success_activation'),
+            );
+          } else {
+            $window.open(order.url, '_blank');
+            goBack(
+              $translate.instant('hosting_dashboard_cdn_order_success', { t0: order.url }),
```
marie-j: I wonder if we really need to try to go on this state to then go back on its parent

packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.routing.js
AxelPeter: It's the name as the parameter of the checkout API.
I've kept it unchanged

packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.routing.js
AxelPeter: I don't understand your comment. About what, the `try / catch` ?
Following the checkout state, we show the success/error message on the dashboard.
I've kept the behaviour of the refactored page for this one.

Were you talking about showing this message directly on the order page ?

packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.html
Jisay: This is specific to locale, so I think a translation is the best thing to do

antleblanc: Thanks for the clarification regarding `wucDuration` filter.

Related to the translation, because it is a  specification based on the locale, we do not have other option that providing it through a new translation.

packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.routing.js
marie-j: I think I might have misunderstood the code, my bad 😅

packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.html
AxelPeter: Ok, I'll check to put that in a translation :)

PR #1684, opened 2019-11-01
fix(telephony.line.last-registration): use last item as sortBy sorts ASC

julieni:
User last item in array, as loadash.sortBy sorts in ASC way


antleblanc:
**Internal ticket reference**: MANAGER-3933

PR #1712, opened 2019-11-05
fix(emailpro): check for space character in password

frenautvh:
DTRSD-3579

Re-open of https://github.com/ovh-ux/ovh-module-emailpro/pull/241


frenautvh:
updated QC : MANAGER-3300

PR #1721, opened 2019-11-06
fix(telephony-livecalls): sort calls and agents for consistent display

julieni:
Signed-off-by: Julien Issler <julien@issler.net>

Sorting the data will prevent flickering every second (BTW I think a refresh every 5 seconds is way enough and will save resources)


julieni:
Still "good first issue" ? ;)

antleblanc:
> Still "good first issue" ? ;)

github com_ovh_manager_pull_1721

The GitHub UI tell me that it's still a 'First-time contributor' :+1:

julieni:
Just kidding, doesn't matter :)

PR #1724, opened 2019-11-06
fix(legacy): fix challenge payment mean method

Jisay:
- DTRSD-5042
- MANAGER-3966


antleblanc:
Seen with Quality check team. Label isn't required.

PR #1730, opened 2019-11-07
fix(telephony): bulk actions

frenautvh:
See : DTRSD-5084

### DIFF
--- a/packages/manager/modules/telecom-universe-components/src/telecom/telephony/bulkAction/modal/telephony-bulk-action-modal.controller.js
+++ b/packages/manager/modules/telecom-universe-components/src/telecom/telephony/bulkAction/modal/telephony-bulk-action-modal.controller.js
@@ -1,4 +1,3 @@
-import bind from 'lodash/bind';
 import filter from 'lodash/filter';
 import find from 'lodash/find';
 import flatten from 'lodash/flatten';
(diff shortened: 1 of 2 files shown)


packages/manager/modules/telecom-universe-components/src/telecom/telephony/bulkAction/modal/telephony-bulk-action-modal.controller.js
```diff
@@ -101,7 +100,8 @@ export default /* @ngInject */ function (
   self.highlightUpdatedServices = function highlightUpdatedServices(services) {
     forEach(services, (service) => {
       forEach(self.billingAccounts, (billingAccount) => {
-        const findService = find(billingAccount.services, bind('serviceName', service.serviceName));
```
frenautvh: another misuse of a lodash function <3

antleblanc: Is the lodash migration to v4 that introduce this bug?

Previous code:
https://github.com/ovh-ux/ng-ovh-telecom-universe-components/commit/d41f5d6cf938454a9854c442b848d99919698950#diff-b379d719f556dad58b8a1008951fbe2eR85-R97

frenautvh: I don't know who wrote that and why, lodash "bind" is expecting a function as first argument but here it's a string so ... it doesn't work

PR #1748, opened 2019-11-08
fix(support): send extra data for Kubernetes

FredericEspiau:
# ⚠️ 
Please note that this is a temporary measure while waiting for the next version of the **Support** section

# 🏠 Internal reference
MANAGER-4004

### DIFF
--- a/packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
+++ b/packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
@@ -176,13 +176,26 @@ export default class SupportNewIssuesFormController {
       name: `answer-finding-${isSuccess ? 'yes' : 'no'}`,
       type: 'action',
     });
+
+    const service = {
+      ...this.currentServiceName,
+      serviceName: this.serviceType.name === 'kube'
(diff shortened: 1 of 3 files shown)


packages/manager/modules/support/src/tickets/new-ticket/issues-form/issues-form.controller.js
```diff
@@ -176,13 +176,26 @@ export default class SupportNewIssuesFormController {
...
+
+    const service = {
+      ...this.currentServiceName,
+      serviceName: this.serviceType.name === 'kube'
+        ? this.currentServiceName.url.match(/\/([a-f0-9]*)\//)[1]
+        : this.currentServiceName.serviceName,
+    };
+
+    const impactedResource = this.serviceType.name === 'kube'
+      ? this.currentServiceName.url.match(/\/((?:[a-f0-9]*-)+[a-f0-9]*)\//)[1]
```
frenautvh: fyi you also have the stateParams attributes that contains [projectId, kubeId], so it's easier to fetch the parameters from this array

FredericEspiau: Indeed 😄

FredericEspiau:
@frenautvh did it better

PR #1757, opened 2019-11-11
fix(cloud): content-language header

radireddy:
do not set content-language header for CORS requests

### DIFF
--- a/packages/manager/modules/core/src/translate/translate.factory.js
+++ b/packages/manager/modules/core/src/translate/translate.factory.js
@@ -1,13 +1,22 @@
+import find from 'lodash/find';
+import startsWith from 'lodash/startsWith';
+
 import {
   HEADER_LOCALE,
+  ROUTES_PREFIX,
 } from './translate.constants';
 
 export default /* @ngInject */ TranslateService => ({
-  request: config => ({
-    ...config,
-    headers: {
-      ...config.headers,
-      [HEADER_LOCALE]: TranslateService.getUserLocale(),
-    },
-  }),
+  request: (config) => {
+    if (find(ROUTES_PREFIX, route => startsWith(config.url, route))) {
+      return {
+        ...config,
+        headers: {
+          ...config.headers,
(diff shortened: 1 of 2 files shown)


packages/manager/modules/core/src/translate/translate.factory.js
```diff
@@ -1,13 +1,22 @@
...
-  }),
+  request: (config) => {
+    const headerConfig = {
+      ...config,
+      headers: {
+        ...config.headers,
+      },
+    };
+    // do not set content-language header for CORS requests
+    if (!startsWith(get(headerConfig, 'url'), 'https://')) {
```
cbourgois: Why not allowing the parameter only on trusted routes like we did for `request-tagger` (https://github.com/ovh/manager/blob/master/packages/manager/modules/request-tagger/src/factory.js#L23 && https://github.com/ovh/manager/blob/master/packages/manager/modules/request-tagger/src/constants.js#L8-L12) ?

packages/manager/modules/core/src/translate/translate.factory.js
```diff
@@ -1,13 +1,22 @@
...
 
 export default /* @ngInject */ TranslateService => ({
-  request: config => ({
-    ...config,
-    headers: {
-      ...config.headers,
-      [HEADER_LOCALE]: TranslateService.getUserLocale(),
-    },
-  }),
+  request: (config) => {
```
cbourgois: Avoid to recreate the `headerConfig` if not needed : 

```javascript
request: (config) => {
  if (!startsWith(get(headerConfig, 'url'), 'https://')) {
    return {
      headers: {	      
        ...config,
        headers: {
          ...config.headers,
          [HEADER_LOCALE]: TranslateService.getUserLocale(),
        },
      },
    };
  }
  return config;
}
```

cbourgois:
Maybe the scope of your commit should be `core`, and you should sign off your commit ;)

PR #1780, opened 2019-11-14
feat(pci): add black friday offer

frenautvh:
See : MANAGER-3932


packages/manager/modules/pci/src/projects/new/payment/payment.scss
```diff
@@ -35,5 +35,15 @@ pci-project-new-payment { /* stylelint-disable-line */
...
+
+    .black-friday {
+      color: black;
+      font-size: 14px;
+      .oui-status {
+        background-color: black;
+        color: white;
+        line-height: 1rem;
+      }
+    }
```
marie-j: I suppose there were no other choice than custom styles ?

packages/manager/modules/pci/src/projects/new/payment/payment.html
```diff
@@ -67,6 +67,13 @@
                         <span data-translate="pci_project_new_payment_use_credits"
                               data-translate-values="{ href: $ctrl.getStateLink('credits') }">
                         </span>
+                        <p class="black-friday mt-2">
```
marie-j: Maybe you could have created a reusable component as you duplicate 
Also how come you have a custom style for the new project but not for credits ?

packages/manager/modules/pci/src/projects/new/payment/payment.scss
antleblanc: Nope!
see: https://ovh-ux.github.io/ovh-ui-kit-documentation/#!/ovh-ui-kit/status

packages/manager/modules/pci/src/projects/new/payment/payment.html
frenautvh: credit is a child element of project that's why the style is only specified once

packages/manager/modules/pci/src/projects/new/payment/credits/credits.html
```diff
@@ -41,6 +41,13 @@
              data-translate="pci_projects_new_credits_amount_min"
              data-translate-values="{ amount: $ctrl.creditMinPrice.text }">
         </div>
+        <p class="black-friday mt-2">
+            <span class="oui-status mr-1"
+                  data-translate="pci_project_new_payment_black_friday">
+            </span>
+            <strong data-translate="pci_project_new_payment_black_friday_info">
```
FredericEspiau: Looks like this is displayed even if `$ctrl.isBlackFridayAsteriskVisible()` is `false` ?

Also, this is not a11y friendly, you could either:
- add a link to the end note (and a link to go back here 😄 )
- add the note as a `aria-label`
- have a link to the note as a `aria-describedby`

I think the best solution is the link, what do you think ? 😄

frenautvh: i used aria-describedby

packages/manager/modules/pci/src/projects/project/vouchers/vouchers.html
```diff
@@ -44,4 +45,5 @@
         </oui-column>
         <oui-column data-title="'cpb_vouchers_available_credit_cell' | translate" data-property="available_credit.text" data-sortable></oui-column>
     </oui-datagrid>
+    <pci-project-black-friday-notice class="mt-4 d-block"></pci-project-black-friday-notice>
```
antleblanc: Is the `d-block` helper really needed?

We tend to apply margin in the "same" direction.

So I would suggest you to add this margin directly on the `<oui-datagrid>` (`mb-4`)

PR #1786, opened 2019-11-15
fix(xdsl.migration.offer): update the sum for selected options

Steffy29:
Ref: UXCT-200

Signed-off-by: Stephanie Moallic <stephanie.moallic@corp.ovh.com>


packages/manager/apps/telecom/src/app/telecom/pack/migration/offers/telecom-pack-migration-offers.controller.js
```diff
@@ -3,69 +3,74 @@ import get from 'lodash/get';
...
-  self.process = null;
-  self.loading = {
-    init: false,
-  };
+  $onInit() {
+    this.process = null;
+    this.loading = {
+      init: false,
+    };
+    this.loading.init = true;
```
marie-j: You could set init directly to `true`

Steffy29: you're right

PR #1788, opened 2019-11-15
Feat/veeam

frenautvh:
ref: MANAGER-4017


ovh-cds:
CDS Report installation#7238.0 ■   
* Stage 1
  * Install ■

packages/manager/modules/veeam/src/template.html
```diff
@@ -1,3 +1,3 @@
-<div class="cui-page">
+<div class="cui-page veam">
```
antleblanc:
```suggestion
<div class="cui-page veeam">
```

packages/manager/modules/veeam/src/storage/add/veeam-storage-add.controller.js
```diff
@@ -0,0 +1,21 @@
+export default class VeeamStorageAddCtrl {
+  constructor($uibModalInstance, serviceName, VeeamService) {
```
antleblanc: Missing ngInject annotation:

```suggestion
  /* @ngInject */
  constructor($uibModalInstance, serviceName, VeeamService) {
```

antleblanc:
> Your README files are empty

I introduced a blank `README.md` file when you want to scaffold a new application/module because we used to copy/paste some other `README.md` but the content wasn't correctly aligned with the application/module itself (e.g. documentation references).

I'm actually working on a template for the `README.md` files for both modules and applications. So in the next weeks or so, theses empty `README.md` files will be updated in a separate Pull Request.

But here we are talking about private packages, so it's not a requirement.

Hope it's fine for you.

antleblanc:
Regarding the API that serve this product, I would prefer name it as `veeam-cloud-connect`.

Is that make sense for you?

frenautvh:
> Regarding the API that serve this product, I would prefer name it as `veeam-cloud-connect`.
> 
> Is that make sense for you?

renamed the module : https://github.com/ovh/manager/pull/1788/commits/abf7bf9f207cc652797c056ef79edf90f6509803

PR #1794, opened 2019-11-18
refactor(beta): create module to handle beta preferences

marie-j:


packages/manager/modules/beta-preference/package.json
```diff
@@ -0,0 +1,15 @@
+{
+  "name": "@ovh-ux/manager-beta-preference",
+  "version": "0.0.0",
```
antleblanc: Could be directly settled as:
```suggestion
  "version": "1.0.0",
```

packages/manager/modules/beta-preference/src/template.html
```diff
@@ -0,0 +1,12 @@
+<p data-translate="user_account_advanced_section_beta_description"></p>
+
+<div class="oui-box oui-box_light mb-5">
+    <oui-spinner data-ng-if="$ctrl.isUpdatingBeta" data-size="s"></oui-spinner>
+    <oui-checkbox
+        data-ng-if="!$ctrl.isUpdatingBeta"
+        aria-label=""
```
FredericEspiau: Is it necessary ?

marie-j: Auto generated with @antleblanc vscode-ovh-ui-kit-snippets and forgot to remove it 😉

antleblanc: Pay attention because all snippets are no more compatible with the ovh-ui-kit@2.0.0 (see: https://github.com/antleblanc/vscode-ovh-ui-kit-snippets/blob/master/CHANGELOG.md#200-2019-09-03).

PR #1796, opened 2019-11-19
Feat/pagination uirouter

marie-j:


packages/manager/modules/telecom-universe-components/src/list-layout-helper/index.js
```diff
@@ -0,0 +1,137 @@
+import get from 'lodash/get';
+
+import ListLayoutCtrl from './list-layout.controller';
+
+const FILTER_OPERATORS = {
```
antleblanc: My points is that I'm not 100% sure that this helpers are exclusively useful for the Telecom Universe Component.

Both `FILTER_OPERATORS` configuration and the corresponding `mapFilterForIceberg` relies on the usage of the Iceberg layer.

Iceberg and his helpers are located into [`@ovh-ux/ng-ovh-api-wrappers`](https://github.com/ovh-ux/ng-ovh-api-wrappers) so why not considering moving this configuration inside this dependency?

cbourgois: Let's focus on `telecom` manager for now, maybe we should refactor it after the main PR (not opened yet) will be merged if theses helpers/configs are needed in every manager.

marie-j: @antleblanc This is a compromise between avoiding duplication and providing a workaround soon enough 
This is a proposal and imo we should iterate in order to find the best technical solution but we have to start somewhere :) 
Also, it might not be the best thing to do but Telecom Universe component is a library and could be imported elsewhere. 
Moreover this is a mapping between oui-datagrid and Iceberg, if it belongs somewhere else, I'm not convinced it should be in @ovh-ux/ng-ovh-api-wrappers

FredericEspiau:
@marie-j for your information, the "containsNot" Iceberg filter will be prodded tomorrow, it's name will be `nlike` 😃

marie-j:
> @marie-j for your information, the "containsNot" Iceberg filter will be prodded tomorrow, it's name will be `nlike` 😃

@FredericEspiau I'll leave that way for now and I'll add it once the feature will be available

PR #1801, opened 2019-11-19
fix(vps): move cui-grid styles to cuc

cbourgois:
### Description of the Change

As `cui-grid` is used in `@ovh-ux/manager-vps`, styles from `cui-grid` should be in `@ovh-ux/ng-ovh-cloud-universe-components`

ce5ce69c34 — fix(vps): move cui-grid styles to cuc

### DIFF
--- a/packages/manager/modules/cloud-universe-components/src/cui/grid.less
+++ b/packages/manager/modules/cloud-universe-components/src/cui/grid.less
@@ -1,4 +1,5 @@
 @cui-grid-margin: 0.9375rem;
+@screen-sm-min: 801px;
 
 .cui-grid-cell  {
   padding: @cui-grid-margin / 2;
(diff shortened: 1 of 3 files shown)


packages/manager/modules/cloud-universe-components/src/cui/grid.less
```diff
@@ -1,4 +1,5 @@
 @cui-grid-margin: 0.9375rem;
+@screen-sm-min: 801px;
```
frenautvh: i'm just curious, why do we set screen-sm-min to 801px? is this legacy code ?

cbourgois: yes, it comes from https://github.com/ovh/manager/blob/master/packages/manager/apps/cloud/client/app/app.less#L43

ovh-cds:
CDS Report deploy_uxci_eu#7306.0 ✘  
* deploy_uxci
  * deploy_uxci ✘

PR #1806, opened 2019-11-19
Feat/freefax list

marie-j:


packages/manager/modules/freefax/src/freefaxes.controller.js
```diff
@@ -0,0 +1,23 @@
+import { ListLayoutHelper } from '@ovh-ux/ng-ovh-telecom-universe-components';
+
+export default class FreefaxesCtrl extends ListLayoutHelper.ListLayoutCtrl {
+  /* @ngInject */
+  constructor($q, ouiDatagridService) {
```
antleblanc: Both `$q` and `ouiDatagridService` are really needed?

I can't find any references.

packages/manager/modules/freefax/src/index.js
```diff
@@ -4,24 +4,38 @@ import 'oclazyload';
...
     'ovhManagerCore',
     'oc.lazyLoad',
+    freefax,
   ])
   .config(/* @ngInject */($stateProvider) => {
-    $stateProvider.state('freefax.**', {
-      url: '/freefax/:serviceName',
-      lazyLoad: ($transition$) => {
-        const $ocLazyLoad = $transition$.injector().get('$ocLazyLoad');
+    $stateProvider.state('freefaxes', {
```
antleblanc: Couldn't be moved into a `*.routing.js` file in order to keep consistency?

PR #1834, opened 2019-11-21
fix(telecom.telephony): hotfix to bypass API error when loading groups

julieni:
Hotfix against master to prevent loading error on groups when API fails

Call from manager: https://www.ovhtelecom.fr/engine/apiv7/telephony?$expand=1
Response, for some groups:
```{"1":{"key":"ovhtel-xxxxxxxxx-1","value":{"message":"Failed to bootstrap api"},"error":null}}```
the needed ```billingAccount``` property is not set and breaks the Javascript

Api error incident reported here: http://travaux.ovh.com/?do=details&id=41617

### DIFF
--- a/packages/manager/apps/telecom/src/components/telecom/telephony/telephony-voip-service.service.js
+++ b/packages/manager/apps/telecom/src/components/telecom/telephony/telephony-voip-service.service.js
@@ -13,7 +13,7 @@ angular.module('managerApp').service('TelephonyVoipService', function TelephonyV
     // fetch all billing accounts
     return OvhApiTelephony.v7().query().expand().execute().$promise.then((result) => {
       forEach(result, (item) => {
-        if (!item.error) { // how should we handle errors ?
+        if (!item.error && item.value.billingAccount) { // how should we handle errors ?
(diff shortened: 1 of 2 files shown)


antleblanc:
Hi @julieni!

Thanks for reporting this hotfix but an OVH Tasks were opened during this incident and the issue
is now fixed.

Could you please confirm that everything is back to normal?

Thanks!

julieni:
Hi @antleblanc 
indeed the issue is now fixed on the API side. But given the fact it lasted for 24 hours and prevented any action on my telephony panel (500 lines and aliases) for all this time, I think the workaround code I wrote is still to be merged, in case the API crashes again one day (hopefully not). Furthermore this code has very few impact, it's just 3 basic tests to prevent whole JS crashing.

FredericEspiau:
Makes sense, it it allows users to use some parts of the manager while the current way we do doesn't

julieni:
> Makes sense, it it allows users to use some parts of the manager while the current way we do doesn't

True. Broken billingAccounts on the API side are still unusable, but at least it doesn't crash the entire list. Thanks

PR #1840, opened 2019-11-22
fix(web.domain): prevent displaying error if it is due to contact rights

marie-j:
This error is used to know if we should display owner update link

ref: DTRSD-4511

### DIFF
--- a/packages/manager/apps/web/client/app/domain/general-informations/general-information.constants.js
+++ b/packages/manager/apps/web/client/app/domain/general-informations/general-information.constants.js
@@ -1,0 +1,5 @@
+export const OWNER_CHANGE_URL = 'useraccount/contact/';
+
+export default {
+  OWNER_CHANGE_URL,
+};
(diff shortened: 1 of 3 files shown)


packages/manager/apps/web/client/app/domain/general-informations/general-information.constants.js
```diff
@@ -0,0 +1,5 @@
+export const OWNER_CHANGE_URL = 'useraccount/contact/';
```
jleveugle: I think it’s always better to start with `/` and not finish by `/` :)

And be careful, because here `OWNER_CHANGE_URL` is not equals to what we had before:

`${OWNER_CHANGE_URL}${domain.name}/${domaine.whosisOwner.id}` is going to print something like:

`useraccount/contact/aaa/bbb` (relative hash)

instead of:

`#/useraccount/contact/aaa/bbb` (absolute hash)

marie-j: @jleveugle I agree that URL constants should be formatted the same way. I'll update that 

According to the second part of your comment, we join Dedicated manager prefix beforehand https://github.com/ovh/manager/pull/1840/files#diff-1114d6771d7a666f998677f99d379a3bR388 which already contains the hash

PR #1843, opened 2019-11-22
fix(nasha): fix max length eslint error

JeremyDec:
Signed-off-by: Jérémy De Césare <jeremy.de-cesare@corp.ovh.com>


JeremyDec:
Ok closing

PR #1849, opened 2019-11-25
fix: replace uib-btn-radio by oui-radio

ganeshkumar1989:
The uib-btn-radio component from bootstrap has been replaced by by oui-radio

MANAGER-4055

### DIFF
--- a/packages/manager/apps/telecom/src/components/telecom/telephony/scheduler/actions/bankHolidays/telephony-scheduler-bank-holidays.html
+++ b/packages/manager/apps/telecom/src/components/telecom/telephony/scheduler/actions/bankHolidays/telephony-scheduler-bank-holidays.html
@@ -42,15 +42,13 @@
                     <div class="btn-group-vertical w-100"
                          role="group"
                          data-translate-attr="{ 'aria-label': 'telephony_scheduler_bank_holidays_year_aria_label' }">
-                        <button class="btn btn-default"
-                                type="button"
-                                data-ng-class="{ active: SchedulerHolidaysCtrl.model.year === year }"
-                                data-ng-repeat="year in SchedulerHolidaysCtrl.yearList track by $index"
-                                data-ng-model="SchedulerHolidaysCtrl.model.year"
-                                data-uib-btn-radio="'{{ year }}'">
-                            <span data-ng-bind="year"></span>
-                            <span data-ng-bind="'(' + SchedulerHolidaysCtrl.filterEvents(year).length + ')'"></span>
-                        </button>
+                        <oui-radio-toggle-group class="year-list" data-model="SchedulerHolidaysCtrl.model.year">
+                            <oui-radio data-ng-repeat="year in SchedulerHolidaysCtrl.yearList track by $index"
+                                        data-value="year">
+                                <span data-ng-bind="year"></span>
+                                <span data-ng-bind="'(' + SchedulerHolidaysCtrl.filterEvents(year).length + ')'"></span>
+                            </oui-radio>
+                        </oui-radio-toggle-group>
                     </div>
                 </div> <!-- YEAR -->
             </div> <!-- "FILTERS" -->
--- a/packages/manager/apps/telecom/src/components/telecom/telephony/scheduler/telephony-scheduler.less
+++ b/packages/manager/apps/telecom/src/components/telecom/telephony/scheduler/telephony-scheduler.less
@@ -48,6 +48,22 @@
   }
 }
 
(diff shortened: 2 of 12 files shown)


packages/manager/apps/telecom/src/components/telecom/telephony/scheduler/telephony-scheduler.less
```diff
@@ -48,6 +48,22 @@
   }
 }
 
+.telecom-scheduler-bank-holidays {
+  .year-list .oui-radio-toggle-group {
+    flex-direction: column;
+
+    .oui-radio-toggle,
+ .oui-radio-toggle:first-of-type,
```
FredericEspiau: Identation

packages/manager/apps/telecom/src/components/telecom/telephony/scheduler/telephony-scheduler.less
```diff
@@ -48,6 +48,22 @@
...
+ .oui-radio-toggle:last-of-type {
```
FredericEspiau: Why is this needed?

packages/manager/apps/telecom/src/components/telecom/telephony/scheduler/actions/bankHolidays/telephony-scheduler-bank-holidays.html
```diff
@@ -42,15 +42,13 @@
...
-                        <button class="btn btn-default"
-                                type="button"
-                                data-ng-class="{ active: SchedulerHolidaysCtrl.model.year === year }"
-                                data-ng-repeat="year in SchedulerHolidaysCtrl.yearList track by $index"
-                                data-ng-model="SchedulerHolidaysCtrl.model.year"
-                                data-uib-btn-radio="'{{ year }}'">
-                            <span data-ng-bind="year"></span>
-                            <span data-ng-bind="'(' + SchedulerHolidaysCtrl.filterEvents(year).length + ')'"></span>
-                        </button>
+                        <oui-radio-toggle-group class="year-list" data-model="SchedulerHolidaysCtrl.model.year">
```
marie-j:
```suggestion
                        <oui-radio-toggle-group data-model="SchedulerHolidaysCtrl.model.year">
```
I don't see the class being used originally

packages/manager/apps/telecom/src/components/telecom/telephony/scheduler/telephony-scheduler.less
ganeshkumar1989: This is done

packages/manager/apps/telecom/src/components/telecom/telephony/scheduler/telephony-scheduler.less
ganeshkumar1989: The UI originally had the `uib-btn-radio` displayed as a column (meaning the buttons, which denotes the years like 2018, 2019 etc are displayed one below the other).  But the `oui-radio-toggle` component does not support this vertical orientation. Only horizontal orientation of the buttons is supported, due to which, the borders are not visible for the right side of the first button, both sides of the buttons in the middle and the left side of the last button. Here we changed the orientation of the buttons to vertical and hence we will have to have uniform borders for each of the buttons, which is what has been done :)

packages/manager/apps/telecom/src/components/telecom/telephony/scheduler/actions/bankHolidays/telephony-scheduler-bank-holidays.html
ganeshkumar1989: The UI originally had the `uib-btn-radio` displayed as a column (meaning the buttons, which denotes the years like 2018, 2019 etc are displayed one below the other).  But the `oui-radio-toggle` component does not support this vertical orientation. This class `year-list` has been added to change the orientation of the buttons to vertical and to handle some css changes, that have been explained in the comment above.

marie-j: Maybe instead of adding custom styles we could see if it makes sense to add that as an option to `oui-radio-toggle-group`.

ganeshkumar1989: Point taken. A new ticket is created for this: UK-125. Once done, the manager can be updated with the new option

PR #1853, opened 2019-11-25
fix(dedicated.billing): check engagement date is a past date

marie-j:
ref:  DTRSD-5758

### DIFF
--- a/packages/manager/apps/dedicated/client/app/models/BillingService.class.js
+++ b/packages/manager/apps/dedicated/client/app/models/BillingService.class.js
@@ -82,7 +82,7 @@ export default class BillingService {
   }
 
   hasEngagement() {
-    return !isNull(this.engagedUpTo);
+    return !isNull(this.engagedUpTo) && moment().isBefore(this.engagedUpTo);
   }
 
   setRenewPeriod(period) {


packages/manager/apps/dedicated/client/app/models/BillingService.class.js
```diff
@@ -82,7 +82,7 @@ export default class BillingService {
   }
 
   hasEngagement() {
-    return !isNull(this.engagedUpTo);
+    return !isNull(this.engagedUpTo) && moment().isBefore(this.engagedUpTo);
```
frenautvh: not a big deal but i think it's safer to use /auth/time api call to get the current time, otherwise we rely on the client's browser date settings

FredericEspiau: You are right, but it looks like a Manager-wide solution is needed for this :)

Also, this is not the scope of this issue

marie-j: @frenautvh indeed sounds like a better solution but as @FredericEspiau said we might need to think of a way to globally use /auth/time

PR #1858, opened 2019-11-27
fix(pci): black friday subsidiaries

frenautvh:
Display black friday depending on user's ovhSubsidiary


packages/manager/modules/pci/src/components/project/black-friday/black-friday-notice.component.js
```diff
@@ -1,5 +1,13 @@
 import template from './black-friday-notice.html';
 
 export default {
+  controller: /* @ngInject */ function ctrl(pciProjectBlackFridayService) {
```
marie-j: It would have been better in a separate file

PR #1864, opened 2019-11-27
fix(web): domain email delegation

frenautvh:
See : DTRSD-2653

### DIFF
--- a/packages/manager/apps/web/client/app/email-domain/delegate/email-domain-delegate.controller.js
+++ b/packages/manager/apps/web/client/app/email-domain/delegate/email-domain-delegate.controller.js
@@ -125,14 +125,16 @@ angular.module('App').controller(
       return this.$q
         .all({
           email: this.WucEmails.getDelegatedEmail(item),
-          usage: this.WucEmails.getEmailDelegatedUsage(item),
+          usage: this.WucEmails.getEmailDelegatedUsage(item).catch(() => null),
         })
         .then(({ email, usage }) => {
           const emailData = clone(email);
 
-          emailData.quota = usage.quota;
-          emailData.emailCount = usage.emailCount;
-          emailData.date = usage.date;
+          if (usage) {
+            emailData.quota = usage.quota;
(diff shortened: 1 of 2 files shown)


packages/manager/apps/web/client/app/email-domain/delegate/email-domain-delegate.controller.js
```diff
@@ -125,14 +125,16 @@ angular.module('App').controller(
       return this.$q
         .all({
           email: this.WucEmails.getDelegatedEmail(item),
-          usage: this.WucEmails.getEmailDelegatedUsage(item),
+          usage: this.WucEmails.getEmailDelegatedUsage(item).catch(() => null),
```
marie-j: This is the wanted behaviour for all API errors ? Even 500 for instance ?

FredericEspiau: @frenautvh yes ? maybe ? I don't know ? Can you repeat the question ?

frenautvh: hello @marie-j @FredericEspiau, my idea was that if the call to usage fail then we simply ignore the value and display a "-" in the datagrid. Do you want me to add an error message? if so, what kind of message : a single one if one or more usage call fail or for example one red pill in the datagrid for each failed call?

antleblanc: @frenautvh @marie-j @FredericEspiau here we are addressing a hotfix so I'm agree with the first suggestion that tends to display the `-` character inside the datagrid.

FredericEspiau: @antleblanc we don't want the remedy to be worse than the poison 😉 

Also, if it's a hotfix, just merge it already, a hotfix can't last for 21 days 😄

PR #1879, opened 2019-12-02
Feat/sms spain

marie-j:
ref: MANAGER-4073


FredericEspiau:
Are you sure there is not at some place an address or a phone number format to change ?

marie-j:
Translations are here : https://github.com/ovh/manager/pull/1923

antleblanc:
> Are you sure there is not at some place an address or a phone number format to change ?

We didn't identify any places within the SMS module.

ovh-cds:
CDS Report installation#8055.0 ■   
* Stage 1
  * Install ■

PR #1881, opened 2019-12-02
fix(server-sidebar): clear action menu before creating order menu

cbourgois:
### Description of the Change

ref: DTRSD-6091

b94ef21ddb — fix(server-sidebar): clear action menu before creating order menu


cbourgois:
In the case where the `server-sidebar` is added, removed and added a second time, the order menu is filled twice. 
To avoid it we clean the `actionsMenuOptions` before fill it.

PR #1336, opened 2019-09-21
feat: add prettier

antleblanc:
# Add Prettier

Goal is to reduce some code style comments in review.

## :sparkles: Feature

cb42a00 - feat: add prettier
dadb7a0 - feat: add stylelint-prettier

**Added following configuration**:

- [eslint-config-prettier@^6.3.0](https://www.npmjs.com/package/eslint-config-prettier).

**Added following plugins**:

- [@typescript-eslint/eslint-plugin@^2.3.0](https://www.npmjs.com/package/@typescript-eslint/eslint-plugin).
- [eslint-plugin-angular@^4.0.1](https://www.npmjs.com/package/eslint-plugin-angular).
- [eslint-plugin-import@^2.18.2](https://www.npmjs.com/package/eslint-plugin-import).
- [eslint-plugin-prettier@^3.1.1](https://www.npmjs.com/package/eslint-plugin-prettier).
- [stylelint-prettier@^1.1.2](https://www.npmjs.com/package/stylelint-prettier)

## :lipstick: Style

342cf8d - style: apply eslint rules
bf29044 - style: apply styelint rules
9695f61 - style: apply prettier to all HTML files

## :eyes: Examples

<details>
 <summary>Array of objects</summary>

```diff
-    this.actions = [{
-      name: 'compose_message',
-      sref: 'sms.service.sms.compose',
-      text: this.$translate.instant('sms_actions_send_sms'),
-    }, {
-      name: 'recredit_options',
-      sref: 'sms.service.order',
-      text: this.$translate.instant('sms_actions_credit_account'),
-    }, {
-      name: 'manage_recipient_new',
-      sref: 'sms.service.receivers',
-      text: this.$translate.instant('sms_actions_create_contact'),
-    }, {
-      name: 'manage_senders',
-      sref: 'sms.service.senders.add',
-      text: this.$translate.instant('sms_actions_create_sender'),
-    }, {
-      name: 'manage_soapi_users',
-      sref: 'sms.service.users',
-      text: this.$translate.instant('sms_actions_create_api_user'),
-    }, {
-      name: 'manage_blacklisted_senders',
-      sref: 'sms.service.receivers',
-      text: this.$translate.instant('sms_actions_clean_contact_list'),
-    }];
+    this.actions = [
+      {
+        name: 'compose_message',
+        sref: 'sms.service.sms.compose',
+        text: this.$translate.instant('sms_actions_send_sms'),
+      },
+      {
+        name: 'recredit_options',
[...]

PR #1892, opened 2019-12-03
feat(sign-up): improve sign-up app

Jisay:
# :notebook: Note:
- [x] : #1887 has to be merged first

# :house_with_garden: Internal:
- ECX-785
- ECX-786
- MANAGER-3648


packages/manager/modules/sign-up/src/form/details/details.controller.js
```diff
@@ -32,6 +32,29 @@ export default class SignUpDetailsCtrl {
         },
       },
     };
+
+    this.zipModel = {
+      value: null,
+      model: (...args) => {
```
frenautvh:
```
model: (zipModel) => {
  if (zipModel) {
    ...
```
?

packages/manager/apps/sign-up/src/incomplete/routing.js
```diff
@@ -0,0 +1,33 @@
...
+    views: {
+      '@': {
+        component: component.name,
+      },
+    },
+    redirectTo: (transition) => {
+      const mePromise = transition.injector().getAsync('me');
+      return mePromise.then(({ state }) => {
+        if (state !== NIC_STATUS_ENUM.INCOMPLETE) {
+          return transition.router.stateService.target('sign-up', {}, {
```
frenautvh: is there a simpler way of doing that ? like ```transition.redirect(...)``` ?

packages/manager/modules/sign-up/src/form/details/details.controller.js
Jisay: `zipModel` can be `null` and then never be setted

packages/manager/apps/sign-up/src/incomplete/routing.js
Jisay: I don't know. The difficulty here was to find a way to redirect without changing the location. This is the only way I found for doing it

packages/manager/modules/sign-up/src/form/details/details.html
```diff
@@ -10,7 +10,8 @@
                             class="oui-select__input"
                             data-ng-options="legalform.value as legalform.label for legalform in $ctrl.signUpFormCtrl.rules.legalform.in"
                             data-ng-model="$ctrl.signUpFormCtrl.model.legalform"
-                            data-ng-required="$ctrl.signUpFormCtrl.rules.legalform.mandatory">
+                            data-ng-required="$ctrl.signUpFormCtrl.rules.legalform.mandatory"
+                            data-ng-blur="$ctrl.signUpFormCtrl.onFieldBlur($ctrl.formCtrl.legalform)">
```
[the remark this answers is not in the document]
Jisay: This controller (signUpFormCtrl) is required by the component, so it is available through the current controller (https://docs.angularjs.org/guide/component#intercomponent-communication).
Maybe inheritance is an other solution, never tested.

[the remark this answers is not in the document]
FredericEspiau: Well it's a bit different:
- inheritance means your controller `IS` like another one
- composition means your controller `HAS` the same behaviour than another one, but they are not the same

ovh-cds:
CDS Report build-monorepository#7901.0 ✘  
* 
  * Build ✘

antleblanc:
Hi @Jisay!

We recently merged #1336 which apply Prettier rules so could you please rebase
your branch on `develop`?

To easily resolve conflicts you can **ignore** `HEAD` changes and only keep yours.

After that you can run:

```sh
$ yarn run lint
```

This command will lint the entire codebase.

or:

Amend your commit to trigger the `pre-commit` hook (prettier will be launched
under the hood).

Thanks!

packages/manager/apps/sign-up/src/incomplete/routing.js
marie-j: I think you can pass a `lcoation: false` if you use a `$state.go`, this will keep the current location

antleblanc:
@Jisay I've put the `blocked` label since we depends on updates on related tickets.

Feel free to resolve existings conflicts if needed. Thanks

PR #1903, opened 2019-12-04
Feat/contracts

marie-j:
Subtree from https://github.com/ovh-ux/ng-ovh-contracts


antleblanc:
Why not using [`lerna import`](https://github.com/lerna/lerna/blob/master/commands/import/README.md)?

marie-j:
> Why not using [`lerna import`](https://github.com/lerna/lerna/blob/master/commands/import/README.md)?
@antleblanc Why use it ? 😄 
A subtree was used to avoid having all the history brought into the repository as commits scope are not accurate anymore. 
Moreover I wonder if there's is an equivalent to `subtree pull`, I haven't find yet

antleblanc:
> @antleblanc Why use it ? 😄

Lerna provide this useful feature that's why I ask if it has been taking in account.

> A subtree was used to avoid having all the history brought into the repository as commits scope are not accurate anymore.

Especially for that reason. "(…) avoid having all the history (…)" is in my opinion a wrong path to follow because we will lack informations regarding how this module has evolve.

packages/components/ng-ovh-contracts/src/template.html
```diff
@@ -0,0 +1,101 @@
...
+                       data-ng-click="ContractsCtrl.scroll($index)"
+                       data-ng-class="{
+                            'active' : $index === ContractsCtrl.currentIndex
+                       }">
+                    </a>
+                </li>
+            </ul>
+            <button type="button" class="btn btn-link pull-right contracts-breadcrumb-navigate contracts-breadcrumb-navigate-next" data-ng-click="ContractsCtrl.next()" data-ng-disabled="!ContractsCtrl.hasNextElement()">
+                <span data-translate="ovh_contract_next"></span>
+                <i class="glyphicon glyphicon-chevron-right" aria-hidden="true"></i>
```
FredericEspiau:
```suggestion
                <span class="glyphicon glyphicon-chevron-right" aria-hidden="true"></span>
```

packages/components/ng-ovh-contracts/src/template.html
```diff
@@ -0,0 +1,101 @@
...
+            </button>
+        </div>
+        <a class="contracts-link"
+           title="{{ contract.name }}"
+           target="_blank"
+           rel="noopener"
+           data-ng-if="ContractsCtrl.currentContract.name"
+           data-ng-href="{{ ContractsCtrl.currentContract.url || ContractsCtrl.currentContract.pdf }}">
+            <i class="fa fa-file-pdf-o" aria-hidden="true"></i>
```
FredericEspiau:
```suggestion
            <span class="fa fa-file-pdf-o" aria-hidden="true"></span>
```

packages/components/ng-ovh-contracts/src/template.html
```diff
@@ -0,0 +1,101 @@
...
+                         data-ng-bind="contract.content || contract.text">
+                    </pre>
+                </uib-accordion-group>
+            </uib-accordion>
+        </div>
+        <div class="contracts-breadcrumb text-center clearfix"
+             role="navigation"
+             data-ng-show="ContractsCtrl.ovhContracts.length > 1">
+            <button type="button" class="btn btn-link pull-left contracts-breadcrumb-navigate contracts-breadcrumb-navigate-previous" data-ng-click="ContractsCtrl.previous()" data-ng-disabled="!ContractsCtrl.hasPreviousElement()">
+                <i class="glyphicon glyphicon-chevron-left" aria-hidden="true"></i>
```
FredericEspiau:
```suggestion
                <span class="glyphicon glyphicon-chevron-left" aria-hidden="true"></span>
```

packages/components/ng-ovh-contracts/src/template.html
```diff
@@ -0,0 +1,101 @@
...
+        <table class="table table-bordered"
+               data-ng-class="{
+                   'table-striped': ContractsCtrl.ovhContracts.length > 1
+               }">
+            <tr data-ng-repeat="contract in ContractsCtrl.ovhContracts track by $index">
+                <td>
+                    <a target="_blank"
+                       rel="noopener"
+                       data-ng-href="{{ contract.url || contract.pdf }}">
+                        <i class="fa fa-file-pdf-o" aria-hidden="true"></i>
```
FredericEspiau:
```suggestion
                        <span class="fa fa-file-pdf-o" aria-hidden="true"></span>
```

packages/components/ng-ovh-contracts/src/template.html
```diff
@@ -0,0 +1,101 @@
...
+                      data-translate="ovh_contract_agree">
+                </span>
+            </label>
+        </div>
+    </div>
+
+    <div data-ng-if="ContractsCtrl.fullText === true">
+        <p data-ng-if="ContractsCtrl.ovhContracts.length > 1"
+           data-translate="ovh_contracts_list_title"
+           data-translate-values="{ 'nbr': ContractsCtrl.ovhContracts.length }">
```
FredericEspiau:
```suggestion
           data-translate-values="{ 'number': ContractsCtrl.ovhContracts.length }">
```
is that really harder to write ? 😄

marie-j:
@FredericEspiau I've only `git subtree pull`

ovh-cds:
CDS Report build-monorepository#7785.0 ✘  
* 
  * Build ✘

cbourgois:
> > A subtree was used to avoid having all the history brought into the repository as commits scope are not accurate anymore.
> 
> Especially for that reason. "(…) avoid having all the history (…)" is in my opinion a wrong path to follow because we will lack informations regarding how this module has evolve.

I think that keeping commits like `:tractor: chore(*): update license and readme years to 2018.`(https://github.com/ovh-ux/ng-ovh-contracts/commit/790c8f2a0ec7cc7d9490ef6e08877069f7721362) is not something we want to keep (don't follow our commits rules).

If we really want to use `lerna import` instead of `git subtree` we have to update the documentation about it (https://github.com/ovh/manager/blob/master/DEVELOP.md#import-branch-from-legacy-repository)

antleblanc:
You're right to point out that some commits aren't aligned with our current guidelines.

It's is the same regarding the two commits needed in a subtree context and I don't know if it could be improved in some way.

But the whole history of `ng-ovh-contracts` component is correct even if conventions changed at some point.

Regarding the documentation you mentioned, I think it was valid because we were migrating some large applications `ovh-manager-(web|dedicated|cloud|telecom)` into the monorepo. It helped us to retrieve more easily some pending works (but also `ovh-module-(emailpro|exchange|office|sharepoint)`).

When it comes to bring an entire (existing) component I still recommend using the `lerna import` command.

marie-j:
@antleblanc we should probably discuss this as I don't really agree with the way `lerna import` is handling it. 
For now, I'll leave this as it is as it's the way we recommend it in https://github.com/ovh/manager/blob/master/DEVELOP.md#import-branch-from-legacy-repository

PR #1904, opened 2019-12-05
feat(wuc): add helper with engine version compare method

JeremyDec:
## 🚀 Feature
Add helper to sort engine version, see MANAGER-4121

### DIFF
--- a/packages/manager/modules/web-universe-components/src/helper/helper.service.js
+++ b/packages/manager/modules/web-universe-components/src/helper/helper.service.js
@@ -1,0 +1,26 @@
+export default class WucHelper {
+  /*
+   * Compare engine version based on Engine X.X format,
+   * from latest to oldest (by default)
+   */
+  static engineVersionCompare(engineA, engineB, splitChar = ' ', desc = true) {
+    const [engineALabel, engineAVersion] = engineA.split(splitChar);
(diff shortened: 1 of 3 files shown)


JeremyDec:
@frenautvh "engine" concerns for example cloud database server engines ("MySQL 5.7"). Ok it is generic, but I guess that not just Cloud db have server engines who have this version format. Maybe "serverEngineVersionCOmpare" would be clearer ?

I've created it in wuc, as this is used for Cloud db.

frenautvh:
> @frenautvh "engine" concerns for example cloud database server engines ("MySQL 5.7"). Ok it is generic, but I guess that not just Cloud db have server engines who have this version format. Maybe "serverEngineVersionCOmpare" would be clearer ?
> 
> I've created it in wuc, as this is used for Cloud db.

The naming should describe what the function does. In your case your function is sorting some data using a label then if it's equals use the version.

It's really a bad practice to create helper functions for such simple tasks

you should delete this code and simply use _.sortBy(data, ["engineLabel", "engineVersion"])

frenautvh:
```_.sortBy(data, ["engineLabel", o => parseFloat(o.engineVersion)])```

JeremyDec:
Ok doing

PR #1908, opened 2019-12-06
Feat/kill uib modal nasha

varun257:
Remove uib modal from nasha module


packages/manager/modules/nasha/src/partition/access/controller.js
```diff
@@ -8,165 +8,146 @@ export default class NashaPartitionAccessCtrl {
...
-        });
-    }
+  transformItem(access) {
+    return this.getAccessForIp(access.ip);
+  }
 
-    $scope.$on('$destroy', () => {
-      Poller.kill({ namespace: 'nasha.access' });
+  removeAccess(access) {
+    this.openModal('nasha/partition/access/delete/template.html', 'NashaPartitionAccessDeleteCtrl', {
```
marie-j: The purpose of the PR is to kill `uibModal` in NASHA, but it could also make sense to migrate this one as part of this scope

packages/manager/modules/nasha/src/partition/access/index.js
```diff
@@ -1,29 +1,34 @@
 import angular from 'angular';
 
 import routing from './routing';
-import NashaPartitionAccessCtrl from './controller';
-import NashaPartitionAccessAddCtrl from './add/controller';
-import NashaPartitionAccessDeleteCtrl from './delete/controller';
+import NashaPartitionAccessComponent from './access.component';
+import NashaPartitionAccessAdd from './add';
+import NashaPartitionAccessDelete from './delete';
+// import NashaPartitionAccessAddCtrl from './add/controller';
```
marie-j: Should be removed if not needed

PR #1910, opened 2019-12-09
Fix/sign up app

Jisay:
- no QC required


marie-j:
Commits are incorrectly signed off

PR #1949, opened 2019-12-13
fix: fix Analytics Data Platform order pipeline

Hugoch:
Current implementation of Analytics Data Platform pipeline wrongly implements service name check when ordering through Agora. We use a Poller to wait for Agora to return updated service name but check for a default of `*` instead of `*001`. Asynchronous process can lead to Manager progressing too fast compared to Agora updates and failing to deploy cluster.


antleblanc:
**Internal ticket reference**: MANAGER-4189

PR #2010, opened 2019-12-20
feat(telephony.alias.config): remove agent number edition in agents list

Steffy29:
## ✨ Feature
feat(telephony.alias.config): remove agent number edition in agents list

## 🏠 Internal

Ref: UXCT-205


frenautvh:
i think it should rather be a fix than a feat though

antleblanc:
> i think it should rather be a fix than a feat though

Will be done with a "Squash and merge"

PR #2011, opened 2019-12-20
Feat/mxplan

marie-j:
- MANAGER-4027
- MANAGER-4028

This is the reapplication of https://github.com/ovh/manager/pull/2003


packages/manager/apps/web/client/app/email-domain/order/email-domain-order.html
```diff
@@ -1,122 +1,99 @@
...
+                    data-ng-repeat="product in $ctrl.products track by product.planCode"
+                    data-label="{{ product.displayName }}"
+                    name="{{:: product.planCode }}"
+                    data-model="$ctrl.model.product"
+                    data-values="[product]"
+                    data-variant="light">
+                    <oui-select-picker-footer>
+                        <ovh-manager-order-catalog-price
+                            data-user="$ctrl.user"
+                            data-price="product.getPrice('installation')"
```
antleblanc: Shouldn't be:

```suggestion
                            data-price="$ctrl.model.product.getPrice('installation')"
```
?

packages/manager/modules/emailpro/src/mailing-list/emailpro-mailing-list-view.html
```diff
@@ -7,10 +7,11 @@
...
         <div>
             <p class="mb-3" data-translate="mailing_list_tab_table_no_quota"></p>
-            <button class="btn btn-info" type="button"
-                    data-translate="hosting_dashboard_service_change_offer"
-                    data-ng-click="setAction('hosting/offer/upgrade/hosting-offer-upgrade', hosting)">
-            </button>
+            <a
+                class="oui-button oui-button_secondary"
+                data-ng-href="#/configuration/email-domain/{{exchange.associatedDomainName}}/upgrade">
+                <span data-translate="emailpro_dashboard_accounts_upgrade_mxPlan"></span>
```
antleblanc: `translate` directive could be applied to the parent element.

packages/manager/apps/web/client/app/email-domain/order/EmailDomainOffer.class.js
```diff
@@ -0,0 +1,27 @@
...
+    });
+  }
+
+  getPrice(capacity) {
+    return get(find(this.prices, ({ capacities }) => includes(capacities, capacity)), 'priceInUcents');
+  }
+
+  get displayName() {
+    const [prefix, offer] = this.planCode.split('_');
+    return `${prefix.toUpperCase()} ${startCase(trimStart(offer, '0'))}`;
```
frenautvh: so sad we have to do stuff like this :/

packages/manager/apps/web/client/app/email-domain/order/email-domain-order.service.js
```diff
@@ -1,67 +1,33 @@
...
-     * @param $q
-     * @param OvhHttp
-     */
-    constructor($q, OvhHttp) {
-      this.$q = $q;
-      this.OvhHttp = OvhHttp;
-    }
+export default class MXPlan {
+  /* @ngInject */
+  constructor($q, OrderService) {
```
frenautvh: so wucOrderCartService?

packages/manager/apps/web/client/app/email-domain/order/order.routing.js
```diff
@@ -0,0 +1,78 @@
...
+export default /* @ngInject */ ($stateProvider) => {
+  $stateProvider.state('app.mx-plan', {
+    url: '/configuration/mx_plan?domain',
+    component: 'emailDomainOrder',
+    params: {
+      domain: null,
+    },
+    onExit: /* @ngInject */(cart, OrderService) => OrderService.deleteCart(cart),
+    resolve: {
+      addConfiguration: /* @ngInject */ (
```
frenautvh: There are lots of function defined in the routing like 'addConfiguration', 'deleteItem' ...
Imho i don't think they belong here, it should exist in the controller or if you need to reuse these, in a service.
Our routing is starting to look like a generic placeholder and the resolve part should only be used for routing related stuff or inputs passed to the component.

packages/manager/apps/web/client/app/email-domain/upgrade/upgrade.html
```diff
@@ -0,0 +1,78 @@
...
+            data-on-submit="$ctrl.displaySelectedOffer = true">
+            <p data-translate="email_domain_upgrade_step_header_offer_description"></p>
+            <div class="row">
+                <oui-select-picker
+                    data-ng-if="$ctrl.displaySelectedOffer"
+                    class="mb-3 col-md-4"
+                    data-label="{{ $ctrl.model.offer.displayName }}"
+                    name="{{:: $ctrl.model.offer.name }}"
+                    data-model="$ctrl.model.offer"
+                    data-values="[$ctrl.model.offer]"
```
frenautvh: is there only one offer to select? why using a select then?

packages/manager/apps/web/client/app/email-domain/upgrade/upgrade.html
```diff
@@ -0,0 +1,78 @@
...
+                    data-variant="light">
+                    <oui-select-picker-footer>
+                        <ovh-manager-order-catalog-price
+                            data-user="$ctrl.user"
+                            data-price="$ctrl.model.offer.duration.prices.withoutTax.value * 100000000"
```
frenautvh: it's a shame that order-catalog-price does not take a "price" from agora as a parameter and handle those things internally

packages/manager/apps/web/client/app/email-domain/order/email-domain-order.controller.js
```diff
@@ -1,176 +1,105 @@
...
+    this.loading = {
+      domain: false,
+      checkout: false,
+    };
+  }
+
+  getProductTaxAmount({ planCode }) {
+    const { plans } = this.catalog;
+    const pricings = get(find(plans, { planCode }), 'pricings');
+    const price = find(pricings, ({ capacities }) => includes(capacities, 'installation'));
```
frenautvh: We are making a lot of assumptions on UX side, like : the product price will be the price of the installation capability.

I know you're not responsible for that but i think it's a huge design flaw from the order api, we should receive on UX side a clear total price.

I even think it could backfire on us in the future (why the hell does UX compute the price of the product from the installation price only? and people will forgot that we had no choices from the begining...)

packages/manager/apps/web/client/app/email-domain/order/email-domain-order.service.js
marie-j: @frenautvh A refactor will be handled in another PR (here master is the target not `develop`

packages/manager/apps/web/client/app/email-domain/order/order.routing.js
marie-j: I don't agree with you. There are some examples when putting this stuff in the controller will lead to duplication. Let's says you have two data provider but want the same display. If you don't pass it as a binding (in the resolve's routing in our case) you'll need to duplicate the two components because you'll have to handle a controller by data provider

packages/manager/apps/web/client/app/email-domain/upgrade/upgrade.html
marie-j: This is the display when the `step-form` is validated to only display the selected offer, it is just for display purpose, when this occurs the user is not selecting the offer.
This is to keep harmonization with the way it is displayed when the steps is being edited. 
Moreover it's the way it's done in public-cloud so I think it's better to have the same experience everywhere

packages/manager/apps/web/client/app/email-domain/upgrade/upgrade.html
marie-j: This was done because of emergency, it will be refactored

packages/manager/apps/web/client/app/email-domain/order/order.routing.js
frenautvh: Then you can use a service if you want but there is no way that the routing file is the proper way to avoid code duplication.

Also i think it's wrong to wrap a simple function call in another function like that :
```
getCheckout: /* @ngInject */ (
        cart,
        OrderService,
      ) => () => OrderService.getCheckoutInformations(cart)
```
you are not factoring anything it's useless, the function you are calling is already coming from a service, if changes have to be made then the service will be updated

packages/manager/apps/web/client/app/email-domain/order/email-domain-order.html
marie-j: We're in a `ng-repeat`

packages/manager/apps/web/client/app/email-domain/order/order.routing.js
marie-j: A good example is the PCI containers https://github.com/ovh/manager/tree/master/packages/manager/modules/pci/src/projects/project/storages/containers
You have two different behaviours/API providers but the component is barely aware of it. It just worries about displaying stuff. 

I'm having trouble understanding how can an injected service could prevent duplication if you need different data providers

cbourgois: IMO, it's a good implementation. 
If we read the code, `cart` is never used in controller or template, because the main goal of this component is to provide a way to order a product whatever what should be used/done to achieve it. 
But yes, some `resolve` methods decorate service's methods but they also provide an easy way to order the product because It's not the component's responsibility to manage the cart.

Abstraction helps us to have reusable components.

frenautvh: i see what you mean, it's still possible to pass a service to the component, something like :
```
resolve : {
  containerApi: (MySpecificContainerApi) => MySpecificContainerApi,
}
```
or yes you can bind every function you need in the routing, maybe it's a matter of preferences

frenautvh: you just made me think that something like this is possible too :
```
resolve: {
  cartApi: (OrderService) => {
    addConfiguration: (cartId) => OrderService.addConfiguration(cartId),
    addItem: (itemId) => OrderService...,
    [...]
  },
}
```
that way we are creating a bridge to OrderService and it makes the component interface cleaner i think

i'm just throwing some ideas, no need to make changes

packages/manager/apps/web/client/app/email-domain/order/email-domain-order.service.js
marie-j: https://github.com/ovh/manager/pull/2013

packages/manager/apps/web/client/app/email-domain/upgrade/upgrade.html
marie-j: https://github.com/ovh/manager/pull/2013

FredericEspiau:
I'm out of the loop here, why does this target `master` ?

packages/manager/apps/web/client/app/email-domain/email-domain.html
```diff
@@ -10,21 +10,20 @@ <h1 class="word-break" data-ng-bind="domain.displayName"></h1>
...
+                  selected-tab="ctrlEmailDomainTabs.selectedTab"
+                  product="email"
+                  change-tab="ctrlEmailDomainTabs.setSelectedTab"
+                  tr="tr"
+                  menu="ctrlEmailDomainTabs.tabMenu"></wuc-ovh-tabs>
+            <div class="tab-content">
+                <div data-ovh-alert="{{alerts.tabs}}"></div>
+                <div class="tab-pane active" data-ui-view>
+                    <div
+                        data-ng-include="ctrlEmailDomainTabs.selectedTab && 'email-domain/' + ctrlEmailDomainTabs.constructor.toKebabCase(ctrlEmailDomainTabs.selectedTab) + '/email-domain-' + ctrlEmailDomainTabs.constructor.toKebabCase(ctrlEmailDomainTabs.selectedTab) + '.html'">
```
FredericEspiau: Goodness 😅

packages/manager/apps/web/client/app/email-domain/general-informations/email-domain-general-informations.controller.js
```diff
@@ -140,5 +136,11 @@ angular.module('App').controller(
           this.loading.urls = false;
         });
     }
+
+    goToUpgrade() {
+      this.$state.go('app.email.domain.upgrade', {
```
FredericEspiau:
```suggestion
      return this.$state.go('app.email.domain.upgrade', {
```

ovh-cds:
CDS Report build-monorepository#8404.0 ✘  
* 
  * Build ✘

packages/manager/apps/web/client/app/email-domain/upgrade/translations/Messages_fr_FR.json
```diff
@@ -1,18 +1,19 @@
...
   "email_domain_upgrade_contracts_accept": "Afin de confirmer votre demande, veuillez accepter les conditions d'utilisation.",
-  "email_domain_upgrade_in_progress": "Modification de l'offre de messagerie électronique en '{{ newOffer }}'",
+  "email_domain_upgrade_in_progress": "Modification de l'offre e-mail en '{{ newOffer }}'",
   "email_domain_upgrade_fetch_duration_error": "Une erreur s'est produite lors de l'obtention des détails de l'offre : {{message}}",
-  "email_domain_upgrade_error": "Impossible de modifier l'offre de messagerie : {{message}}",
-  "email_domain_upgrade_success": "Le changement d'offre par e-mail est en cours.",
-  "email_domain_upgrade_success_follow": "<a href=\"{{ orderTrackURL }}\" target=\"_blank\" rel=\"noopener\"> Cliquez ici pour suivre la progression.</a>"
+  "email_domain_upgrade_error": "Impossible de modifier l'offre e-mail : {{message}}",
+  "email_domain_upgrade_success": "Le bon de commande <a href=\"{{orderUrl}}\" target=\"_blank\">{{orderId}}</a> a été créé avec succès.",
+  "email_domain_upgrade_price_without_tax": "Prix HT : ",
```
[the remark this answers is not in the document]
marie-j: Won't fit with the current display unfortunately but might not fit with all translations I agree

ovh-cds:
CDS Report deploy_uxci_eu#8507.0 ✘  
* deploy_uxci
  * deploy_uxci ✘

ovh-cds:
CDS Report deploy_uxci_eu#8507.1 ✘  
* deploy_uxci
  * deploy_uxci ✘

ovh-cds:
CDS Report deploy_uxci_eu#8507.2 ✘  
* deploy_uxci
  * deploy_uxci ✘

PR #2012, opened 2019-12-20
refactor(email-domain): share upgrade component 

marie-j:
ref: MANAGER-4224
ref: MANAGER-4225


packages/manager/modules/emailpro/src/emailpro.constants.js
```diff
@@ -1,3 +1,5 @@
+export const EMAILPRO_MX_UPGRADE = '#/configuration/email-domain/{{ domain }}/upgrade';
```
frenautvh: i think it would be cleaner to export a function instead of a string to interpolate, i don't like the idea that the caller has to perform the interpolation
```javascript
export const EMAILPRO_MX_UPGRADE = (domain) => '#/configuration/email-domain/_domain_/upgrade'.replace('_domain_', domain);
```

packages/manager/modules/emailpro/src/emailpro.html
```diff
@@ -113,8 +113,10 @@ <h1 class="d-inline-block mb-0" data-ng-bind="exchange.displayName"></h1>
                         data-ng-if="messageDetails.length > 0"></strong>
             </div>
         </div>
-        <div class="tab-pane active platform"
-             data-ng-include="selectedTab && 'emailpro/' + kebabCase(selectedTab) + '/emailpro-' + kebabCase(selectedTab) + '.html'"></div>
+        <div ui-view>
```
antleblanc: To keep consistency:
```suggestion
        <div data-ui-view>
```

PR #2013, opened 2019-12-20
Refactor/miaw mxplan

marie-j:


packages/manager/apps/web/client/app/components/manager-order-catalog-price/manager-order-catalog-price.controller.js
```diff
@@ -22,6 +22,12 @@ export default class {
       this.block = true;
     }
 
+    if (!isUndefined(this.$attrs.convertToUcents) && this.$attrs.convertToUcents === '') {
```
FredericEspiau: I don't get this ?

If it's defined and the value is empty ? I feel like booleans would make things easier to write

`<my-component convert-to-ucent="{{ $ctrl.shouldConvertToUcents }}"></my-component>`

For consistency with HTML5 you would also have `convert-to-ucent="convert-to-ucent"` but who wants that, really

marie-j: This is to act like `disabled` or `required`, you don't have to pass any value if you want this attribute always active

FredericEspiau: And if I want to disable it dynamically I should do this

`<my-component convert-to-ucent="{{ $ctrl.shouldConvertToUcents && 'thisIsDisabled' }}"></my-component>`

😢

marie-j: Ok I understand what you mean now 😅

marie-j: @FredericEspiau I've now updated so both syntax can be used, you can have 
```suggestion
    <my-component convert-to-ucent></my-component>
```
or 
```suggestion
    <my-component convert-to-ucent="{{ $ctrl.shouldConvertToUcents }}"></my-component>
```

FredericEspiau: Genius

PR #2016, opened 2019-12-23
fix(dedicated): add email change validation token request action

frenautvh:
See : DTRSD-6593

Issue :
Currently when a user wants to change his e-mail, a token is sent to his previous e-mail address. If the e-mail is not received, the user has no way in the manager to send the validation token again and so he is stuck.

Fix :
Add a button to send a new validation token.


packages/manager/apps/dedicated/client/app/account/user/infos/user-infos.controller.js
```diff
@@ -212,6 +214,22 @@ angular.module('UserAccount').controller('UserAccount.controllers.Infos', [
...
     };
 
+    $scope.requestChangeEmailToken = function requestChangeEmailToken(email) {
+      $scope.requestingToken = true;
+      return UseraccountInfos.changeEmail(email)
+        .then(({ id }) => {
+          $scope.controls.taskEmailChangeTodo.id = id;
+          // reset requestingToken after a long delay to avoid the user
+          // being tempted to make too many token requests
+          $timeout(() => { $scope.requestingToken = false; }, 60000);
```
FredericEspiau: This should be handled by the API, not by us

packages/manager/apps/dedicated/client/app/account/user/translations/Messages_fr_FR.json
```diff
@@ -395,6 +395,9 @@
   "useraccount_contacts_change_contact_valid": "Valider",
   "user_account_email_pending_validation": "{{t0}} est en cours de validation.",
   "user_account_email_token_received": "J’ai reçu mon code de validation.",
+  "user_account_email_token_resend": "Renvoyer un code de validation",
+  "user_account_email_token_resend_success": "Un nouveau code de validation vient dêtre envoyé. Veuillez noter qu'un délai de quelques minutes peut être nécessaire avant réception de celui-ci.",
```
FredericEspiau:
```suggestion
  "user_account_email_token_resend_success": "Un nouveau code de validation vient d'être envoyé. Veuillez noter qu'un délai de quelques minutes peut être nécessaire avant réception de celui-ci.",
```

packages/manager/apps/dedicated/client/app/account/user/infos/user-infos.controller.js
frenautvh: Not really, it's just a frontside trick to avoid the user being tempted to click repeatedly on the button and make him wait (because the email might takes some time to be received).
It's not a security/limitation feature implemented on api side.

marie-j: Maybe we could see to have a task to poll instead or something like that ?

frenautvh: it's not possible, we cannot know when the user will receive the email (network latency etc)

FredericEspiau: Why does it matter if the user clicks repeatedly ? 🤔 I feel like it shouldn't matter to us

If it really matters because the API can't handle it, just disable the button, what do you think ?  :)

antleblanc:
Hi @frenautvh!

We recently merged #1336 which apply Prettier rules so could you please rebase
your branch on `develop`?

To easily resolve conflicts you can **ignore** `HEAD` changes and only keep yours.

After that you can run:

```sh
$ yarn run lint
```

This command will lint the entire codebase.

or:

Amend your commit to trigger the `pre-commit` hook (prettier will be launched
under the hood).

Thanks!

PR #2027, opened 2019-12-30
fix(dedicated.billing): add checks for automatic renew

marie-j:
DTRSD-6740

### DIFF
--- a/packages/manager/apps/dedicated/client/app/models/BillingService.class.js
+++ b/packages/manager/apps/dedicated/client/app/models/BillingService.class.js
@@ -46,7 +46,11 @@ export default class BillingService {
   }
 
   hasAutomaticRenew() {
-    return this.renew.automatic;
+    return (
+      this.renew.automatic ||
+      this.renewalType === 'automaticV2012' ||
+      this.renewalType === 'automaticV2016'
+    );
   }
 
   hasForcedRenew() {


packages/manager/apps/dedicated/client/app/models/BillingService.class.js
```diff
@@ -46,7 +46,11 @@ export default class BillingService {
   }
 
   hasAutomaticRenew() {
-    return this.renew.automatic;
+    return (
```
antleblanc: @marie-j Could you please amend your commit to apply Prettier rules (see below the expected diff)

```js
  hasAutomaticRenew() {
    return (
      this.renew.automatic ||
      this.renewalType === 'automaticV2012' ||
      this.renewalType === 'automaticV2016'
    );
  }
```

packages/manager/apps/dedicated/client/app/models/BillingService.class.js
```diff
@@ -46,7 +46,11 @@ export default class BillingService {
...
+      this.renew.automatic ||
```
frenautvh: I am wondering, you can have renew.automatic === false and still have automatic renew activated?

marie-j: Currently waiting an API response about that

antleblanc: Since this ticket has been validated then I process to the merge.
Feel free to open a new Pull Request if you have updates to apply.

PR #2041, opened 2020-01-06
fix(ip): refresh correctly ip if not ip block

JeremyDec:

### DIFF
--- a/packages/manager/apps/dedicated/client/app/ip/ip/ip-ip.controller.js
+++ b/packages/manager/apps/dedicated/client/app/ip/ip/ip-ip.controller.js
@@ -528,12 +528,26 @@ angular
           $scope.getReverseDelegations(ipBlock);
         }
 
-        Ip.getIpsForIpBlock(
-          ipBlock.ipBlock,
-          ipBlock.service.serviceName,
-          ipBlock.service.category,
-        ).then((data) => {
-          set(ipBlock, 'ips', data);
+        const isUniqueIp = ipBlock.isUniq;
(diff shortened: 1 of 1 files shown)


packages/manager/apps/dedicated/client/app/ip/ip/ip-ip.controller.js
```diff
@@ -647,10 +647,26 @@ angular
         set(ipBlock, 'refreshing', true);
         IpReverse.updateReverse(ipBlock, ip.ip, ip.reverseEditValue).then(
           () => {
-            $rootScope.$broadcast('ips.table.refreshBlock', ipBlock);
-            Alerter.success(
-              $translate.instant('ip_table_manage_reverse_success'),
-            );
+            if (!ipBlock.isUniq) {
```
frenautvh: it's better to avoid double negation : if (ipBlock.isUniq) { ...

frenautvh: sorry not a double negation but it's common practice to put the "true" block in the if and the "false" block in the else. it has better readability

cbourgois: 👍  and `else` is unnecessary.
```js
if (ipBlock.isUniq) {
  return loadService(ipBlock.service) // ...
}

$rootScope.$broadcast('ips.table.refreshBlock', ipBlock);
Alerter.success($translate.instant('ip_table_manage_reverse_success'));
return null;
```

PR #2043, opened 2020-01-07
fix(wuc): order cart service

frenautvh:
Related : https://github.com/ovh/manager/pull/2044


marie-j:
The `OrderService` is probably impacted too

frenautvh:
> The `OrderService` is probably impacted too

thanks, i opened #2044

Issue #2036, opened 2020-01-04
Translations for cPanel order a licence


tadhglewis:
**Describe the bug**
No translations for `2. Select your options` dropdowns on order a licence page for cPanel

**To Reproduce**
- Go to Dedicated/Server > Licences > Order > Select IP Block > Select cPanel > See Dropdown

**Screenshots**
![image](https://user-images.githubusercontent.com/53756558/71757720-0a687980-2eec-11ea-8152-554cd6e55ea7.png)

**Additional context**
Translations required.
This may also be an issue with manager/packages/manager/apps/dedicated/client/app/license/order/license-order.html line 136

antleblanc:
Hi @tadhglewis!

Thanks for reporting this issue.
I will keep you posted as soon as I have updates.

**Internal ticket reference**: MANAGER-4254

PR #2059, opened 2020-01-10
fix(licence): cPanel options are not human readable

radireddy:
cPanel options are not human-readable

ref: #2036, MANAGER-4254

### DIFF
--- a/packages/manager/apps/dedicated/client/app/license/order/license-order.controller.js
+++ b/packages/manager/apps/dedicated/client/app/license/order/license-order.controller.js
@@ -1,8 +1,26 @@
 import assign from 'lodash/assign';
+import capitalize from 'lodash/capitalize';
+import filter from 'lodash/filter';
+import get from 'lodash/get';
 import head from 'lodash/head';
 import includes from 'lodash/includes';
+import join from 'lodash/join';
+import map from 'lodash/map';
+import parseInt from 'lodash/parseInt';
+import replace from 'lodash/replace';
 import set from 'lodash/set';
+import sortBy from 'lodash/sortBy';
+import split from 'lodash/split';
 import values from 'lodash/values';
+
+import {
+  CPANEL_PREMIER_TYPES,
+  CPANEL_ADMIN_TYPE,
+  CPANEL_PRO_TYPE,
+  MAX_CPANEL_PRO_ACCOUNTS,
+  MAX_CPANEL_ADMIN_ACCOUNTS,
+  LICENCE_TYPES,
+} from './license-order.constants';
 
 angular
   .module('Module.license')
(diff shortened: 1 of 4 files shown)


packages/manager/apps/dedicated/client/app/license/order/license-order.controller.js
```diff
@@ -231,6 +236,34 @@ angular
             $scope.selected.version =
               $scope.types[$scope.selected.licenseType].options[0].options[0];
           }
+          const { options } = $scope.types[
+            $scope.selected.licenseType
+          ].options[0];
+          if (
```
antleblanc: Could be turned into:

```js
if (['CPANEL', 'WORKLIGHT'].includes($scope.selected.licenseType)) {
  // …
}
```

packages/manager/apps/dedicated/client/app/license/order/license-order.html
```diff
@@ -154,7 +154,7 @@ <h2 data-translate="license_order_step2_title"></h2>
                         class="form-control"
                         id="optionVersion"
                         name="optionVersion"
-                        data-ng-options="optionVersion as (('license_designation_' + selected.licenseType + '_version_' + optionVersion.value) | translate) for optionVersion in types[selected.licenseType].options[0].options"
+                        data-ng-options="optionVersion as optionVersion.displayName for optionVersion in formatedOptions"
```
marie-j: Can the translations keys used previously be remove then ? Or are they still used elsewhere ?

radireddy: these are used by other options. I have moved them from template to controller to handle them conditionally.

packages/manager/apps/dedicated/client/app/license/order/license-order.controller.js
radireddy: since there are only two, I have used || operator.

marie-j: Yes but it will be easier if we have to add another one

packages/manager/apps/dedicated/client/app/license/order/license-order.controller.js
```diff
@@ -231,6 +248,65 @@ angular
...
+              }
+              set(option, 'displayName', displayName);
+              set(option, 'cPanelCount', parseInt(count));
+              return option;
+            });
+            $scope.formatedOptions = sortBy(
+              $scope.formatedOptions,
+              'cPanelCount',
+            );
+          } else if ($scope.selected.licenseType === 'WORKLIGHT') {
```
JeremyDec: This can be in a constant too maybe ?

ovh-cds:
CDS Report build-monorepository#8709.0 ✘  
* 
  * Build ✘

PR #2063, opened 2020-01-10
[PCI-5407] update rescue message with password

ticapix:


frenautvh:
hello @ticapix thanks for contributing, don't you also need to change this translation or is it just for linux ? https://github.com/ovh/manager/blob/c0a55ac1a3d082192c2d46144eb7eb01d3cf57bd/packages/manager/modules/pci/src/projects/project/instances/instance/rescue/translations/Messages_fr_FR.json#L15

ticapix:
I haven't tested with windows. If the password is not correct, I don't know where it will be available.

ticapix:
I've tried with a windows vm and filled another bug DTRSD-7133.
Since the password recovery seems to be the same, I'm also updating the message for windows too.

ticapix:
I confirm that the password is not the correct one for windows too.
image

PR #2069, opened 2020-01-13
refactor(emailpro): use oui-header-tabs

marie-j:


packages/manager/modules/emailpro/src/emailpro.controller.js
```diff
@@ -7,6 +7,7 @@ export default /* @ngInject */ function EmailProCtrl(
   $q,
   $rootScope,
   $scope,
+  $state,
```
antleblanc: Could you please confirm that it's needed for achieving a refactor?
Because we used to avoid dealing with `$state` from a Controller right?

marie-j: As I'm changing the tab system yes it is. But you're right I'll try to pass it through resolve

PR #2076, opened 2020-01-14
fix(cuc.price): get price according to capacity

marie-j:
ref: DTRSD-7106

### DIFF
--- a/packages/manager/modules/cloud-universe-components/src/price/service.js
+++ b/packages/manager/modules/cloud-universe-components/src/price/service.js
@@ -1,6 +1,7 @@
 import find from 'lodash/find';
 import forEach from 'lodash/forEach';
 import get from 'lodash/get';
+import includes from 'lodash/includes';
 
 export default class OvhCloudPriceHelper {
   /* @ngInject */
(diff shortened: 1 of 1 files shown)


packages/manager/modules/cloud-universe-components/src/price/service.js
```diff
@@ -44,7 +45,12 @@ export default class OvhCloudPriceHelper {
         forEach(projectPlan.addonFamilies, (family) => {
           forEach(family.addons, (planCode) => {
             pricesMap[planCode] = this.transformPrice(
-              get(find(catalog.addons, { planCode }), 'pricings[0]'),
+              find(
+                get(find(catalog.addons, { planCode }), 'pricings'),
+                ({ capacities }) =>
+                  includes(capacities, 'renew') ||
```
Jisay:
```
capacities.includes('renew') || capacities.includes('consumption')
```
? To avoid an import ^^

marie-j: As capacities is in the structure of the object returned by API, we'll need to check it's defined first

marie-j: @Jisay yes but it implies checking if capacities is an array and other things :)

PR #2087, opened 2020-01-16
feat(web): private sql upgrade prorata

frenautvh:
See : MANAGER-4242

### DIFF
--- a/packages/manager/apps/web/client/app/private-database/database/ram/update/private-database-database-ram-update.controller.js
+++ b/packages/manager/apps/web/client/app/private-database/database/ram/update/private-database-database-ram-update.controller.js
@@ -167,5 +167,9 @@ angular.module('App').controller(
           this.loading.validation = false;
         });
     }
+
+    static isProrataDuration({ duration }) {
+      return /^upto/.test(duration);
+    }
   },
 );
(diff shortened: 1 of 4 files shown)


packages/manager/apps/web/client/app/private-database/database/ram/update/private-database-database-ram-update.controller.js
```diff
@@ -167,5 +167,9 @@ angular.module('App').controller(
           this.loading.validation = false;
         });
     }
+
+    static isProrataDuration({ duration }) {
+      return /^upto/.test(duration);
```
marie-j: This could be handled by a generic model (perhaps extended depending on the service) 
something like: 
```suggestion
     class OrderableService {
       hasProrataDuration() {
         return /^upto/.test(this.duration)
        }
     }
```

antleblanc: @marie-j I agree this could be a nice enhancement.

Since this ticket has already been validated I will merge it to the `develop` branch.

Feel free to open an Issue to address and track the reported suggestion.

Thanks

PR #2088, opened 2020-01-16
fix(web): disable correctly switch if they are unavailable

JeremyDec:
See : DTRSD-7165

ℹ️ In production, we're currently displaying a tooltip to inform the customer that the action switches, are unavailable.
This PR will disable correctly the switches, if they aren't available.
Tooltips are removed.


antleblanc:
You might investigate but it seems that some extra styles can be removed.

```sh
$ git grep v6-ui-switch
packages/manager/apps/web/client/app/domain/general-informations/GENERAL_INFORMATIONS.html:                            data-wuc-v6-ui-switch
packages/manager/apps/web/client/app/domain/general-informations/GENERAL_INFORMATIONS.html:                            data-wuc-v6-ui-switch
packages/manager/modules/web-universe-components/src/v6UiSwitch/index.js:import './v6-ui-switch.less';
packages/manager/modules/web-universe-components/src/v6UiSwitch/v6-ui-switch.less:wuc-v6-ui-switch,
packages/manager/modules/web-universe-components/src/v6UiSwitch/v6-ui-switch.less:[data-wuc-v6-ui-switch],
packages/manager/modules/web-universe-components/src/v6UiSwitch/v6-ui-switch.less:[wuc-v6-ui-switch] {
```

To be confirmed.

packages/manager/apps/web/client/app/domain/general-informations/domain-general-informations.controller.js
```diff
@@ -93,8 +97,21 @@ export default class DomainTabGeneralInformationsCtrl {
...
     this.vm = {
-      protection: { uiSwitch: {} },
-      dnssec: { uiSwitch: {} },
+      protection: {
+        disabled: [
+          PROTECTION_TYPES.LOCKING,
+          PROTECTION_TYPES.UNLOCKING,
+          PROTECTION_TYPES.UNAVAILABLE,
+        ].includes(this.domain.protection),
+        uiSwitch: {},
```
marie-j: Not yours but isn't `uiSwitch` just a boolean ? Moreover could it be something less related to UI implemation ? Something like `enabled` ?

JeremyDec:
@antleblanc after deep search it seems not to be used elsewhere, so yes let's remove it

packages/manager/apps/web/client/app/domain/general-informations/domain-general-informations.controller.js
JeremyDec: Yes you're right, I'll change the name

packages/manager/apps/web/client/app/domain/dns/sec/domain-dns-sec.controller.js
```diff
@@ -2,10 +2,19 @@ import get from 'lodash/get';
 
 angular.module('controllers').controller(
   'Domain.controllers.DnsSec',
+  /* @ngInject */
```
FredericEspiau: We usually set it above the `constructor`, don't we ?

JeremyDec: Yes,  making the change, to keep consistency :)

packages/manager/modules/web-universe-components/src/index.js
```diff
@@ -48,7 +47,6 @@ angular.module(moduleName, [
   wucServiceStatusAction,
   wucString,
   wucTabs,
-  wucV6UiSwitch,
```
antleblanc: Removing this directive should be marked as a `BREAKING CHANGE`.

Could you please update the associated commit?

Thanks!

packages/manager/apps/web/client/app/domain/general-informations/GENERAL_INFORMATIONS.html
```diff
@@ -308,15 +320,27 @@
                     data-term="{{ ::'domain_dashboard_dnssec' | translate }}"
                 >
                     <oui-tile-description>
+                        <oui-switch
+                            class="mt-2 d-inline-block"
+                            data-disabled="$ctrl.vm.dnssec.isUnavailable"
```
FredericEspiau: You can remove the `data-` here

JeremyDec: Hmm this one should be removed ?  (https://github.com/ovh/manager/pull/2188#discussion_r375674733 ?)

FredericEspiau: I think I'm victim of a bug from Github because my comments don't match the line

`data-`should only be kept:
- if not on a component
- if on an attribute redefined by `ui-kit` and that already existed for `HTML`

PR #2100, opened 2020-01-17
Update description

antleblanc:

1f60201 - build: add or update description for packages
50836aa - build(docs): update packages list

Note: `docs/.vuepress/public/assets/json/packages.json` is an autogenerated file.

## :house: Internal

- No quality check required.


packages/manager/apps/carrier-sip/package.json
```diff
@@ -2,6 +2,7 @@
   "name": "@ovh-ux/manager-carrier-sip-app",
   "version": "1.0.4",
   "private": true,
+  "description": "Carrier SIP Trunk standalone application.",
```
FredericEspiau: I don't think we need `.`, do we ?

antleblanc: It's not mandatory but It could be helpful I would say.

Description can be enhanced.

PR #2109, opened 2020-01-20
fix(telecom-telephony): called number property update

ganeshkumar1989:
Updated correct property to be used in the incoming calls table

ref: DTRSD-7456


antleblanc:
Added the `question` label because the initial request isn't clear.

ganeshkumar1989:
The bug is closed

PR #2115, opened 2020-01-20
refactor(adp): remove uib modal

varun257:
quota error model is not used anywhere
remove quota error model

Signed-off-by: Varun <varun.shivaprasad@corp.ovh.com>


varun257:
QC not required

PR #2153, opened 2020-01-27
Refactor/wuc order cart documentation

JeremyDec:
## 🚜 Refactor
c532716 - Add missing service documentation
7d1990b - Store constant in its own file

### DIFF
--- a/packages/manager/modules/web-universe-components/src/order-cart/order-cart.service.js
+++ b/packages/manager/modules/web-universe-components/src/order-cart/order-cart.service.js
@@ -1,3 +1,5 @@
+import { FIDELITY_ACCOUNT } from './order-cart.constants';
+
 /**
  * Cart order management service, based on API /order/cart
  * It is, for the moment, aimed to handle order of product, or existing product
@@ -16,6 +18,7 @@ export default class OrderCartService {
    * Will create a new cart for the user
    *
    * @param  {string} ovhSubsidiary OVHcloud Subsidiary in order to propose
+   * localized offers
    * @return {Promise<Object>}      Promise of the new created cart
    */
   createNewCart(ovhSubsidiary) {
@@ -218,7 +221,7 @@ export default class OrderCartService {
             orderId: order.orderId,
           },
           {
-            paymentMean: 'fidelityAccount',
+            paymentMean: FIDELITY_ACCOUNT,
           },
         ).$promise;
     }
--- a/packages/manager/modules/web-universe-components/src/order-cart/order-cart.service.js
+++ b/packages/manager/modules/web-universe-components/src/order-cart/order-cart.service.js
@@ -1,3 +1,9 @@
+/**
+ * Cart order management service, based on API /order/cart
+ * It is, for the moment, aimed to handle order of product, or existing product
+ * upgrades.
+ * All carts must be assigned before they are used.
+ */
 export default class OrderCartService {
   /* @ngInject */
   constructor($q, OvhApiMe, OvhApiOrder) {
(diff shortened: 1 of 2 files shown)


packages/manager/modules/web-universe-components/src/order-cart/order-cart.service.js
```diff
@@ -6,24 +14,46 @@ export default class OrderCartService {
     this.OvhApiOrder = OvhApiOrder;
   }
 
+  /**
```
antleblanc: Are you using a convention for documentation block comments?

In my opinion, I don't find useful to repeat the method name and then describe it.

Why not considering something like (template):

```txt
  /**
   * [createNewCart description]
   * @param  {[type]} ovhSubsidiary [description]
   * @return {[type]}               [description]
   */
```

Notice here I'm also mentioning the `@return`.

FYI I'm using the following extension: https://atom.io/packages/docblockr

JeremyDec: @antleblanc, I'm just applying the auto-generated template from https://atom.io/packages/atom-easy-jsdoc. 

So I've let the method name, and just removed the `@return {type}` line, for those calls which returns $promise, because I don't see much benefits to say to user : `@return Promise<Object>`, as it is binded to the API response.

Let me know your point of view.

FredericEspiau: Well telling that you can chain your method call and not use its value directly without looking at the code is lit 🔥

Jisay: I think we should have an uniformized way to comment our codes :-)

FredericEspiau: You mean like `js-doc` ? 😄 

I agree with you, it would make it easier to have a component library yes

JeremyDec: Ok, adding the return line

packages/manager/modules/web-universe-components/src/order-cart/order-cart.service.js
```diff
@@ -1,3 +1,11 @@
+import { FIDELITY_ACCOUNT } from './order-cart.constants';
+
+/**
+ * Cart order management service, based on API /order/cart
+ * It is, for the moment, aimed to handle order of product, or existing product
+ * upgrades.
+ * All carts must be assigned before to be used.
```
marie-j:
```suggestion
 * All carts must be assigned before they are used.
```

packages/manager/modules/web-universe-components/src/order-cart/order-cart.service.js
```diff
@@ -58,6 +111,15 @@ export default class OrderCartService {
       }).$promise;
   }
 
+  /**
+   * Add an option for a new product tp order
```
marie-j:
```suggestion
   * Add an option for a new product to order
```

packages/manager/modules/web-universe-components/src/order-cart/order-cart.service.js
```diff
@@ -98,6 +180,12 @@ export default class OrderCartService {
       }).$promise;
   }
 
+  /**
+   * Get the informations for checkout a order
```
marie-j:
```suggestion
   * Get the information for checkout a order
```
"Fun" fact : `information` is invariable :)

packages/manager/modules/web-universe-components/src/order-cart/order-cart.service.js
```diff
@@ -169,20 +290,39 @@ export default class OrderCartService {
...
+   */
   deleteItem(cartId, itemId) {
     return this.OvhApiOrder.Cart()
       .Item()
       .v6()
       .delete({ cartId, itemId }).$promise;
   }
 
+  /**
+   * Delete all items of the cart
```
marie-j:
```suggestion
   * Delete all items from the cart
```

packages/manager/modules/web-universe-components/src/order-cart/order-cart.service.js
```diff
@@ -126,14 +221,23 @@ export default class OrderCartService {
...
+            paymentMean: FIDELITY_ACCOUNT,
           },
         ).$promise;
     }
 
     return order;
   }
 
+  /**
+   * Get the details of product, in the public catalog, for a given subsidiary
```
marie-j:
```suggestion
   * Get the details for a product, in the public catalog, for a given subsidiary
```

packages/manager/modules/web-universe-components/src/order-cart/order-cart.service.js
```diff
@@ -86,6 +158,16 @@ export default class OrderCartService {
       ).$promise;
   }
 
+  /**
+   * Delete a configuration item of the product
```
marie-j:
```suggestion
   * Delete a configuration item for the product
```

PR #2188, opened 2020-02-03
feat(components.choice): return object if payment method is selected

JeremyDec:
BREAKING CHANGE: the two-way binded model will now store an object representing default payment method, instead of just setting a boolean value

See MANAGER-3328 for the context

### DIFF
--- a/packages/components/ng-ovh-payment-method/src/components/choice/template.html
+++ b/packages/components/ng-ovh-payment-method/src/components/choice/template.html
@@ -11,7 +11,7 @@
     <oui-radio
         data-name="paymentMethod"
         data-model="$ctrl.model"
-        data-value="true"
+        data-value="$ctrl.defaultPaymentMethod"
         data-disabled="!$ctrl.defaultPaymentMethod"
         data-required
     >
--- a/packages/components/ng-ovh-payment-method/src/components/choice/controller.js
+++ b/packages/components/ng-ovh-payment-method/src/components/choice/controller.js
@@ -16,7 +16,9 @@ export default class {
       .getDefaultPaymentMethod()
       .then((defaultPaymentMethod) => {
         this.defaultPaymentMethod = defaultPaymentMethod;
-        this.model = true;
+        this.model = this.defaultPaymentMethod
+          ? this.defaultPaymentMethod
+          : false;
       })
       .catch((error) => {
         this.$log.error(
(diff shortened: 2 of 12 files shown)


packages/components/ng-ovh-payment-method/src/components/choice/template.html
```diff
@@ -42,7 +42,7 @@
     <oui-radio
         data-name="paymentMethod"
         data-model="$ctrl.model"
-        data-value="false"
+        data-value="null"
```
radireddy: I think no need to use data- for name, value and required as they are standard HTML attributes

packages/components/ng-ovh-payment-method/src/components/choice/template.html
```diff
@@ -11,7 +11,7 @@
     <oui-radio
         data-name="paymentMethod"
         data-model="$ctrl.model"
-        data-value="true"
+        data-value="$ctrl.defaultPaymentMethod"
```
radireddy: I think no need to use data- for name, value, disabled and required as they are standard HTML attributes

packages/components/ng-ovh-payment-method/src/components/choice/template.html
FredericEspiau: You're right !

But `ui-kit` overrides some default `html` attributes, this is why it's needed :(

packages/components/ng-ovh-payment-method/src/components/choice/template.html
ganeshkumar1989: This will break the API call done here: https://github.com/ovh/manager/blob/master/packages/manager/apps/web/client/app/private-database/order/clouddb/private-database-order-clouddb.controller.js#L145
The API call at https://github.com/ovh/manager/blob/master/packages/manager/modules/web-universe-components/src/order-cart/order-cart.service.js#L117 expects a boolean for `autoPayWithPreferredPaymentMethod`. Not sure if passing an object would work here.
Similarly, we would have to check out the other places where the `ovh-payment-method-choice` component is used :)

JeremyDec: Yes you're right @ganeshkumar1989, that's why I've set the commit as BREAKING CHANGE.

Once this PR will be merged, the `@ovh-ux/ng-ovh-payment-method` package version will be increased to `6.0.0`, so it won't break anything until we explicitly use the new version, and make the adaptation ;)

marie-j: Do we have a migration plan for this ?

JeremyDec: @marie-j you mean adaptation of application using this module ?

JeremyDec: @marie-j here it is : https://github.com/ovh/manager/pull/2243

PR #2208, opened 2020-02-06
fix(dedicated-support-level): enable 4ls support ww

ganeshkumar1989:
4ls support enable ww except us, tn and ma

ref: MANAGER-4228

Due date: Feb 13th

### DIFF
--- a/packages/manager/apps/dedicated/client/app/account/user/support-level/support-level.constants.js
+++ b/packages/manager/apps/dedicated/client/app/account/user/support-level/support-level.constants.js
@@ -4,7 +4,28 @@ export const INACTIVE_SUPPORT_LEVEL = ['premium-accredited'];
 
 export const AVAILABLE_SUPPORT_LEVEL = ['business', 'enterprise'];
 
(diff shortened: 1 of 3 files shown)


packages/manager/apps/dedicated/client/app/account/user/support-level/support-level.constants.js
```diff
@@ -4,7 +4,28 @@ export const INACTIVE_SUPPORT_LEVEL = ['premium-accredited'];
...
+  'LT',
+  'NL',
+  'PL',
+  'PT',
+  'QC',
+  'SG',
+  'SN',
+  'WE',
+  'WS',
+];
```
marie-j: Why not reverse the logic here and instead of listing the AVAILABLE SUBSIDIARIES, list the UNAVAILABLE_SUBS which will be a shorter and will ease the maintenance ?

ganeshkumar1989: I thought about it... But decided not to disturb the logic that is already prodded

PR #2226, opened 2020-02-10
fix(enterprise-cloud-database.create): add missing label and contracts

Jisay:
ref:
* DTRSD-8137

### DIFF
--- a/packages/manager/modules/enterprise-cloud-database/src/create/create.html
+++ b/packages/manager/modules/enterprise-cloud-database/src/create/create.html
@@ -366,13 +366,14 @@
                 >
                 </enterprise-cloud-database-payment-info-component>
                 <div class="row">
-                    <div
-                        class="col-md-6"
-                        data-ng-if="$ctrl.order"
-                        data-ovh-contracts="$ctrl.order.contracts"
-                        data-full-text="false"
-                        data-ovh-contracts-validated="$ctrl.contractsAccepted"
-                    ></div>
+                    <div class="col-md-6" data-ng-if="$ctrl.order">
+                        <ovh-contracts-summary
+                            data-items="$ctrl.order.contracts"
+                            data-model="$ctrl.contractsAccepted"
+                            data-name="agreements"
+                        >
+                        </ovh-contracts-summary>
+                    </div>
                 </div>
                 <!-- deployment in progress -->
                 <div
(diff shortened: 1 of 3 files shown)


packages/manager/modules/enterprise-cloud-database/src/create/create.html
```diff
@@ -366,13 +366,14 @@ <h5>
...
                 </enterprise-cloud-database-payment-info-component>
                 <div class="row">
-                    <div
-                        class="col-md-6"
-                        data-ng-if="$ctrl.order"
-                        data-ovh-contracts="$ctrl.order.contracts"
-                        data-full-text="false"
-                        data-ovh-contracts-validated="$ctrl.contractsAccepted"
-                    ></div>
+                    <div class="col-md-6" data-ng-if="$ctrl.order">
```
FredericEspiau: Hm if `$ctrl.order` is null, then the `class="row"` right above will be incorrect so you can remove the row altogether

packages/manager/modules/enterprise-cloud-database/src/payment-info/payment-info.html
```diff
@@ -2,14 +2,15 @@
...
-            data-ng-src="{{::$ctrl.defaultPaymentMethod.original.icon.data}}"
-            alt=""
-        />
+        <img data-ng-src="{{::$ctrl.defaultPaymentMethod.icon.data}}" alt="" />
         <span data-ng-bind="$ctrl.defaultPaymentMethod.label"></span>
-        <span class="text-success"
-            >[{{ ::$ctrl.defaultPaymentMethod.paymentType.text }}]</span
-        >
+        <span class="text-success">
+            <span>[</span>
```
FredericEspiau:
```suggestion
            [
```
should be enough ;)

packages/manager/modules/enterprise-cloud-database/src/create/create.html
antleblanc: You're right! We will improve it.

packages/manager/modules/enterprise-cloud-database/src/payment-info/payment-info.html
antleblanc: Correct!

PR #2236, opened 2020-02-11
Feat/cancellation forms

marie-j:
ref: MANAGER-2414

Related : https://github.com/ovh-ux/ovh-api-services/pull/280


packages/manager/apps/dedicated/client/app/billing/confirmTerminate/billing-confirmTerminate.html
```diff
@@ -1,218 +1,58 @@
...
-                </table>
-            </div>
-
-            <div
-                class="mt-5"
-                data-ng-if="!TerminateServiceCtrl.loading &&
-                             TerminateServiceCtrl.serviceState !== 'suspending' &&
-                             TerminateServiceCtrl.serviceState !== 'suspended'"
+<oui-page-header
+    data-heading="{{ 'billing_confirm_termination_resiliate' | translate:{
```
JeremyDec:
```suggestion
    data-heading="{{:: 'billing_confirm_termination_resiliate' | translate:{
```

packages/manager/apps/dedicated/client/app/billing/confirmTerminate/billing-confirmTerminate.html
```diff
@@ -1,218 +1,58 @@
...
+        serviceName: $ctrl.service.resource.name
+    } }}"
+></oui-page-header>
+
+<div class="container-fluid mt-3">
+    <div class="mt-3" data-ovh-alert></div>
+
+    <oui-message data-type="warning" data-ng-if="$ctrl.hasSpecialConditions()"
+        ><span
+            data-translate="{{ 'billing_confirm_termination_information_text_' + $ctrl.user.ovhSubsidiary | translate }}"
```
JeremyDec:
```suggestion
            data-translate="{{:: 'billing_confirm_termination_information_text_' + $ctrl.user.ovhSubsidiary }}"
```

packages/manager/apps/dedicated/client/app/billing/confirmTerminate/Service.class.js
```diff
@@ -0,0 +1,13 @@
+export default class Service {
+  constructor(resource) {
```
FredericEspiau: Wasn't the correct pattern 
```suggestion
  constructor({ property1, property2 }) {
```
?

I don't remember why we had to do this

packages/manager/apps/dedicated/client/app/billing/confirmTerminate/Service.class.js
```diff
@@ -0,0 +1,13 @@
...
+    Object.assign(this, resource);
+  }
+
+  get name() {
+    return this.resource.name;
+  }
+
+  get path() {
+    return this.route.url;
```
FredericEspiau: ? where does the `this.route` comes from ?

If someone goes `this.path = a` then you won't have the same value when you `const a = this.path`

packages/manager/apps/dedicated/client/app/billing/confirmTerminate/billing-confirmTerminate.html
```diff
@@ -1,218 +1,58 @@
...
-                </table>
-            </div>
-
-            <div
-                class="mt-5"
-                data-ng-if="!TerminateServiceCtrl.loading &&
-                             TerminateServiceCtrl.serviceState !== 'suspending' &&
-                             TerminateServiceCtrl.serviceState !== 'suspended'"
+<oui-page-header
+    data-heading="{{:: 'billing_confirm_termination_resiliate' | translate:{
```
FredericEspiau: you could safely

```suggestion
    heading="{{:: 'billing_confirm_termination_resiliate' | translate:{
```

packages/manager/apps/dedicated/client/app/billing/confirmTerminate/billing-confirmTerminate.html
```diff
@@ -1,218 +1,58 @@
...
+    <oui-message data-type="warning" data-ng-if="!$ctrl.hasSpecialConditions()"
+        ><span
+            data-translate="billing_confirm_termination_information_text"
+        ></span
+    ></oui-message>
+
+    <form name="termination" novalidate class="mt-3">
+        <oui-field
+            data-ng-repeat="question in $ctrl.questions track by question.question"
+            data-label="{{:: 'billing_confirm_termination_question_' + question.question  | translate}}"
```
FredericEspiau: Questions have questions inside them ? 😱 

Is this inception ?

packages/manager/apps/dedicated/client/app/billing/confirmTerminate/Service.class.js
marie-j: This comes from API but it's to abstract the way the property is handled by API

packages/manager/apps/dedicated/client/app/billing/confirmTerminate/Service.class.js
marie-j: I think I used this shortcut because I wasn't sure the property I was going to need. It might be better with your proposal as it is explicit

packages/manager/apps/dedicated/client/app/billing/confirmTerminate/billing-confirmTerminate.html
marie-j: Habits 😄

packages/manager/apps/dedicated/client/app/billing/confirmTerminate/billing-confirmTerminate.html
marie-j: Property in payload is `question`

packages/manager/apps/dedicated/client/app/billing/confirmTerminate/billing-confirmTerminate.html
FredericEspiau: It's not like it's imperative, do whatever you want !

packages/manager/apps/dedicated/client/app/billing/confirmTerminate/Service.class.js
cbourgois: Yes it is better to have explicit properties. 
We got some trouble in past with remote paging in the datagrid from `ovh-ui-angular` if `resource` contains a `$promise` property (cf: https://github.com/ovh-ux/ovh-ui-angular/blob/master/packages/oui-datagrid/src/paging/datagrid-paging-abstract.js#L88)

radireddy:
[`ovh-api-services@9.37.0`](https://github.com/ovh-ux/ovh-api-services/blob/master/CHANGELOG.md#9370-2020-02-14) has been released and published

packages/manager/apps/dedicated/client/app/billing/confirmTerminate/Service.class.js
FredericEspiau: I understand but it means that the setter and the getter don't work on the same value :)

marie-j: I see your point, I think you are raising a bigger issue because those are not supposed to be updated and we should be able to have readonly properties 
This could be a solution : https://www.typescriptlang.org/docs/handbook/classes.html#readonly-modifier

FredericEspiau: Or you could use a sane alternative to JavaScript which enforces strict immutability 😅

marie-j:
@antleblanc This is done, I had to make tiny changes to the logic to adapt to fixes made

ovh-cds:
CDS Report build-monorepository#10019.0 ✘  
* 
  * Build ✘

ovh-cds:
CDS Report translator-push#5785.0 ✘  
* 
  * push ✘

ovh-cds:
CDS Report translator-push#5790.0 ✘  
* 
  * push ✘

marie-j:
Rebased onto master

PR #2239, opened 2020-02-12
fix(web): private database activation

frenautvh:
See : MANAGER-4357

### DIFF
--- a/packages/manager/apps/web/client/app/hosting/hosting.controller.js
+++ b/packages/manager/apps/web/client/app/hosting/hosting.controller.js
@@ -7,6 +7,7 @@ import isString from 'lodash/isString';
 import kebabCase from 'lodash/kebabCase';
 import map from 'lodash/map';
 import merge from 'lodash/merge';
+import reduce from 'lodash/reduce';
 import remove from 'lodash/remove';
 import set from 'lodash/set';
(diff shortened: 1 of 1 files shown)


packages/manager/apps/web/client/app/hosting/hosting.controller.js
```diff
@@ -265,7 +266,13 @@ export default class {
         this.$scope.hosting.sqlPriveInfo.nbDataBaseInclude = this.$scope.hosting.offerCapabilities.privateDatabases.length;
         this.$scope.hosting.sqlPriveInfo.nbDataBaseActive =
           this.$scope.hosting.sqlPriveInfo.nbDataBaseInclude -
-          privateDbCapabilities.length;
+          reduce(
```
Jisay: why don't use `Array.reduce` instead of `lodash/reduce`?

FredericEspiau: What about https://lodash.com/docs/4.17.15#sumBy ?

antleblanc: We can keep it as is since it's has been validated.

But you're right, we have multiple way to address this fix.

Thanks for your suggestions!

PR #2242, opened 2020-02-12
Fix/manager 4343

radireddy:
The language field is disabled in signup page if the country selected is Singapure

ref: MANAGER-4343


radireddy:
Regarding setting the first language, I thought about it. But, we can not do it because the field is optional and a customer can select it or ignore it.

marie-j:
> Regarding setting the first language, I thought about it. But, we can not do it because the field is optional and a customer can select it or ignore it.

I'm having trouble understanding the need here, because the field is disabled if there is only one choice though the customer cannot select the optional field. Shouldn't it be disabled if there is no choice ?

radireddy:
@marie-j If there are no languages and fields is disabled, it might be very confusing to the user. I think hiding the field if there are no languages makes sense. I think every country has at least one language, for Singapur it is English. But I am not very sure about this.

antleblanc:
Rebased on following commit: a8471875e37973fd3807012229bb89345bd8224e

PR #2243, opened 2020-02-12
refactor(web): adapt payment method choice use to new version

JeremyDec:
Refactor following #2188 

### DIFF
--- a/packages/manager/apps/web/client/app/domain/webhosting/order/steps/payment/domain-webhosting-order-steps-payment.controller.js
+++ b/packages/manager/apps/web/client/app/domain/webhosting/order/steps/payment/domain-webhosting-order-steps-payment.controller.js
@@ -39,6 +39,7 @@ export default class {
   }
 
   preparePayment() {
-    this.stepper.autoPayWithPreferredPaymentMethod = this.autoPayWithPreferredPaymentMethod;
+    this.stepper.autoPayWithPreferredPaymentMethod = !!this
+      .defaultPaymentMethod;
   }
 }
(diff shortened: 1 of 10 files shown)


packages/manager/apps/web/client/app/domain/webhosting/order/steps/payment/domain-webhosting-order-steps-payment.controller.js
```diff
@@ -39,6 +39,7 @@ export default class {
   }
 
   preparePayment() {
-    this.stepper.autoPayWithPreferredPaymentMethod = this.autoPayWithPreferredPaymentMethod;
+    this.stepper.autoPayWithPreferredPaymentMethod = !!this
```
frenautvh: i don't like the ```!!``` much, what about ```angular.isDefined``` ?

packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
```diff
@@ -49,9 +49,7 @@ export default class {
 
     this.checkoutOrderCart(
       // Autovalidate if the order is free
-      this.isOptionFree
-        ? this.isOptionFree
-        : this.autoPayWithPreferredPaymentMethod,
+      this.isOptionFree ? this.isOptionFree : !!this.defaultPaymentMethod,
```
frenautvh: ```this.isOptionFree || angular.isDefined(this.defaultPaymentMethod)``` ?

packages/manager/apps/web/client/app/domain/webhosting/order/steps/payment/domain-webhosting-order-steps-payment.controller.js
JeremyDec: Hmm in fact, `this.defaultPaymentMethod` will be either an object if the user select it, or will equal to `false` if user don't want to use the payment method.
So the `!!` is used as a shortcut to test `if( typeof this.defaultPaymentMethod === 'object')`. Where `angular.isDefined` won't be appropriate.

PR #2245, opened 2020-02-12
Refactor/remove cancellation forms

marie-j:
ref: MANAGER-4396


marie-j:
Closing this as there is no ETA on when it will be unblocked. Will be reopened then

PR #2256, opened 2020-02-13
Refactor/pcc/responsive

FredericEspiau:
MANAGER-4409

# Validation policy
It's been validated by two peers already, but I've rebase in the meantime

# Merging policy
Please don't squash, the commits should be good enough 🙏


packages/manager/apps/dedicated/client/app/dedicatedCloud/dashboard/index.js
```diff
@@ -21,6 +22,11 @@ angular
     'ui.router',
   ])
   .component(component.name, component)
+  .config(
```
cbourgois: Please use the same pattern as usually `.config(routing)`

packages/manager/apps/dedicated/client/app/dedicatedCloud/dashboard/dedicatedCloud-dashboard.routes.js
```diff
@@ -0,0 +1,7 @@
+import component from './dedicatedCloud-dashboard.component';
+
+export default /* @ngInject */ ($stateProvider) => {
+  $stateProvider.state('app.dedicatedClouds.dashboard', {
+    component,
```
marie-j: No need to use the angular declared component ?

packages/manager/apps/dedicated/client/app/dedicatedCloud/dedicatedCloud.html
```diff
@@ -7,85 +7,74 @@
...
+            <div class="oui-page-header__container d-block">
+                <div class="oui-page-header__content">
+                    <div class="d-flex">
+                        <h1
+                            class="oui-page-header__title"
+                            data-ng-bind="dedicatedCloud.description"
+                        ></h1>
+                        <button
+                            type="button"
+                            class="btn btn-icon ml-2"
```
marie-j: `oui-button` class instead maybe ?

FredericEspiau: Added some nasting padding that I couldn't get rid of unless I create a new stylesheet 😢

packages/manager/apps/dedicated/client/app/dedicatedCloud/dashboard/dedicatedCloud-dashboard.routes.js
FredericEspiau: Good point, I'll check

FredericEspiau: It was unused so I removed the declaration ! Good find @marie-j

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/dedicatedCloud-datacenter.html
```diff
@@ -1,118 +1,118 @@
-<div class="dedicatedCloud-datacenter">
-    <div
-        class="text-center mt-5"
-        data-ng-if="$ctrl.loadingInformations && !$ctrl.loadingError"
-    >
-        <oui-spinner data-size="l"></oui-spinner>
-    </div>
+<div
```
FredericEspiau: Please don't look at this whole file, it's mostly whitespaces, you can suppress them by clicking on "Hide whitespace changes" 🙏

packages/manager/apps/dedicated/client/app/dedicatedCloud/dashboard/dedicatedCloud-dashboard.routes.js
```diff
@@ -0,0 +1,7 @@
+import component from './dedicatedCloud-dashboard.component';
```
antleblanc: We tends to name file that defined a route in something like: `foo.routing.js`.

What do you think about it?

In comparison:
```sh
$ find . -type f -name "*.routing.js" | wc -l
378

$ find . -type f -name "*.routes.js" | wc -l
106
```

PR #2225, opened 2020-02-10
feat(hub): add user panel

frenautvh:

PR #2269, opened 2020-02-17
feat(hub): add billing summary tile

frenautvh:


packages/manager/modules/hub/src/components/billing-summary/controller.js
```diff
@@ -0,0 +1,72 @@
...
+      [this.billingPeriod] = this.periods;
+    });
+
+    return this.$q.all([loadLanguage, loadBills, loadDebt]);
+  }
+
+  onPeriodChange() {
+    this.bills = null;
+    return this.fetchBills(this.billingPeriod.value).then(({ data }) => {
+      this.bills = data;
```
radireddy: return this.bills maybe

packages/manager/modules/hub/src/components/billing-summary/controller.js
```diff
@@ -0,0 +1,72 @@
...
+    this.$q = $q;
+    this.$translate = $translate;
+    this.OvhHttp = OvhHttp;
+  }
+
+  $onInit() {
+    const loadLanguage = this.$q
+      .when(this.me ? this.me.language : this.fetchLanguage())
+      .then((language) => {
+        this.language = language;
```
marie-j: Why base yourself on the `me.language` and not the manager settting ?

packages/manager/modules/hub/src/components/billing-summary/index.scss
```diff
@@ -0,0 +1,69 @@
+.manager-hub-billing-summary {
```
marie-j: Could some styles be put in common with other features ?

packages/manager/modules/hub/src/components/billing-summary/template.html
```diff
@@ -0,0 +1,35 @@
...
+    <oui-select
+        data-model="$ctrl.billingPeriod"
+        data-items="$ctrl.periods"
+        data-on-change="$ctrl.onPeriodChange()"
+        data-match="label"
+    >
+    </oui-select>
+    <span
+        class="manager-hub-billing-summary_bill-total"
+        data-ng-bind-html="$ctrl.bills && $ctrl.language ? $ctrl.getFormattedPrice($ctrl.bills.total) : '…'"
```
radireddy: getFormattedPrice will be called multiple times based on angular lifecycle dirty check. Can you populate it after fetching bills data.

packages/manager/modules/hub/src/components/billing-summary/template.html
```diff
@@ -0,0 +1,35 @@
+<div class="manager-hub-billing-summary">
+    <h3 data-translate="hub_billing_summary_title"></h3>
+    <oui-select
```
radireddy: don't you think loading is required when bills and other data are being fetched?

packages/manager/modules/hub/src/components/billing-summary/controller.js
```diff
@@ -0,0 +1,72 @@
...
+      });
+    const loadBills = this.$q
+      .when(this.bills ? this.bills : this.fetchBills())
```
ganeshkumar1989: I guess we can make it simple by deciding whether `bills` (and other similar data) should be made mandatory or should it be fetched inside the component itself :)
What do you think?

packages/manager/modules/hub/src/components/billing-summary/controller.js
```diff
@@ -0,0 +1,72 @@
...
+        this.bills = data;
+      });
+    const loadDebt = this.$q
+      .when(this.debt ? this.debt : this.fetchDebt())
+      .then(({ data }) => {
+        this.debt = data;
+      });
+
+    this.$translate.refresh().then(() => {
+      this.periods = [1, 3, 6].map((months) => ({
```
ganeshkumar1989: Should `[1,3,6]` go into constants?

packages/manager/modules/hub/src/components/billing-summary/index.js
```diff
@@ -0,0 +1,21 @@
+import angular from 'angular';
+import ngOvhHttp from '@ovh-ux/ng-ovh-http';
+import '@ovh-ux/ng-translate-async-loader';
+import 'angular-translate';
+import component from './component';
```
ganeshkumar1989: Sort the imports please :)

packages/manager/modules/hub/src/components/billing-summary/component.js
```diff
@@ -0,0 +1,13 @@
...
+import template from './template.html';
+
+export default {
+  controller,
+  template,
+  bindings: {
+    me: '<?',
+    bills: '<?',
+    debt: '<?',
+    goToBills: '<',
```
ganeshkumar1989: Guess the component can internally handle the redirection, since the target url is not going to changeif passed into the component or if handled by the component itself :)
What do you think about it?

frenautvh: yes i'm waiting for https://github.com/ovh/manager/pull/2225 to be merged and i'll refactor it with RedirectionService

packages/manager/modules/hub/src/components/billing-summary/template.html
frenautvh: yes, i'm not sure how we should handle loading in the hub tiles though. oui-skeleton doesn't looks good with the current tile style & color

packages/manager/modules/hub/src/components/billing-summary/index.js
JeremyDec: @ganeshkumar1989, in this case, vendor imports should always be at top level, so the import of the component comes after, as above

packages/manager/modules/hub/src/components/billing-summary/index.scss
frenautvh: yes, i was thinking MANAGER-4306 was here to handle the style refactoring and merging of all the hub components

packages/manager/modules/hub/src/components/billing-summary/controller.js
frenautvh: since this code is not duplicated anywhere i'm not sure

packages/manager/modules/hub/src/components/billing-summary/controller.js
frenautvh: i like the idea that you can either provide or not the data to the component but you are right that the component would be simpler if we choose for example to provides data and the component would just act as a template

i would not bother overthinking it because it's still an angularjs component so it's not something that is meant to be reused in the future

if it's ok for you i'll let it like that otherwise please share your thoughts :)

ganeshkumar1989: I am ok with it @frenautvh. Sounds good :)

packages/manager/modules/hub/src/components/billing-summary/index.js
ganeshkumar1989: Yeah got it @JeremyDec . Thanks :)

PR #2270, opened 2020-02-17
fix(public-cloud): show price for object storage

radireddy:
show price for object storage

#MANAGER-4040


ovh-cds:
CDS Report build-monorepository#9374.0 ✘  
* 
  * Build ✘

packages/manager/modules/pci/src/projects/project/storages/containers/add/add.controller.js
```diff
@@ -99,4 +113,13 @@ export default class PciStoragesContainersAddController {
         this.isLoading = false;
       });
   }
+
+  getPrice(price) {
```
frenautvh: maybe rename to ```getFormattedPrice``` for clarity

packages/manager/modules/pci/src/projects/project/storages/containers/add/add.controller.js
```diff
@@ -63,6 +64,19 @@ export default class PciStoragesContainersAddController {
...
   onTypesFocus() {
     this.displaySelectedType = false;
+    this.loadingPrice = true;
+    this.PciProjectStorageContainersService.getPriceEstimation(
+      this.user.ovhSubsidiary,
+    )
+      .then((price) => {
+        this.price = price;
+        if (this.price !== null) {
+          this.price.formatedPrice = this.getPrice(price.price / 100000000);
```
frenautvh: where does 10000000 comes from? is the price in micro cents?

radireddy: yes, it needs to be converted to show in Euros

packages/manager/modules/pci/src/projects/project/storages/containers/add/add.controller.js
```diff
@@ -63,6 +64,21 @@ export default class PciStoragesContainersAddController {
 
   onTypesFocus() {
     this.displaySelectedType = false;
+    this.loadingPrice = true;
+    this.PciProjectStorageContainersService.getPriceEstimation(
+      this.user.ovhSubsidiary,
+    )
+      .then((price) => {
+        this.price = price;
```
ganeshkumar1989:
```suggestion
        this.price = {
            ...price,
            formatedPrice: this.getFormatedPrice(
                get(price, 'price', 0) / 100000000,
            )
        };
```

packages/manager/modules/pci/src/projects/project/storages/containers/containers.service.js
```diff
@@ -368,4 +374,29 @@ export default class PciStoragesContainersService {
...
+
+  getPriceEstimation(ovhSubsidiary) {
+    return this.OvhApiOrderCatalogPublic.v6()
+      .get({
+        productName: PUBLIC_CLOUD_PRODUCT_NAME,
+        ovhSubsidiary,
+      })
+      .$promise.then((catalog) => {
+        let priceObj = null;
+        find(get(catalog, 'addons', []), (addon) => {
```
ganeshkumar1989: Can we use `reduce` instead, to find `priceObj`? We are not storing the result of `find` anywhere :)

packages/manager/modules/pci/src/projects/project/storages/containers/add/add.controller.js
```diff
@@ -63,6 +64,21 @@ export default class PciStoragesContainersAddController {
...
+        if (this.price !== null) {
+          this.price.formatedPrice = this.getFormatedPrice(
+            price.price / 100000000,
```
marie-j: We might need to check we don't have a component/filter/service that doesn't already handle microcents prices

radireddy: I think we need to add a price format filter or component

packages/manager/modules/pci/src/projects/project/storages/containers/containers.service.js
radireddy: I don't think, I can use reduce here. What I am doing here is filtring array based on two nested properties and at the same time return the nested object back to avoid one more iteration.

packages/manager/modules/pci/src/projects/project/storages/containers/add/add.controller.js
radireddy: I get "Shorthand property assignments are valid only in destructuring patterns (74:24)" when I make this change

FredericEspiau: I probably made a syntaxic mistake because this should work :)

ovh-cds:
CDS Report translator-push#5226.0 ✘  
* 
  * push ✘

packages/manager/modules/pci/src/projects/project/storages/containers/add/add.controller.js
radireddy: I tried again with no syntax error and I still get this eslint error.

PR #2272, opened 2020-02-17
feat(cloud.logs): input allowed networks - bulk

pdepaepe:
Hello,

This PR add following changes on LDP manager:

- New way to handle allowed networks on data gathering tools. Instead to be in real time, we now allow them to be declared in bulk mode. Internal reference : MANAGER-4132
- Facilitate the way to copy tokens. Internal reference: MANAGER-4429

As always, thank you for the time spent on this.

Signed-off-by: Pierre De Paepe <pierre.de-paepe@corp.ovh.com>


packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/configure/logs-inputs-add-configure.controller.js
```diff
@@ -145,7 +145,7 @@ class LogsInputsAddConfigureCtrl {
       return this.$q.reject();
     }
     if (!this.flowggerForm.$dirty) {
-      return this.goToNetworkPage();
+      return this.$state.go('dbaas.logs.detail.inputs');
```
marie-j: Why not keep the function and update it to suit the needs ? It seems it will be easier to maintain

FredericEspiau: At least change the function name and the state it goes to

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/configure/logs-inputs-add-configure.controller.js
```diff
@@ -155,8 +155,8 @@ class LogsInputsAddConfigureCtrl {
           this.input.data,
           this.configuration.flowgger,
         )
-          .then(() => this.goToNetworkPage())
-          .finally(() => this.CucControllerHelper.scrollPageToTop()),
+          .then(() => this.$state.go('dbaas.logs.detail.inputs'))
+          .catch(() => this.CucControllerHelper.scrollPageToTop()),
```
FredericEspiau: So it shouldn't be done everytime ?

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/configure/logs-inputs-add-configure.controller.js
pdepaepe: Since the third panel (network) disappear, i don't see the point to keep this function.

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/configure/logs-inputs-add-configure.controller.js
pdepaepe: Good point, moving back to finally.

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/configure/logs-inputs-add-configure.controller.js
marie-j: It does not make sense to keep this function as is, but you could rename it to `goToInputs` and go to the state `dbaaas.logs.details.inputs`

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/configure/logs-inputs-add-configure.controller.js
```diff
@@ -190,27 +190,26 @@ class LogsInputsAddConfigureCtrl {
...
 
   getDelimiters() {
     return this.LogsInputsService.getDelimiters();
   }
+
+  goToInputs() {
+    this.$state.go('dbaas.logs.detail.inputs', {
+      serviceName: this.serviceName,
+    });
+    return this.$q.resolve();
```
JeremyDec: Why using a `$q.resolve` here ? If you want to return a promise, you can directly :
```suggestion
  return this.$state.go
```

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/edit/logs-inputs-add-edit.controller.js
```diff
@@ -95,9 +96,13 @@ class LogsInputsAddEditCtrl {
...
       .then((successData) => {
-        this.gotToNextStep(this.inputId || successData[0].item.inputId);
+        if (successData[0].item) {
+          this.gotToNextStep(this.inputId || successData[0].item.inputId);
+        } else {
+          this.$q.reject();
+        }
       })
-      .catch(() => this.CucControllerHelper.scrollPageToTop());
+      .finally(() => this.CucControllerHelper.scrollPageToTop());
```
JeremyDec: It seems that you could kept the `catch` call, as you either call `$q.reject` or a `$state.go`

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/edit/logs-inputs-add-edit.controller.js
```diff
@@ -95,9 +96,13 @@ class LogsInputsAddEditCtrl {
     return this.inputAddEdit
       .load()
       .then((successData) => {
-        this.gotToNextStep(this.inputId || successData[0].item.inputId);
+        if (successData[0].item) {
+          this.gotToNextStep(this.inputId || successData[0].item.inputId);
```
JeremyDec:
```suggestion
          return this.gotToNextStep(this.inputId || successData[0].item.inputId);
```

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/logs-inputs.service.js
```diff
@@ -1,4 +1,7 @@
 import set from 'lodash/set';
+import has from 'lodash/has';
+import map from 'lodash/map';
+import trim from 'lodash/trim';
```
JeremyDec: Can these imports be sorted please ?

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/logs-inputs.service.js
```diff
@@ -629,14 +642,18 @@ class LogsInputsService {
   }
 
   static transformInputToSave(input) {
+    if (trim(input.info.allowedNetworks) !== '') {
+      this.allowedNetworks = map(
```
JeremyDec: You can remove the static keyword if using `this` object in the method.

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/logs-inputs.service.js
```diff
@@ -411,6 +414,12 @@ class LogsInputsService {
       [input.info.engine.name, input.info.engine.version].join(' '),
     );
     set(input, 'info.exposedPort', parseInt(input.info.exposedPort, 10));
+    if (
+      has(input, 'info.allowedNetworks.length') &&
```
FredericEspiau: Isn't it the same ?

```js
has(input, 'info.allowedNetworks.length')  === Array.isArray(input.info.allowedNetworks)
```

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/logs-inputs.service.js
```diff
@@ -628,15 +641,19 @@ class LogsInputsService {
     return details;
   }
 
-  static transformInputToSave(input) {
+  transformInputToSave(input) {
+    if (trim(input.info.allowedNetworks) !== '') {
```
FredericEspiau:
```suggestion
    import isEmptry from "lodash/isEmpty";
    
    ...
    
    if (isEmpty(trim(input.info.allowedNetworks))) {
```
Is more expressive than `!== ''` :)

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/edit/logs-inputs-add-edit.html
```diff
@@ -83,6 +83,21 @@
             </oui-numeric>
         </oui-field>
 
+        <oui-field
```
FredericEspiau: This is good 🙂

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/edit/logs-inputs-add-edit.controller.js
```diff
@@ -95,9 +96,14 @@ class LogsInputsAddEditCtrl {
     return this.inputAddEdit
       .load()
       .then((successData) => {
-        this.gotToNextStep(this.inputId || successData[0].item.inputId);
+        if (successData[0].item) {
```
FredericEspiau: What is supposed to be in `successData[0]` and in `successData[0].item` ?

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/edit/logs-inputs-add-edit.controller.js
```diff
@@ -95,9 +96,14 @@ class LogsInputsAddEditCtrl {
...
+          return this.gotToNextStep(
+            this.inputId || successData[0].item.inputId,
+          );
+        }
+        return this.$q.reject();
```
FredericEspiau: This will have no impact, what are you trying to do here ?

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/logs-inputs.service.js
marie-j: @FredericEspiau a string could have a length

FredericEspiau: @marie-j because a string is an Array 😅 

What I meant is: isn't `Array.isArray` enough ?

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/logs-inputs.service.js
pdepaepe: Not valid, since isEmpty only handles Arrays and Objects. Here we check a String.

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/logs-inputs.service.js
pdepaepe: You are right, Array.isArray is enough.

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/edit/logs-inputs-add-edit.controller.js
pdepaepe: If related operation succeed, item is given in the promise successData[0]. 
successData[0].item is the LDP operation.
When API reject the call due to a Validation Error, item is not set and so i let the user on the current page to let him fix the incorrect value.

They are probably a better "Lexi" trick to handle that, but it's beyond my skills

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/edit/logs-inputs-add-edit.html
pdepaepe: Not sure to understand.

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/edit/logs-inputs-add-edit.controller.js
pdepaepe: Same answer.

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/add/edit/logs-inputs-add-edit.html
FredericEspiau: It's a compliment 🙂

packages/manager/apps/cloud/client/app/dbaas/logs/detail/inputs/logs-inputs.service.js
FredericEspiau: You are absolutely right, but `string` are `array-like` 🙂 

In `JavaScript`, `array-like` means "has a `.length` property and indexed elements", for example, both those are valid:
- `"hello".length`
- `"hello"[0]`

This way, you can use `_.isEmpty` on a `string`!

antleblanc:
@pdepaepe I've added the label `blocked` due to MANAGER-4132 which hasn't be validated yet.

I will keep you posted as soon as it could be unblocked.

Thanks!

PR #2267, opened 2020-02-14
feat(tuc.editable_service_name): make compatible with ui-kit header

FredericEspiau:
MANAGER-4419

PR #2275, opened 2020-02-18
Refactor/editable service name

FredericEspiau:
MANAGER-4419 (related to #2267)

# Merging policy
Please don't squash, the commits are good I think 🙏


FredericEspiau:
@frenautvh 😅

FredericEspiau:
@frenautvh done

frenautvh:
> @frenautvh done

What are the bootstrap 4 class that you are using ? I see you are importing both bootstrap 3 & 4 i think something is wrong

FredericEspiau:
@frenautvh imported both `bootstrap3` and `bootstrap4` is done pretty much everywhere so I didn't care much about it 🙂 

I'm using the `boostrap4` utilities

frenautvh:
> @frenautvh imported both `bootstrap3` and `bootstrap4` is done pretty much everywhere so I didn't care much about it 🙂
> 
> I'm using the `boostrap4` utilities

it's not the same, it's imported that way : https://github.com/ovh/manager/blob/master/packages/manager/modules/support/src/support.scss#L2 (scoped sass imports)

ovh-cds:
CDS Report installation#9418.0 ■   
* Stage 1
  * Install ■

ovh-cds:
CDS Report build-monorepository#9424.0 ✘  
* 
  * Build ✘

packages/manager/apps/freefax/src/index.js
```diff
@@ -1,7 +1,9 @@
+import 'bootstrap4';
```
frenautvh: you should remove every bootstrap4 js imports, since you imported the utilities function within a sass file

FredericEspiau: You're killing me 😅

FredericEspiau:
@frenautvh after 35 tries, I've done what you wanted I think

packages/manager/apps/telecom/src/app/telecom/telephony/line/telecom-telephony-line-main.view.html
```diff
@@ -5,15 +5,18 @@
...
-        <header class="page-header" role="banner">
-            <!-- GROUP TITLE -->
-            <tuc-editable-service-name
-                data-tuc-editable-service-name-title="LineCtrl.line.getDisplayedName()"
-                data-tuc-editable-service-name-title-on-save="LineCtrl.lineNameSave"
-                data-tuc-editable-service-name-service-name="LineCtrl.line.serviceName"
-                data-maxlength="25"
-            >
-            </tuc-editable-service-name>
+        <header class="oui-page-header">
```
marie-j: I suppose component doesn't support custom component transclusion ?

FredericEspiau: You suppose well 🙂

packages/manager/modules/telecom-universe-components/package.json
```diff
@@ -16,6 +16,9 @@
     "lodash": "^4.17.15",
     "ovh-ui-kit": "^2.41.1"
   },
+  "devDependencies": {
+    "bootstrap4": "twbs/bootstrap#v4.0.0"
```
antleblanc, line 20: Why `bootstrap4` is defined as a devDependency?

Why not considering having it as a peer Dependency instead?

FredericEspiau: 😅 

Two months ago, you, me, @marie-j  and @frenautvh we determined that the best was to have it in `devDependencies`. I don't care if you want to change, just settle on a way of doing things once and for all

antleblanc: Sorry but I can't find mention that is saying that `devDependencies` is/was the **best** solution regarding the context.

Same from the previous/original Pull Request:  https://github.com/ovh/manager/pull/2267#pullrequestreview-359691462

So could you please argue about **why** you've decided to turn it into a `devDependencies`? It might be really helpful to understand the needs from there.

cbourgois: _devDependencies_ are dependencies that a module needs to run in **development**.

_Peer_ are dependencies that a package needs to run in production (like dependencies) but expects dependencies to be provided by its dependant (we define that all AngularJS modules are in peer).

As here `bootstrap4` parts are imported and bundled, following our definitions it could stay in _devDependencies_

cbourgois: **Update**: this package has no dist (no build script), so shouldn't be in `devDependencies`.

FredericEspiau: @antleblanc we can't keep on having the same talk over and over again, let's write the procedure somewhere, so we never have to wonder again :)

antleblanc: We covered all the case via the comment feeds.
Let us know if something isn't clear.

Based on the comments I will push a commit (fixup) for `e7cebf6` in order to be able to merge this Pull Request.

Thanks!

PR #2284, opened 2020-02-19
fix(translations): clean translations files

cbourgois:
### Description of the Change

Clean translations to sync translations files between source (`fr_FR`) and targets: 
* Some translations are no more present in source translations files (`fr_FR`) and are still in target translations files.
* Some translations moved from one path to another and are still present in the previous path. It could be an issue depending on the translations loading (if translations were changed, they are not changed in old path).

Example of moved translations : 
https://github.com/ovh/manager/blob/master/packages/manager/modules/vrack/src/translations/Messages_en_GB.json#L29 moved to https://github.com/ovh/manager/blob/master/packages/manager/modules/vrack/src/move-dialog/translations/Messages_fr_FR.json#L2

58724ae29 — fix(translations): clean translations files


antleblanc:
I'm confused by additions in this Pull Request (+10,115).

github com_ovh_manager_pull_2284

Is the `translations-cleaner` isn't supposed to **only** clean translations? (meaning only remove duplicate ones?

FredericEspiau:
It also reindents @antleblanc

antleblanc:
Thanks @FredericEspiau!

FredericEspiau:
What do you mean @antleblanc ? That we can't prove that we didn't loose any translation ?

cbourgois:
I think an example could be more explicit (pure example): 

```
$ cat packages/manager/modules/product/translations/Messages_fr_FR.json
{ 
  "product_title": "Produit",
  "product_description": "Description"
}

$ cat packages/manager/modules/product/translations/Messages_en_GB.json
{ 
  "product_title": "Product",
  "product_description": "Description"
}
```

And at one point we move `product_title` to another file after some refactor : 
```
$ cat packages/manager/modules/product/translations/Messages_fr_FR.json
{ 
  "product_description": "Description"
}

$ cat packages/manager/modules/product/translations/Messages_en_GB.json
{ 
  "product_title": "Product",
  "product_description": "Description"
}

$ cat packages/manager/modules/product/details/translations/Messages_fr_FR.json
{ 
  "product_title": "Produit"
}
```
At this point `en_GB` translation for `product_title` is still available (and maybe loaded).

After cleaning translations files, we removed `product_title` from `packages/manager/modules/product/translations/Messages_en_GB.json`, and `product_title` is no more present for `en_GB`

We could say that is OK because translations weren't as the good place,
but there were chances that the translation has continued to be loaded and available for the user before cleaning.

PR #1955, opened 2019-12-14
fix(telephony): disable some features when nic has tech role only

julieni:
Disable some features when user has only tech role

NB: eslint changed some foreign code, hope it's OK

PR #1833, opened 2019-11-21
feat(module.server-sidebar): hide empty sections

julieni:
Signed-off-by: Julien Issler <julien@issler.net>

Needs module update: https://github.com/ovh-ux/ng-ovh-sidebar-menu/pull/111

PR #2291, opened 2020-02-19
fix(server-sidebar): restore items loading on click

cbourgois:
### Description of the Change

81cc265a9 — fix(server-sidebar): restore items loading on click

### DIFF
--- a/packages/manager/modules/server-sidebar/src/controller.js
+++ b/packages/manager/modules/server-sidebar/src/controller.js
@@ -198,7 +198,6 @@ export default class OvhManagerServerSidebarController {
 
         if (has(service, 'types')) {
           menuItem.onLoad = () => this.loadServices(service, menuItem);
-          menuItem.loadSubItems();
         } else {
           this.addItems(get(service, 'children'), menuItem);
         }
@@ -234,7 +233,6 @@ export default class OvhManagerServerSidebarController {
             },
             parent,
           );
-          parent.hide();
         } else {
           each(typesServices, (typeServices) => {
             let items = get(typeServices, 'items');


FredericEspiau:
This is basically reversing https://github.com/ovh/manager/pull/1833, so does this mean empty sections will be kept open ? Can you take this into account maybe ?

cbourgois:
@FredericEspiau Empty sections will be kept visible. If we want to display/hide sections at the app initialization we should retrieve all services at the bootstrap even if we don't need them.

FredericEspiau:
@cbourgois but why did we access the original PR then ? We should wonder about this

julieni:
> @cbourgois but why did we access the original PR then ? We should wonder about this

Hi, as the contributor of the original PR, I have to say it was a change requested by users on mailing lists, and it took some time to implement as it requested changes in submodule too.

I think the best way to go is to provide an API call to get services count by section. 1 request only, and this would also allow to display a badge with services count on every section.

julieni:
And also, you accepted the PR because it improved UX, I think. It's not because the proper, lightweight API calls are missing that we can't improve UX.

It makes me think of https://github.com/ovh/manager/pull/1955, the API is not ready (in this case not even safe), so all improvements should go to trash ? ...

cbourgois:
Hello @julieni,

The original feature for the sidebar was to load services only when the user click on a product section (_load only what you needs_).

I understand that the original PR improved UX for some customers who don't have all products.
They saw all the products in the sidebar and when they clicked to open a product section they saw "_No services_".

But we also have customers who have a lot of different services, and regarding the performances we don't want to make 15 requests to the API to retrieve all the products to detect if we need to show / hide them. 

I think we should restore the original feature (loading services when the user wants to see them) for now, and we will open an issue. IMO, this is the best compromise.

In this issue we should address this sidebar UX improvement for **all** customers based on your needs. 
As you said, we could have an API to get number of services to initialize the sidebar (hide/show products sections), and keep lazy-loading the services.

Thanks,

julieni:
> I think we should restore the original feature (loading services when the user wants to see them) for now, and we will open an issue. IMO, this is the best compromise.
> In this issue we should address this sidebar UX improvement for **all** customers based on your needs.
> As you said, we could have an API to get number of services to initialize the sidebar (hide/show products sections), and keep lazy-loading the services.

I totally agree on this, but due to the lack of such an API endpoint, the only way was to preload the services list.
Please provide such an API endpoint and I'll be happy to improve my PR using it ;)

cbourgois:
The issue is created : #2360

PR #2295, opened 2020-02-20
feat(core): add redirection service

frenautvh:
Extracted redirection service as it will be used from multiple tile components


packages/manager/modules/core/src/redirection/redirection.constants.js
```diff
@@ -0,0 +1,47 @@
+const managerRoot = {
```
FredericEspiau: I once did that and people got mad at me 😅 

Good if it gets merged

PR #2296, opened 2020-02-20
fix(pci.project.sidebar): set the right color for the beta status badge

antleblanc:
## :bug: Bug Fix

7aa6d56 - fix(pci.project.sidebar): set the right color for the beta status badge

## :house: Internal

ref: DTRSD-8627

### DIFF
--- a/packages/manager/apps/public-cloud/src/assets/theme/default/_sidebar.less
+++ b/packages/manager/apps/public-cloud/src/assets/theme/default/_sidebar.less
@@ -182,6 +182,16 @@ cloud-sidebar {
     .oui-sidebar-link[aria-expanded='true'] {
       background-color: #fff;
     }
+
+    .oui-status_beta {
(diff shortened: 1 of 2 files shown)


packages/manager/apps/public-cloud/src/assets/theme/default/_sidebar.less
```diff
@@ -182,6 +182,16 @@ cloud-sidebar {
     .oui-sidebar-link[aria-expanded='true'] {
       background-color: #fff;
     }
+
+    .oui-status_beta {
+      padding: 0 8px;
+      line-height: 15px;
+      font-size: 12px;
+      font-weight: 600;
+      background-color: #47ff78;
```
antleblanc: Both background-color and color must be retrieved from the ODS that will brings theming.

Since we are relying on a override I let them as is.

PR #2299, opened 2020-02-20
refactor(core): remove ovh-api-services

frenautvh:
replace ovh-api-services with $http

### DIFF
--- a/packages/manager/modules/core/src/session/session.service.js
+++ b/packages/manager/modules/core/src/session/session.service.js
@@ -1,11 +1,11 @@
 export default class SessionService {
   /* @ngInject */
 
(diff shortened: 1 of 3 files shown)


packages/manager/modules/core/src/session/session.service.js
```diff
@@ -1,11 +1,11 @@
...
 
-  constructor(OvhApiMe) {
-    this.OvhApiMe = OvhApiMe.v6();
+  constructor($http) {
+    this.$http = $http;
   }
 
   getUser() {
-    return this.OvhApiMe.get().$promise;
+    return this.$http.get('/me').then(({ data }) => data);
```
FredericEspiau: 😅

PR #2300, opened 2020-02-20
fix(pci.project.quota): use ngOvhPaymentMethod

Jisay:
# 🏡 Internal

ref:
- DTRSD-8582

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/quota/quota.service.js
+++ b/packages/manager/modules/pci/src/projects/project/quota/quota.service.js
@@ -9,6 +9,12 @@ export default class {
     this.OvhApiMe = OvhApiMe;
   }
 
+  /**
+   * WARNING: should not be used anymore!!!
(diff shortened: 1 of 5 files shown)


packages/manager/modules/pci/src/projects/project/quota/quota.service.js
```diff
@@ -9,6 +9,10 @@ export default class {
     this.OvhApiMe = OvhApiMe;
   }
 
+  /**
+   * WARNING: should not be used anymore!!!
+   * Need to use ovhPaymentMethod.getDefaultPaymentMethod instead.
+   */
```
FredericEspiau: https://jsdoc.app/tags-deprecated.html ?

marie-j: Moreover, does this log a message to be aware of it while developing ?

FredericEspiau: Ah, good idea

PR #2308, opened 2020-02-21
fix(iplb): manage links by region

Jisay:
ref: 
- DTRSD-8554

### DIFF
--- a/packages/manager/modules/iplb/src/home/iplb-home.controller.js
+++ b/packages/manager/modules/iplb/src/home/iplb-home.controller.js
@@ -1,4 +1,5 @@
 import filter from 'lodash/filter';
+import get from 'lodash/get';
 import set from 'lodash/set';
 import groupBy from 'lodash/groupBy';
 import head from 'lodash/head';
@@ -16,6 +17,7 @@ export default class IpLoadBalancerHomeCtrl {
     $state,
     $stateParams,
     $translate,
+    coreConfig,
     CucControllerHelper,
     CucCloudMessage,
     CucFeatureAvailabilityService,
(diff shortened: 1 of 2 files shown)


packages/manager/modules/iplb/src/home/iplb-home.controller.js
```diff
@@ -1,11 +1,14 @@
 import filter from 'lodash/filter';
+import get from 'lodash/get';
 import set from 'lodash/set';
 import groupBy from 'lodash/groupBy';
 import head from 'lodash/head';
 import isArray from 'lodash/isArray';
 import map from 'lodash/map';
 import values from 'lodash/values';
 
+import { Environment } from '@ovh-ux/manager-config';
```
antleblanc: Prefer using `coreConfig.getRegion` from `@ovh-ux/manager-core` module.

packages/manager/modules/iplb/src/iplb-url.constants.js
```diff
@@ -1,7 +1,17 @@
...
+    'https://www.ovh.com/manager/dedicated/#/billing/autoRenew?selectedType={serviceType}&searchText={serviceName}',
+  CA:
+    'https://ca.ovh.com/manager/#/billing/autorenew?selectedType={serviceType}&searchText={serviceName}',
+};
+
+export const CONTACTS_URL = {
+  EU:
+    'https://www.ovh.com/manager/dedicated/#/useraccount/contacts?tab=SERVICES&serviceName={serviceName}',
+  CA:
+    'https://ca.ovh.com/manager/#/useraccount/contacts?tab=SERVICES&serviceName={serviceName}',
```
marie-j: I'm not sure contacts are available in CA

antleblanc: It could be enhance into a separate Pull Request.

It will be also reviewed by the new URL bring by the Hub application.

Thanks for mentioning it!

PR #2315, opened 2020-02-24
feat: extract BillingService class

marie-j:
This will allow to share it between apps and modules

ref: MANAGER-4474
Signed-off-by: Marie JONES <marie.jones@corp.ovh.com>


ovh-cds:
CDS Report build-monorepository#9564.0 ✘  
* 
  * Build ✘

FredericEspiau:
Any reason to use `api-models` instead of `models` ? I mean it's models for the front isn't it ? :)

marie-j:
> Any reason to use `api-models` instead of `models` ? I mean it's models for the front isn't it ? :)

I was more focus on building the module than its name so I didn't put any extra thought to it but it makes sense to change

FredericEspiau:
@marie-j that's what code reviews are for I guess !

packages/manager/modules/models/BillingService/BillingService.class.js
```diff
@@ -3,6 +3,7 @@ import head from 'lodash/head';
 import includes from 'lodash/includes';
 import isNull from 'lodash/isNull';
 import snakeCase from 'lodash/snakeCase';
+import 'moment';
```
FredericEspiau: @cbourgois in lateral security position

yarn.lock
```diff
@@ -3236,11 +3236,6 @@ ansi-regex@^2.0.0:
   resolved "https://registry.yarnpkg.com/ansi-regex/-/ansi-regex-2.1.1.tgz#c3b33ab5ee360d86e0e628f0468ae7ef27d654df"
   integrity sha1-w7M6te42DYbg5ijwRorn7yfWVN8=
 
-ansi-regex@^3.0.0:
```
FredericEspiau: This is driving me crazy

marie-j: We definitely need to check why we have this strange behaviour

packages/manager/modules/models/BillingService/BillingService.class.js
marie-j: @FredericEspiau we could think on a refactoring for this model later on, here the purpose is just to allow reusability

FredericEspiau: Yeah it wasn't a criticism, you are entirely right to do this

marie-j:
Splitted commits but they appear in the wrong order due to https://help.github.com/en/github/committing-changes-to-your-project/why-are-my-commits-in-the-wrong-order

PR #2317, opened 2020-02-24
Feat/hub support

frenautvh:


packages/manager/modules/core/src/redirection/redirection.constants.js
```diff
@@ -21,6 +21,7 @@ export default {
     paymentMethod: `${managerRoot.EU}/manager/dedicated/${URI.paymentMethod}`,
     services: `${managerRoot.EU}/manager/dedicated/${URI.services}`,
     support: `${managerRoot.EU}/manager/dedicated/${URI.support}`,
+    ticket: `${managerRoot.EU}/manager/dedicated/#/support/tickets/:ticketId`,
```
FredericEspiau: Looks like you could abstract the `/manager/dedicated/` part too, can't you ?

ie

```js
${eu.manager.dedicated}/${URI.support}
```

for example

frenautvh: the problem is that it's just ```/manager``` for CA region

packages/manager/modules/hub/src/components/support/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,12 @@
+{
+  "hub_support_title": "Dernières demandes d'assitance",
```
FredericEspiau:
```suggestion
  "hub_support_title": "Dernières demandes d'assistance",
```

packages/manager/modules/hub/src/components/support/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,12 @@
...
+  "hub_support_see_more": "Voir tout",
+  "hub_support_read": "Lire",
+  "hub_support_state_open": "Ouvert",
+  "hub_support_state_closed": "Fermé",
+  "hub_support_state_unknown": "Inconnu",
+  "hub_support_account_management": "Gestion du compte",
+  "hub_support_need_help": "Besoin d'aide?",
```
FredericEspiau:
```suggestion
  "hub_support_need_help": "Besoin d'aide ?",
```

packages/manager/modules/hub/src/components/support/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,12 @@
...
+  "hub_support_need_help_more": "Des difficultés pour utiliser vos produits et services? Consultez dès maintenant nos guides.",
```
FredericEspiau:
```suggestion
  "hub_support_need_help_more": "Des difficultés pour utiliser vos produits et services ? Consultez dès maintenant nos guides.",
```

FredericEspiau: Why are you so mad with French today @frenautvh ?

packages/manager/modules/core/src/redirection/redirection.constants.js
FredericEspiau: `${eu.manager.dedicated}`=== `https://www.ovh.com/manager/dedicated`

and

`${ca.manager.dedicated}`=== `https://www.ovh.com/manager`

Is that an issue ?

packages/manager/modules/hub/src/components/support/controller.js
```diff
@@ -0,0 +1,56 @@
...
+    this.CORE_URLS = CORE_URLS;
+    this.RedirectionService = RedirectionService;
+    this.SessionService = SessionService;
+  }
+
+  $onInit() {
+    this.isLoading = true;
+    this.tickets = [];
+    return this.$q
+      .when(this.tickets ? this.tickets : this.fetchTickets())
```
JeremyDec: Hmm you're setting `this.tickets = []` above, so the test will always return true no ?

packages/manager/modules/hub/src/components/support/index.js
```diff
@@ -0,0 +1,19 @@
...
+import component from './component';
+
+import './index.scss';
+
+const moduleName = 'ovhManagerHubSupport';
+
+angular
+  .module(moduleName, [
+    'ngTranslateAsyncLoader',
+    'oui',
```
JeremyDec: Need `import ovh-ui-angular` ?

packages/manager/modules/hub/src/components/support/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,12 @@
+{
+  "hub_support_title": "Dernières demandes d'assistance",
+  "hub_support_see_more": "Voir tout",
+  "hub_support_read": "Lire",
+  "hub_support_state_open": "Ouvert",
+  "hub_support_state_closed": "Fermé",
+  "hub_support_state_unknown": "Inconnu",
+  "hub_support_account_management": "Gestion du compte",
+  "hub_support_need_help": "Besoin d'aide ?",
```
FredericEspiau:
```suggestion
  "hub_support_need_help": "Besoin d'aide ?",
```
non-breaking space

packages/manager/modules/hub/src/components/support/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,12 @@
...
+  "hub_support_need_help_more": "Des difficultés pour utiliser vos produits et services ? Consultez dès maintenant nos guides.",
```
FredericEspiau:
```suggestion
  "hub_support_need_help_more": "Des difficultés pour utiliser vos produits et services ? Consultez dès maintenant nos guides.",
```
Non breaking space

packages/manager/modules/hub/src/components/support/controller.js
```diff
@@ -0,0 +1,49 @@
...
+        }
+      })
+      .then(() => {
+        if (this.tickets.length <= 0) {
+          return this.fetchGuide().then((guideURL) => {
+            this.guideURL = guideURL;
+          });
+        }
+        return null;
+      })
```
FredericEspiau: You could also

```suggestion
      .then(() => isEmpty(this.tickets) && this.fetchGuide())
      .then((guideURL) => {
            this.guideURL = guideURL;
      })
```

packages/manager/modules/hub/src/components/support/controller.js
```diff
@@ -0,0 +1,49 @@
...
+    this.isLoading = true;
+    return this.$q
+      .when(this.tickets ? this.tickets : this.fetchTickets())
+      .then((tickets) => {
+        if (tickets) {
+          this.tickets = tickets
+            .slice(0, MAX_TICKETS_TO_DISPLAY)
+            .map((ticket) => new SupportTicket(ticket));
+        }
+      })
```
FredericEspiau:
```suggestion
      .then((tickets = []) => {
          this.tickets = tickets
            .slice(0, MAX_TICKETS_TO_DISPLAY)
            .map((ticket) => new SupportTicket(ticket));
      })
```

frenautvh: it's almost the same but would break if ticket is ```null``` i guess the correct way would be to check with ```if (Array.isArray(tickets) ```

packages/manager/modules/core/src/redirection/redirection.constants.js
frenautvh: yep refactoring will be done in another PR

packages/manager/modules/core/src/redirection/redirection.service.js
```diff
@@ -1,14 +1,33 @@
...
+        if (keyValidator.test(key)) {
+          url = url.replace(`:${key}`, encodeURIComponent(value));
+        }
+      });
+    }
+    return url;
+  }
+
+  getGuideURL(id) {
+    return this.SessionService.getUser().then(({ ovhSubsidiary }) => {
```
marie-j: Do you think maybe the `subsidiary` could be passed as a parameter ? Because for quite a few URLs we will have different values depending on the subs and we will have to fetch `/me` each time

packages/manager/modules/hub/src/components/support/controller.js
```diff
@@ -0,0 +1,46 @@
...
+      .finally(() => {
+        this.isLoading = false;
+      });
+  }
+
+  fetchGuide() {
+    return this.RedirectionService.getGuideURL('home');
+  }
+
+  fetchTickets() {
```
marie-j: I'm quite puzzled about this, on one hand I found a good idea that the component fetches its data if none is provided but on the other hand I'm wondering if it is not doing too much then handling that logic

packages/manager/modules/hub/src/components/support/ticket.class.js
```diff
@@ -0,0 +1,24 @@
...
+  getStateCategory() {
+    switch (this.state) {
+      case 'open':
+        return 'success';
+      case 'closed':
+        return 'info';
+      case 'unknown':
+        return 'warning';
+      default:
+        return 'error';
```
marie-j: Maybe we should reuse https://github.com/ovh/manager/blob/master/packages/manager/modules/support/src/tickets/ticket/ticket.class.js#L3 and add what its needed

packages/manager/modules/core/src/redirection/redirection.service.js
frenautvh: Yes i was not happy with the getGuideURL method we should have a single getURL method that handle all the cases. I'm going to refactor it

packages/manager/modules/hub/src/components/support/ticket.class.js
frenautvh: yep please see : https://github.com/ovh/manager/pull/2363

packages/manager/modules/hub/src/components/support/controller.js
FredericEspiau: It's doing too much but it may be premature to have the alternative, which is a generic way to inject data from the API into our components

packages/manager/modules/hub/src/components/support/template.html
```diff
@@ -0,0 +1,66 @@
+<div class="manager-hub-support" data-ng-if="!$ctrl.isLoading">
+    <div data-ng-if="$ctrl.tickets.length > 0">
+        <h3 data-translate="hub_support_title"></h3>
```
FredericEspiau: No `h2` above?

marie-j: @FredericEspiau the `h2` will be brought by the component's parent when integrated I guess. 
Imo it's quite complicated because components shouldn't always have a `h1` but it's hard to know if the headings hierarchy will be valid

FredericEspiau: Yes this is why I ask if there is a `h2` above because you neven know with `html`

At some point, to fix this, `html` implemented an outline algorithm but it didn't work properly so nobody actually implemented it: https://developer.mozilla.org/en-US/docs/Web/Guide/HTML/Using_HTML_sections_and_outlines

PR #2320, opened 2020-02-25
feat(veeam-backup): add backup module, backup page and offers component

radireddy:
Below features are added as part of this PR
1. New veeam backup page
2. Veeam backup home page
3. Veeam backup offers component
4. Price component


packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/price/price.component.js
```diff
@@ -0,0 +1,15 @@
+import controller from './price.controller';
```
radireddy: This component needs to be moved to modules folder as commented by Fred and Jeremy earlier. And also needs to be merged with exiting price component that Jeremy has developed.

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/offers/translations/Messages_fr_FR.json
```diff
@@ -11,5 +11,6 @@
   "dedicatedCloud_offer_upgrade_choice_backup_time": "Choix de l’heure de début des sauvegardes",
   "dedicatedCloud_offer_upgrade_long_term_retention": "11 points de rétention longue durée (GPS rétention)",
   "dedicatedCloud_offer_upgrade_proxy_backup": "1 backup proxy par hôte",
-  "dedicatedCloud_offer_current_offer": "Offre actuelle"
+  "dedicatedCloud_offer_current_offer": "Offre actuelle",
+  "dedicatedCloud_offer_price_per_vm": "/VM"
```
FredericEspiau:
```suggestion
  "dedicatedCloud_offer_price_per_vm": "{{price}}/VM"
```
or else the translators won't understand the context :)

packages/manager/apps/dedicated/client/app/app.js
```diff
@@ -231,6 +234,7 @@ angular
       });
 
       $state.defaultErrorHandler((error) => {
+        console.log(error);
```
radireddy: remove

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup-offer.class.js
```diff
@@ -0,0 +1,52 @@
+import get from 'lodash/get';
+import { BACKUP_OFFER_PREMIUM } from './backup.constants';
+
+export default class BackupOffer {
+  constructor(offer) {
```
marie-j: Maybe offer could be destructured ? That way we can easily know what property are expected and also we only assign the properties the class needs

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.class.js
```diff
@@ -0,0 +1,49 @@
...
+      ? this.restorePointInReport
+      : false;
+    this.diskSizeInReport = this.diskSizeInReport
+      ? this.diskSizeInReport
+      : false;
+    this.encryption = this.encryption ? this.encryption : false;
+    this.backupDurationInReport = this.backupDurationInReport
+      ? this.backupDurationInReport
+      : false;
+    this.sendReportToCustomeEmail = this.mailAddress !== null;
```
marie-j: Maybe use a getter for this 
```suggestion
   get sendReportToCustomerEmail() {
     return this.mailAddress !== null
    }
```

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.class.js
```diff
@@ -0,0 +1,49 @@
...
+} from './backup.constants';
+
+export default class Backup {
+  constructor(backup) {
+    Object.assign(this, backup);
+    this.initializeBackup();
+  }
+
+  initializeBackup() {
+    this.backupSizeInReport = this.backupSizeInReport
```
marie-j: Maybe this could be done with destructuring in the constructor and something like 
```suggestion
    this.backupSizeInReport = backupSizeInReport || false
```

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
```diff
@@ -0,0 +1,73 @@
...
+    $translate,
+    Alerter,
+    datacenterBackupService,
+  ) {
+    this.$anchorScroll = $anchorScroll;
+    this.$location = $location;
+    this.$translate = $translate;
+    this.Alerter = Alerter;
+    this.backup = {};
+    this.alerter = Alerter;
```
marie-j: alerter is defined twice https://github.com/ovh/manager/pull/2320/files#diff-f1f66957f52691f22d275f3fcec7a002R15

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
```diff
@@ -0,0 +1,73 @@
...
+    $location,
+    $translate,
+    Alerter,
+    datacenterBackupService,
+  ) {
+    this.$anchorScroll = $anchorScroll;
+    this.$location = $location;
+    this.$translate = $translate;
+    this.Alerter = Alerter;
+    this.backup = {};
```
marie-j: This is usually done in the `$onInit` hook

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
```diff
@@ -0,0 +1,73 @@
...
+
+  $onInit() {
+    this.alertMain = 'app.dedicatedClouds.datacenter.backup';
+    this.loader = {
+      updatingCapabilities: false,
+    };
+  }
+
+  scrollToTop() {
+    this.$location.hash('backupMainAlert');
```
marie-j: I'm not sure the component should be aware of the `$location`

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
```diff
@@ -0,0 +1,73 @@
...
+      backupDurationInReport: this.backup.backupDurationInReport,
+      backupSizeInReport: this.backup.backupSizeInReport,
+      diskSizeInReport: this.backup.diskSizeInReport,
+      fullDayInReport: this.backup.fullDayInReport,
+      restorePointInReport: this.backup.restorePointInReport,
+      mailAddress: this.backup.mailAddress,
+      backupOffer: this.backup.backupOffer,
+    };
+    this.datacenterBackupService
+      .updateBackupCapabilities(this.productId, this.datacenterId, capabilities)
```
marie-j: Why not just pass `this.backup` ?

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.constants.js
```diff
@@ -0,0 +1,22 @@
+export const BACKUP_OFFER_LEGACY = 'legacy';
+export const BACKUP_OFFER_PREMIUM = 'premium';
+export const BACKUP_OFFER_ADVANCED = 'advanced';
+export const BACKUP_OFFER_CLASSIC = 'classic';
```
marie-j: Do we need to copy the enum values ? Maybe if needed we could generate them, `BACKUP_OFFER_${offer.toUpperCase()}`

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
```diff
@@ -0,0 +1,103 @@
+<h1 data-translate="datacenter_backup_title"></h1>
+<div class="datacenter-backup" data-ui-view>
+    <div data-ovh-alert="{{ $ctrl.alertMain }}" id="backupMainAlert"></div>
+    <oui-switch data-model="true" data-on-change="$ctrl.deactivate()">
```
marie-j: A static model ?

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
```diff
@@ -0,0 +1,103 @@
...
+        <h4 class="mb-3" data-translate="datacenter_backup_choose_offers"></h4>
+        <dedicated-cloud-datacenter-backup-offers
+            data-backup-offers="$ctrl.backupOffers"
+            data-backup="$ctrl.backup"
+            data-preview-mode="false"
+            data-user="$ctrl.currentUser"
+        ></dedicated-cloud-datacenter-backup-offers>
+        <oui-button
+            data-variant="primary"
+            data-text="{{ 'datacenter_backup_update_offer' | translate }}"
```
marie-j: I think `text` attribute is deprecated, we should use a nested `span` instead

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.less
```diff
@@ -0,0 +1,14 @@
...
+  section {
+    border-top: 1px solid #bbbdbf;
+    margin: 25px 0;
+  }
+
+  .font-weight-normal .oui-checkbox__label {
+    font-weight: normal;
+  }
+
+  .font-weight-bold .oui-field .oui-label {
```
marie-j: `font-weight-bold` is a bootstrap utility, why do we need to override it ?

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.service.js
```diff
@@ -0,0 +1,166 @@
...
+          cartItem.itemId,
+          'offer_type',
+          offerType,
+        ).$promise,
+      ])
+      .then(() => cartItem);
+  }
+
+  assignCart(cart) {
+    return this.WucOrderCartService.assignCart(cart.cartId);
```
marie-j: Why add another layer above `WucOrderCartService` ? I think it already provides the `/order/cart` api abstraction and it seems like there is no more specific logic added

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.service.js
```diff
@@ -0,0 +1,166 @@
...
+  addOptionToCart(option, cart, productId) {
+    return this.WucOrderCartService.addProductServiceOptionToCart(
+      cart.cartId,
+      'privateCloud',
+      productId,
+      {
+        cartId: cart.cartId,
+        duration: get(option, 'prices[0].duration'),
+        planCode: option.planCode,
+        pricingMode: get(option, 'prices[0].pricingMode'),
```
marie-j: Please use the `capacity` property to get prices, we already had the case where some prices were added and so the first one wasn't the one to take anymore

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/price/translations/Messages_pl_PL.json
```diff
@@ -0,0 +1,8 @@
+{
+  "price_ht_label": "{{price}} netto",
+  "price_ttc_label": "{{price}} brutto",
+  "price_free": "W ofercie",
+  "price_gst_excl_label": "{{price}} netto",
+  "price_gst_incl_label": "{{price}} brutto",
+  "price_per_month": "/m-c"
```
marie-j: Just a reminder that translations added shoud only be `fr_FR`

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/dedicatedCloud-datacenter-backup.routes.js
```diff
@@ -1,7 +1,7 @@
 angular.module('App').config(
   /* @ngInject */ ($stateProvider) => {
-    $stateProvider.state('app.dedicatedClouds.datacenter.backup', {
-      url: '/backup',
+    $stateProvider.state('app.dedicatedClouds.datacenter.backup-old', {
+      url: '/backup-old',
```
marie-j:
```suggestion
      url: '/backup-legacy',
```
sounds better

packages/manager/apps/dedicated/package.json
```diff
@@ -125,7 +125,7 @@
     "oclazyload": "^1.1.0",
     "office-ui-fabric-core": "^11.0.0",
     "ovh-angular-responsive-tabs": "^4.0.0",
-    "ovh-api-services": "^9.38.0",
+    "ovh-api-services": "^9.39.0",
```
marie-j: This should be done for all packages using `ovh-api-services`
(`yarn upgrade-interactive --latest ovh-api-services`)

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.constants.js
FredericEspiau: - An `enum` makes more sense and is easier to read imo 🙂 
- I'm becoming less and less a fan of `BACKUP_OFFER_ + something` as time goes by because I means I will never be able to `ctrl + f` all the instances of `BACKUP_OFFER_CLASSIC` for example

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/offers/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,16 @@
+{
+  "dedicatedCloud_offer_classic": "Veeam Backup Standard",
```
radireddy:
```suggestion
  "dedicatedCloud_datacenter_backup_offer_classic": "Veeam Backup Standard",
```

all translations must start with "dedicatedCloud_datacenter_backup_"

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup-offer.class.js
ganeshkumar1989: `offer` is the object returned by the API and all the properties are required. Could you please elaborate on why de-structuring is required?

cbourgois: A long time ago, we used to declare class like this, but in some cases it fails working with the datagrid because the Object was initialized with data containing a `$promise` property.
 
So we updated our code to declare only properties needed like this:
```suggestion
  constructor({propertyA, propertyB, propertyC}) {
    Object.assign(this, {propertyA, propertyB, propertyC});
  }
```
Thanks,

ovh-cds:
CDS Report build-monorepository#9807.0 ✘  
* 
  * Build ✘

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
ganeshkumar1989: These information are not available in the backup object

ganeshkumar1989: `capabilities` is being passed to the API as such. Passing `this.backup` would result in a API call with data more that what is required

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.service.js
ganeshkumar1989: Its actually not required. But now the code returns the cart object again, to enabling chaining of calls

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/dedicatedCloud-datacenter-backup.routes.js
ganeshkumar1989: This is just a temporary rename and these files will be removed at the end of dev for the new ranges. Further, we have another route for legacy, which we are working on

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup-offer.class.js
ganeshkumar1989: Makes sense. Thanks for the explanation @cbourgois

packages/manager/apps/dedicated/client/app/app.js
ganeshkumar1989: removed

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.class.js
ganeshkumar1989: This is done

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
ganeshkumar1989: Additional definition removed. Thanks

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
ganeshkumar1989: This is not required and has been removed

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
ganeshkumar1989: This is moved to a span

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.less
ganeshkumar1989: This is not required to be over-ridden. This is fixed

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.service.js
ganeshkumar1989: This is done. Thanks for the input

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/price/translations/Messages_pl_PL.json
ganeshkumar1989: Other language translations removed

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.constants.js
```diff
@@ -0,0 +1,22 @@
...
+
+export const BACKUP_STATE_DISABLED = 'disabled';
+export const CATALOG_INFO = {
+  ADDON_FAMILY_NAME: 'backup',
+  ADDON_PLAN_CODE: 'pcc-option-backup-managed',
+  ADDON_BACKUP_JOB_OFFER_FAMILY_NAME: 'backupjob-offer',
+};
+export const VEEAM_TARIFF_DETAILS_URL = 'https://www.ovhcloud.com/';
```
cbourgois: No dependencies on `ovhSubsidiary` or `region` ?

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
ganeshkumar1989: This is not completely implemented. Its a part of https://github.com/ovh/manager/pull/2350
:)

packages/manager/apps/dedicated/package.json
ganeshkumar1989: Like @FredericEspiau said, it can be done at the end. Thanks for pointing this out @marie-j  :)

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup-offer.class.js
ganeshkumar1989: This is done :)

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.class.js
ganeshkumar1989: This is done :)

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
ganeshkumar1989: > This is due to use of `$anchorScroll` I guess. @radireddy I know I did recommend to use `$anchorScroll` in another PR.
> Can you try to do please:

This works and is incorporated in the code @JeremyDec . Thanks a lot!

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/offers/translations/Messages_fr_FR.json
ganeshkumar1989: This is done

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
```diff
@@ -0,0 +1,71 @@
...
+    const capabilities = {
+      backupDurationInReport: this.backup.backupDurationInReport,
+      backupSizeInReport: this.backup.backupSizeInReport,
+      diskSizeInReport: this.backup.diskSizeInReport,
+      fullDayInReport: this.backup.fullDayInReport,
+      restorePointInReport: this.backup.restorePointInReport,
+      mailAddress: this.backup.mailAddress,
+      backupOffer: this.backup.backupOffer,
+    };
+    this.datacenterBackupService
```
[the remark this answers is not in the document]
ganeshkumar1989: This is done. Thanks :)

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
```diff
@@ -0,0 +1,111 @@
+<h1 data-translate="datacenter_backup_title"></h1>
+<div data-ui-view>
+    <div class="datacenter-backup">
+        <div data-ovh-alert="{{ $ctrl.alertMain }}" id="backupMainAlert"></div>
```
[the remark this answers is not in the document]
ganeshkumar1989: This is done

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.constants.js
ganeshkumar1989: We got the urls just recently for each subsidiary and the same has been updated. Thanks

packages/manager/apps/dedicated/package.json
ganeshkumar1989: Solved after the rebase from develop

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup-offer.class.js
```diff
@@ -0,0 +1,80 @@
...
+      tax,
+    });
+  }
+
+  isPremium() {
+    return get(this, 'offerName', null) === BACKUP_OFFER_PREMIUM;
+  }
+
+  hasCustomReport() {
+    return get(this, 'customReport', false);
```
FredericEspiau: I think there are two issues:
- the function is named `has` but returns the content of the property, so maybe we should name it `getCustomReport` instead
- also, why do this ? These are mostly getters, users could do

```js
const backupOffer = new BackupOffer({ customReport: "this is the custom reports" });
if (has(backupOffer, 'customReport')) {
    console.log(backupOffer.customReport);
}
```

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.class.js
```diff
@@ -0,0 +1,69 @@
...
+      scheduleHour,
+      state,
+      vmwareVmId,
+    });
+
+    this.backupSizeInReport = backupSizeInReport || false;
+    this.fullDayInReport = fullDayInReport || false;
+    this.restorePointInReport = restorePointInReport || false;
+    this.diskSizeInReport = diskSizeInReport || false;
+    this.encryption = encryption || false;
```
FredericEspiau: Looks like those properties can contain more than one type, maybe we should only have one type for each property, it would make code easier to read I think

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.constants.js
FredericEspiau: Why did you resolve ? Will you do it or not ?

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
```diff
@@ -0,0 +1,111 @@
+<h1 id="dedicatedCloud_datacenter_backup_header" data-translate="dedicatedCloud_datacenter_backup_title"></h1>
+<div data-ui-view>
+    <div class="datacenter-backup">
+        <div data-ovh-alert="{{ ::$ctrl.alertMain }}"></div>
+        <oui-switch data-model="true" data-on-change="$ctrl.deactivate()">
+        </oui-switch>
+        <span data-translate="dedicatedCloud_datacenter_backup_enabled"></span>
+
+        <section class="datacenter-backup-offers">
+            <h4
```
FredericEspiau: No `h1`, no `h2` and not `h3` above

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
```diff
@@ -0,0 +1,111 @@
...
+                class="mb-3"
+                data-translate="dedicatedCloud_datacenter_backup_choose_offers"
+            ></h4>
+            <dedicated-cloud-datacenter-backup-offers
+                data-backup-offers="$ctrl.backupOffers"
+                data-backup="$ctrl.backup"
+                data-preview-mode="false"
+                data-user="$ctrl.currentUser"
+            ></dedicated-cloud-datacenter-backup-offers>
+            <oui-button data-variant="primary">
```
FredericEspiau: Nothing happens on clicking on the button ?

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.module.js
```diff
@@ -0,0 +1,33 @@
...
+    'ngOvhSwimmingPoll',
+    'ngTranslateAsyncLoader',
+    'oui',
+    'ovh-api-services',
+    'pascalprecht.translate',
+    backupOffers,
+    backupNew,
+  ])
+  .config(routing)
+  .component('ovhManagerDedicatedCloudBackup', component)
```
FredericEspiau:
```suggestion
  .component('dedicatedCloudBackup', component)
```

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/new/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,11 @@
...
+  "dedicatedCloud_datacenter_backup_new_description2": "Vous pouvez choisir entre 3 nouvelles offres pour protéger vos machines virtuelles (VM) : <strong>Standard</strong>, <strong>Advanced</strong> ou <strong>Premium</strong>.",
+  "dedicatedCloud_datacenter_backup_new_tariff_url": "Consulter en détail la nouvelle gamme Veeam Backup",
+  "dedicatedCloud_datacenter_backup_new_choose_offer": "Choix de la solution de sauvegarde",
+  "dedicatedCloud_datacenter_backup_new_choose_offer_action": "Choisir l'offre",
+  "dedicatedCloud_datacenter_backup_new_validate_conditions": "Validation des conditions d’utilisation",
+  "dedicatedCloud_datacenter_backup_new_activate_offer_action": "Activer l’offre",
+  "dedicatedCloud_datacenter_backup_new_create_success": "Votre offre a bien été activée. Vous êtes maintenant abonné·e à la solution Veeam Backup {{ offerType }}. Si vous souhaitez changer d’offre, merci d’attendre 15 jours pour tout autre changement.<br/><a href='{{ guideLink }}' target='_blank' class='oui-link_icon'>Consulter le guide sur Veeam Backup afin de configurer votre sauvegarde <span class='oui-icon oui-icon-arrow-right_extra-thin' aria-hidden='true'></span></a>",
```
FredericEspiau: Please split into two translations instead, `br` has a semantic meaning: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/br

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/new/new.module.js
```diff
@@ -0,0 +1,24 @@
...
+import offers from '../components/offers';
+import offerDetails from '../components/offer-details';
+import routing from './new.routing';
+import termsOfUse from '../components/termsOfUse';
+
+const moduleName = 'ovhManagerDedicatedCloudBackupNewModule';
+
+angular
+  .module(moduleName, [
+    offers,
```
FredericEspiau: No translations, no ui-router, no ui-kit ?

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/new/new.html
```diff
@@ -0,0 +1,58 @@
...
+        ></span>
+        <span
+            class="oui-icon oui-icon-arrow-right_extra-thin"
+            aria-hidden="true"
+        ></span>
+    </a>
+</strong>
+<oui-stepper class="mt-3" data-on-finish="$ctrl.orderBackup()" data-current-index="$ctrl.data.currentStep">
+    <oui-step-form
+        data-header="{{ 'dedicatedCloud_datacenter_backup_new_choose_offer' | translate }}"
```
FredericEspiau:
```suggestion
        data-header="{{:: 'dedicatedCloud_datacenter_backup_new_choose_offer' | translate }}"
```
We should have this everywhere  when it's required

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/dedicatedCloud-datacenter-backup.routes.js
FredericEspiau: I've unresolved so we know that this is to be done 🙂

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/termsOfUse/terms-of-use.html
```diff
@@ -0,0 +1,43 @@
+<p
+    data-ng-bind="::('dedicatedCloud_datacenter_backup_terms_of_use_description1_' + $ctrl.mode) | translate: {
+        datacenter: $ctrl.datacenterId
+    }"
+></p>
+<p
+    data-ng-bind="::('dedicatedCloud_datacenter_backup_terms_of_use_description2_' + $ctrl.mode) | translate"
+></p>
+<strong class="mb-3">
```
FredericEspiau: `strong` is a `display: inline` property so I doubt adding a `.mb-3` will do anything

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/termsOfUse/terms-of-use.html
```diff
@@ -0,0 +1,43 @@
...
+    <a
+        target="_blank"
+        rel="noopener"
+        class="oui-link_icon"
+        data-ng-href="{{ ::$ctrl.backupTariffUrl }}"
+    >
+        <span
```
FredericEspiau:
```suggestion
        <strong
```

instead :)

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/price/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,8 @@
+{
```
FredericEspiau: Please use https://github.com/ovh/manager/tree/develop/packages/manager/modules/catalog-price instead

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/price/price.component.js
FredericEspiau: Yes you can use https://github.com/ovh/manager/tree/develop/packages/manager/modules/catalog-price it has been merged !

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/offers/index.js
```diff
@@ -0,0 +1,14 @@
+import angular from 'angular';
+
+import component from './component';
+import price from '../price';
+import './styles.less';
+
+const moduleName = 'ovhManagerDedicatedCloudBackupOffers';
+
+angular
+  .module(moduleName, [price])
```
FredericEspiau: Translations, ui-router, ui-kit

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.service.js
```diff
@@ -0,0 +1,172 @@
+import find from 'lodash/find';
```
FredericEspiau: Please use `https://github.com/ovh/manager/tree/develop/packages/manager/modules/product-offers` instead

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.constants.js
marie-j: @FredericEspiau I agree about readability, but we are basically copy/pasting enum from the API, meaning if an enum is added we have to add it in the constant file which means more maintenance

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
marie-j: Ok then maybe we could `_.pick` the properties needed from `this.backup` instead

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/offers/translations/Messages_fr_FR.json
ganeshkumar1989: But the price is being displayed by the `catalog-price` component from https://github.com/ovh/manager/tree/develop/packages/manager/modules/catalog-price

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.constants.js
ganeshkumar1989: Sorry, I resolved it by mistake. But we we do not have the enums defined, and if a enum is modified on the API side, we will have to change it in the entire code-base (we will have to do it only in one place if the enum is changed, for example, they rename `classic` to `veeam-classic` later on). IMO, having enums increases readability by defining certain things more explicitly

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.class.js
ganeshkumar1989: These are all boolean properties returned by the API. They can also be null sometimes (which should be taken as false) and here we convert them to boolean values (null becomes false)

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
ganeshkumar1989: There is a `h1` at the top of the page. `h2` or `h3` makes the text bigger than the required size

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
ganeshkumar1989: This is being implemented still :)

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/price/price.component.js
ganeshkumar1989: This is done

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
ganeshkumar1989: This is done. Thanks

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup-offer.class.js
ganeshkumar1989: You are right @FredericEspiau ... We were not using the destructing in the class's constructor, and hence we had to expose getter methods to explicitly tell the user about the properties available. Now that we are using the object destructing in the constructor, the available properties are clearly visible to the user and these getters are no longer required and have been removed. Thanks

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.module.js
ganeshkumar1989: This is done

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/price/translations/Messages_fr_FR.json
ganeshkumar1989: This is done

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/new/translations/Messages_fr_FR.json
ganeshkumar1989: The link was not required. Hence it has been removed. Thanks

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/new/new.module.js
ganeshkumar1989: Added the required dependencies

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/new/new.html
ganeshkumar1989: This is done in all places where required :)

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/termsOfUse/terms-of-use.html
ganeshkumar1989: You are right... Changed this tag to a `<p>`

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/termsOfUse/terms-of-use.html
ganeshkumar1989: This is done

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/components/offers/index.js
ganeshkumar1989: Added the required dependencies

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.service.js
ganeshkumar1989: Thanks for letting me know about this component. I had a look at it and I see that the component is capable of displaying the various pricing for an addon. But our requirement is that we display different addons for the user to choose 1 from. Unfortunately, this component would not fit our requirement

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/dedicatedCloud-datacenter-backup.routes.js
ganeshkumar1989: The old files have been removed at https://github.com/ovh/manager/pull/2385
Thanks :)

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
ganeshkumar1989: The upgrade offers feature has been implemented and merged. Thanks

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
ganeshkumar1989: The PR has been merged and this is complete now

ovh-cds:
CDS Report build-monorepository#9951.0 ✘  
* 
  * Build ✘

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.class.js
```diff
@@ -0,0 +1,81 @@
...
+  isAdvanced() {
+    return this.backupOffer === BACKUP_OFFER_ADVANCED;
+  }
+
+  isPremium() {
+    return this.backupOffer === BACKUP_OFFER_PREMIUM;
+  }
+
+  setState(state) {
+    this.state = state;
```
marie-j: Why not directly use the setter ?

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
marie-j: Maybe a class should be used if this is for style purpose

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.routing.js
```diff
@@ -0,0 +1,117 @@
...
+              'app.dedicatedClouds.datacenter.backup',
+            );
+          });
+        }
+        return promise;
+      },
+      scrollToTop: () => () =>
+        document
+          .getElementById('dedicatedCloud_datacenter_backup_header')
+          .scrollIntoView(),
```
marie-j: Just a wonder, maybe we could have an anchor on the page and go to that anchor ? Instead of having to use DOM selectors

packages/manager/apps/dedicated/package.json
```diff
@@ -25,6 +25,7 @@
     "@bower_components/angular-ui-utils": "angular-ui/ui-utils#~0.2.3",
     "@ovh-ux/manager-banner": "^1.1.3",
     "@ovh-ux/manager-beta-preference": "^1.0.0",
+    "@ovh-ux/manager-catalog-price": "^0.0.0 || ^1.0.0",
```
marie-j: If component has been released we can  
```suggestion
    "@ovh-ux/manager-catalog-price": "^1.0.0",
```

ovh-cds:
CDS Report deploy_uxci_us#10146.0 ✘  
* deploy_uxci
  * deploy_uxci ✘

ovh-cds:
CDS Report deploy_uxci_us#10249.0 ✘  
* deploy_uxci
  * deploy_uxci ✘

ovh-cds:
CDS Report deploy_uxci_eu#10325.0 ✘  
* deploy_uxci
  * deploy_uxci ✘

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
```diff
@@ -15,6 +23,22 @@ export default class {
     this.loader = {
       updatingCapabilities: false,
     };
+    this.$timeout(() => this.createProgressAlert());
```
JeremyDec: Is there a reason to trigger it after a delay? In the constructor also?

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
```diff
@@ -5,26 +5,21 @@
...
-                data-on-change="$ctrl.goToDeleteBackup()"
-            >
-            </oui-switch>
+            <div class="oui-switch">
+                <label class="oui-switch__label-container">
+                    <input
+                        type="checkbox"
+                        class="oui-switch__checkbox"
+                        checked="checked"
+                        onchange="this.checked=true"
```
JeremyDec: Native events won't be digested by angular ;)
```suggestion
                        ng-change="this.checked=true"
```

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/upgrade/translations/Messages_fr_FR.json
```diff
@@ -3,11 +3,11 @@
...
+  "dedicatedCloud_datacenter_backup_upgrade_description_features": "Vous aurez désormais les paramètres suivants :",
   "dedicatedCloud_datacenter_backup_upgrade_description_confirmation": "Êtes-vous sûr(e) de vouloir changer d’offre ?",
   "dedicatedCloud_datacenter_backup_upgrade_description_billing": "Vous serez facturé(e) selon le nombre et la taille des machines virtuelles sauvegardées.",
-  "dedicatedCloud_datacenter_backup_upgrade_success_1": "Votre offre de sauvegarde a bien été activée. Vous êtes maintenant abonné à l'offre {{offer}}.",
-  "dedicatedCloud_datacenter_backup_upgrade_success_2": "Si vous souhaitez à nouveau Offre de sauvegarde, merci d’attendre 15 jours.",
+  "dedicatedCloud_datacenter_backup_upgrade_replication_condition": "Je valide la réplication des données",
+  "dedicatedCloud_datacenter_backup_upgrade_success_1": "La demande de changement de votre offre Veeam {{actualOffer}} à {{newOffer}} est en cours et prendra un peu de temps.",
+  "dedicatedCloud_datacenter_backup_upgrade_success_2": "Vous pouvez suivre l'avancement de l'opération <a href='{{operationsUrl}}'>ici</a>.",
```
JeremyDec: FYI, you can do:
```suggestion
  "dedicatedCloud_datacenter_backup_upgrade_success_2": "Vous pouvez suivre l'avancement de l'opération <a href=\"{{operationsUrl}}\">ici</a>.",
```
;)

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
ganeshkumar1989: This method creates the alerts required for the page. If done without the `$timeout`, the messages are created before the dom is fully rendered and it causes the messages to not be displayed

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
ganeshkumar1989: The requirement is that the switch should appear enabled always and we should not be able to switch it off. When the user clicks on it, we display a pop-up to deactivate the backup. The above code just sets the ui state of the checkbox to checked (and does not change the model). So it is not in angular scope. The `this.checked` is the native checkbox's checked property. There is no model associated with this control and we just need the on-click to trigger the pop-up

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.html
ganeshkumar1989: The class has been added. Thanks for the suggestion :)

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.class.js
ganeshkumar1989: The method was used earlier, since we did not know what properties the class had. But now, since we are destructuring in the constructor, we can set the property value directly and even a setter would not be needed. Please let me know your thought

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.routing.js
ganeshkumar1989: I agree with you that the DOM selectors have been removed. Since the code has been moved to the routes, am now using the `$location.hash()` and `$anchorScroll()`. What do you think about this?

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/upgrade/translations/Messages_fr_FR.json
ganeshkumar1989: This is done. Thanks for the suggestion :)

packages/manager/apps/dedicated/package.json
ganeshkumar1989: The version has been updated

JeremyDec:
@radireddy / @varun257 there are conflicts

packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.controller.js
JeremyDec: Maybe handling this in a `$postLink` hook would work? (PR is opened since more than one month, so no matter if you let this as is)

JeremyDec:
@radireddy / @varun257 conflict

PR #2331, opened 2020-02-25
Feat/hub product listing component

frenautvh:
As discussed, you can either instantiate the component :

- with an api route, results will be raw from api :
```
<ovh-manager-hub-user-product-tile data-product="'/hosting/web/{serviceName}'">
</ovh-manager-hub-user-product-tile>
```

- with a pre-configured product id, columns will be formatted :
```
<ovh-manager-hub-user-product-tile data-product="'domain'">
</ovh-manager-hub-user-product-tile>
```

Currently only 2 products configuration have been added (domain & web hosting). Other configuration might be added later depending on the needs

### Includes 
- [x] https://github.com/ovh/manager/pull/2400
- [x] https://github.com/ovh/manager/pull/2404


packages/manager/modules/core/src/redirection/redirection.service.js
```diff
@@ -7,8 +8,17 @@ export default class RedirectionService {
     this.coreConfig = coreConfig;
   }
 
-  getURL(id) {
+  getURL(id, params = {}) {
```
FredericEspiau: Could you give an example using jsdoc please ?

packages/manager/modules/hub/src/components/product-list/config/domain.config.js
```diff
@@ -0,0 +1,50 @@
+export default {
+  route: '/domain/{domain}',
+  columns: [
```
FredericEspiau: Looks like transclusion with extra steps

```js
 <ovh-manager-hub-user-product-tile how-to-go-to-product-page="$ctrl.goToDomain">
    <ovh-manager-hub-user-product-tile-column>
        <span data-translate="'hub_product_listing_offer_' + $row.offer"></span>
    </ovh-manager-hub-user-product-tile-column>

...and so on
</ovh-manager-hub-user-product-tile>
```

packages/manager/modules/hub/src/components/product-list/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,23 @@
+{
+  "hub_product_listing_domain": "Domaine",
+  "hub_product_listing_offer": "Offre",
+  "hub_product_listing_expiration_date": "Date d'expiration",
+  "hub_product_listing_offer_diamond": "Diamant",
+  "hub_product_listing_offer_gold": "Or",
+  "hub_product_listing_offer_platinum": "Platinum",
+  "hub_product_listing_dns_type": "DNS",
+  "hub_product_listing_dns_type_hosted": "Hégerbé",
```
FredericEspiau:
```suggestion
  "hub_product_listing_dns_type_hosted": "Hébergé",
```
😆

packages/manager/modules/hub/src/components/product-list/config/domain.config.js
frenautvh: no, those are the columns passed to the oui-datagrid 'columns' attribute ; this is how dynamic columns are handled in the oui-datagrid

packages/manager/modules/core/src/redirection/redirection.constants.js
```diff
@@ -7,6 +7,8 @@ const managerRoot = {
 const URI = {
   billing: '#/billing/history',
   contacts: '#/contacts/services',
+  domain: '#/configuration/domain/:domain',
```
marie-j: Those are already handled by the `service` 2API and accessible in `/hub/billingServices` payload. I think we should avoid duplicating those as it will be a pain to maintain if we have them in different places

packages/manager/modules/hub/src/components/product-list/controller.js
```diff
@@ -0,0 +1,92 @@
...
+      .get(this.endpoint, {
+        headers: {
+          'X-Pagination-Mode': 'CachedObjectList-Pages',
+          'X-Pagination-Size': 50000,
+        },
+      })
+      .then(({ data }) => data);
+  }
+
+  fetchProductsBilling(products) {
```
marie-j: I don't think the component should make any calls to add logic to its data. 
Also we already have this call made in the root state app

packages/manager/modules/hub/src/components/product-list/controller.js
```diff
@@ -0,0 +1,92 @@
...
+          'X-Pagination-Mode': 'CachedObjectList-Pages',
+          'X-Pagination-Size': 50000,
+        },
+      })
+      .then(({ data }) => data)
+      .then((services) => {
+        return map(products, (product) => {
+          const productId = get(product, this.resource);
+          const productService = find(services, (service) => {
+            return get(service, 'resource.name') === productId;
```
marie-j: Why not use a filter header ?

packages/manager/modules/hub/src/components/product-list/controller.js
frenautvh: We need to link the billing part to the product in order to display product's infos & service expiration on the same row.
I see that the hub is returning only the first 4 results of billing services. I have two options : making a call to /billing/autorenew/services here or adding an option parameter to the hub 2api endpoint to fetch all the billing infos. Either way i'll have to do some logic to link the billing service to the product.
What do you think?

frenautvh: And i'll also have to make a call to either /billing/autorenew/services or /hub here because the root hub call in the app won't fetch all the billingServices

marie-j: @frenautvh The url is agreggated to `/services` call in 2API, It has been taken in account in https://github.com/ovh/manager/pull/2400

ovh-cds:
CDS Report build-monorepository#10145.0 ✘  
* 
  * Build ✘

marie-j:
Rebased this PR on `develop` which is an aggregation of PR

marie-j:
I've rewritten history

PR #2363, opened 2020-02-28
refactor(support): move support tickets to manager-models

frenautvh:
extract ticket & ticket.constant from manager-support to manager-models (in order to reuse it in manager-hub)


ovh-cds:
CDS Report installation#9809.0 ✘  
* Stage 1
  * Install ✘

packages/manager/modules/support/package.json
```diff
@@ -17,6 +17,7 @@
   },
   "peerDependencies": {
     "@ovh-ux/manager-core": "^7.6.2",
+    "@ovh-ux/manager-models": "^0.0.0 || ^1.0.0",
```
marie-j: I'm not sure it is needed as a peer in fact, it's not an `angularjs` module

frenautvh: damn i'm so used to work with angularjs modules that i forgot it wasn't one!

packages/manager/modules/models/support/ticket/ticket.class.js
```diff
@@ -1,10 +1,15 @@
...
+import { STATUS } from './ticket.constants';
 
-export default class SupportTicket {
+export default class {
   constructor(ticket) {
     Object.assign(this, ticket);
   }
 
+  static get Status() {
+    return { ...STATUS };
```
FredericEspiau: I don't get what this is supposed to do

frenautvh: maybe not, but some people like having a lots of files ^^

PR #2364, opened 2020-02-28
build(support): ovh-api-service peer deps

frenautvh:


frenautvh:
> Good catch!
> 
> I'm wondering if it's not the case for other application since #2299 has been merged.

yes probably,

i think the issue is that ovh-api-services is broken it should import ng-ovh-api-wrappers and swimming-poll, i'm going to fix it an open a PR

frenautvh:
> > Good catch!
> > I'm wondering if it's not the case for other application since #2299 has been merged.
> 
> yes probably,
> 
> i think the issue is that ovh-api-services is broken it should import ng-ovh-api-wrappers and swimming-poll, i'm going to fix it an open a PR

PR is here : https://github.com/ovh-ux/ovh-api-services/pull/285

once it's merged i should close this one and simply bump ovh-api-services

antleblanc:
@frenautvh [`ovh-api-services@9.39.1`](https://github.com/ovh-ux/ovh-api-services/blob/master/CHANGELOG.md#9391-2020-02-28) has been release and published :tada:.

PR #2214, opened 2020-02-07
feat(modules): add product offers module

JeremyDec:
### Feature 🚀
New module to provide: 
- Components to handle duration and payment steps
- Service to get information for detach actions

Signed-off-by: Jérémy De-Cesare <jeremy.de-cesare@corp.ovh.com>
Co-authored-by: Frédéric Espiau <frederic.espiau@corp.ovh.com>

PR #2368, opened 2020-02-28
feat(modules.product.offers): display renew extra pricing even if free

JeremyDec:
ref: MANAGER-3328

related to #2214 

### DIFF
--- a/packages/manager/modules/product-offers/src/pricing/pricing.class.js
+++ b/packages/manager/modules/product-offers/src/pricing/pricing.class.js
@@ -3,7 +3,7 @@ import moment from 'moment';
 
 import GenericPricingHelper from './generic-pricing.helper';
 
-import { ISO_DURATION_FORMAT } from './pricing.constants';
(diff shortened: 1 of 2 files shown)


packages/manager/modules/product-offers/src/pricing/pricing.class.js
```diff
@@ -46,4 +46,8 @@ export default class Pricing {
   hasExtraPricing() {
     return !!this.extraPricing;
   }
+
+  mustBeDisplayed() {
```
frenautvh: This shouldn't be part of pricing class. "Must be displayed" in which context ? This has no meaning at the product level.

The method should be called something like ```hasRenew ``` regarding it's body

JeremyDec: Sure, the method is renamed

JeremyDec:
Approved by @marie-j  and @FredericEspiau before force-push.
Missing a word in method description. Now merging

PR #2374, opened 2020-03-02
fix(instances.add): prevent error if call to /credit crash

marie-j:
ref: DTRSD-9035


packages/manager/modules/pci/src/projects/project/instances/add/add.routing.js
```diff
@@ -25,6 +25,7 @@ export default /* @ngInject */ ($stateProvider) => {
                   ? deals
                   : null,
               )
+              .catch(() => null)
```
FredericEspiau: No error message ?

marie-j: No this is to prevent instance creation crashing just to display an offer. 
An error message could indeed improve this but currently we have no way to display it and it will require translations

PR #2383, opened 2020-03-03
feat(dashboard): add ovh-manager-banner

marie-j:
Add ovh-manager-banner to hub dashboard


ovh-cds:
CDS Report build-monorepository#9925.0 ✘  
* 
  * Build ✘

packages/manager/apps/hub/src/index.scss
```diff
@@ -3,3 +3,7 @@
 @import '~bootstrap4/scss/_mixins';
 @import '~bootstrap4/scss/_utilities.scss';
 @import '~bootstrap4/scss/_grid.scss';
+
+.minw-0 {
```
antleblanc: Is this helper couldn't be scoped at least?

Related to: https://github.com/ovh/manager/blob/master/packages/manager/modules/pci/src/index.scss#L12-L14

FredericEspiau: No, it should be the opposite of scoped and be available to everyone, it's an utility, you can even add `!important` as it's supposed to be used for utilities only

antleblanc: For such a kind of utilities (if needed), it shouldn't be a part of our Application but could be part of  something that is shared.

We can consider support it in our UI framework for instance but I think here we are relying on an layout issue with several helpers combination that leads to a side-effect.

We have a parent element which combines all these helpers:

- `d-flex`
- `flex-wrap`
- `flex-lg-nowrap`
- `w-100`
- `align-items-center`
- `justify-content-between`

With nested component that also rely of the following helper:

- `w-100`

To be able to move forward it could be kept as is and we can iterate on the layout.
That's why I've suggested to at least scope it.

marie-j: @antleblanc I agree that helpers should be shared but currently our UI framework doesn't provide any of those, maybe we could discuss with @AxelPeter on how we could add it
I didn't scope it because as @FredericEspiau said a helper should be available to everyone and this utility might be needed for the whole hub app. 
(FYI: https://dfmcphee.com/flex-items-and-min-width-0/ explains the reason why this is needed)

FredericEspiau: Maybe we could use https://tailwindcss.com/ instead of Boostrap at some point, we don't need Bootstrap anymore as flex and grids are available (I mean CSS-wise, I now we still use Boostrap related components)

marie-j:
Rebased to fix conflicts

PR #2396, opened 2020-03-04
Fix/telephon/alias/configuration

FredericEspiau:
DTRSD-8463

# Note
I'm sending this to QA right away as I'm not sure this is the required behavior


FredericEspiau:
Nop

PR #2399, opened 2020-03-04
Fix/pci bm lab

lizardK:
# PCI Bare Metal instances improvement

## Description of the change

- Move Bare Metal instances from beta to alpha
- Check if region has been activated through the form registration in order to create instances

### DIFF
--- a/packages/manager/apps/public-cloud/src/assets/theme/default/_sidebar.less
+++ b/packages/manager/apps/public-cloud/src/assets/theme/default/_sidebar.less
@@ -183,6 +183,16 @@ cloud-sidebar {
       background-color: #fff;
     }
 
+    .oui-status_alpha {
+      background-color: #fbff47;
+      padding: 0 8px;
+      line-height: 15px;
+      font-size: 12px;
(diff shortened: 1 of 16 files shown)


packages/manager/modules/pci/src/projects/project/project.less
```diff
@@ -1,6 +1,18 @@
 @background-color: rgb(254, 234, 134);
 @text-color: rgb(139, 97, 17);
 @link-color: rgb(0, 80, 215);
+@alpha-background-color: #fbff47;
+@alpha-font-color: #00185e;
+
+.oui-status_alpha {
```
antleblanc: I've recently updated the color from the badge `beta`.

Since we are relying on a override maybe it could be moved right next the `beta` styles.

Related to: https://github.com/ovh/manager/pull/2296/files#diff-a7b979d7d31f21e2905d7f31c8b3823bR186

So changes can be applied here: `packages/manager/apps/public-cloud/src/assets/theme/default/_sidebar.less`.

Is that make sense for you?

PR #2400, opened 2020-03-04
Feat/hub listing

marie-j:


yarn.lock
```diff
@@ -3236,6 +3236,11 @@ ansi-regex@^2.0.0:
   resolved "https://registry.yarnpkg.com/ansi-regex/-/ansi-regex-2.1.1.tgz#c3b33ab5ee360d86e0e628f0468ae7ef27d654df"
   integrity sha1-w7M6te42DYbg5ijwRorn7yfWVN8=
 
+ansi-regex@^3.0.0:
```
frenautvh: i often see this dep popping up, is it because of different node js version?

packages/manager/modules/hub/src/components/products/products.component.js
```diff
@@ -3,6 +3,7 @@ import template from './products.html';
 export default {
   bindings: {
     products: '<',
+    seeAll: '&',
```
frenautvh: Just a side note, i think something like 'onProductSelect' is better because the component should not have the information on the action that will occur when a product is selected. Knowing the "seeAll" action is already on a higher level (unless it's labelled "see all' in the component, in that case it's valid). Just being picky, not mandatory

yarn.lock
marie-j:
```
node -v
v12.16.0
yarn -v 
1.22.0
```

packages/manager/modules/hub/src/components/products/products.component.js
marie-j: @frenautvh I agree with you it makes sense to change to something less parent specific. I'll update as the change is quite easy to make :)

yarn.lock
FredericEspiau: We've done our researched with @marie-j for the last two weeks and we weren't able to a culprit yet

packages/manager/modules/hub/src/components/product-list/controller.js
```diff
@@ -1,90 +1,3 @@
...
-          });
-          return {
-            ...product,
-            expirationDate: get(productService, 'billing.expirationDate'),
-          };
-        });
-      });
-  }
-}
+export default class ManagerHubBillingProductList extends ListLayoutHelper.ListLayoutCtrl {}
```
FredericEspiau: So how do you add another behavior to this component ? You can only extend one `class`

packages/manager/apps/hub/src/dashboard/products-list/routing.js
```diff
@@ -0,0 +1,109 @@
...
+      columns: /* @ngInject */ (dataModel, displayedColumns, propertyId) => {
+        const columns = mapApiProperties(dataModel.properties).filter(
+          ({ title, type }) => type && title !== propertyId,
+        );
+        columns.unshift(getFirstColumnTemplate(propertyId));
+        return columns.map((column, index) => ({
+          ...column,
+          name: startCase(column.name),
+          hidden: isEmpty(displayedColumns)
+            ? index > 4
```
JeremyDec: Why `4` and not `10` for example ? Maybe you can name this constant ?

packages/manager/modules/hub/src/components/products/products.html
```diff
@@ -7,11 +7,12 @@
             heading="{{:: 'manager_hub_products_' + product | translate }}"
             badge-text="{{ data.count }}"
             class="oui-list"
+            data-on-click="$ctrl.onProductSelect({ productType: product })"
         >
             <ul class="oui-list__items">
                 <li
                     class="oui-list__item"
-                    data-ng-repeat="service in data.data track by service.serviceId"
+                    data-ng-repeat="service in data.data  | limitTo: 4 track by service.serviceId"
```
JeremyDec: This limit, could be stored in constant to be shared with the one used in the routing

packages/manager/modules/hub/src/components/product-list/controller.js
marie-j: Add typescript and use an `interface` instead maybe ?

packages/manager/modules/hub/src/components/products/products.html
marie-j: Not the same constant used, they do not impact the same component and don"t have the same purpose but I agree it could be more explicit

packages/manager/modules/hub/src/components/product-list/controller.js
FredericEspiau: No please don't use inheritance 😅

marie-j:
Fixed conflicts

PR #2402, opened 2020-03-05
feat(navbar): manager-hub sidebar integration

frenautvh:
Integrate manager hub sidebar with the navbar


packages/manager/apps/hub/src/index.html
```diff
@@ -32,11 +32,11 @@
                 universe: 'hub',
                 version: 'beta'
             }"
+            data-sidebar-expand="true"
         >
         </ovh-manager-navbar>
-        <div class="d-flex">
+        <div class="d-flex hub-main-view">
```
marie-j: I wonder if we still need this, maybe we can remove and apply the class on the children

packages/manager/modules/navbar/src/index.less
```diff
@@ -44,3 +45,44 @@
...
 }
+
+.oui-navbar-backdrop {
+  display: none;
+}
+
+.ovh-sidebar-wrapper {
+  position: absolute;
+  right: 0;
+  z-index: 99;
```
marie-j: We probably have no other choices but this looks like we are applying a random z-index

frenautvh: i don't know, maybe trying to reuse : https://github.com/twbs/bootstrap/blob/148421be129ed05a01f4f8086ae0c4f7b019226d/scss/_variables.scss#L784 ?
for example z-index = $zindex-dropdown ? also we have to make sure that the z-index is not higher than our manager preload ...

marie-j: https://github.com/ovh/manager/blob/master/packages/manager/apps/hub/src/components/manager-preload/manager-preload.less#L10 
It can stay as is because we don't know the impact of other elements having z-indexes (some component in oui-kit sometimes)

marie-j:
@frenautvh I think this PR needs a rebase on `feat/manager-hub`

PR #2406, opened 2020-03-05
fix(constants): redirect to exchange root state

marie-j:
No internal reference but related to 2API prod


antleblanc:
@marie-j Is this relates to a ticket reference?

PR #2412, opened 2020-03-06
[RFC] docs: add PULL_REQUEST_TEMPLATE.md file

antleblanc:
## GitHub Pull Request template

We recently initiate discussions about having a `PULL_REQUEST_TEMPLATE`.

Here is a suggestion where we can discuss about it and iterate over it to cover
the right balance between having Pull Request well described and easy to fill.

## :link: Resources

- https://github.com/devspace/awesome-github-templates
- https://github.com/ovh/cds/blob/master/.github/PULL_REQUEST_TEMPLATE.md
- https://github.com/symfony/symfony/blob/master/.github/PULL_REQUEST_TEMPLATE.md

### DIFF
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,3 +1,9 @@
+<!--
+Hello 👋 Thank you for submitting a Pull Request.
+
+Have any questions? Check out the contributing docs at https://github.com/ovh/manager/blob/master/CONTRIBUTING.md
+-->
+
 | Question         | Answer
 | ---------------- | ---
 | Branch?          | `master` for hotfix / `develop` for bug fixes <!-- see below -->
@@ -6,3 +12,7 @@
 | Breaking change? | yes/no
 | Tickets          | Fix #... <!-- prefix each issue number with "Fix #", if any -->
 | License          | BSD 3-Clause
+
+## Description
+
+<!-- Write a brief description of the changes introduced by this PR -->
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -1,0 +1,8 @@
+| Question         | Answer
+| ---------------- | ---
+| Branch?          | `master` for hotfix / `develop` for bug fixes <!-- see below -->
+| Bug fix?         | yes/no
+| New feature?     | yes/no
+| Breaking change? | yes/no
+| Tickets          | Fix #... <!-- prefix each issue number with "Fix #", if any -->
+| License          | BSD 3-Clause


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,8 @@
+| Question         | Answer
+| ---------------- | ---
+| Branch?          | `master` for hotfix / `develop` for bug fixes <!-- see below -->
+| Bug fix?         | yes/no
+| New feature?     | yes/no
+| Breaking change? | yes/no
```
FredericEspiau: Maybe we could have more details 😄 

I think about `Potential side-effects`

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,8 @@
+| Question         | Answer
+| ---------------- | ---
+| Branch?          | `master` for hotfix / `develop` for bug fixes <!-- see below -->
+| Bug fix?         | yes/no
```
FredericEspiau:
```suggestion
| Bug fix?         | [ ]
```
?

.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,8 @@
...
+| New feature?     | yes/no
+| Breaking change? | yes/no
+| Tickets          | Fix #... <!-- prefix each issue number with "Fix #", if any -->
+| License          | BSD 3-Clause
```
FredericEspiau: I think about:

```
| I have read the rules before submitting this PR          | [ ]
```

antleblanc: That's the purpose of the CONTRIBUTING.md file I would say.

.github/PULL_REQUEST_TEMPLATE.md
antleblanc: I use the `Breaking change` to speak the same as our Git commit convention.

https://github.com/angular/angular/blob/master/CONTRIBUTING.md#footer

.github/PULL_REQUEST_TEMPLATE.md
antleblanc: We cannot draw easily a checkbox.

See: https://stackoverflow.com/questions/47344571/how-to-draw-checkbox-or-tick-mark-in-github-markdown-table

.github/PULL_REQUEST_TEMPLATE.md
FredericEspiau: `CONTRIBUTING.md` states the rules, the rules states that you have to read them, but the user never says he actually read them before this step 😉

.github/PULL_REQUEST_TEMPLATE.md
FredericEspiau: I mean, maybe we can encourage having more details than mere `yes/no` and say

```
There will be breaking changes at those locations:
- location/breaking change
- location/breaking change
```

.github/PULL_REQUEST_TEMPLATE.md
FredericEspiau: Well maybe we don't need a table then 😄

.github/PULL_REQUEST_TEMPLATE.md
antleblanc: Users are invited to read them when they contributes:

https://help.github.com/en/github/building-a-strong-community/setting-guidelines-for-repository-contributors

.github/PULL_REQUEST_TEMPLATE.md
antleblanc: I've added a `## Description` section

.github/PULL_REQUEST_TEMPLATE.md
antleblanc: A Todolist can be confusing because a Pull Request will never be seen as **completed**.
That's why using a table with questions like `yes/no` look more affordable.
Let me know what do you think.

Thanks!

.github/PULL_REQUEST_TEMPLATE.md
FredericEspiau: And they never tell us that they did :)

PR #2414, opened 2020-03-09
fix(web-emailpro): add server field

varun257:
add a server field in the connection menu for professional email
DTRSD-6572


antleblanc:
I'm having trouble to understand **why** it's a `fix`?
Isn't supposed to be at least an enhancement regarding your changes?

Thanks!

PR #2415, opened 2020-03-09
build: add start script

antleblanc:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | no

### :construction_worker_man: Build

enhance development experience with a command line user interface
that give the ability to start an application.

usage: `$ yarn start`

a1c8c23 - build: add start script

### :camera_flash: Preview

[preview](https://gifyu.com/image/IRkS)

### :house: Internal

No quality check required.


scripts/start/index.js
```diff
@@ -0,0 +1,68 @@
+import execa from 'execa';
+import inquirer from 'inquirer';
+
+const applications = new Set([
+  '@ovh-ux/manager-hub',
+  '@ovh-ux/manager-dedicated',
+  '@ovh-ux/manager-cloud',
+  '@ovh-ux/manager-public-cloud',
```
JeremyDec: No need to add `@ovh-ux/manager-telecom` or `@ovh-ux/manager-web` ? As it is proposed in the questipns ?

cbourgois: I think this array is only here to detect if we  have to ask for a region (cf : https://github.com/ovh/manager/pull/2415/files#diff-cdf6d8d06626995121777296083524dbR47)

cbourgois:
IMO we should add all apps in this script, could be done after.

scripts/start/index.js
antleblanc: You're right! I'm using it to be able to determine if it's related to a `region`.

antleblanc:
> IMO we should add all apps in this script, could be done after.

I agree. Maybe we can consider using an autocomplete plugin to facilitate the way to start to appropriate application.

Resource: https://github.com/mokkabonna/inquirer-autocomplete-prompt

PR #2416, opened 2020-03-09
fix(secondary.dns): add resolve to parameter to the modal

varun257:
add resolve to paramter to the modal
DTRSD-9235

### DIFF
--- a/packages/manager/modules/vps/src/secondary-dns/delete/delete-secondary-dns.routing.js
+++ b/packages/manager/modules/vps/src/secondary-dns/delete/delete-secondary-dns.routing.js
@@ -1,11 +1,17 @@
 export default /* @ngInject */ ($stateProvider) => {
   $stateProvider.state('vps.detail.secondary-dns.delete', {
     url: '/delete',
+    params: {
+      domain: null,
+    },
     views: {
       modal: {
         component: 'vpsSecondaryDnsDelete',
       },
     },
     layout: 'modal',
+    resolve: {
+      domain: /* @ngInject */ ($transition$) => $transition$.params().domain,
+    },
   });
 };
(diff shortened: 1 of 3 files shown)


packages/manager/modules/vps/src/secondary-dns/delete/delete-secondary-dns.routing.js
```diff
@@ -1,11 +1,17 @@
 export default /* @ngInject */ ($stateProvider) => {
   $stateProvider.state('vps.detail.secondary-dns.delete', {
     url: '/delete',
```
marie-j: No need for
```suggestion
    url: '/delete?domain',
```
?

FredericEspiau: I suppose `domain` isn't a string

FredericEspiau:
Why don't you target `master` ? Isn't it a bug in prod ?

packages/manager/modules/vps/src/secondary-dns/delete/delete-secondary-dns.routing.js
varun257: Its an `object`

varun257:
> Why don't you target `master` ? Isn't it a bug in prod ?

Yes it is in `prod`. I thought only `quickfix` targets master ?

FredericEspiau:
@varun257 this was my question, I don't know if it's a bug in prod or not 😄

packages/manager/modules/vps/src/secondary-dns/delete/delete-secondary-dns.routing.js
marie-j: @FredericEspiau @varun257 I'm having trouble understanding the answer here 😄 
What I meant is you don't need to declare that your url is taking a queryParameter ?

varun257: No, since i add resolve parameter, its not required to add domain as `queryParams`

antleblanc:
@varun257 I will change the target branch.

PR #2434, opened 2020-03-09
fix(ip): disable old iplb tab

lizardK:
# Remove old iplb section

ref: MANAGER-4350

Signed-off-by: Cyril Biencourt <cyril.biencourt@corp.ovh.com>


antleblanc:
No quality check required.

PR #2446, opened 2020-03-10
fix(pci.failover.ips): make filter works

varun257:
rename the field so that filter works
DTRSD-8917


ovh-cds:
CDS Report installation#10222.0 ✘  
* Stage 1
  * Install ✘

PR #2447, opened 2020-03-10
fix(sidebar): add dbaas.logs for ca region

lizardK:
# Add dbaas logs product in CA region

ref: DTRSD-9409

Signed-off-by: Cyril Biencourt <cyril.biencourt@corp.ovh.com>


antleblanc:
No quality check required.

PR #2452, opened 2020-03-10
fix(hub): add tracking for assistance component

frenautvh:


packages/manager/apps/hub/src/dashboard/dashboard.html
```diff
@@ -25,6 +25,7 @@ <h2 data-translate="manager_hub_dashboard_overview"></h2>
             <ovh-manager-hub-support
                 data-me="$ctrl.me"
                 data-tickets="$ctrl.tickets"
+                data-tracking-prefix="hub::dashboard::activity"
```
marie-j: I'm wondering if this could be injected instead as it seems it will be the same for all dashboard components

frenautvh: yes sure maybe a last refactoring once every component has tracking? because i see some weird stuff like for example a 'trackingPrefix' only for payment tile in app hub's routing. Also the hub sidebar has a specific prefix

PR #2484, opened 2020-03-11
feat(hub): add debt pay on billing component

frenautvh:
Internal reference : MANAGER-4650

### DIFF
--- a/packages/manager/modules/hub/src/components/billing-summary/template.html
+++ b/packages/manager/modules/hub/src/components/billing-summary/template.html
@@ -20,13 +20,23 @@
         </span>
         <span data-translate="hub_billing_summary_debt_null"></span>
     </p>
-    <p
-        data-ng-if="$ctrl.debt.dueAmount.value > 0"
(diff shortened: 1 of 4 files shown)


packages/manager/modules/hub/src/components/billing-summary/template.html
```diff
@@ -20,13 +20,22 @@ <h3 data-translate="hub_billing_summary_title"></h3>
...
+            data-translate-values="{
+                debt: $ctrl.formattedDebtPrice
+              }"
+        >
+        </span>
+        <a
+            class="d-block"
+            data-ng-href="{{:: $ctrl.RedirectionService.getURL('debtPay') }}"
+            data-translate="hub_billing_summary_debt_pay"
+            target="_blank"
```
marie-j:
```suggestion
            target="_blank"
            rel="noopener"
```
I'm wondering why not open in the same window ? this is the behaviour for all other links (see my bills, see my services etc)

PR #2488, opened 2020-03-12
feat(domain-dkim): fix preview of dkim

varun257:
init values of checkbox
assign model value to checkbox

DTRSD-9480

### DIFF
--- a/packages/manager/apps/web/client/app/domain/zone/record/DKIM/domain-zone-record-DKIM.html
+++ b/packages/manager/apps/web/client/app/domain/zone/record/DKIM/domain-zone-record-DKIM.html
@@ -12,7 +12,7 @@
                 id="v"
                 name="v"
                 data-model="ctrl.model.target.v.DKIM1"
-                data-on-change="ctrl.setTargetValue('DKIM')"
+                data-on-change="ctrl.bindModelValue('DKIM', 'v.DKIM1', modelValue)"
                 data-text="DKIM1"
             >
             </oui-checkbox>
@@ -53,7 +53,7 @@
                 name="h"
                 id="h-sha1"
                 data-model="ctrl.model.target.h.sha1"
-                data-on-change="ctrl.setTargetValue('DKIM')"
+                data-on-change="ctrl.bindModelValue('DKIM', 'h.sha1', modelValue)"
                 data-text="SHA-1"
             >
             </oui-checkbox>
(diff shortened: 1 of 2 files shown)


packages/manager/apps/web/client/app/domain/zone/record/DKIM/domain-zone-record-DKIM.html
```diff
@@ -12,7 +12,7 @@
                 id="v"
                 name="v"
                 data-model="ctrl.model.target.v.DKIM1"
-                data-on-change="ctrl.setTargetValue('DKIM')"
+                data-on-change="ctrl.model.target.v.DKIM1 = modelValue; ctrl.setTargetValue('DKIM')"
```
FredericEspiau: Maybe you could extract to a function 🙂

radireddy: can you add this in controller and call that method in all other places plz, also add 'DKIM' in contants

packages/manager/apps/web/client/app/domain/zone/record/DKIM/domain-zone-record-DKIM.html
```diff
@@ -53,15 +53,15 @@
                 name="h"
                 id="h-sha1"
                 data-model="ctrl.model.target.h.sha1"
-                data-on-change="ctrl.setTargetValue('DKIM')"
+                data-on-change="ctrl.model.target.h.sha1 = modelValue; ctrl.setTargetValue('DKIM')"
```
radireddy: call the reusable method in controller

packages/manager/apps/web/client/app/domain/zone/record/DKIM/domain-zone-record-DKIM.html
```diff
@@ -53,15 +53,15 @@
...
                 data-text="SHA-1"
             >
             </oui-checkbox>
             <oui-checkbox
                 name="h"
                 id="h-sha256"
                 data-model="ctrl.model.target.h.sha256"
-                data-on-change="ctrl.setTargetValue('DKIM')"
+                data-on-change="ctrl.model.target.h.sha256 = modelValue; ctrl.setTargetValue('DKIM')"
```
radireddy: call the reusable method in controller and same in below 2 places

packages/manager/apps/web/client/app/domain/zone/record/DKIM/domain-zone-record-DKIM.html
varun257: There are a lot of constants and the whole code needs to be rewritten in future so we should do it when we refactor :)

FredericEspiau: Ok @varun257 but please create a JIRA for this and add it to `TO_GROOM` sprint, this way we don't forget about this 🙂

PR #2489, opened 2020-03-12
fix(hub): add filter to billing redirection

frenautvh:
Internal reference : MANAGER-4652

### DIFF
--- a/packages/manager/modules/hub/src/components/billing-summary/template.html
+++ b/packages/manager/modules/hub/src/components/billing-summary/template.html
@@ -29,7 +29,7 @@
     ></p>
     <a
         class="manager-hub-billing-summary_billing-link"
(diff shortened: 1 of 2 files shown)


packages/manager/modules/hub/src/components/billing-summary/template.html
```diff
@@ -29,7 +29,7 @@ <h3 data-translate="hub_billing_summary_title"></h3>
     ></p>
     <a
         class="manager-hub-billing-summary_billing-link"
-        data-ng-href="{{:: $ctrl.RedirectionService.getURL('billing') }}"
+        data-ng-href="{{ $ctrl.getBillingURL() }}"
```
radireddy: this method will be called multiple times because of the digest loop. maybe you can construct the URL and store it in a variable.

PR #2493, opened 2020-03-12
fix(products-list): handle pagination change with static rows

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          |   MANAGER-4657

## Description

Prevent pagination from always reseting to page 1 when rows are all loaded. Also prevent reloading everything in that case


marie-j:
Fixed conflicts

PR #2505, opened 2020-03-13
refactor(account-sidebar): apply hub styles

frenautvh:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | 

## Description

Apply hub styles to account-sidebar

### DIFF
--- a/packages/manager/modules/account-sidebar/src/payment-mean/index.scss
+++ b/packages/manager/modules/account-sidebar/src/payment-mean/index.scss
@@ -1,4 +1,6 @@
 .manager-hub-payment-mean {
+  @import '@ovh-ux/manager-hub/src/variables.scss';
+
   $payment-bg-color: #bef1ff;
 
   > a {
@@ -16,5 +18,9 @@
   &_status {
     font-size: 0.8rem;
     font-weight: bold;
+  }
+
+  .oui-icon {
+    color: $hub-title-color;
   }
 }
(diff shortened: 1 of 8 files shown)


packages/manager/modules/account-sidebar/src/payment-mean/index.scss
```diff
@@ -17,4 +19,8 @@
     font-size: 0.8rem;
     font-weight: bold;
   }
+
+  .oui-icon {
+    color: $hub-title-color;
```
marie-j: Is this different from all  the hub app ? If not it could be globalized

packages/manager/modules/account-sidebar/src/user-infos/controller.js
```diff
@@ -45,7 +45,7 @@ export default class ManagerHubUserInfosCtrl {
   }
 
   getDisplayNicHandle() {
-    return toUpper(this.me.nichandle.replace(/-ovh$/, ''));
+    return this.me.nichandle;
```
marie-j: Is a function needed ? I agree it provides abstraction but in that case I think it would be better in a `User` model or somehting like that

packages/manager/modules/account-sidebar/src/payment-mean/index.scss
frenautvh: yes this color is specific to the payment mean tile

marie-j:
@frenautvh LGTM but there are conflicts

PR #2509, opened 2020-03-13
Import @ovh-ux/ng-at-internet

antleblanc:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-4499

## Description

Import **@ovh-ux/ng-at-internet** :truck:

Using following command:

```sh
lerna import ~/Code/github/ovh-ux/ng-at-internet \
  --dest=packages/components \
  --preserve-commit \
  --flatten
```

### :construction_worker_man: Build

10bd505 - build: remove extra parts

### :memo: Documentation

5b6410c - docs: update README.md file

### :bookmark: TODOs

- [x] Update CONTRIBUTORS file.


marie-j:
Why not use a git subtree instead ? Do we really want to keep history here ?

antleblanc:
We are using the `lerna import` built-in command to easily import packages.

In that way we can keep the Git Commit History which could be helpful to understand changes between commits I would say.

Using a Git-subtree doesn't allow us to delete later on the repository.

PR #2525, opened 2020-03-16
Feat/vps new range price

FredericEspiau:
| Question         | Answer
| ---------------- | ---
| Branch?          | feat/vps-new-range-upgrade
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-4318

## Description

Components to be used in the VPS upscale screen

### DIFF
--- a/packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.component.js
+++ b/packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.component.js
@@ -1,0 +1,24 @@
+import controller from './configuration-price.controller';
+import template from './configuration-price.html';
+
+/**
+ * Examples:
+ * <vps-upscale-configuration-price price="{ value: 32.30, unit: 'microcents', currency: 'EUR' }" language="'fr_FR'"></vps-upscale-configuration-price>
+ * <vps-upscale-configuration-price price="{ value: 32.30, unit: 'cents', currency: 'USD' }" language="'fr_CA'"></vps-upscale-configuration-price>
+ * <vps-upscale-configuration-price price="{ value: 32.30, unit: 'cents', currency: 'EUR' }" language="'en_US'"></vps-upscale-configuration-price>
+ */
+export default {
+  bindings: {
+    /**
+     * string
+     * formatted the OVHcloud way
+     */
+    language: '<',
+    /**
+     * @ovh-ux/manager-models:Price
+     */
+    price: '<',
+  },
+  controller,
+  template,
+};
--- a/packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.controller.js
+++ b/packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.controller.js
@@ -1,0 +1,13 @@
+import { Price } from '@ovh-ux/manager-models';
+import UpscaleService from '../upscale.service';
+
+export default class {
+  $onInit() {
+    UpscaleService.validateLanguage(this.language);
+  }
+
+  buildPrice() {
+    const price = new Price(this.price);
+    return UpscaleService.buildPriceToDisplay(price, this.language);
+  }
+}
--- a/packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.scss
+++ b/packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.scss
@@ -1,0 +1,12 @@
+vps-upscale-configuration-price {
+  .vps-upscale-summary-price__price {
+    display: block;
+    font-size: 2rem;
+    font-weight: 400;
+  }
+
+  .vps-upscale-summary-price__frequency {
+    display: block;
+    font-size: 0.8rem;
+  }
+}
--- a/packages/manager/modules/vps/src/upscale/model-price/model-price.component.js
+++ b/packages/manager/modules/vps/src/upscale/model-price/model-price.component.js
@@ -1,0 +1,24 @@
+import controller from './model-price.controller';
+import template from './model-price.html';
+
+/**
+ * Examples:
+ * <vps-upscale-model-price price="{ value: 32.30, unit: 'microcents', currency: 'EUR' }" language="'fr_FR'"></vps-upscale-model-price>
+ * <vps-upscale-model-price price="{ value: 32.30, unit: 'cents', currency: 'USD' }" language="'fr_CA'"></vps-upscale-model-price>
+ * <vps-upscale-model-price price="{ value: 32.30, unit: 'cents', currency: 'EUR' }" language="'en_US'"></vps-upscale-model-price>
+ */
+export default {
+  bindings: {
+    /**
+     * string
+     * formatted the OVHcloud way
+     */
+    language: '<',
+    /**
+     * @ovh-ux/manager-models:Price
+     */
+    price: '<',
+  },
+  controller,
+  template,
+};
--- a/packages/manager/modules/vps/src/upscale/model-price/model-price.controller.js
+++ b/packages/manager/modules/vps/src/upscale/model-price/model-price.controller.js
@@ -1,0 +1,13 @@
+import { Price } from '@ovh-ux/manager-models';
+import UpscaleService from '../upscale.service';
+
+export default class {
+  $onInit() {
+    UpscaleService.validateLanguage(this.language);
+  }
+
+  buildPrice() {
+    const price = new Price(this.price);
+    return UpscaleService.buildPriceToDisplay(price, this.language);
+  }
+}
--- a/packages/manager/modules/vps/src/upscale/summary-price/index.js
+++ b/packages/manager/modules/vps/src/upscale/summary-price/index.js
@@ -1,0 +1,17 @@
+import angular from 'angular';
+
+import 'angular-translate';
+import 'ovh-ui-angular';
+
+import './summary-price.scss';
+
+import component from './summary-price.component';
+
+const moduleName = 'ovhManagerVpsUpscaleSummaryPrice';
+
+angular
+  .module(moduleName, ['oui', 'pascalprecht.translate'])
+  .component('vpsUpscaleSummaryPrice', component)
+  .run(/* @ngTranslationsInject:json ../translations */);
+
+export default moduleName;
--- a/packages/manager/modules/vps/src/upscale/summary-price/summary-price.controller.js
+++ b/packages/manager/modules/vps/src/upscale/summary-price/summary-price.controller.js
@@ -1,0 +1,44 @@
+import includes from 'lodash/includes';
+import isNumber from 'lodash/isNumber';
+import isObject from 'lodash/isObject';
+import values from 'lodash/values';
+
+import { Price } from '@ovh-ux/manager-models';
+
+import { PAYMENT_TYPES } from './summary-price.constants';
+import UpscaleService from '../upscale.service';
+
+export default class {
+  $onInit() {
+    UpscaleService.validateLanguage(this.language);
+    this.validatePaymentType();
+
+    this.paymentTypes = PAYMENT_TYPES;
+  }
+
+  validatePaymentType() {
+    if (!isObject(this.paymentType)) {
+      throw new TypeError('Attribute "paymentType" should be of type "object"');
+    }
+
+    if (!includes(values(PAYMENT_TYPES), this.paymentType.value)) {
+      throw new RangeError(
+        `Attribute "paymentType.value" has an invalid value: ${this.paymentType.value}`,
+      );
+    }
+
+    if (
+      this.paymentType.value === PAYMENT_TYPES.UPFRONT &&
+      !isNumber(this.paymentType.monthNumber)
+    ) {
+      throw new TypeError(
+        'Attribute "paymentType.monthNumber" should be of type "number"',
+      );
+    }
+  }
+
+  buildPrice() {
+    const price = new Price(this.price);
+    return UpscaleService.buildPriceToDisplay(price, this.language);
+  }
+}
--- a/packages/manager/modules/vps/src/upscale/summary-price/summary-price.html
+++ b/packages/manager/modules/vps/src/upscale/summary-price/summary-price.html
@@ -1,0 +1,29 @@
+<span class="oui-box d-block py-2 border-0 rounded-0">
+    <span class="d-block text-right vps-upscale-summary-price__price-frequency">
+        <span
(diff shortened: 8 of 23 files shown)


packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.controller.js
```diff
@@ -0,0 +1,20 @@
...
+
+    // the variables used in the view
+    this.v = {
+      price: PriceService.buildPriceToDisplay(this.price, this.language),
+    };
+  }
+
+  validateLanguage() {
+    if (!isString(this.language)) {
+      throw new TypeError('Attribute "language" should be of type "string"');
```
radireddy: if the language is of type string, why not use @ binding?

packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.controller.js
```diff
@@ -0,0 +1,20 @@
+import isString from 'lodash/isString';
+import PriceService from '../price/price.service';
+
+export default class {
+  $onInit() {
+    this.validateLanguage();
+    PriceService.validatePrice(this.price);
+
+    // the variables used in the view
+    this.v = {
```
radireddy: can you rename v to something meaningful?

packages/manager/modules/vps/src/upscale/model-price/model-price.component.js
```diff
@@ -0,0 +1,24 @@
+import controller from './model-price.controller';
+import template from './model-price.html';
+
+/**
+ * Examples:
+ * <vps-upgrade-model-price price="{ value: 32.30, unit: 'microcents', currency: 'EUR' }" language="'fr_FR'"></vps-upgrade-model-price>
+ * <vps-upgrade-model-price price="{ value: 32.30, unit: 'cents', currency: 'USD' }" language="'fr_CA'"></vps-upgrade-model-price>
+ * <vps-upgrade-model-price price="{ value: 32.30, unit: 'cents', currency: 'EUR' }" language="'en_US'"></vps-upgrade-model-price>
+ */
+export default {
```
radireddy: I see both model-price and configuration-price are identical (component, controller and template). Is it possible to reuse the same component?

packages/manager/modules/vps/src/upscale/summary-price/summary-price.controller.js
```diff
@@ -0,0 +1,47 @@
...
+    PriceService.validatePrice(this.price);
+    this.validatePaymentType();
+
+    // the variables used in the view
+    this.v = {
+      price: PriceService.buildPriceToDisplay(this.price, this.language),
+    };
+  }
+
+  validateLanguage() {
```
radireddy: I see validateLanguage function defined in multiple places, is it possible to write once and reuse?

packages/manager/modules/vps/src/upscale/summary-price/summary-price.html
```diff
@@ -0,0 +1,29 @@
+<span class="oui-box d-block py-2 border-0 rounded-0">
+    <span class="d-block text-right vps-upscale-summary-price__price-frequency">
+        <span
+            data-ng-if="$ctrl.paymentType.value === 'monthly'"
```
radireddy: payment type values can be constants?

packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.component.js
```diff
@@ -0,0 +1,24 @@
...
+ * <vps-upgrade-configuration-price price="{ value: 32.30, unit: 'cents', currency: 'USD' }" language="'fr_CA'"></vps-upgrade-configuration-price>
+ * <vps-upgrade-configuration-price price="{ value: 32.30, unit: 'cents', currency: 'EUR' }" language="'en_US'"></vps-upgrade-configuration-price>
+ */
+export default {
+  bindings: {
+    /**
+     * string
+     * formatted the OVHcloud way
+     */
+    language: '<',
```
marie-j: Why do you have to pass the language ? Why not get it in the component ?

packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.controller.js
```diff
@@ -0,0 +1,20 @@
+import isString from 'lodash/isString';
+import PriceService from '../price/price.service';
+
+export default class {
+  $onInit() {
+    this.validateLanguage();
+    PriceService.validatePrice(this.price);
+
+    // the variables used in the view
```
marie-j: Do we really want to encapsulate variables into others ?

packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.scss
```diff
@@ -0,0 +1,12 @@
+vps-upscale-configuration-price {
+  .vps-upscale-summary-price__price {
```
marie-j: I don't see any class applied in your component, is it still useful ? https://github.com/ovh/manager/pull/2525/files#diff-60d175817534df7f5a90ac0825b899bcR1

packages/manager/modules/vps/src/upscale/price/price.class.js
```diff
@@ -0,0 +1,12 @@
+export default class Price {
```
marie-j: Does it make sense to add it to @ovh-ux/manager-models module instead ? Looks like this could be used elsewhere

packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.component.js
JeremyDec: Maybe it is because the lanaguage correspond to the user language, not the`universe_selected_language` ?

packages/manager/modules/vps/src/upscale/model-price/model-price.component.js
JeremyDec: +1 Yes, only the translate key used in the template differs

packages/manager/modules/vps/src/upscale/price/price.class.js
JeremyDec: I can't see where this is instanciated in the components. Is this class useful ?

packages/manager/modules/vps/src/upscale/summary-price/summary-price.html
JeremyDec: or at least use a method to avoid exposing constants. @radireddy there are defined here : packages/manager/modules/vps/src/upscale/summary-price/summary-price.constants.js

packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.controller.js
FredericEspiau: Because it's a chore as long as you have to use bindings

```js
<configuration-price language="{{ $ctrl.theLanguageToUse}}"></configuration-price>
```
what a waste of time compared to
```js
<configuration-price language="$ctrl.theLanguageToUse"></configuration-price>
```

Also, you will never ever in your entire life

```js
<configuration-price language="fr_FR"></configuration-price>
```

packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.controller.js
FredericEspiau: It's the conventional name for the view bindings

packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.scss
FredericEspiau: Translations 🙂

packages/manager/modules/vps/src/upscale/price/price.class.js
FredericEspiau: @marie-j yes indeaed it should be there, didn't know this existed

FredericEspiau: @JeremyDec for documentation purpose, so yes it has the most important purpose 🙂

JeremyDec: Hmm even for documentation, it is weird as you suppose that the price used in (where?) has the same representation as this class, without any explicit instance.

packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.scss
JeremyDec: @marie, used here : https://github.com/ovh/manager/blob/feat/vps-new-range-upgrade/packages/manager/modules/vps/src/upscale/translations/Messages_fr_FR.json#L4

packages/manager/modules/vps/src/upscale/model-price/model-price.component.js
FredericEspiau: I do see what you mean but they are semantically different and could evolve in different directions, this would be a bad abstraction to refactor only for the sake of being DRY 🙂 

I invite you to read this article and to tell me what you think about this: https://overreacted.io/goodbye-clean-code/ 🙂

packages/manager/modules/vps/src/upscale/summary-price/summary-price.controller.js
FredericEspiau: It's very very linked to the component so it would be a bad idea imo

packages/manager/modules/vps/src/upscale/summary-price/summary-price.html
FredericEspiau: Yup good idea

packages/manager/modules/vps/src/upscale/price/price.class.js
FredericEspiau: JavaScript is not statistically typed and I don't want to enforce the user the create an instance of Price each time they use the component, it would prevent them from using it in pure HTML

It's specified in the documentation that the type is Price

packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.component.js
marie-j: I'm having trouble understanding. Do you mean user language selected for other apps that is equal to `me.language` from `/me` ?

packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.scss
marie-j: This saddens me ... But I guess we don't have any better solution

packages/manager/modules/vps/src/upscale/price/price.class.js
marie-j: @FredericEspiau Why not just add the documentation part where the object should be manipulated ? Given the huge code we already have to maintain, it looks to me like adding unused code is not a good idea

packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.scss
FredericEspiau: Yes, the translations should be directly in the file, it would be easier for everyone (us included)

packages/manager/modules/vps/src/upscale/configuration-price/configuration-price.component.js
JeremyDec: Exactly ! :)

FredericEspiau: I wanted it to be as stupid as stupid does, this way it the less dependencies possible (no constructor is needed), but I can include it if it makes more sense to you

packages/manager/modules/vps/src/upscale/model-price/model-price.controller.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default class {
+  $onInit() {
+    UpscaleService.validateLanguage(this.language);
+  }
+
+  buildPrice() {
+    const price = new Price(this.price);
+    return UpscaleService.buildPriceToDisplay(price, this.language);
+  }
```
marie-j: Looks like both controllers http://github.com/ovh/manager/pull/2525/files#diff-e51ca0dx0a33d8b5ffb38ab635b7b0e57R11 are doing the same thing , like the template (except for translation). Maybe we could have a single component in which the translation is injected instead so we don't need to duplicate ?

packages/manager/modules/vps/src/upscale/summary-price/index.js
```diff
@@ -0,0 +1,15 @@
...
+import 'angular-translate';
+
+import './summary-price.scss';
+
+import component from './summary-price.component';
+
+const moduleName = 'ovhManagerVpsUpscaleSummaryPrice';
+
+angular
+  .module(moduleName, ['oui', 'pascalprecht.translate'])
```
marie-j:
```suggestion
import 'ovh-ui-angular';
```
is missing

packages/manager/modules/vps/src/upscale/model-price/model-price.controller.js
FredericEspiau: I already answered this question, both components are actually pretty similar but differents in their intents and it would be bad idea to DRY them (more information here: https://www.amazon.com/Pragmatic-Programmer-journey-mastery-Anniversary/dp/0135957052/ref=sr_1_1?crid=3O2JZTUCLVHAS&keywords=pragmatic+programmer&qid=1584952852&sprefix=pragma%2Caps%2C243&sr=8-1 😄 )

ovh-cds:
CDS Report build-monorepository#10751.0 ✘  
* 
  * Build ✘

ovh-cds:
CDS Report build-monorepository#10751.0 ✘  
* 
  * Build ✘

packages/manager/modules/vps/src/upscale/model-price/model-price.controller.js
marie-j: I'm missing context here, so what I see is only two component doing the exact same thing. Let's say I want to reuse one of them, I won't be able to determine the one suiting my needs. 
If they are meant to evolve differently but still have a common base maybe we can temporarily have an inheritance from a parent controller or something that will avoid duplicating this logic

FredericEspiau: It would not be the right abstraction, they look like the same but they are not :)

PR #2529, opened 2020-03-16
fix(hub): view max width

frenautvh:

| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | internal MANAGER-4681

### DIFF
--- a/packages/manager/apps/hub/src/catalog/catalog.scss
+++ b/packages/manager/apps/hub/src/catalog/catalog.scss
@@ -1,4 +1,11 @@
 hub-catalog {
+  @import '../assets/theme/index.scss';
+
+  > div {
+    max-width: $hub-view-max-width;
+    margin: auto;
+  }
+
   .right-0 {
     right: 0;
   }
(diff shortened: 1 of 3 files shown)


packages/manager/apps/hub/src/catalog/catalog.scss
```diff
@@ -1,4 +1,11 @@
 hub-catalog {
```
antleblanc: Maybe the component `<hub-catalog>` could be centered by default (instead of targeting direct `div` elements).

Is that make sense?

packages/manager/apps/hub/src/dashboard/dashboard.scss
```diff
@@ -1,4 +1,11 @@
 hub-dashboard {
```
antleblanc: Same suggestion as the `<hub-catalog>` component.

packages/manager/apps/hub/src/catalog/catalog.scss
frenautvh: i tried to apply the style directly on the component but it didn't work :/

PR #2531, opened 2020-03-16
feat(hub): clickable order id

frenautvh:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | internal MANAGER-4645

## Description

Make last order clickable


packages/manager/modules/core/src/redirection/redirection.constants.js
```diff
@@ -79,6 +81,7 @@ export default {
       },
     },
     help: 'https://help.ovhcloud.com',
+    order: `${managerRoot.CA}/manager/${URI.order}`,
```
antleblanc: Correct me if I'm wrong, isn't supposed to be:

```suggestion
    order: `${managerRoot.CA}/manager/dedicated/${URI.order}`,
```

frenautvh: @antleblanc i don't know, currently dedicated CA in on /manager, is it planned to be moved ?

antleblanc: From my point of view I would answer **Yes**.

cc @marie-j to confirm.

marie-j: @frenautvh This is future URL for hub, so dedicated in CA has to move

PR #2548, opened 2020-03-17
feat(billing.orders): enable order tracking is us

lizardK:
| Question         | Answer
| ---------------- | ---
| Branch?          | master
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | OVHR-741

## Description

Enable order tracking feature in US

### DIFF
--- a/packages/manager/apps/dedicated/client/app/billing/billing-feature-availability.js
+++ b/packages/manager/apps/dedicated/client/app/billing/billing-feature-availability.js
@@ -5,7 +5,7 @@ export default class {
   }
 
   allowOrderTracking() {
-    return this.deny('US');
+    return this.allow('CA', 'EU', 'US');
   }
 
   allow(...args) {


packages/manager/apps/dedicated/client/app/billing/billing-feature-availability.js
```diff
@@ -5,7 +5,7 @@ export default class {
   }
 
   allowOrderTracking() {
-    return this.deny('US');
+    return this.allow('CA', 'EU', 'US');
```
marie-j:
```suggestion
    return true;
```
even better we could maybe remove `if` where this is applied

antleblanc: What if we decided to introduce a new region? 

Looks good to me.

PR #2553, opened 2020-03-18
fix: handle overflow

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-4712

## Description

- Handle overflow and fix position issues

### DIFF
--- a/packages/manager/apps/hub/src/index.scss
+++ b/packages/manager/apps/hub/src/index.scss
@@ -1,10 +1,16 @@
-@import '~bootstrap4/scss/_functions';
-@import '~bootstrap4/scss/_variables';
-@import '~bootstrap4/scss/_mixins';
-@import '~bootstrap4/scss/_utilities.scss';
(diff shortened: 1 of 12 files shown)


packages/manager/apps/hub/src/index.scss
```diff
@@ -6,6 +6,20 @@
 @import '@ovh-ux/manager-hub/src/variables.scss';
 @import './assets/theme/index.scss';
 
+.hub-main-view {
+  > * {
+    max-width: $hub-view-max-width;
+  }
+}
+
+.overflow-hidden {
```
antleblanc: Since version 4.2 these helpers are available.

Maybe we can consider a bump to the version 4.4.

marie-j: You're right ! this is done :)

PR #2563, opened 2020-03-18
fix(hub): expand sidebar only for dashboard

frenautvh:

| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | internal MANAGER-4683

## Description

The sidebar should only be expanded in the dashboard state ...
I tried to avoid using the rootScope but couldn't because you need to detect that the state changed to 'app.dashboard' then updates attributes outside of the dashboard template (for the sidebar and the main view margin).

### DIFF
--- a/packages/manager/apps/hub/src/index.html
+++ b/packages/manager/apps/hub/src/index.html
@@ -45,7 +45,7 @@
                     event: 'sidebar:loaded',
                 },
             }"
-            data-sidebar-expand="true"
+            data-sidebar-expand="shouldExpandSidebar"
             data-sidebar-links="{}"
         >
         </ovh-manager-navbar>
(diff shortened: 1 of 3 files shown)


packages/manager/apps/hub/src/index.less
```diff
@@ -2,7 +2,9 @@
 @import '@ovh-ux/manager-account-sidebar/src/variables.less';
 
 .hub-main-view {
-  margin-right: @ovh-sidebar-width;
+  &_sidebar_expanded {
```
antleblanc: Nested selector isn't needed I think.

This should work: `.hub-main-view_sidebar_expanded`

packages/manager/apps/hub/src/index.html
```diff
@@ -45,7 +45,7 @@
                     event: 'sidebar:loaded',
                 },
             }"
-            data-sidebar-expand="true"
+            data-sidebar-expand="isDashboardState"
```
marie-j:
```suggestion
            data-sidebar-expand="shouldExpandSidebar"
```
we might want in future cases to expand the sidebar

PR #2603, opened 2020-03-20
fix(ng-ovh-payment-method): default payment method

frenautvh:

| Question         | Answer
| ---------------- | ---
| Branch?          | master
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | 

## Description

The attribute for new payment mean call is named 'defaultPaymentMean' and not 'default' (which is used for legacy payment means)

### DIFF
--- a/packages/components/ng-ovh-payment-method/src/payment-method.service.js
+++ b/packages/components/ng-ovh-payment-method/src/payment-method.service.js
@@ -62,9 +62,10 @@ export default class OvhPaymentMethodService {
       transform: true,
     }).then(
       (paymentMethods) =>
-        find(paymentMethods, {
-          default: true,
-        }) || null,
+        find(
+          paymentMethods,
+          (method) => method.default || method.defaultPaymentMean,
+        ) || null,
     );
   }
 


packages/components/ng-ovh-payment-method/src/payment-method.service.js
```diff
@@ -62,9 +62,10 @@ export default class OvhPaymentMethodService {
       transform: true,
     }).then(
       (paymentMethods) =>
-        find(paymentMethods, {
-          default: true,
-        }) || null,
+        find(
+          paymentMethods,
+          (method) => method.default || method.defaultPaymentMean,
+        ) || null,
```
FredericEspiau: `undefined` wasn't good enough for you ?

frenautvh: it's the original behavior i didn't want to change it because it could have some side effects

FredericEspiau: Good point

PR #2608, opened 2020-03-23
fix(hub): remove view max width

frenautvh:

| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? |no
| Tickets          | 

## Description

Remove the max width of 1024px on manager hub as it causes display issues on large devices

### DIFF
--- a/packages/manager/apps/hub/src/index.less
+++ b/packages/manager/apps/hub/src/index.less
@@ -2,7 +2,8 @@
 @import '@ovh-ux/manager-account-sidebar/src/variables.less';
 
 .hub-main-view_sidebar_expanded {
-  margin-right: @ovh-sidebar-width;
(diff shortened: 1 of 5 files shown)


packages/manager/apps/hub/src/index.less
```diff
@@ -2,7 +2,8 @@
 @import '@ovh-ux/manager-account-sidebar/src/variables.less';
 
 .hub-main-view_sidebar_expanded {
-  margin-right: @ovh-sidebar-width;
+  // -0.75 rem is to take the scrollbar width into account
+  margin-right: calc(@ovh-sidebar-width - 0.75rem);
```
FredericEspiau: well maybe it could be in a `@scrollbar-width` then, it looks like a value we could reuse :)

PR #2610, opened 2020-03-23
Feat/hub enterprise links

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? |no
| Tickets          | none

## Description

Update links for enterprise customers


packages/manager/apps/dedicated/client/app/components/user/user.service.js
```diff
@@ -29,18 +30,20 @@ angular.module('services').service('User', [
...
-                    ) === -1,
-                };
+                user = new User(
+                  {
+                    ...result.me,
+                    firstName: result.me.firstname,
+                    lastName: result.me.name,
+                    billingCountry: result.me.country,
+                    customerCode: result.me.customerCode,
+                    isVATNeeded:
```
cbourgois: IMO, if we have a `User` model, `isVATNeeded` could be a method of `User` as same conditions will always applies. What do you think ?

marie-j: @cbourgois I agree but I didn't spend much time on checking and refactoring the usage of `User`. Is that ok with you if we handle it in another PR ?

cbourgois: It's ok ;)

marie-j:
Rebased and fixed conflicts

marie-j:
I've added missing import and updated some content due to CX review

PR #2310, opened 2020-02-21
Feat/manager hub

marie-j:
| Question         | Answer
| ---------------- | ---
| Branch?          | `develop`
| Bug fix?         | yes
| New feature?     | yes
| Breaking change? | yes
| Tickets          | MANAGER-4134
| License          | BSD 3-Clause

### 🚀 Feature

Release Home Universes Board 1.0

### 💥 Breaking Changes

a595155 - Language menu is no longer available from navbar 
2dc906e - init @ovh-ux/manager-account-sidebar module 
362d238 - CORE_MANAGER_URLS is now filtered according to region 
09642f0 - create stable version for hub 

### 🔗 Related

- [x] MANAGER-4292
- [x] MANAGER-4575
- [x] MANAGER-4613
- [x] MANAGER-4614
- [x] MANAGER-4615
- [x] MANAGER-4617
- [x] MANAGER-4622
- [x] MANAGER-4622
- [x] MANAGER-4623
- [x] MANAGER-4628
- [x] MANAGER-4630
- [x] MANAGER-4640
- [x] MANAGER-4641
- [x] MANAGER-4647
- [x] MANAGER-4654
- [x] MANAGER-4657
- [x] MANAGER-4658
- [x] MANAGER-4661
- [x] MANAGER-4673
- [x] MANAGER-4677
- [x] MANAGER-4711
- [x] MANAGER-4712
- [x] MANAGER-4724
- [x] MANAGER-4727
- [x] MANAGER-4728
- [x] MANAGER-4729
- [x] MANAGER-4733
- [x] MANAGER-4743
- [x] MANAGER-4744
- [x] MANAGER-4746
- [x] MANAGER-4750
- [x] MANAGER-4773
- [x] MANAGER-4804

PR #2612, opened 2020-03-23
Feat/vps new range configuration

Jisay:
| Question         | Answer
| ---------------- | ---
| Branch?          | feat/vps-new-range-information 
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-4444 

## Description

Add the configuration tile for VPS new range.

## Dependencies

- [ ] #2310 (for order tracking URLs)


packages/components/ng-ovh-stop-event/src/directive.js
```diff
@@ -1,9 +1,12 @@
-export default function () {
+export default function() {
```
marie-j: Did we forget to run prettier on those files ?

packages/manager/modules/vps/src/dashboard/tile/configuration/controller.js
```diff
@@ -0,0 +1,13 @@
...
+
+export default class VpsDashboardTileConfigurationCtrl {
+  constructor() {
+    this.currentPlan = VpsConfigurationTile.currentPlan;
+  }
+
+  $onInit() {
+    this.isMaxEliteVcore = this.vps.vcore === 8;
+    this.isMaxEliteRam = this.isMaxEliteVcore && this.vps.ram.value === 32;
+    this.isMaxEliteStorage = this.isMaxEliteVcore && this.stateVps.model.disk === 640;
```
marie-j: This could be handled with a `VPS` model

packages/manager/modules/vps/src/dashboard/tile/configuration/controller.js
```diff
@@ -0,0 +1,13 @@
+import VpsConfigurationTile from './service';
+
+export default class VpsDashboardTileConfigurationCtrl {
+  constructor() {
+    this.currentPlan = VpsConfigurationTile.currentPlan;
+  }
+
+  $onInit() {
+    this.isMaxEliteVcore = this.vps.vcore === 8;
```
marie-j:
```suggestion
    this.hasMaxEliteVcore = this.vps.vcore === 8;
```
maybe instead

packages/manager/modules/vps/src/dashboard/tile/configuration/service.js
```diff
@@ -0,0 +1,164 @@
...
+      value,
+    };
+  }
+
+  /**
+   *  Set the vps informations
+   *
+   *  @param {Object} vps      The informations provided by 2API
+   *  @param {Object} vpsModel The model attribute provided by
+   *                           GET /vps/{servicName} (not present in 2API...)
```
marie-j:
```suggestion
   *                           GET /vps/{serviceName} (not present in 2API...)
```
'Not present in 2API' > Would it make sense to add it to avoid one more call ?

packages/manager/modules/vps/src/dashboard/tile/configuration/service.js
```diff
@@ -0,0 +1,164 @@
...
+      }
+
+      const cpu = get(blobs, 'technical.cpu.cores');
+      const ram = get(blobs, 'technical.memory.size');
+      const storage = get(blobs, 'technical.storage.disks[0].capacity');
+
+      return (
+        cpu === this.vps.vcore &&
+        ram === this.vps.ram.value &&
+        storage === this.vpsModel.disk * 2
```
marie-j: The `* 2` looks like magic to me. Maybe a comment on which rule is applied could help to understand the logic here

packages/manager/modules/vps/src/dashboard/tile/configuration/upgrade/component.js
```diff
@@ -0,0 +1,11 @@
+import controller from './controller';
+import template from './index.html';
+
+export default {
+  name: 'vpsDashboardTileConfigurationUpgrade',
+  controller,
+  template,
+  bindings: {
+    resolve: '<',
```
marie-j: It sounds confusing to have a binding call `resolve`  with ui router. What is its purpose ?

packages/manager/modules/vps/src/dashboard/tile/configuration/upgrade/constants.js
```diff
@@ -0,0 +1,9 @@
+export const ORDER_TRACKING_URLS = {
+  EU: 'https://www.ovh.com/manager/dedicated/#/billing/order',
+  CA: 'https://ca.ovh.com/manager/#/billing/order',
+  US: 'https://us.ovhcloud.com/manager/dedicated/#/billing/order',
```
marie-j: Maybe we should use `RedirectionService` or at least CORE_MANAGER_URLS. Especially URL for Dedicated in CA is about to move soon

packages/manager/modules/vps/src/dashboard/tile/configuration/upgrade/routing.js
```diff
@@ -0,0 +1,222 @@
...
+
+import { Environment } from '@ovh-ux/manager-config';
+import VpsConfigurationTile from '../service';
+import { ORDER_TRACKING_URLS } from './constants';
+
+import component from './component';
+
+export default /* @ngInject */ ($stateProvider) => {
+  $stateProvider.state('vps.detail.dashboard.configuration.upgrade', {
+    url: '/upgrade/{upgradeType:memory|storage}?upgradeStatus&upgradeOrderId',
```
marie-j: I didn't know this was possible, nice :)

packages/manager/modules/vps/src/dashboard/tile/configuration/upgrade/routing.js
```diff
@@ -0,0 +1,222 @@
...
+
+          const upgradeModelPlan = get(
+            configurationTile,
+            `model[${upgradeType}]`,
+          );
+
+          // check if the model is setted correctly => if we access to the state
+          // directly (without clicking on oui-radio in dashboard state)
+          if (upgradeModelPlan.planCode !== upgradeToPlan.planCode) {
+            set(configurationTile.model, upgradeType, upgradeToPlan);
```
marie-j: To me it is not a good idea to update the value of `configurationTile.model` especially the value being injected. This is causing side effects and updating the injection without the programmer/user knowing. 
What is the need ?

JeremyDec:
@Jisay it seems that you need to update branch `feat/vps-new-range-information` and rebase your PR branch on it

packages/manager/modules/vps/src/import/vps.service.js
```diff
@@ -31,6 +31,8 @@ export default /* @ngInject */ function VpsService(
 
   const vpsTabVeeamCache = $cacheFactory('UNIVERS_WEB_VPS_TABS_VEEAM');
 
+  const apiCatalogProductName = 'virtualprivateserver';
```
JeremyDec: Reminder: in production, we will have to use /order/catalog/public/vps

Jisay: I didn't know that. Thanks ^^

packages/components/ng-ovh-stop-event/src/directive.js
Jisay: I had some problem running husky hook on commit. Seems to be fixed now. I will run it to correct it.

packages/manager/modules/vps/src/dashboard/tile/configuration/upgrade/component.js
Jisay: This is because the component is used by a `oui-modal` layout state. To be compatible with new version of `uibModal` resolves are accessible through resolve binding of the component. See https://github.com/ovh/manager/blob/feat/vps-new-range-configuration/packages/components/ng-ui-router-layout/src/oui-modal/index.js#L61

packages/manager/modules/vps/src/dashboard/tile/configuration/upgrade/routing.js
Jisay: I would like to reset the model in order to have the previous value in case of modal closing (click on cancel button, click on dismiss button or click outside of the modal). It was possible to reload the state but as we have a lot of call to make, it was not very fluid for customer. So I use this approach to directly reset the models that control the toggle group present in parent state view (the dashboard view).

packages/manager/modules/vps/src/dashboard/tile/configuration/upgrade/constants.js
Jisay: I didn't know there was this constant. I will check that. Thx

packages/manager/modules/vps/src/dashboard/tile/configuration/service.js
Jisay: Ok. I will add it

packages/manager/modules/vps/src/dashboard/tile/configuration/service.js
Jisay: The call is already done in fact :-)
And in my opinion the VPS information 2API is bad because it changes too much the response of the APIv6 call. The 2API should be rewritten in order to have something coherent.

ovh-cds:
CDS Report build-monorepository#10810.0 ✘  
* 
  * Build ✘

ovh-cds:
CDS Report build-monorepository#10810.0 ✘  
* 
  * Build ✘

packages/manager/modules/vps/src/import/vps.service.js
JeremyDec: we will handle it once all features has been validated by QA

packages/manager/modules/vps/src/dashboard/tile/configuration/upgrade/constants.js
Jisay: I will wait for #2310 in order to have global const.

packages/manager/modules/vps/src/dashboard/tile/configuration/upgrade/routing.js
marie-j: Could it be possible to create a copy when entering the modal, update that copy and if you need to on modal close set that copy as the new model for its parent ?

packages/manager/modules/vps/src/dashboard/tile/configuration/upgrade/routing.js
```diff
@@ -0,0 +1,222 @@
...
+                currentStorage: stateVps.model.disk,
+                nextStorage: stateVps.model.disk * 2,
+              };
+
+        return $translate.instant(translationKey, translationValues);
+      },
+
+      primaryLabel: /* @ngInject */ ($translate, upgradeStatus) => {
+        const translationKey =
+          upgradeStatus === 'success'
```
JeremyDec: Maybe `upgradeStatus === 'success' ` should be stored in a resolve variable, to be injected, as you use it more than once

packages/manager/modules/vps/src/dashboard/tile/configuration/upgrade/routing.js
```diff
@@ -0,0 +1,222 @@
...
+
+        return vpsUpgrade.getUpgrade(
+          serviceName,
+          get(configurationTile.upgrades, `${upgradeType}.plan.planCode`),
+          { quantity: 1 },
+        );
+      },
+
+      hasDefaultPaymentMethod: /* @ngInject */ (ovhPaymentMethod) =>
+        ovhPaymentMethod.hasDefaultPaymentMethod(),
```
JeremyDec: This resolve is available here: https://github.com/ovh/manager/blob/feat/vps-new-range-upgrade/packages/manager/modules/vps/src/routing.js#L36

packages/components/ng-ui-router-layout/src/oui-modal/index.js
```diff
@@ -153,8 +153,17 @@ angular
         };
 
         // close state modal when exiting the state
-        state.onExit = () => {
+        state.onExit = (transition) => {
           modalInstance.close();
+          // call state declaration onExit function if present
+          let onExitFn = state.self.onExit;
+          if (isArray(onExitFn)) {
```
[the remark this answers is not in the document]
Jisay: No, in fact, you have some injections. Once your state declared, the onExit function becomes an array if you have params in declaration. The last index is the function itself and the others indexes are the params that will be injected.
It's like you define an angular service:
```
angular.service('myService', ($log, $q) => { ... });
```

This will be annoted to something like:
```
angular.service('myService', ['$log', '$q', function ($log, $q) { ... }];
```

That's what is done by babel-plugin-angularjs-annotate and how angularJS worked at the beginning :-)

packages/manager/modules/vps/src/dashboard/tile/configuration/upgrade/routing.js
Jisay: We need first to have your code merged in order for me to use this resolve.

JeremyDec: Yes sorry, forgot to write it as a reminder ;)

packages/manager/modules/vps/src/dashboard/tile/configuration/upgrade/routing.js
Jisay: Mmmmmh, I don't really understand why changing the reference of the model object will change. The parent model object reference will also change.

PR #2620, opened 2020-03-24
feat(hub): add testing ids

frenautvh:

| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | internal MANAGER-4630

## Description

Add ids for e2e tests

### DIFF
--- a/packages/manager/modules/account-sidebar/src/payment-mean/template.html
+++ b/packages/manager/modules/account-sidebar/src/payment-mean/template.html
@@ -1,4 +1,7 @@
-<div class="manager-hub-payment-mean">
+<div
+    class="manager-hub-payment-mean"
+    data-navi-id="account-sidebar-paymentMean-block"
+>
     <a
         class="d-flex flex-row align-items-center p-2"
         data-ng-href="{{ :: $ctrl.RedirectionService.getURL('paymentMethod') }}"
--- a/packages/manager/modules/account-sidebar/src/template.html
+++ b/packages/manager/modules/account-sidebar/src/template.html
@@ -16,7 +16,11 @@
         >
         </ovh-manager-hub-shortcuts>
     </div>
-    <div data-ng-if="$ctrl.links" class="manager-hub-user-panel_links mb-5">
+    <div
+        data-ng-if="$ctrl.links"
+        class="manager-hub-user-panel_links mb-5"
+        data-navi-id="account-sidebar-usefulLinks-block"
+    >
         <h3 data-translate="hub_links_title"></h3>
         <div data-ng-repeat="link in $ctrl.links track by $index">
             <span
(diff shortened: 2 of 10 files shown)


packages/manager/modules/account-sidebar/src/template.html
```diff
@@ -13,7 +13,11 @@
         <ovh-manager-hub-shortcuts data-notifications="$ctrl.notifications">
         </ovh-manager-hub-shortcuts>
     </div>
-    <div data-ng-if="$ctrl.links" class="manager-hub-user-panel_links mb-5">
+    <div
+        data-ng-if="$ctrl.links"
+        class="manager-hub-user-panel_links mb-5"
+        data-navi-id="usefullLinks-block"
```
marie-j:
```suggestion
        data-navi-id="usefulLinks-block"
```

packages/manager/modules/account-sidebar/src/payment-mean/template.html
```diff
@@ -1,4 +1,4 @@
-<div class="manager-hub-payment-mean">
+<div class="manager-hub-payment-mean" data-navi-id="paymentMean-block">
```
FredericEspiau: How does it work ?

frenautvh: The attributes are given by QA team and are used for end to end tests

FredericEspiau: Is there a documentation about this ?

This looks like an extra unrequired overload, so could easily be fixed by coding semantically once again

cbourgois: I think we shoudn't need this for targeting element in tests (some great tools are here to help us to target elements as the customer target them: using text and global type (object, input...) like [Testing Library](https://testing-library.com/))
And IMO, this `navi-id` is ok in the context of the hub, but if you are in a global payment mean section, is `paymentMean-block` ok (because you could have more than one block in paymentMean page for example)? I think we should be more precise, here it's a `paymentMean-block` in a user sidebar.

antleblanc: cc @fchabrolle

fchabrolle: i think it's a block for payments inside the sidebar, if we can have multiple payments listed, there is still only one payment block containging them all. text shouldn't be used (it can be translated for example) theses attributes are here to find inputs easily

antleblanc: Could be handled right after: MANAGER-4630.

thanks all for your feedback

PR #2622, opened 2020-03-24
Feat/vps new range capabilities

JeremyDec:
| Question         | Answer
| ---------------- | ---
| Branch?          | feat/vps-new-range-information
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-4605

## Description

First VPS launched in US just had rebuild capability.
VPS new range ones will have all capabilities available, also in US, so we need to make difference between first vps launched (best-value), and the new coming ones.

### DIFF
--- a/packages/manager/modules/vps/src/routing.js
+++ b/packages/manager/modules/vps/src/routing.js
@@ -27,10 +27,15 @@ export default /* @ngInject */ ($stateProvider) => {
       resolve: {
         connectedUser: /* @ngInject */ (OvhApiMe) =>
           OvhApiMe.v6().get().$promise,
-        capabilities: /* @ngInject */ (serviceName, OvhApiVpsCapabilities) =>
-          OvhApiVpsCapabilities.Aapi()
-            .query({ serviceName })
-            .$promise.then((capabilities) =>
+        capabilities: /* @ngInject */ ($http, serviceName, stateVps) =>
+          $http
+            .get(`/vps/capabilities/${serviceName}`, {
+              serviceType: 'aapi',
+              params: {
+                modelName: stateVps.model.name,
+              },
+            })
+            .then((capabilities) =>
               capabilities.map((capability) => kebabCase(capability)),
             ),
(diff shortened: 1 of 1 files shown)


packages/manager/modules/vps/src/routing.js
```diff
@@ -27,11 +33,22 @@ export default /* @ngInject */ ($stateProvider) => {
...
+          OvhApiVpsCapabilities,
+        ) =>
           OvhApiVpsCapabilities.Aapi()
             .query({ serviceName })
             .$promise.then((capabilities) =>
-              capabilities.map((capability) => kebabCase(capability)),
+              capabilities
+                .filter((capability) =>
+                  !isVpsNewRange && coreConfig.isRegion('US')
+                    ? capability === DASHBOARD_FEATURES.rebuild
```
FredericEspiau: Why aren't capabilities only handled by the `vps-capabilities` 2api ?

JeremyDec: I choose that the `vps-capabilities` just returns all items for US, and then it is filtered here (as it is meant to be removed). Also it is to avoid to send to 2api serviceName + modelName (which contains information of vps is new range or not)

FredericEspiau: No I meant that the 2API could tell you when a capability is available for a new range or a legacy, but don't change anything

packages/manager/modules/vps/src/routing.js
```diff
@@ -56,6 +65,8 @@ export default /* @ngInject */ ($stateProvider) => {
           }
           return promise;
         },
+        isVpsNewRange: /* @ngInject */ (stateVps) =>
+          stateVps.model.version === NEW_RANGE_VERSION,
```
marie-j, line 69: Another way could have been to have a `Vps` class and manipulate instances with a `isNewRange()` method

JeremyDec: I'll keep this in mind for a later refactor or changes on vps section

PR #2646, opened 2020-03-26
fix(hub): billing style

frenautvh:

| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-4799

## Description

fix hub billing tile style

### DIFF
--- a/packages/manager/modules/hub/src/variables.scss
+++ b/packages/manager/modules/hub/src/variables.scss
@@ -1,4 +1,4 @@
-@import 'oui4/dist/scss/tokens/_colors.scss';
+@import 'oui4/dist/scss/_tokens';
 
 $hub-brand-color: #001191;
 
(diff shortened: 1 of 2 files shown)


packages/manager/modules/hub/src/variables.scss
```diff
@@ -43,6 +44,8 @@ $hub-border-radius-default: 6pt;
 
   &:hover {
     color: $hub-button-fg-color;
+    background-color: $p-700;
+    box-shadow: 0 4px 6px 0 rgba(0, 14, 156, 0.2);
```
marie-j: I think there is a variable for this but it can be done in a second time

frenautvh: looks like p-700 with 20% opacity

marie-j: $base-box-shadow-secondary

PR #2678, opened 2020-03-31
fix(dedicated): legay bare metal servers

radireddy:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-9904

## Description

Customers are not able to access network interface of servers that are not migrated to the agora. This is because we call /order API of agora to get OLA price details. I made a change to only call this API for servers migrated to Agora and OLA is enabled.


cbourgois:
Missing `--signoff` in commit.

PR #2712, opened 2020-04-01
Optimize assets minification and moment locales

antleblanc:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-4864

### ⚡️ Performace

555f3de - perf: add plugin to optimize css asset minification

uses: `$ yarn workspace @ovh-ux/manager-webpack-config add add 
optimize-css-assets-webpack-plugin terser-webpack-plugin`

**Before**:

```sh
$ ls -alh packages/manager/apps/hub/dist/ | grep css
-rw-r--r--    1 antleblanc  staff   444K Apr  1 21:08 287.css
-rw-r--r--    1 antleblanc  staff   398K Apr  1 21:08 288.css
-rw-r--r--    1 antleblanc  staff   174B Apr  1 21:08 289.css
-rw-r--r--    1 antleblanc  staff    31B Apr  1 21:08 290.css
````

**After**:

```sh
$ ls -alh packages/manager/apps/hub/dist/ | grep css
-rw-r--r--    1 antleblanc  staff   380K Apr  1 21:00 287.css
-rw-r--r--    1 antleblanc  staff   390K Apr  1 21:00 288.css
-rw-r--r--    1 antleblanc  staff   172B Apr  1 21:00 289.css
-rw-r--r--    1 antleblanc  staff    29B Apr  1 21:00 290.css
```

### 🔗 Related

- https://webpack.js.org/plugins/mini-css-extract-plugin/#minimizing-for-production

### 🔖 TODO

- [x] #2755 

Signed-off-by: Antoine Leblanc <ant.leblanc@gmail.com>


cbourgois:
Does-it apply only on `production` or also in `development` ?

antleblanc:
> Does-it apply only on `production` or also in `development` ?

Applied only in mode `production`.

packages/manager/tools/webpack-config/src/webpack.common.ts
```diff
@@ -6,6 +6,9 @@ import CopyWebpackPlugin from 'copy-webpack-plugin';
 import get from 'lodash/get';
 import set from 'lodash/set';
 import MiniCssExtractPlugin from 'mini-css-extract-plugin';
+import TerserJSPlugin from 'terser-webpack-plugin';
```
FredericEspiau, line 9: Why not `TerserWebpackPlugin` ?

antleblanc: Roughly inspired by the webpack documentation (see: https://webpack.js.org/plugins/mini-css-extract-plugin/#minimizing-for-production).

marie-j: @antleblanc I don't understand, link should be https://webpack.js.org/plugins/terser-webpack-plugin/ right ?

antleblanc: Please reach the section called "Minimizing For Production" (anchor doesn't work correctly).

packages/manager/apps/iplb/package.json
```diff
@@ -54,7 +54,7 @@
     "chart.js": "chartjs/Chart.js#^2.0",
     "flatpickr": "~4.5.2",
     "jquery": "^2.1.3",
-    "moment": "^2.19",
+    "moment": "^2.24.0",
```
marie-j, line 57: No need to change the import or the webpack config for standalone apps ?

antleblanc: You're right! We have to find solutions for standalone applications too.

As we were mainly focused on the "Universe" (Large) applications I've only updated the impacted webpack configurations.

Last weeks some discussions has started around a shift to date-fns or other libraries.

We can keep investigating to a suitable solution for the long-term and then apply it across all applications.

PR #2717, opened 2020-04-02
fix(private-registry): wrong translation in create page

radireddy:
rename "Nom du cluster" to "Nom du Private Registry"

Closes #MANAGER-4852

| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-4852

## Description

When creating a registry the name field is called "Nom du cluster" should be "Nom du Private Registry". Renamed it to "Nom du Private Registry"


antleblanc:
@radireddy Could you please signed-off your commit?

Thanks!

PR #2732, opened 2020-04-03
fix(navbar): notification text not visible in dark mode

radireddy:
Closes #DTRSD-10442

| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-10442

## Description

Notification content is not visible after adding hub because of same background color. I have added required CSS changes to make it visible.

### DIFF
--- a/packages/manager/apps/public-cloud/src/assets/theme/dark/index.less
+++ b/packages/manager/apps/public-cloud/src/assets/theme/dark/index.less
@@ -1000,5 +1000,26 @@
   manager-error-page .ovh-manager-error-page {
     background: #181818;
   }
+
+  /* hub navbar */
+  .oui-navbar-menu-hub {
+    @notification-bcg-color: #def8ff;
+
+    background-color: @notification-bcg-color;
+    border: 1px solid @notification-bcg-color;
+
+    .oui-navbar-notification {
+      .oui-status_error {
+        color: #c11b1b;
+      }
+      .oui-status_info {
+        color: #4d6f92;
+      }
+      a.oui-navbar-notification__content,
+      a.oui-navbar-notification__content a {
+        color: @dark-900;
+      }
+    }
+  }
 }
 /* stylelint-enable */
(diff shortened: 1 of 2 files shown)


FredericEspiau:
Looks like a hotfix, isn't it ?

antleblanc:
> Looks like a hotfix, isn't it ?

🤔 Dark mode.

This _feature_ is in alpha mode so it doesn't deserve an _hotfix_, especially for the affected parts I guess ;)

FredericEspiau:
> > Looks like a hotfix, isn't it ?
> 
> 🤔 Dark mode.
> 
> This _feature_ is in alpha mode so it doesn't deserve an _hotfix_, especially for the affected parts I guess ;)

Didn't know it wasn't fully implemented yet

packages/manager/apps/public-cloud/src/assets/theme/dark/index.less
```diff
@@ -1000,5 +1000,26 @@
...
+
+    .oui-navbar-notification {
+      .oui-status_error {
+        color: #c11b1b;
+      }
+      .oui-status_info {
+        color: #4d6f92;
+      }
+      a.oui-navbar-notification__content,
+      a.oui-navbar-notification__content a {
```
antleblanc, line 1019: Nested `<a>` selector? 🤔 

Is there an issue with the component template?

radireddy: yes, HTML has nested <a> tag

radireddy: @antleblanc there is no issue with "a.oui-navbar-notification__content a". It required because HTML has nested anchor tag. I verified and it is working properly.

antleblanc: @radireddy Having nested anchor tags is an issue. This is not valid.
We have to fix the navbar notification template first.
Could you please have a look? Thanks!

@marie-j Is this template has recently been modified to remove this nested anchor tags issue?

marie-j: @antleblanc The issue is the component is an anchor tag but we sometimes have anchor tags in the content returned by 2API notification (not always but for one or two)

PR #2755, opened 2020-04-06
perf: import dynamically moment locales

antleblanc:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feature/webpack-config-optimize-css`
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | no

### ⚡️ Performance

46fa030 - perf: import dynamically moment locales

Signed-off-by: Antoine Leblanc <ant.leblanc@gmail.com>

### DIFF
--- a/packages/manager/apps/hub/webpack.config.js
+++ b/packages/manager/apps/hub/webpack.config.js
@@ -70,6 +70,11 @@ module.exports = (env = {}) => {
       mainFields: ['module', 'browser', 'main'],
     },
     plugins: [
+      new webpack.ContextReplacementPlugin(
+        /moment[/\\]locale$/,
+        /cs|de|en-gb|es|es-us|fi|fr-ca|fr|it|lt|pl|pt/,
+      ),
+
       new webpack.DefinePlugin({
         __NODE_ENV__: process.env.NODE_ENV
           ? `'${process.env.NODE_ENV}'`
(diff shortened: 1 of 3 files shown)


packages/manager/apps/hub/webpack.config.js
```diff
@@ -70,6 +70,11 @@ module.exports = (env = {}) => {
       mainFields: ['module', 'browser', 'main'],
     },
     plugins: [
+      new webpack.ContextReplacementPlugin(
+        /moment[/\\]locale$/,
+        /cs|de|en-gb|es|es-us|fi|fr-ca|fr|it|lt|pl|pt/,
```
antleblanc, line 75: Based on the following list: https://github.com/ovh/manager/blob/3297580738182ce4284efa189840c30c9d02041e/packages/manager/tools/component-rollup-config/src/plugins/translation-utils.ts#L7-L20

marie-j: Could this be declared in the common webpack config ? so it can be reused in all managers

antleblanc: Since our webpack configuration is already glued with the Rollup configuration I think I will be able to get access to this list but I think it's not a good solution.

So as you've suggested I will integrate this list directly into the webpack configuration.
Thanks for the feedback.

antleblanc: After digging around it I didn't any option to factorize this code snippet.

Let me know if you have some ideas to improve it.

For now I will distribute this enhancement across other webpack configuration.

Thanks!

PR #2768, opened 2020-04-08
fix(sign-up): use differents labels depending on selected legalform

Jisay:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-4812

## Description

Use differents translations for vat and organisation fields depending on the selected legalform and selected country (for vat field).


packages/manager/modules/sign-up/src/form/activity/constants.js
```diff
@@ -0,0 +1,71 @@
...
+  GT: 'NIT',
+  HN: 'RTN',
+  MX: 'RFC',
+  NI: 'RUC',
+  PA: 'RUC',
+  PY: 'RUC',
+  PE: 'RUC',
+  DO: 'RNC',
+  UY: 'RUT',
+  VE: 'RIF',
```
marie-j, line 66: No library or API that already has this kind of knowledge I guess

Jisay: Indeed, I don't find a library that handle it.
And for API, they have such informations but it's not exposed and don't plan to expose it

PR #2772, opened 2020-04-09
fix(public-cloud): onboarding redirect

ganeshkumar1989:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-10648

## Description

When visiting public cloud manager from any other managers (like web, dedicated etc), the user should be redirected to the onboarding page, when he does not have any project. But this is not happening. A search parameter `onboarding` is already being passed, but this parameter was not being considered properly. This has been handled

### DIFF
--- a/packages/manager/apps/public-cloud/src/index.routes.js
+++ b/packages/manager/apps/public-cloud/src/index.routes.js
@@ -8,8 +8,9 @@ export default /* @ngInject */ ($stateProvider, $urlRouterProvider) => {
         .injector()
         .get('publicCloud')
         .getDefaultProject()
-        .then((projectId) =>
-          projectId
+        .then((projectId) => {
+          const $window = trans.injector().get('$window');
+          return projectId
             ? {
                 state: 'pci.projects.project',
                 params: {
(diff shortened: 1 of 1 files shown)


ganeshkumar1989:
Sure @cbourgois . I will fix that

packages/manager/apps/public-cloud/src/index.routes.js
```diff
@@ -8,20 +8,22 @@ export default /* @ngInject */ ($stateProvider, $urlRouterProvider) => {
...
                 },
               }
             : {
-                state: trans.params().onboarding
-                  ? 'pci.projects.onboarding'
-                  : 'pci.projects.new',
-              },
-        ),
+                state:
+                  $window.location.search.search('onboarding') !== -1
```
FredericEspiau, line 22: It didn't work with `$transition$.params` ? Do you know why ?

ganeshkumar1989: No it does not work... :) 
`$transition$.params` has access to the route params. But here, it is a search param outside of angular routes.

ganeshkumar1989:
Closing this Pull request since the public cloud url has been updated to pass onboarding as a param at navbar-universes: 14

PR #2792, opened 2020-04-15
feat(cloud.logs): remove options concern on roles

pdepaepe:
Hello,

This PR introduce following changes:

- An option or a plan is no more required on roles
- Initiate content for a new incoming LDP plan: ldp-account

Related internal tickets: OB-3225, OB-3226 

As always, thanks for your review, merge & prod time

Signed-off-by: Pierre De Paepe <pierre.de-paepe@ovhcloud.com>


packages/manager/apps/cloud/client/app/dbaas/logs/detail/logs-detail.less
```diff
@@ -15,6 +15,19 @@
   }
 }
 
+ul.overview {
```
antleblanc, line 18: The `ul` HTML element already provide this behavior.
The issue here is that the padding left is removed via the following helper: `pl-0`.
Maybe you can simply remove it. Is that make sense for you?

In that way, you will be able to get rid of all the styles added here.

packages/manager/apps/cloud/client/app/dbaas/logs/detail/roles/overview/logs-role-overview.html
```diff
@@ -27,36 +27,44 @@
             <div data-ng-if="ctrl.role.permissions.length > 0">
                 <div data-ng-if="ctrl.permissions.stream.length > 0">
                     <span class="pl-0 preview-list__list-heading">Stream</span>
-                    <ul
-                        data-ng-repeat="stream in ctrl.permissions.stream track by $index"
-                    >
-                        {{stream.title}}
+                    <ul class="overview pl-0">
```
antleblanc, line 30: Please refer to my previous comment.
Something like this should be enough:

```suggestion
                    <ul>
```

packages/manager/apps/cloud/client/app/dbaas/logs/detail/options/home/logs-options-home.html
```diff
@@ -4,16 +4,16 @@
...
-    <p class="oui-paragraph">
-        <span data-translate="log_options_intro_text"></span>
-        <span data-translate="logs_options_description_manage"></span>
-        <a
-            data-translate="logs_options_manage_link"
-            data-ui-sref="dbaas.logs.detail.options.manage"
-        ></a
-        >.
-    </p>
+    <div class="row mb-4">
```
antleblanc, line 7: Using a `.row` implies that the children elements  must have a `.col-x`.
But here it is not useful.
Why not simply considering using a regular `<p>` element like:

```html
<p class="oui-paragraph">
    <span data-translate="logs_options_description_manage"></span>
    <a
        data-translate="logs_options_manage_link"
        data-ui-sref="dbaas.logs.detail.options.manage"
    ></a
    >.
</p>
```

packages/manager/apps/cloud/client/app/dbaas/logs/detail/logs-detail.less
```diff
@@ -15,6 +15,19 @@
...
+  list-style-type: none;
+}
+
+ul.overview > li {
+  text-indent: -5px;
```
frenautvh, line 23: we prefer to use rem sizing if possible

packages/manager/apps/cloud/client/app/dbaas/logs/detail/logs-detail.less
pdepaepe: You are right, additional style is not useful here. I removed it.

packages/manager/apps/cloud/client/app/dbaas/logs/detail/roles/overview/logs-role-overview.html
pdepaepe: Fixed yes.

packages/manager/apps/cloud/client/app/dbaas/logs/detail/options/home/logs-options-home.html
pdepaepe: If it's not in this div, line is not aligned with the content following in the page. 
You will see in the source page that next content is also in a `<div class="row mb-4">`.

Also please consider that this page will disappear in a short term.

packages/manager/apps/cloud/client/app/dbaas/logs/detail/logs-detail.less
pdepaepe: No more a subject since additional style has been removed from this PR.

packages/manager/apps/cloud/client/app/dbaas/logs/detail/options/home/logs-options-home.html
antleblanc: Ok so it means that the next element is applying some margin which is responsible of some inconsistency.

We can take it into account after if necessary so it's not mandatory then.

antleblanc:
**Internal ticket reference**: MANAGER-4960

PR #2793, opened 2020-04-15
[WIP] Feat/ui consistency

AxelPeter:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-4953

## Description

- Standardize version of UI Kit dependencies across Managers
- Apply OVHcloud light theme on the Sidebar.
- Clean CSS overrides that has been centralized in the component package

**Need a release of this package before merge**: https://github.com/ovh-ux/ng-ovh-sidebar-menu/pull/120

ref: MANAGER-4953


ovh-cds:
CDS Report build#11452.0 ■   
* 
  * Build ■

ovh-cds:
CDS Report build#11452.0 ■   
* 
  * Build ■

packages/components/ng-ovh-contacts/package.json
```diff
@@ -52,8 +52,8 @@
     "flag-icon-css": "^3.2.1",
     "moment": "^2.20.1",
     "ovh-api-services": "^9.0.0",
-    "ovh-ui-angular": "^2.23.x",
-    "ovh-ui-kit": "^2.23.x",
+    "ovh-ui-angular": "^3.16.0",
+    "ovh-ui-kit": "^2.41.1",
```
marie-j, line 56: I think this was already done on `develop` branch

packages/manager/modules/navbar/src/index.scss
```diff
@@ -3,7 +3,7 @@
 
   background-color: $p-000-white;
   box-shadow: $base-box-shadow-primary;
-  z-index: 100;
+  z-index: 1032;
```
marie-j, line 6: I'm not sure we should play with `z-index`, it oftens look like we put a random number so it works

packages/manager/modules/server-sidebar/src/template.html
```diff
@@ -1,36 +1,8 @@
...
-        </div>
-    </div>
-    <div
-        data-sidebar-menu
-        data-ng-class="{
-            'nav-open': $ctrl.isOpen
-        }"
-        class="col-md-3 col-lg-2"
-    ></div>
-</div>
```
marie-j, line 33: I think this was meant to handle responsive navigation. Has it been moved to soemwhere else ?

AxelPeter: In fact, the way it was handled didn't work at all, and it was missing in Telecom.
This should have been managed directly form the main component.
So for now, I remove it to correct it later directly in `ng-ovh-sidebar-menu`.

packages/manager/modules/navbar/src/index.scss
AxelPeter: It looks like a random number, but it's set compared to the navbar and the backdrop.

navbar: 1032
sidebar: 1031
backdrop: 1030

Why we've started so high, I don't know. I suppose it was based on the value from Bootstrap.

packages/components/ng-ovh-contacts/package.json
AxelPeter: That's not really what was saying the `yarn.log` on my side when I've started it from `develop`. But it's not a bad idea to correct all of this. There was some `~` and fixed versions in some `package.json` :)

packages/manager/modules/navbar/src/index.scss
Jisay: So we can maybe create a var which expose the z-index of the navbar?

AxelPeter: Or maybe directly use the values of the UI kit :P
https://github.com/ovh/ovh-ui-kit/blob/v2.x/packages/oui-navbar/_variables.less#L11

I'll update it, it should be clearer if all if regrouped in a single place ;)

AxelPeter: I've didn't see that the file was a `SCSS` one. So it'll be difficult to use UI-kit variables for now.

marie-j: @AxelPeter I think you can import the variables from v4, I don't know if that is of any help with the `z-index`

AxelPeter: IMO, I think that importing a new package as big as the UI Kit, just to use some variables, is a bit to much (I mean here with v2 + v4). Of course we lose the centralized side of the thing by doing like that.

To see a little farther than just the `z-index`. It should be done as a whole with the way we manage the main page layout of all the managers. But it's something that will be planned very soon :)

AxelPeter:
I'll close this PR and recreate it later, since we want to test the UI consistency changes as an entire batch. The additions of the others tickets should normally be only small changes and packages updates :)

PR #2798, opened 2020-04-16
feat(web): add private db detach action

JeremyDec:

| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-3327

### DIFF
--- a/packages/manager/apps/web/client/app/hosting/database/detach-private/index.js
+++ b/packages/manager/apps/web/client/app/hosting/database/detach-private/index.js
@@ -1,0 +1,22 @@
+import angular from 'angular';
+import productOffers from '@ovh-ux/manager-product-offers';
+import '@uirouter/angularjs';
+import 'angular-translate';
+import 'ovh-ui-angular';
+
+import component from './component';
+import routing from './routing';
+
+const moduleName = 'ovhManagerHostingDatabaseDetachPrivateModule';
+
+angular
+  .module(moduleName, [
+    productOffers,
+    'oui',
+    'ui.router',
+    'pascalprecht.translate',
+  ])
+  .component(component.name, component)
+  .config(routing);
+
+export default moduleName;
--- a/packages/manager/apps/web/client/app/hosting/database/detach-private/template.html
+++ b/packages/manager/apps/web/client/app/hosting/database/detach-private/template.html
@@ -1,0 +1,27 @@
+<div data-ovh-alert="detach"></div>
+
+<oui-back-button on-click="$ctrl.goBack()">
+    <span data-translate="hosting_database_private_detach_option_title"></span>
+</oui-back-button>
+
+<p>
+    <span
+        data-translate="hosting_database_private_detach_option_description"
+        data-translate-values="{
+          'domain': '<strong>' + $ctrl.workflow.options.optionId + '</strong>',
+          'hosting': '<strong>' + $ctrl.serviceName + '</strong>',
+        }"
+    ></span>
+    <strong
+        data-translate="hosting_database_private_detach_option_warning_irreversible"
+    ></strong>
+</p>
+
+<ovh-manager-product-offers
+    pricing-type="$ctrl.pricingType"
+    user="$ctrl.user"
+    on-error="$ctrl.onError(error)"
+    on-success="$ctrl.onSuccess(detachResult)"
+    workflow-type="$ctrl.workflow.type"
+    workflow-options="$ctrl.workflow.options"
+></ovh-manager-product-offers>
--- a/packages/manager/apps/web/client/app/hosting/hosting.routes.js
+++ b/packages/manager/apps/web/client/app/hosting/hosting.routes.js
@@ -58,10 +58,45 @@ export default /* @ngInject */ ($stateProvider) => {
           ),
       pendingTasks: /* @ngInject */ (HostingTask, serviceName) =>
         HostingTask.getPending(serviceName).catch(() => []),
+      privateDatabasesIds: /* @ngInject */ (HostingDatabase, serviceName) =>
+        HostingDatabase.getPrivateDatabaseIds(serviceName).catch(() => []),
+      privateDatabasesDetachable: /* @ngInject */ (
+        $q,
+        ovhManagerProductOffersDetachService,
+        PrivateDatabase,
+        privateDatabasesIds,
+      ) =>
+        $q
+          .all(
+            privateDatabasesIds.map((id) =>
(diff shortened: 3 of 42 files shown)


packages/manager/apps/web/client/app/hosting/database/detach-private/template.html
```diff
@@ -0,0 +1,26 @@
+<div data-ovh-alert="detach"></div>
+
+<oui-back-button on-click="$ctrl.goBack()">
```
radireddy, line 3:
```suggestion
<oui-back-button data-on-click="$ctrl.goBack()">
```

packages/manager/apps/web/client/app/hosting/database/detach-private/template.html
```diff
@@ -0,0 +1,26 @@
...
+          'hosting': '<strong>' + $ctrl.serviceName + '</strong>',
+        }"
+    ></span>
+    <strong
+        data-translate="hosting_database_private_detach_option_warning_irreversible"
+    ></strong>
+</p>
+
+<ovh-manager-product-offers
+    pricing-type="$ctrl.pricingType"
```
radireddy, line 20: data-

packages/manager/apps/web/client/app/hosting/database/detach-private/index.js
```diff
@@ -0,0 +1,14 @@
+import angular from 'angular';
+import '@uirouter/angularjs';
+
+import component from './component';
+import routing from './routing';
+
+const moduleName = 'ovhManagerHostingDatabaseDetachPrivateModule';
+
+angular
+  .module(moduleName, ['ui.router'])
```
frenautvh, line 10: some missing deps like angular-translate, ui-kit etc

packages/manager/apps/web/client/app/hosting/hosting.routes.js
```diff
@@ -52,10 +52,41 @@ export default /* @ngInject */ ($stateProvider) => {
           .then((servicesInformation) => servicesInformation.flatten()),
       pendingTasks: /* @ngInject */ (HostingTask, serviceName) =>
         HostingTask.getPending(serviceName).catch(() => []),
+      privateDatabasesIds: /* @ngInject */ (HostingDatabase, serviceName) =>
+        HostingDatabase.getPrivateDatabaseIds(serviceName),
+      privateDatabasesDetachable: /* @ngInject */ (
```
frenautvh, line 57: It might be a performance issue to add this resolve at the "app.hosting" state. It means every time you go the a child state of app.hosting you will fetch all the services infos of privates databases of the hosting

Do every sub-states of app.hosting requires this information? if no you should move this resolve in the corresponding child state

FredericEspiau: No every time, each time :)

Because if you go from state A to state A.B, it won't be fetched between both, because it has already been fetch, won't it ?

FredericEspiau:
It's not true, private SQL database aren't linked to a hosting but to a datacenter, maybe we should explain this instead

packages/manager/apps/web/client/app/hosting/hosting.routes.js
frenautvh: Yes if you go from A to A.B it's already fetched but if the data is only required by B then there is no need to load it in A. If everybody is putting data loading at a high level state such as "app.hosting" then the whole application will takes more time to load and it could be easily avoided.

I don't know in the case of this PR if it makes sense but i had the feeling that putting the database fetching queries in the routing of "app.hosting" could be avoided. Hope it makes sense :)

FredericEspiau: I was nitpicking your sentence, your reasoning was correct !

JeremyDec: @frenautvh, problem is, hosting dashboard requires main routing of hosting, and the actions are in the dashboard so...
But you're right, it may causes performance issues.
To fix that, the general informations dashboard should have its own routing.

I will create a task to do the refactor

JeremyDec: @frenautvh => MANAGER-4966

JeremyDec:
@FredericEspiau has discussed and confirm with PU, the link has to be in hosting dashboard, as it concerns only private SQL databases included in specific hosting offer.

ovh-cds:
CDS Report installation#11881.0 ■   
* Stage 1
  * Install ■

ovh-cds:
CDS Report installation#11881.0 ■   
* Stage 1
  * Install ■

JeremyDec:
Blocked until translations are made

ovh-cds:
CDS Report translator-push#7636.0 ✘  
* 
  * push ✘

ovh-cds:
CDS Report translator-push#7639.0 ✘  
* 
  * push ✘

ovh-cds:
CDS Report translator-push#7640.0 ✘  
* 
  * push ✘

ovh-cds:
CDS Report translator-push#7641.0 ✘  
* 
  * push ✘

ovh-cds:
CDS Report translator-push#7642.0 ✘  
* 
  * push ✘

ovh-cds:
CDS Report translator-push#7643.0 ✘  
* 
  * push ✘

ovh-cds:
CDS Report translator-push#7644.0 ✘  
* 
  * push ✘

packages/manager/apps/web/client/app/hosting/hosting.routes.js
```diff
@@ -58,10 +58,45 @@ export default /* @ngInject */ ($stateProvider) => {
...
+                  .getAvailableDetachPlancodes(serviceId)
+                  .catch(() => [])
+                  .then((plancodes) => ({
+                    optionId: domain,
+                    serviceId,
+                    detachPlancodes: plancodes,
+                  })),
+              ),
+            ),
+          ),
```
marie-j, line 93: Maybe we can add a `catch` to avoid crashing the whole hosting dashboard and adapt the display inside ? (can be done in another PR as this is already QC ok)

JeremyDec: If I'm not wrong, L85 does the job to secure the call, as it is the only API request done in this resolve :).
But you make me think of securing the call to retrieve private databases for a hosting.
And with all those issues we had with hosting, I'll do it in this PR

PR #2810, opened 2020-04-17
telephony line groups: preload all groups, sort by name, skip current

julieni:


packages/manager/apps/telecom/src/app/telecom/telephony/billingAccount/administration/linesGroup/telecom-telephony-billing-account-administration-lines-group.controller.js
```diff
@@ -54,14 +54,26 @@ angular
 
         return $q
           .all({
-            billingAccounts: OvhApiTelephony.v6().query().$promise,
+            billingAccounts: OvhApiTelephony.v7()
```
antleblanc, line 57: Thanks @julieni for suggesting code changes to the Control Panel, however the
usage of the apiv7 wrapper is deprecated.

The reason is that the team who has working on it has deprecated the usage of
theses features due to some issues with the cache mechanism and they've started
to exposed a new solution called **Iceberg**.

You can refer to the api wrapper to find this all theses informations.
(see: https://github.com/ovh-ux/ng-ovh-api-wrappers#api-v7).

In the same way we noticed that the documentation is not clearly mentioning it
so we will fix that.

So we are sorry to inform you that we won't be able to integrate your changes
but we will take into an account your suggestions in order to improve the
Billing Account Administration section.

Thank you!

julieni: no problem, just switched to iceberg ;)

antleblanc:
**Internal ticket reference**: MANAGER-5053

PR #2826, opened 2020-04-20
fix(server-sidebar): update web constants for CA

cbourgois:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     |no
| Breaking change? | no
| Tickets          | 

## Description

Fix web sidebar for CA (product listing + order links).

### DIFF
--- a/packages/manager/modules/server-sidebar/src/order.constants.js
+++ b/packages/manager/modules/server-sidebar/src/order.constants.js
@@ -254,6 +254,12 @@ export const ORDER_URLS = {
     },
   },
   CA: {
+    orderDomain: {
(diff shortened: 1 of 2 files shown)


packages/manager/modules/server-sidebar/src/order.constants.js
```diff
@@ -254,6 +254,12 @@ export const ORDER_URLS = {
     },
   },
   CA: {
+    orderDomain: {
+      CA: 'https://www.ovh.com/ca',
```
cbourgois, line 258: We don't have links for now, but it's to avoid a display bug if no links are provided to the order menu.

PR #2827, opened 2020-04-21
Model version number

FredericEspiau:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | None

## Description
- **Dedicated**: fix duplicated dependency declaration
- **Hub** & **Support**: use the same version number as the other packages

### DIFF
--- a/packages/manager/apps/dedicated/package.json
+++ b/packages/manager/apps/dedicated/package.json
@@ -143,7 +143,6 @@
     "whatwg-fetch": "^3.0.0"
   },
   "devDependencies": {
-    "@ovh-ux/manager-models": "^0.0.0 || ^1.0.0",
     "@ovh-ux/manager-webpack-config": "^3.1.1",
     "eslint-plugin-angular": "^4.0.1",
     "glob": "^7.1.4",
--- a/packages/manager/modules/hub/package.json
+++ b/packages/manager/modules/hub/package.json
@@ -12,7 +12,7 @@
   "author": "OVH SAS",
   "main": "./src/index.js",
   "dependencies": {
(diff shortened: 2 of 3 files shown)


packages/manager/apps/dedicated/package.json
```diff
@@ -143,7 +143,6 @@
     "whatwg-fetch": "^3.0.0"
   },
   "devDependencies": {
-    "@ovh-ux/manager-models": "^0.0.0 || ^1.0.0",
```
FredericEspiau, line 146: There already is a declaration in the `dependencies`

marie-j: If we only  use `import { foo } from @ovh-ux/manager-models` maybe we can leave it in the `devDependencies`

FredericEspiau: Don't know, it's in the `dependencies` everywhere, you shouldn't have to worry about how a package is used to know how you declare it, or the day you change the way you use it, then you have to change the way you declare it 🤔

antleblanc: @marie-j The business logic contained inside the `@ovh-ux/manager-models` module looks like a dependency from the Dedicated application point of view, so I'm ok with theses changes.

Is that make sense for you?

marie-j: @FredericEspiau indeed, however in the current context there is a difference between `devDependencies` and `dependencies`. If you use a dependency to only `import` things it can be required as a `devDependencies` as it is used to only build your app.
As it is already done that way, it's ok to leave it that way but we have to be sure that we don't install/bring unnecessary dependencies at runtime

PR #2832, opened 2020-04-21
feat(cloud.logs): remove options concern from dashboards

pdepaepe:
Hello,

As we did it previously for roles, this PR will allow LDP customers to handle dashboards without  quota (plan & options) concerns.

Thanks for your review, merge & prod time on this subject.

Signed-off-by: Pierre De Paepe <pierre.de-paepe@ovhcloud.com>


antleblanc:
**Internal ticket reference**: MANAGER-4975

PR #2842, opened 2020-04-23
fix(emailpro): canada api calls

ganeshkumar1989:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/remove-options-dashboard`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-4912

## Description

Avoid calls to `/email/domain` Apis for CA region

ref: MANAGER-4912

Signed-off-by: Ganesh Kumar <ganesh.kumar@corp.ovh.com>

### DIFF
--- a/packages/manager/modules/emailpro/src/emailpro.service.js
+++ b/packages/manager/modules/emailpro/src/emailpro.service.js
@@ -2,6 +2,7 @@ import angular from 'angular';
 import punycode from 'punycode';
 import camelCase from 'lodash/camelCase';
 import find from 'lodash/find';
+import get from 'lodash/get';
 import isEmpty from 'lodash/isEmpty';
 import keys from 'lodash/keys';
 import map from 'lodash/map';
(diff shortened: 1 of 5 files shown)


packages/manager/modules/emailpro/src/emailpro.service.js
```diff
@@ -32,6 +35,11 @@ export default class EmailPro {
     this.$timeout = $timeout;
     this.$translate = $translate;
 
+    this.apiRoutes = get(
+      EMAILPRO_CONFIG.API_ROUTES,
+      Environment.getRegion(),
```
cbourgois, line 40: In `Angular.js` you could use `CoreConfig.getRegion()` provided by `@ovh-ux/manager-core`

ganeshkumar1989: I used `manager-config` since it is a smaller package. But it makes sense to use `manager-core` in angular context. I have done the change. Thanks :)

PR #2847, opened 2020-04-24
fix: fix styles to match ui specs

AxelPeter:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-4992

## Description

- Update `ovh-ui-kit` and `ovh-ui-angular`
- Fix sidebar positionning
- Fix navbar icons spacing
- Fix button styles based of the UI specs (use of `oui-button_ghost` and `oui-button_primary`)
- Fix small stylelint errors
- Fix BEM naming on some classnames: it should be `block-name__element-name` instead of `block-name_element-name`


packages/manager/modules/navbar/src/index.less
```diff
@@ -158,8 +158,10 @@ ovh-manager-navbar {
   }
 
   .ovh-sidebar-wrapper {
-    position: absolute;
+    position: fixed;
+    top: 2.75rem; /* Navbar height */
```
JeremyDec, line 162: There is no variable which stores this height into ui-kit maybe ?

AxelPeter: Probably, I'll take a look

AxelPeter: You're right, nice catch ! Updated ;)

AxelPeter:
@frenautvh Sorry, the new commit dismissed your approve

antleblanc:
Updating both `ovh-ui-kit` and `ovh-ui-angular` dependencies must be done for all packages.

Could be done with the following command:
```sh
$ yarn upgrade-interactive --latest ovh-ui-kit
$ yarn upgrade-interactive --latest  ovh-ui-angular
```

Upgrading these dependencies is not like fixing an Issue so I think we can stick with a dedicated commit type like:
```txt
build(deps): upgrade ovh-ui-kit  to vX.X.X
build(deps): upgrade ovh-ui-angular to vX.X.X
```

If it's related to some devDependencies then:
```txt
build(dev-deps): upgrade ovh-ui-kit  to vX.X.X
```

Could you please then rewrite the git commit history?

Apart from that, all changes looks good to me.
Well done!

Thank you!

AxelPeter:
@antleblanc In fact, it was done with this PR https://github.com/ovh/manager/pull/2848 before I cherry-pick these changes into a new branch. I'll update this.

Careful about using `--latest` because this could result of having `ui-kit@v4` before the real migration :p

AxelPeter:
I push the new commit with the packages update and reorder it as first.
I've create a single commit because `ovh-ui-kit` and `ovh-ui-angular` need to be at this version together (shared changes).

antleblanc:
> Careful about using `--latest` because this could result of having `ui-kit@v4` before the real migration :p

You're right! But here I'm pretty sure it won't appears because package name are different.

AxelPeter:
Updated the deps with the new fix releases.
Got 2 approvals dismissed before that ;)

antleblanc:
@AxelPeter Can you please resolve existing conflicts in order to be able to merge it?
Thank you!

PR #2852, opened 2020-04-27
Refactor/sso jquery

FredericEspiau:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | `MANAGER-4997`

## Description

Reduce the use of `jQuery` in order to get rid of it when possible

### DIFF
--- a/packages/components/ng-ovh-sso-auth-modal-plugin/package.json
+++ b/packages/components/ng-ovh-sso-auth-modal-plugin/package.json
@@ -39,9 +39,6 @@
     "start:dev": "lerna exec --stream --scope='@ovh-ux/ng-ovh-sso-auth-modal-plugin' --include-dependencies -- yarn run dev",
     "start:watch": "lerna exec --stream --parallel --scope='@ovh-ux/ng-ovh-sso-auth-modal-plugin' --include-dependencies -- yarn run dev:watch"
   },
-  "dependencies": {
-    "lodash": "~4.17.11"
-  },
   "devDependencies": {
     "@ovh-ux/component-rollup-config": "^7.0.0"
   },
--- a/packages/components/ng-ovh-sso-auth-modal-plugin/package.json
+++ b/packages/components/ng-ovh-sso-auth-modal-plugin/package.json
@@ -46,7 +46,9 @@
     "@ovh-ux/ng-ovh-sso-auth": "^4.2.3",
     "angular": "^1.3.x",
     "angular-translate": "^2.18.1",
-    "angular-ui-bootstrap": "1.3.3",
-    "jquery": "^2.1.3"
+    "angular-ui-bootstrap": "1.3.3"
+  },
+  "optionalDependencies": {
+    "whatwg-fetch": "^3.0.0"
   }
 }
--- a/packages/components/ng-ovh-sso-auth-modal-plugin/src/controller.js
+++ b/packages/components/ng-ovh-sso-auth-modal-plugin/src/controller.js
@@ -1,5 +1,3 @@
-import $ from 'jquery';
-
 /**
  * @ngdoc controller
  * @name ng-ovh-sso-auth-modal-plugin.SsoAuthModalController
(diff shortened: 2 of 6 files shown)


packages/components/ng-ovh-sso-auth/package.json
```diff
@@ -43,6 +43,6 @@
   "peerDependencies": {
     "angular": "^1.5.0",
     "angular-cookies": "^1.5.0",
-    "jquery": "^2.1.3"
+    "whatwg-fetch": "^3.0.0"
```
marie-j, line 46: This is not an angular dependency, would it make sense to add it as `dependency` instead ?

FredericEspiau: oh yeah

packages/components/ng-ovh-sso-auth-modal-plugin/package.json
```diff
@@ -40,7 +40,8 @@
     "start:watch": "lerna exec --stream --parallel --scope='@ovh-ux/ng-ovh-sso-auth-modal-plugin' --include-dependencies -- yarn run dev:watch"
   },
   "dependencies": {
-    "lodash": "~4.17.11"
+    "lodash": "^4.17.15",
```
frenautvh, line 43: bonus points : lodash could easily be removed too it's only used here (for a "set"...) https://github.com/ovh/manager/blob/master/packages/components/ng-ovh-sso-auth-modal-plugin/src/index.js#L28
one could say that since we only import a single function it's fine but in reality in the case of lodash it adds a lots of weight to the dist

FredericEspiau: Good find !

packages/components/ng-ovh-sso-auth/src/provider.js
```diff
@@ -1,3 +1,5 @@
+import 'whatwg-fetch';
```
antleblanc, line 1: Why not consider using it as a PeerDependency instead? (same from ng-ovh-sso-auth-modal-plugin component).

Because this dependency is always imported from the host application.

```sh
$ git grep "whatwg-fetch" -- "*.js"
packages/manager/apps/dedicated/client/app/index.js:import 'whatwg-fetch';
packages/manager/apps/hub/src/index.js:import 'whatwg-fetch';
packages/manager/apps/public-cloud/src/index.js:import 'whatwg-fetch';
packages/manager/apps/telecom/src/app/index.js:import 'whatwg-fetch';
packages/manager/apps/web/client/app/index.js:import 'whatwg-fetch';
```

Missing `packages/manager/apps/cloud/..`

cbourgois: The `cloud` manager don't use the `bootstrapApplication` from https://github.com/ovh/manager/blob/master/packages/manager/modules/core/src/index.js#L261, so no `fetch` used.

`whatwg-fetch`is a `fetch` polyfill so I added it in applications (not required, it depends on the application)

packages/components/ng-ovh-sso-auth-modal-plugin/src/controller.js
```diff
@@ -1,4 +1,4 @@
-import $ from 'jquery';
+import 'whatwg-fetch';
```
cbourgois, line 1: Is it required to work ? Or is it required because we want that the application that use it support old browser versions that don't provide `fetch` ?

FredericEspiau:
@cbourgois sure I don't care

cbourgois:
I think we should add `whatwg-fetch` in `cloud` application : 

```bash
git grep "@ovh-ux/ng-ovh-sso-auth-modal-plugin" -- "*.js"
packages/manager/apps/cloud/client/app/app.js:import ngOvhSsoAuthModalPlugin from '@ovh-ux/ng-ovh-sso-auth-modal-plugin';
packages/manager/apps/dedicated/client/app/app.js:import ngOvhSsoAuthModalPlugin from '@ovh-ux/ng-ovh-sso-auth-modal-plugin';
packages/manager/apps/telecom/src/app/app.js:import ngOvhSsoAuthModalPlugin from '@ovh-ux/ng-ovh-sso-auth-modal-plugin';
packages/manager/apps/web/client/app/app.js:import ngOvhSsoAuthModalPlugin from '@ovh-ux/ng-ovh-sso-auth-modal-plugin';
```

```bash 
git grep "whatwg-fetch" -- "*package.json"
packages/manager/apps/dedicated/package.json:    "whatwg-fetch": "^3.0.0"
packages/manager/apps/hub/package.json:    "whatwg-fetch": "^3.0.0"
packages/manager/apps/public-cloud/package.json:    "whatwg-fetch": "^3.0.0"
packages/manager/apps/telecom/package.json:    "whatwg-fetch": "^3.0.0"
packages/manager/apps/web/package.json:    "whatwg-fetch": "^3.0.0"
```

FredericEspiau:
done @cbourgois

PR #2871, opened 2020-04-29
fix(exchange): export navigation service

cbourgois:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-5006

## Description

Expose `navigation` service when `@ovh-ux/manager-exchange` root AngularJS module is used, instead of declaring it when the first exchange route is lazy loaded.
This service is actually used by email pro for Emails footers configuration.

### DIFF
--- a/packages/manager/modules/exchange/src/index.js
+++ b/packages/manager/modules/exchange/src/index.js
@@ -7,6 +7,7 @@ import billingAccountRenew from './billing/account-renew/renew.module';
 import APIExchange from './exchange.api';
 import Exchange from './exchange.service';
 import ExchangePassword from './exchange.password.service';
+import navigation from './services/exchange.navigation.service'; // used by emailpro
 
 const moduleName = 'ovhManagerExchangeLazyLoading';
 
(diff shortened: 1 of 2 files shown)


packages/manager/modules/exchange/src/index.js
```diff
@@ -28,6 +29,7 @@ angular
   )
   .service('APIExchange', APIExchange)
   .service('Exchange', Exchange)
-  .service('ExchangePassword', ExchangePassword);
+  .service('ExchangePassword', ExchangePassword)
+  .service('navigation', navigation);
```
frenautvh, line 33: the service name is too generic to be exported like that no ?

cbourgois: We could rename but actually it's already exported as this. 
It was available before adding lazy loading to the exchange root state, and now this `navigation` service is only available after reaching the first exchange route.

cbourgois:
Removed `universe-server` because it is a bug in `exchange` that breaks `emailpro`

PR #2897, opened 2020-05-05
fix(plugin.translation-ui-router): annotate dependency injection

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | none

## Description

When state resolves are using spread to extend objects, the annotation is  not made which causes some unknown provider error

### DIFF
--- a/packages/manager/tools/component-rollup-config/src/plugins/translation-ui-router.ts
+++ b/packages/manager/tools/component-rollup-config/src/plugins/translation-ui-router.ts
@@ -20,14 +20,16 @@ const removeProperty = (code, magicString, start, end) => {
   }
 };
 
-export = (opts:any = {}) => {
+export = (opts: any = {}) => {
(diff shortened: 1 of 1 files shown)


packages/manager/tools/component-rollup-config/src/plugins/translation-ui-router.ts
```diff
@@ -92,10 +91,14 @@ export = (opts:any = {}) => {
                 format,
               );
 
-              inject = `translations: ($q, $translate, asyncLoader) => { ${inject} }`;
+              inject = `translations: /* @ngInject */ ($q, $translate, asyncLoader) => { ${inject} }`;
```
marie-j, line 94: Only human change made is here, the rest is due to linters being applied

PR #2900, opened 2020-05-05
fix(exchange): add lazy loading futures states

cbourgois:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | 

## Description

Before : when we load the exchange manager with a bookmarked url we are redirected to `#/configuration`
After: it loads as expected.

### DIFF
--- a/packages/manager/modules/exchange/src/index.js
+++ b/packages/manager/modules/exchange/src/index.js
@@ -15,15 +15,52 @@ angular
   .module(moduleName, [billingAccountRenew, 'ui.router', 'oc.lazyLoad'])
   .config(
     /* @ngInject */ ($stateProvider) => {
+      const lazyLoad = ($transition$) => {
+        const $ocLazyLoad = $transition$.injector().get('$ocLazyLoad');
+
+        return import('./exchange.module').then((mod) =>
+          $ocLazyLoad.inject(mod.default || mod),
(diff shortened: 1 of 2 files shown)


packages/manager/modules/exchange/src/index.js
```diff
@@ -15,16 +15,53 @@ angular
...
+        url: '/configuration/exchange_provider/:organization/:productId?tab',
+        lazyLoad,
+      });
+
+      $stateProvider.state('app.microsoft.exchange.order.**', {
+        url: '/configuration/exchange/order',
+        lazyLoad,
+      });
     },
   )
```
marie-j, line 66: Maybe we could also add this state https://github.com/ovh/manager/blob/master/packages/manager/modules/exchange/src/exchange.routes.js#L19 as it is used for redirections

cbourgois: Already lazy loaded (only this one was before this fix) https://github.com/ovh/manager/pull/2900/files#diff-aca744aac0d7981ce8543e63484f2537R26-R29

marie-j: My bad, I didn't see it ! LGTM then :)

PR #2947, opened 2020-05-14
fix(webpack): improve js loaders

jleveugle:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | None -->

## Description

<ins>Before:</ins>

If sources files need to be compiled with Babel plugins and don't have a native native compliance with JS, Webpack is not able to compile the source, even if the right plugins to compile are defined.

This behavior is due to multiple loader for JS files: Babel / translation-inject / translation-ui-router. In fact, translation-inject and translation-ui-router try to parse and compile the code, without any compilation by Babel. The source code is not valid with JS compiler, this 2 loaders are crashing.

The solution consist to use babel-loader before translation-inject and translation-ui-router. Babel does the job, and the 2 following loaders are able to do their own.

But, another issue then arises, babel remove comment like @ngTranslationsInject and prevent translations loader to do thir job. We should let this commit when Babel is used.

<ins>After:</ins>

Babel-loader is used before translation-inject and translation-ui-router and let the comment like @ngTranslationsInject.

Webpack is able to compile with custom plugins which replace custom syntax that JS engine doesn't understand ;)

### DIFF
--- a/packages/manager/tools/webpack-config/src/webpack.common.ts
+++ b/packages/manager/tools/webpack-config/src/webpack.common.ts
@@ -208,7 +208,7 @@ export = (opts) => {
                   require.resolve('babel-plugin-angularjs-annotate'), // ng annotate
                 ],
                 shouldPrintComment: (val) =>
-                  !/@ngInject|@ngTranslationsInject/.test(val),
+                  !/@ng(Translations)?Inject/.test(val),
               },
             },
           ],
--- a/packages/manager/tools/webpack-config/src/webpack.common.ts
+++ b/packages/manager/tools/webpack-config/src/webpack.common.ts
@@ -20,7 +20,7 @@ const cacheLoader = {
 
 // The common webpack configuration
 
-export = opts => {
+export = (opts) => {
   const lessLoaderOptions = {
     sourceMap: true,
     plugins: [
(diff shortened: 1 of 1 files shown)


packages/manager/tools/webpack-config/src/webpack.common.ts
```diff
@@ -180,51 +175,40 @@ export = opts => {
...
-                filtering: false,
+                presets: [
+                  require.resolve('@babel/preset-env'), // babel preset configuration
+                ],
+                plugins: [
+                  require.resolve('@babel/plugin-syntax-dynamic-import'), // dynamic es6 imports
+                  require.resolve('babel-plugin-angularjs-annotate'), // ng annotate
+                ],
+                shouldPrintComment: (val) =>
+                  !/@ngInject|@ngTranslationsInject/.test(val),
```
antleblanc, line 211:
```suggestion
                  !/@ng(Translations)?Inject/.test(val),
```

jleveugle:
@antleblanc It's not a fix for our users, this PR fix something wrong at the build time.

So for me, we should let "build" and not "fix" :)

jleveugle:
> Edit: Needed to publish at least a new patch on the npm registry.

Yes ! When can we schedule it?

antleblanc:
> > Edit: Needed to publish at least a new patch on the npm registry.
> 
> @antleblanc Yes ! When can we schedule it?

As soon as possible but npm is suffering a Major Outage currently. We will keep you posted.

See: https://status.npmjs.org/

PR #2970, opened 2020-05-18
Feat/geoloc ips

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-4902

## Description

- Remove IP selection if option is not available
- Create component to handle common configuration for webhosting domain

## Related 
- [x] TRANS-25635


marie-j:
All the feature would need a refactoring (still a wizard, handling everything with `$scope`...) but this is too huge to be handle yet and in this PR

packages/manager/apps/web/client/app/hosting/multisite/domain-configuration/configuration.html
```diff
@@ -0,0 +1,138 @@
...
+    label="{{:: 'hosting_multisite_domain_configuration_runtime' | translate }}"
+    data-size="l"
+>
+    <oui-select
+        data-items="$ctrl.runtimes"
+        data-model="$ctrl.runtime"
+        name="runtime"
+        match="name"
+    >
+        <span ng-bind="$item.name"></span>
```
antleblanc, line 32: 👀 On custom elements we tend to drop the `data` prefix but we not used to apply
this rule to the native HTMLElement. Is it intentional?

marie-j: Oopsie ! I think I mixed both rules, I've updated it

packages/manager/apps/web/client/app/hosting/multisite/domain-configuration/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,20 @@
+{
+  "hosting_multisite_domain_configuration_home": "Dossier racine",
+  "hosting_multisite_domain_configuration_myfolder": "mondossier",
```
FredericEspiau, line 3: Intentional ?

packages/manager/apps/web/client/app/hosting/multisite/domain-configuration/configuration.component.js
```diff
@@ -0,0 +1,23 @@
+import controller from './configuration.controller';
+import template from './configuration.html';
+
+export default {
+  bindings: {
+    cdn: '=',
```
FredericEspiau, line 6: Just so you know, I hate this, it means I have to go to the `*.component.js` file to know how to use the component, whereas is everything was in `'<'`, we would be so fine 😅

marie-j:
I forgot to move some translations keys so I just updated the new translations file and remove old key

packages/manager/apps/web/client/app/hosting/multisite/domain-configuration/translations/Messages_fr_FR.json
marie-j: I've just extracted the translations so component can have its own. No content update made

packages/manager/apps/web/client/app/hosting/multisite/domain-configuration/configuration.component.js
marie-j: It won't work well in the case of a form though 😬

marie-j:
Rebased on `feat/web-ca`

marie-j:
Rebased on feat/web-ca

PR #2971, opened 2020-05-18
fix(hosting): fetch email service info only if available

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-5114

## Description

Fetch email service info only if available as email domain are not available for every region

### DIFF
--- a/packages/manager/apps/web/client/app/hosting/hosting.routes.js
+++ b/packages/manager/apps/web/client/app/hosting/hosting.routes.js
@@ -34,23 +34,25 @@ export default /* @ngInject */ ($stateProvider) => {
         $q,
         emailOptionIds,
         hostingEmailService,
+        isEmailDomainAvailable,
         OvhApiEmailDomain,
         serviceName,
       ) =>
-        $q
-          .all(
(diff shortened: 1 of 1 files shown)


packages/manager/apps/web/client/app/hosting/hosting.routes.js
```diff
@@ -34,23 +34,25 @@ export default /* @ngInject */ ($stateProvider) => {
...
-                .then(
-                  ({ resource }) =>
-                    OvhApiEmailDomain.v6().serviceInfos({
-                      serviceName: resource.name,
-                    }).$promise,
-                ),
-            ),
-          )
-          .then((servicesInformation) => servicesInformation.flatten()),
+        (isEmailDomainAvailable
```
antleblanc, line 41: Duplicate of #2960?

marie-j: Not entirely, as here we prevent making the call if we know it will throw an error

PR #2981, opened 2020-05-20
feat: prevent calling /email/pro APIs in CA

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-5012

## Description

Prevent calling /email/pro APIs in CA. 
This is due to mxplan  and emailpro sharing the same module and components

### DIFF
--- a/packages/manager/modules/emailpro/src/emailpro.module.js
+++ b/packages/manager/modules/emailpro/src/emailpro.module.js
@@ -31,6 +31,14 @@ angular
   .constant('EMAILPRO_CONFIG_URL', EMAILPRO_CONFIG_URL)
   .constant('EMAILPRO_CONFIG', EMAILPRO_CONFIG)
   .config(routing)
+  .provider(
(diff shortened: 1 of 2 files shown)


packages/manager/modules/emailpro/src/emailpro.module.js
```diff
@@ -31,6 +31,14 @@ angular
   .constant('EMAILPRO_CONFIG_URL', EMAILPRO_CONFIG_URL)
   .constant('EMAILPRO_CONFIG', EMAILPRO_CONFIG)
   .config(routing)
+  .provider(
```
FredericEspiau, line 34: Why a `provider` instead of a `constant` ?

marie-j: Because you can't declare constant depending on dynamic values

PR #3028, opened 2020-05-28
Improve build script

cbourgois:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix 

## Description

* Add `--num-workers` option to define number of worker to use when launch the worskapces packages build :
_Example:_ `yarn run build --num-workers 2`
* Stop the execution of the build script if a package build script fails instead of logging the error and continue to build others packages.
Examples:_ 
  * if `@ovh-ux/manager-server-sidebar` failed to build, applications using this package (`@ovh-ux/manager-cloud`, `@ovh-ux/manager-dedicated`, `@ovh-ux/manager-web`) will failed too
  * if an app package (like `@ovh-ux/manager-web`) failed to build, we can consider that the `build` script failed (its objective is to build all the workspaces packages)

### DIFF
--- a/scripts/build.js
+++ b/scripts/build.js
@@ -51,7 +51,10 @@ function unstack() {
       (dep) => !modules.done.find((el) => el.name === dep.name),
     );
 
-    if (deps.length === 0) {
+    if (
+      deps.length === 0 &&
+      (!program.numWorkers || modules.doing.length < program.numWorkers)
+    ) {
       const workerData = remove(modules.todo, (p) => p.name === pck.name)[0];
       modules.doing.push(workerData);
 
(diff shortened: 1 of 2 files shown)


scripts/build.js
```diff
@@ -10,6 +10,7 @@ EventEmitter.defaultMaxListeners = 100; // Increase the default limit to avoid m
 
 const myEmitter = new EventEmitter();
 
+let limitNumWorkers = false;
```
frenautvh, line 13: i don't think you need this variable, just check if program.numWorkers > 0 to see if you need to apply a limit

cbourgois:
Other parts of the scripts could be updated :
* optimize `lerna` calls (https://github.com/ovh/manager/blob/master/scripts/build.js#L17-L23)
* logs of built packages (`--verbose` ?) when `unstack`(https://github.com/ovh/manager/blob/master/scripts/build.js#L36-L48)

scripts/build.js
cbourgois: Fixed : https://github.com/ovh/manager/blob/4e79984e20ebc2f3cf687cdc8dfee89da6348d8a/scripts/build.js#L56

PR #3033, opened 2020-05-29
feat(public-cloud): default payment method choice

Jisay:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-4345

## Description

The customers that have a validated payment mean but not chosen as the "by default" one see their first valid payment mean taken automatically, and considered as "by default", without any notification. This feature enable users to select an existing payment method or by creating a new one and set it as default.

## 🔖 TODOs

- [x] TRANS-26177


packages/manager/modules/pci/src/projects/new/classes/eligibility.class.js
```diff
@@ -33,4 +35,12 @@ export default class PciEligibility {
   isVerifyPaypalRequired() {
     return this.actionsRequired.includes(ELIGIBILITY_ACTION_ENUM.VERIFY_PAYPAL);
   }
+
+  setValidPaymentMethods(validPaymentMethods) {
+    this.validPaymentMethods = validPaymentMethods;
+  }
+
+  isDefaultPaymentMethodChoiceRequired() {
```
Jisay, line 43: I set it there in order to have all requirements at the same place. It simply extends a little bit the eligibility API response in order to add a new test.

packages/manager/modules/pci/src/projects/new/payment/components/add/component.js
```diff
@@ -0,0 +1,14 @@
+import controller from './controller';
```
Jisay, line 1: Use a component as payment method registration can be done in two situations : if no payment method at all or no default payment method is setted as default.

packages/manager/modules/pci/src/projects/new/payment/components/add/index.html
```diff
@@ -0,0 +1,48 @@
...
+            data-type="warning"
+            data-ng-if="$ctrl.model.paymentMethod.paymentType === 'BANK_ACCOUNT'"
+        >
+            <span
+                data-translate="pci_project_new_payment_method_add_warning_bank_account"
+            ></span>
+        </oui-message>
+
+        <p>
+            <small
```
FredericEspiau, line 40: `small` has a semantic meaning: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/small

is it he case here ?

packages/manager/modules/pci/src/projects/new/payment/components/add/translations/messages_FR_fr.json
```diff
@@ -0,0 +1,8 @@
+{
+  "pci_project_new_payment_method_add_explain": "L'enregistrement d'un moyen de paiement est nécessaire afin de valider votre identité. Tant que vous ne consommez rien vous ne serez pas débité ou prélevé.",
+  "pci_project_new_payment_method_add_info_credit_card": "En cliquant sur « Créer mon projet » vous allez être redirigé sur l’interface sécurisée de notre prestataire pour enregistrer une nouvelle carte bancaire. Une empreinte financière est faite auprès de votre organisme bancaire afin de valider la saisie et la validité de votre carte. Aucun montant ne sera prélevé.",
```
FredericEspiau, line 3:
```suggestion
  "pci_project_new_payment_method_add_info_credit_card": "En cliquant sur « Créer mon projet » vous allez être redirigé vers l’interface sécurisée de notre prestataire pour enregistrer une nouvelle carte bancaire. Une empreinte financière est faite auprès de votre organisme bancaire afin de valider la saisie et la validité de votre carte. Aucun montant ne sera prélevé.",
```
?

packages/manager/modules/pci/src/projects/new/payment/components/choose/index.scss
```diff
@@ -0,0 +1,76 @@
+.pci-project-new-payment-choose {
```
Jisay, line 1: Need to override some styles from ui-kit/ods in order to fit as more as possible the mockups

packages/manager/modules/pci/src/projects/new/payment/components/register/index.html
```diff
@@ -4,50 +4,11 @@
         data-translate="pci_project_new_payment_register_title"
     ></h2>
 
-    <div
-        class="row"
-        data-ng-repeat="authorizedPaymentMethods in $ctrl.authorizedPaymentMethods.chunks track by $index"
+    <pci-project-new-payment-method-add
```
Jisay, line 7: Centralize payment method registration in an other component

packages/manager/modules/pci/src/projects/new/payment/components/add/translations/messages_FR_fr.json
Jisay: This content is not yet defined. I have asked the right one ;-)

packages/manager/modules/pci/src/projects/new/payment/components/add/index.html
```diff
@@ -0,0 +1,48 @@
+<div class="pci-project-new-payment-method-add">
+    <div
+        class="row"
+        data-ng-repeat="authorizedPaymentMethods in $ctrl.authorizedPaymentMethods.chunks track by $index"
```
marie-j, line 4: The `col-md-6` isn't enough to have two elements per `row` ?

packages/manager/modules/pci/src/projects/new/payment/components/add/index.html
```diff
@@ -0,0 +1,48 @@
...
+                data-disabled="$ctrl.globalLoading.finalize"
+            >
+                <span
+                    data-ng-bind="$ctrl.getPaymentMethodTypeText(authorizedPaymentMethod)"
+                ></span>
+            </oui-radio>
+        </div>
+    </div>
+
+    <div data-ng-if="$ctrl.model.paymentMethod.paymentType !== 'CREDIT'">
```
marie-j, line 27: Maybe some utilities function like `isCredit()`, `isBankAccount()` in the `PaymentMethod` model could be useful to avoid using constants in the view

packages/manager/modules/pci/src/projects/new/payment/components/add/index.js
```diff
@@ -0,0 +1,14 @@
+import angular from 'angular';
+import '@ovh-ux/ng-translate-async-loader';
+import 'angular-translate';
+
+import component from './component';
+
+const moduleName = 'pciProjectNewPaymentMethodAdd';
+
+angular
+  .module(moduleName, ['ngTranslateAsyncLoader', 'pascalprecht.translate'])
```
marie-j, line 10:
```suggestion
  .module(moduleName, ['oui', 'ngTranslateAsyncLoader', 'pascalprecht.translate'])
```

packages/manager/modules/pci/src/projects/new/payment/components/choose/index.scss
```diff
@@ -0,0 +1,76 @@
...
+        .oui-radio__label {
+          margin-bottom: 0;
+
+          span:nth-child(1) {
+            display: flex;
+          }
+
+          .payment-method-image-container {
+            width: 44px;
+            max-width: 44px;
```
varun257, line 56: Can use `rem` instead

packages/manager/modules/pci/src/projects/new/payment/controller.js
```diff
@@ -153,9 +154,26 @@ export default class PciProjectNewPaymentCtrl {
...
+      this.globalLoading.setDefaultPaymentMehtod = true;
+
+      defaultPaymentMethodPromise = this.ovhPaymentMethod
+        .setPaymentMethodAsDefault(this.model.defaultPaymentMethod)
+        .finally(() => {
+          this.globalLoading.setDefaultPaymentMehtod = false;
+        });
+    }
+
+    return Promise.all([challengePromise, defaultPaymentMethodPromise]).then(
```
varun257, line 170: Is there is no error handling for this ?

packages/manager/modules/pci/src/projects/new/payment/components/choose/index.scss
marie-j: I can understand that some overriding is needed now and then, but shouldn't the latest mockup design then be integrated in ui-kit/ods and therefore the UI changes automatically applied ?

packages/manager/modules/pci/src/projects/new/payment/controller.js
marie-j, line 170: No issue with `$scope` digest ?

packages/manager/modules/pci/src/projects/new/payment/routing.js
```diff
@@ -38,8 +50,24 @@ export default /* @ngInject */ ($stateProvider) => {
       paymentStatus: /* @ngInject */ ($transition$) =>
         $transition$.params().paymentStatus,
 
-      defaultPaymentMethod: /* @ngInject */ (ovhPaymentMethod) =>
-        ovhPaymentMethod.getDefaultPaymentMethod(),
+      paymentMethods: /* @ngInject */ (ovhPaymentMethod) =>
+        ovhPaymentMethod.getAllPaymentMethods(),
```
marie-j, line 54: I don't know how much impact there is to get all methods but maybe we should only do so if we know it's needed (like when there is no default payment method for instance)

packages/manager/modules/pci/src/projects/new/routing.js
```diff
@@ -152,6 +153,7 @@ export default /* @ngInject */ ($stateProvider) => {
       },
 
       globalLoading: () => ({
+        setDefaultPaymentMehtod: false,
```
marie-j, line 156:
```suggestion
        setDefaultPaymentMethod: false,
```
😄

packages/manager/modules/pci/src/projects/new/payment/index.html
```diff
@@ -80,11 +80,13 @@ <h1 data-translate="pci_project_new_payment_title"></h1>
...
-                        ($ctrl.eligibility.isChallengePaymentMethodRequired() && !$ctrl.model.challenge.isValid($ctrl.defaultPaymentMethod.paymentType)
+                        $ctrl.eligibility.isChallengePaymentMethodRequired() && !$ctrl.model.challenge.isValid($ctrl.defaultPaymentMethod.paymentType)
                         || (!$ctrl.model.paymentMethod && $ctrl.eligibility.isAddPaymentMethodRequired())
+                        || (!$ctrl.model.defaultPaymentMethod && $ctrl.eligibility.isDefaultPaymentMethodChoiceRequired())
                         || $ctrl.model.challenge.checking
-                        || $ctrl.globalLoading.finalize)"
-                    data-ng-if="$ctrl.defaultPaymentMethod || !$ctrl.model.paymentMethod"
+                        || $ctrl.globalLoading.finalize
+                        || $ctrl.globalLoading.setDefaultPaymentMehtod"
+                    data-ng-if="$ctrl.defaultPaymentMethod || !$ctrl.model.paymentMethod || ($ctrl.eligibility.isDefaultPaymentMethodChoiceRequired() && !$ctrl.model.paymentMethod)"
```
cbourgois, line 89: (hard to read)

packages/manager/modules/pci/src/projects/new/payment/controller.js
Jisay: Not here, but when the promises are created. 
Note that I have to add it for set as default payment method promise cause I didn't add it ^^

packages/manager/modules/pci/src/projects/new/payment/components/add/index.html
Jisay: I'm not sure to understand: you mean that there is no need to have this `ng-repeat` to create rows for the cols?
I think that a `div.row` parent is required for `div.col-*`

packages/manager/modules/pci/src/projects/new/payment/controller.js
Jisay: Not here. Because `manageProjectCreation` is called in `then` callback and this method is handled with `$q`

packages/manager/modules/pci/src/projects/new/payment/routing.js
Jisay: To know if there is a default payment method, we have to get all payment methods. So here I don't use `getDefaultPaymentMethod` directly because I need to list payment methods. So I get all payment methods to first know if there is or not a default payment method and if not display a list of valid payment methods. But in all the cases we have to get all payment methods to have their details...

packages/manager/modules/pci/src/projects/new/payment/components/add/index.html
Jisay: This is originally a simple move of the code in different files but yes you are right this is more readable and easier

packages/manager/modules/pci/src/projects/new/payment/index.html
Jisay: Yes, hard to not forget a case too :-D

packages/manager/modules/pci/src/projects/new/payment/components/choose/index.scss
Jisay: That should be the rule indeed and I prefer this approach too

packages/components/ng-ovh-payment-method/src/payment-method-type.class.js
```diff
@@ -96,4 +99,42 @@ export default class OvhPaymentMethodType {
...
     ].includes(this.integration);
   }
+
+  /**
+   * Determine if payment method type is a Bank Account.
+   *
+   * @return {Boolean}
+   */
+  isBankAccount() {
+    return this.paymentType === PAYMENT_METHOD_TYPE_ENUM.BANK_ACCOUNT;
```
marie-j, line 109: Why do this needs to be duplicate ? Isn't there a parent class ?

packages/manager/modules/pci/src/projects/new/payment/components/add/component.js
```diff
@@ -0,0 +1,14 @@
...
+
+export default {
+  name: 'pciProjectNewPaymentMethodAdd',
+  controller,
+  template,
+  bindings: {
+    eligibility: '<',
+    model: '<',
+    registerablePaymentMethods: '<',
+    globalLoading: '<',
```
ganeshkumar1989, line 12: Would be great if the dependencies are sorted :)

packages/components/ng-ovh-payment-method/src/payment-method-type.class.js
Jisay: Mmmmh, can be done but we talk about two different concept here:
- payment methods which have payment method type ;
- payment method types which are used to add a new payment method of given type.

So I don't think it's a good solution to have a mix of concepts. 
Maybe a static method to centralize it or maybe define `paymentType` attribute of `OvhPaymentMethod` instance as an instance of `OvhPaymentMethodType`.

packages/manager/modules/pci/src/projects/new/payment/components/add/index.html
marie-j: What I meant is that it might not be needed to split the initial array into chunks of 2 arrays each

packages/components/ng-ovh-payment-method/src/payment-method-type.class.js
marie-j: Yes Maybe `OvhPaymentMethod.type` should be an instance of OvhPaymentMethodType

packages/components/ng-ovh-payment-method/src/available-payment-method.class.js
```diff
@@ -0,0 +1,108 @@
+import { isNull } from 'lodash-es';
```
Jisay, line 1: This file is just a move from previous `payment-method-type.class.js` file

packages/components/ng-ovh-contacts/LICENSE
```diff
@@ -1,29 +0,0 @@
-BSD 3-Clause License
```
marie-j, line 1: This change doesn't look related to the PR's purpose

packages/components/ng-ovh-payment-method/src/legacy/mean/payment-mean-type.class.js
```diff
@@ -1,15 +1,15 @@
...
 
 export default class OvhPaymentMeanType {
   constructor(options = {}) {
     this.meanType = options.value;
     this.registerable = options.registerable;
   }
 
-  toPaymentMethodType() {
-    return new OvhPaymentMethodType({
+  toAvailablePaymentMethod() {
```
marie-j, line 11: I didn't see any `BREAKING CHANGE` comment but this change looks like one as package is published

Jisay: set `toPaymentMethodType` to deprecated to avoid breaking change.

antleblanc:
Pull Request has been rebase on `develop` in order to be able to resolve existing conflicts!
Also the git commit history has been rewritten to remove some extra commits and the translation has been retrieved.

I'm waiting the status of all check in order to be able to merge it on `develop`.

packages/components/ng-ovh-contacts/LICENSE
antleblanc: Thanks for pointing out this. It has been resolved via the rebase.

PR #2303, opened 2020-02-21
refactor(ng-ovh-payment-method): remove ovh-api-services

frenautvh:
Remove ovh-api-services peer dep from ng-ovh-payment-method

Internal MANAGER-5001

- [ ] Validates changes by QA

PR #3039, opened 2020-06-01
perf(ng-ovh-payment-method): use lodash-es instead of lodash 

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | 

## Description

Use lodash-es instead of lodash in order to reduce bundle size (few Ko gains)

### DIFF
--- a/packages/components/ng-ovh-payment-method/src/legacy/payment-method-legacy.js
+++ b/packages/components/ng-ovh-payment-method/src/legacy/payment-method-legacy.js
@@ -1,9 +1,4 @@
-import filter from 'lodash/filter';
-import flatten from 'lodash/flatten';
-import get from 'lodash/get';
-import has from 'lodash/has';
-import map from 'lodash/map';
-import startCase from 'lodash/startCase';
+import { filter, flatten, get, has, map, startCase } from 'lodash-es';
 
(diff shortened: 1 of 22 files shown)


packages/components/ng-ovh-payment-method/src/legacy/payment-method-legacy.js
```diff
@@ -1,9 +1,9 @@
-import filter from 'lodash/filter';
-import flatten from 'lodash/flatten';
-import get from 'lodash/get';
-import has from 'lodash/has';
-import map from 'lodash/map';
-import startCase from 'lodash/startCase';
+import { filter } from 'lodash-es';
```
frenautvh, line 1: yes it could be merged in a single import but i used a script to batch rename imports and it's not handled for now :)

Jisay:
Should be great to merge with #2303

antleblanc:
@frenautvh Regarding the changes (and the label chosen) do you think it could deserve a commit type [`perf`](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#type) instead?.

PR #3043, opened 2020-06-01
feat(ng-ovh-responsive-tabs): replace lodash with lodash-es

frenautvh:

| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          |

## Description

Replace lodash with lodash-es to reduce bundle size

### DIFF
--- a/packages/components/ng-ovh-responsive-tabs/package.json
+++ b/packages/components/ng-ovh-responsive-tabs/package.json
@@ -41,7 +41,7 @@
   "dependencies": {
     "bootstrap": "~3.3.7",
     "jquery": "^2.1.3",
-    "lodash": "^4.17.15",
+    "lodash-es": "^4.17.15",
     "ovh-ui-kit": "^2.42.7",
     "ovh-ui-kit-bs": "^2.4.6"
   },
--- a/packages/components/ng-ovh-responsive-tabs/src/controller.js
+++ b/packages/components/ng-ovh-responsive-tabs/src/controller.js
@@ -1,5 +1,4 @@
-import map from 'lodash/map';
-import set from 'lodash/set';
+import { map, set } from 'lodash-es';
 
 export default class {
   /* @ngInject */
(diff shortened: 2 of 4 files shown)


antleblanc:
@frenautvh Regarding the changes (and the label chosen) do you think it could deserve a commit type [`perf`](https://github.com/angular/angular/blob/master/CONTRIBUTING.md#type) instead?.

FredericEspiau:
@antleblanc I don't think so because it's not only a change on the `perf` but also on the `build` system, so I would have gone for `build` myself

frenautvh:
> @antleblanc I don't think so because it's not only a change on the `perf` but also on the `build` system, so I would have gone for `build` myself

@antleblanc @FredericEspiau perf / build it's tricky both seems to work. I would just build just because i don't perform any performance tweaks on the code and just replace a dependency. Is that ok for both of you?

antleblanc:
> @antleblanc I don't think so because it's not only a change on the `perf` but also on the `build` system, so I would have gone for `build` myself

Since we are decreasing the bundle size, it feels more natural to be marked as perf.
It is the main topic that this Pull Request is addressing.

antleblanc:
> > @antleblanc I don't think so because it's not only a change on the `perf` but also on the `build` system, so I would have gone for `build` myself
> 
> @antleblanc @FredericEspiau perf / build it's tricky both seems to work. I would chooose build just because i don't perform any performance tweaks on the code and just replace a dependency. Is that ok for both of you?

The `build` commit type won't trigger a new version on the affected package.

I encourage you to use a commit type `perf` to be able to publish a new version on the npm registry.

antleblanc:
@frenautvh Could you please resolve existing conflict in order to be able to merge it? Thank you!

FredericEspiau:
@antleblanc 😅 indeed `perf` is better then

PR #3104, opened 2020-06-10
perf(hub): remove unecessary debt api call

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | 

## Description

Avoid making an unnecessary call to /hub/debt by reusing the /hub call data


frenautvh:
> Hello @frenautvh, ok with the changes, but I see that title of the PR is "remove unnecessary debt call", but you've just added a binding of debt. Is there another place where the call to retrieve the debt is made ?

If the component does not receive a debt data it will perform a calls to /hub/debt to fetch it. Since we are now passing data to it, that call is avoided.

PR #3107, opened 2020-06-10
fix: fix actions menu colors

AxelPeter:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-5249

- Apply ODS on the actions-menu
- Move style from sidebar to actions-menu (concern the actions-menu link)
- Update old icons


AxelPeter:
@marie-j For now, the goal is to fix this issue. I agree on the fact we could use some UI Kit part here. But there's some custom parts that are not covered and need to be adapted :)

@antleblanc I don't agree.

My point is that a commit should work at his state (even with cross-components modifications). All is needed to work together (except maybe the svg). But seriously, I won't split a commit of this size for the sake of the git log. It's not like there is multiple big fixes and features in this one. The body message is there for that too. If you prefer, I'll add the description in it.

antleblanc:
> @antleblanc I don't agree.
> 
> My point is that a commit should work at his state (even with cross-components modifications). All is needed to work together (except maybe the svg). But seriously, I won't split a commit of this size for the sake of the git log. It's not like there is multiple big fixes and features in this one. The body message is there for that too. If you prefer, I'll add the description in it.

We shouldn't take in account the _size_ of the diff to determine the way that
commits can be written but I get your point and it is probably a personnal
perference.

Because I'm thinking of a _better_ commit message that could be used espacially
from the Telecom Changelog application point of view.

Apart from that, it looks good to me. Could you please rebase on `master` and
add all the detailed information into the commit body?

Thank you!

AxelPeter:
I rebased this branch to `master`
Got 3 approvals before the force-push

PR #3116, opened 2020-06-15
fix(dedicatedcloud): change the way to fetch the catalog

antleblanc:
| Question         | Answer
| ---------------- | ---
| Branch?          | `release/2020-w25`
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | no

### 🐛 Bug Fixes

d884fe6 - fix(dedicatedcloud): change the way to fetch the catalog

### 🏠 Internal

No quality check required.

### DIFF
--- a/packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.service.js
+++ b/packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.service.js
@@ -119,7 +119,8 @@ export default class BackupService {
 
   getCatalog(ovhSubsidiary) {
     return this.$http
-      .get(`2api/sws/dedicatedcloud/catalog`, {
+      .get('/sws/dedicatedcloud/catalog', {
+        serviceType: 'aapi',
         params: {
           ovhSubsidiary,
         },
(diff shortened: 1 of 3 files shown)


packages/manager/apps/dedicated/client/app/dedicatedCloud/datacenter/backup/backup.service.js
```diff
@@ -117,8 +118,13 @@ export default class BackupService {
   }
 
   getCatalog(ovhSubsidiary) {
-    return this.catalogApi.get({ catalogName: 'privateCloud', ovhSubsidiary })
-      .$promise;
+    return this.$http
+      .get(`2api/sws/dedicatedcloud/catalog`, {
```
frenautvh, line 122:
```js
get('/sms/dedicatedcloud/catalog', {
  serviceType: 'aapi',
  params: { ... },
})

antleblanc: @frenautvh I can confirm that the URL is the correct one.

antleblanc: You mean passing the `serviceType` parameter?

frenautvh: yes i though we had to set the serviceType but in core, an interceptor is setting it automatically if your url starts with 2api so i guess it's fine :)
see : https://github.com/ovh/manager/blob/master/packages/manager/modules/core/src/index.js#L176

antleblanc: Thanks for pointing out this.

I've added the `serviceType` parameter and remove extra template strings.

PR #3144, opened 2020-06-22
fix(private.database): update graph label information

wiseflat:
Signed-off-by: Mathieu GARCIA <mathieu.garcia@corp.ovh.com>


kedare:
Would not it be more interesting to update the translations keys too as they don't match what is really is (also connexion is French not not English :P) ?

antleblanc:
@kedare Thanks for pointing out this!

Please let me know if you'd like to open a Pull Request to handle this change?

Thank you.

kedare:
But I think 'mysql_connexion_count' is the name of the metrics on the monitoring system ? So that would maybe more interesting to have it updated there too to avoid inconsistencies between backend and frontend

antleblanc:
> But I think 'mysql_connexion_count' is the name of the metrics on the monitoring system ? So that would maybe more interesting to have it updated there too to avoid inconsistencies between backend and frontend

@kedare As you wish. We only need to make sure that the translation key and value are consistent. Hope it helps you!

PR #3154, opened 2020-06-25
feat(telephony): split endpoint into ip list and list with incoming call

JeremyDec:

| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-5286

## Description

Split endpoints details into two lists:

- One of all endpoint ips
- One of endpoint with incoming call routing enabled

### DIFF
--- a/packages/manager/apps/carrier-sip/src/endpoints/routing.js
+++ b/packages/manager/apps/carrier-sip/src/endpoints/routing.js
@@ -3,6 +3,10 @@ export default /* @ngInject */ ($stateProvider) => {
     url: '/endpoints',
     component: 'carrierSipEndpoints',
     resolve: {
+      endpointIpList: /* @ngInject */ (endpoints) =>
+        endpoints.map(({ ip }) => ip),
+      endpointsWithIncomingCallsAllowed: /* @ngInject */ (endpoints) =>
+        endpoints.filter(({ enableIncomingCalls }) => enableIncomingCalls),
       endpoints: /* @ngInject */ (
         billingAccount,
         CarrierSipService,
--- a/packages/manager/modules/carrier-sip/src/endpoints/endpoints.html
+++ b/packages/manager/modules/carrier-sip/src/endpoints/endpoints.html
@@ -1,37 +1,44 @@
 <section>
-    <h3 data-translate="carrier_sip_endpoints"></h3>
-    <oui-datagrid data-rows="$ctrl.endpoints">
+    <h3 data-translate="carrier_sip_endpoints_list_IP"></h3>
+    <oui-datagrid class="small-width" rows="$ctrl.endpointIpList">
+        <oui-column title=":: 'carrier_sip_endpoints_ip_sip' | translate">
+            <span data-ng-bind="::$row"></span>
+        </oui-column>
+    </oui-datagrid>
+
+    <h3 data-translate="carrier_sip_endpoints_with_incoming_call_enabled"></h3>
+    <oui-datagrid rows="$ctrl.endpointsWithIncomingCallsAllowed">
         <oui-column
-            data-title=":: 'carrier_sip_endpoints_id' | translate"
-            data-property="id"
+            title=":: 'carrier_sip_endpoints_id' | translate"
+            property="id"
         >
         </oui-column>
         <oui-column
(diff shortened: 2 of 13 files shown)


packages/manager/modules/carrier-sip/src/endpoints/endpoints.html
```diff
@@ -1,37 +1,44 @@
 <section>
-    <h3 data-translate="carrier_sip_endpoints"></h3>
-    <oui-datagrid data-rows="$ctrl.endpoints">
+    <h3 data-translate="carrier_sip_endpoints_list_IP"></h3>
+    <oui-datagrid class="w-25" rows="$ctrl.endpointIpList">
```
antleblanc, line 3: Please let me know if it could suit your needs because using the `w-25` helper from small devices won't work efficiently.

```suggestion
    <oui-datagrid rows="$ctrl.endpointIpList">
```

packages/manager/apps/carrier-sip/src/endpoints/routing.js
```diff
@@ -3,6 +3,10 @@ export default /* @ngInject */ ($stateProvider) => {
     url: '/endpoints',
     component: 'carrierSipEndpoints',
     resolve: {
+      endpointIpList: /* @ngInject */ (endpoints) =>
```
cbourgois, line 6: I'm a bit confused to see that we duplicate resolve code, but seems to be legit as the way we develop `carrier-sip` module (don't embed the routing).

JeremyDec: Yes good point, I'll check if I can set the routing directly in the module. (In another PR)

packages/manager/modules/carrier-sip/src/endpoints/endpoints.html
JeremyDec: Hello @antleblanc , I'm curious about what won't work efficiently about this class helepr ? (I've check on Bootstrap https://getbootstrap.com/docs/4.0/utilities/sizing/ and it seems to work correctly)

antleblanc: `w-25` in portrait mode can lead to a non-stretched datagrid.

JeremyDec: Component class has been added to add min-width property to this element

packages/manager/modules/carrier-sip/src/endpoints/translations/Messages_fr_FR.json
```diff
@@ -1,5 +1,6 @@
 {
-  "carrier_sip_endpoints": "SIP Endpoints",
+  "carrier_sip_endpoints_list_IP": "Liste d'IP autorisées pour les appels sortants",
```
cbourgois, line 2: Sometimes we write `IPs` sometimes not...
Cf: https://github.com/ovh/manager/blob/7d34718274f5eae1335bd0b8f91af8745cb094d8/packages/manager/modules/carrier-sip/src/dashboard/translations/Messages_fr_FR.json#L11

JeremyDec: Changed in linked file :)

antleblanc: **nitpick:** It is probably a personal preference but I find `IP address(es)` way more explicit.

packages/manager/apps/carrier-sip/src/endpoints/routing.js
marie-j: Maybe the module could expose the service to avoid duplicating code

JeremyDec:
Since previous approvals, just non FR translations files have been pushed and update with FR translations. (No translation demand will be made)

packages/manager/apps/carrier-sip/src/endpoints/routing.js
JeremyDec: @cbourgois / @marie-j a task has been created to address your point. (MANAGER-5427)

PR #3233, opened 2020-07-09
Feat/read on close

AxelPeter:
| Question         | Answer
| ---------------- | ---
| Branch?          | `develop` (see notes below)
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-5363

## Description

When we close the notifications menu, it will read automatically all unread notifications.

### DIFF
--- a/packages/manager/modules/notifications-sidebar/src/controller.js
+++ b/packages/manager/modules/notifications-sidebar/src/controller.js
@@ -42,7 +42,7 @@ export default class NotificationsCtrl {
       this.toggle = false;
 
       // Automatically set all unread messages to read
-      // When we close the notifications menu
+      // when we close the notifications menu
       this.NavbarNotifications.readAllNotifications(
         this.getActiveNotifications(),
       ).then(() => {
@@ -59,7 +59,7 @@ export default class NotificationsCtrl {
     this.$rootScope.$on('ovh::notifications::toggle', () => {
       this.toggle = !this.toggle;
       if (this.toggle) {
-        // Handle the click outside the menu
+        // Handle the click outside the notifications menu
         this.$document.on('click', this.readAllNotifications);
 
         this.atInternet.trackClick({
@@ -112,11 +112,11 @@ export default class NotificationsCtrl {
   }
 
   $postLink() {
-    // Avoid click propagation inside the menu
+    // Avoid click propagation inside the notifications menu
     // Since we're binding a click function of the document
     this.$timeout(() => {
-      this.$element.on('click', (e) => {
-        e.stopPropagation();
+      this.$element.on('click', (event) => {
+        event.stopPropagation();
       });
     });
   }
@@ -126,7 +126,7 @@ export default class NotificationsCtrl {
   }
 
   getActiveNotifications() {
-    return filter(this.sublinks, (notification) => notification.isActive);
+    return filter(this.sublinks, 'isActive');
   }
 
   toggleSublinkAction(toUpdate, linkClicked) {
--- a/packages/manager/modules/notifications-sidebar/src/controller.js
+++ b/packages/manager/modules/notifications-sidebar/src/controller.js
@@ -7,7 +7,10 @@ import { MAX_NOTIFICATIONS } from './constants';
 export default class NotificationsCtrl {
   /* @ngInject */
   constructor(
+    $document,
+    $element,
     $q,
+    $timeout,
     $rootScope,
     $translate,
     atInternet,
@@ -15,7 +18,10 @@ export default class NotificationsCtrl {
     ovhManagerNavbarMenuHeaderBuilder,
     ouiNavbarConfiguration,
   ) {
+    this.$document = $document;
+    this.$element = $element;
     this.$q = $q;
+    this.$timeout = $timeout;
     this.$rootScope = $rootScope;
     this.$translate = $translate;
     this.atInternet = atInternet;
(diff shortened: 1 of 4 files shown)


ovh-cds:
CDS Report test-manager_2#13463.0 ✘  
* Stage 1
  * e2e -   
  * lint ✔ 
  * SemVer ✘

packages/manager/modules/freefax/package.json
```diff
@@ -26,7 +26,7 @@
     "@ovh-ux/ng-ovh-telecom-universe-components": "^5.0.0",
     "@ovh-ux/ng-ovh-ui-confirm-modal": "^2.0.0",
     "@ovh-ux/ng-ui-router-title": "^3.0.0",
-    "@ovh-ux/ui-kit": "^4.1.9",
+    "@ovh-ux/ui-kit": "^4.1.10",
```
marie-j, line 29: Technically this peer doesn't need an update because the module doesn't require change on the navbar. Also a patch ? Wasn't a feature added ?

packages/manager/modules/navbar/src/notifications-menu/controller.js
```diff
@@ -87,10 +124,18 @@ export default class NotificationsCtrl {
...
+      this.$document.on('click', this.readAllNotifications);
+
+      this.atInternet.trackClick({
+        name: 'navbar::action::notifications',
+        type: 'action',
+      });
+    } else {
+      // We unbind the click event in this function
+      this.readAllNotifications();
+    }
```
marie-j, line 139: This looks complicated for just handling a click on opening a menu

packages/manager/modules/freefax/package.json
AxelPeter: In first place, I've done it only for the navbar package. But it doesn't update at all to the needed version.
So I update it everywhere

marie-j: To update apps I usually run `yarn upgrade-interactive`, and then take care of the modules (for the peer dependencies)

packages/manager/modules/navbar/src/notifications-menu/controller.js
AxelPeter: What's complicated ? It supports the click on the trigger button and when we close the menu by clicking outside the panel.

The most complicated part I can see, is the click binding on the `$document`.
We need to unbind the function to avoid polluting `$document` with unnecessary function.

marie-j: Is this a wanted behaviour though ? Because the user panel behaves differently and close only if you click on the button in the navbar. Shouldn't we have a consistency between those two behaviours ?

AxelPeter: Yes, like the dropdown component, it closes the menu/panel when we click outside. It's implemented in the navbar: 
https://ovh.github.io/ovh-ui-kit/?path=/story/legacy-components-navbar-webcomponent--simple

I agree that we should have a consistency between all panels of the navbar. I could take a look at the user panel if you want.

AxelPeter:
Updated with the changes from https://github.com/ovh/manager/pull/3331

packages/manager/modules/notifications-sidebar/src/controller.js
```diff
@@ -31,18 +37,46 @@ export default class NotificationsCtrl {
     this.isLoading = true;
     this.numberOfActiveNotifications = 0;
 
+    // Will be bound to the click event on $document
+    this.readAllNotifications = () => {
+      this.toggle = false;
+
+      // Automatically set all unread messages to read
+      // When we close the notifications menu
```
antleblanc, line 45: **suggestion:** There is no benefit of starting new comment line with a capital letter.

```suggestion
      // when we close the notifications menu
```

packages/manager/modules/notifications-sidebar/src/controller.js
```diff
@@ -31,18 +37,46 @@ export default class NotificationsCtrl {
     this.isLoading = true;
     this.numberOfActiveNotifications = 0;
 
+    // Will be bound to the click event on $document
+    this.readAllNotifications = () => {
```
antleblanc, line 41: **question:** No `return` needed for this arrow function? (I'm asking this question because a promise is launched)

packages/manager/modules/notifications-sidebar/src/controller.js
```diff
@@ -31,18 +37,46 @@ export default class NotificationsCtrl {
...
+      this.toggle = false;
+
+      // Automatically set all unread messages to read
+      // When we close the notifications menu
+      this.NavbarNotifications.readAllNotifications(
+        this.getActiveNotifications(),
+      ).then(() => {
+        this.numberOfActiveNotifications = this.getActiveNotifications().length;
+        this.$rootScope.$emit(
+          'ovh::notifications::count',
```
antleblanc, line 51: **thought:** Maybe events' name can be sorted into a constant file (not mandatory for now but since they are used at multiple place, it might be interesting).

packages/manager/modules/notifications-sidebar/src/controller.js
```diff
@@ -31,18 +37,46 @@ export default class NotificationsCtrl {
...
+        );
+      });
+
+      this.$document.off('click', this.readAllNotifications);
+    };
+
     this.$rootScope.$on('ovh::notifications::toggle', () => {
       this.toggle = !this.toggle;
       if (this.toggle) {
+        // Handle the click outside the menu
```
antleblanc, line 62: **suggestion:** `menu` term looks like too generic. Why not prefer using `notifications menu` (already used L45)?

```suggestion
        // Handle the click outside the notifications menu
```

packages/manager/modules/notifications-sidebar/src/controller.js
```diff
@@ -77,15 +111,32 @@ export default class NotificationsCtrl {
       });
   }
 
-  getNumberOfActiveNotifications() {
-    return filter(this.sublinks, (notification) => notification.isActive)
-      .length;
+  $postLink() {
+    // Avoid click propagation inside the menu
```
antleblanc, line 115: **suggestion:** Based on my previous comment
```suggestion
    // Avoid click propagation inside the notifications menu
```

packages/manager/modules/notifications-sidebar/src/controller.js
```diff
@@ -77,15 +111,32 @@ export default class NotificationsCtrl {
...
+    // Since we're binding a click function of the document
+    this.$timeout(() => {
+      this.$element.on('click', (e) => {
+        e.stopPropagation();
+      });
```
antleblanc, line 120: **nitpick:** Using mangled variable name is not convenient to deal with. Why not prefer using `event` over `e`?

```suggestion
      this.$element.on('click', (event) => {
        event.stopPropagation();
      });
```

packages/manager/modules/notifications-sidebar/src/controller.js
```diff
@@ -77,15 +111,32 @@ export default class NotificationsCtrl {
...
+    });
+  }
+
+  $onDestroy() {
+    this.$document.off('click', this.readAllNotifications);
+  }
+
+  getActiveNotifications() {
+    return filter(this.sublinks, (notification) => notification.isActive);
```
antleblanc, line 129: **thought:** Existing a shorthand.
```suggestion
    return filter(this.sublinks, 'isActive');
```

antleblanc:
**Side question:** Ticket reference mentioned is marked as a `bug` whereas here we are relying on a feature.

Is the `navbar` (or `notifications-sidebar`) shouldn't received a bugfix instead?

packages/manager/modules/notifications-sidebar/src/controller.js
AxelPeter: Could be interesting to do it global with the micro front-end architecture :)

packages/manager/modules/notifications-sidebar/src/controller.js
AxelPeter: No, nothing is waited in return with this function.

AxelPeter:
@antleblanc Since it was never done and fixes nothing, I put it as a feature :)

AxelPeter:
Got 1 approve before rebase

PR #3277, opened 2020-07-16
Refactor/listing page

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-5369

## Description

- Extract layout to corresponding module
- Use component in hub app

## Related 
- [x] https://github.com/ovh/ovh-ui-kit/pull/641
- [x] https://github.com/ovh/ovh-ui-kit/pull/642


ovh-cds:
CDS Report build#13618.0 ✘  
* 
  * Build ✘

marie-j:
> **thought:** I've noticed that the code from the following module `@ovh-ux/manager-ng-layout-helpers` isn't located into a `src` folder. It also miss a `README.md` file maybe to provide some documentation.


@antleblanc Yes I think this went unnoticed when the module was created. I think that we could also add a rollup config to have a dist for this module among other things (maybe merge with the error page layout also). As it is out of scope here, I'll have a look to handle it in another PR if that sounds fine :) (and hopefully provide a README with some documentation maybe)

marie-j:
Updated due to UI-Kit v4 migration

JeremyDec:
@marie-j there's conflict on yarn.lock file

marie-j:
@JeremyDec Fixed conflicts and adapted file architecture due to recent changes

packages/manager/apps/hub/src/dashboard/products-list/config.js
```diff
@@ -99,7 +99,22 @@ export const resolves = {
...
+
+  description: /* @ngInject */ (notifications, productType) => {
+    const filteredNotifications = notifications.filter(
+      ({ type }) => type === productType,
+    );
+    return `<ovh-manager-hub-carousel
+        class="w-100"
+        data-ng-if="${filteredNotifications.length}"
+        data-items="${filteredNotifications}"
+    ></ovh-manager-hub-carousel>`;
```
cbourgois, line 115: Not a big fan of this, But I think it's something to deprecate in future.

packages/manager/modules/ng-layout-helpers/src/list/list-layout.utils.js
```diff
@@ -0,0 +1,134 @@
...
+    value: 'ASC',
+    squash: true,
+  },
+  filter: {
+    value: '[]',
+    squash: true,
+  },
+};
+
+export const urlQueryParams = 'page&pageSize&sort&sortOrder&filter';
```
cbourgois, line 123: NP: Could be something like:
```suggestion
export const urlQueryParams = Object.keys(stateParams).join('&');
```

packages/manager/apps/hub/src/dashboard/products-list/config.js
marie-j: I did that way to avoid a transclude as it was added for VPS migration. If the need persists, we can update to use a transclude instead

PR #3307, opened 2020-07-23
build: add rollup to build dist

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | none

## Description

Add rollup and build scripts to provide a dist 

Following https://github.com/ovh/manager/pull/3277#pullrequestreview-453859908


packages/manager/modules/ng-layout-helpers/package.json
```diff
@@ -10,12 +10,27 @@
   },
   "license": "BSD-3-Clause",
   "author": "OVH SAS",
-  "main": "index.js",
+  "main": "./dist/cjs/index.js",
+  "module": "./dist/esm/index.js",
+  "browser": "./dist/umd/manager-ng-layout-helpers.js",
```
antleblanc, line 15: Isn't supposed to be:
```suggestion
  "browser": "./dist/umd/ng-layout-helpers.js",
```

cbourgois: So the package name should be updated to `@ovh-ux/ng-layout-helpers` ? Seems to be manager specific no ?

marie-j: @cbourgois package name is different from the name generated while building the dist: it takes the name of the folder and not the name of the package

PR #3346, opened 2020-07-30
Fix/environnment region

cbourgois:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | yes
| Tickets          | Fix 

## Description

* add restriction on `Environment.setRegion` to limit region to `EU`, `CA` & `US` (BC)
* bump `@ovh-ux/manager-config` usages


packages/manager/modules/config/src/environment/environment.service.js
```diff
@@ -1,11 +1,15 @@
...
-    this.region = 'EU';
+    this.region = DEFAULT_REGION;
     this.version = null;
   }
 
-  setRegion(region = 'EU') {
-    this.region = region;
+  setRegion(region = DEFAULT_REGION) {
+    if (ALLOWED_REGIONS.includes(region)) {
+      this.region = region;
```
marie-j, line 11: **thought**: Maybe we could throw an error or add a log if region is not supported

PR #3355, opened 2020-07-31
fix(page.header): update to uniformize with ODS styles

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | none

## Description

Update cui-page-header to have a look and feel more or less like ODS

### DIFF
--- a/packages/manager/modules/cloud-universe-components/src/cui/page/header/template.html
+++ b/packages/manager/modules/cloud-universe-components/src/cui/page/header/template.html
@@ -1,8 +1,15 @@
-<header class="cui-page__header">
-    <cui-page-header-title
-        data-text="$ctrl.titleText"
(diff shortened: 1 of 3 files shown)


packages/manager/modules/cloud-universe-components/src/cui/page/header/template.html
```diff
@@ -1,8 +1,15 @@
-<header class="cui-page__header">
-    <cui-page-header-title
-        data-text="$ctrl.titleText"
-        data-subtitle-text="$ctrl.subtitleText"
-        data-guides="$ctrl.guides"
-    ></cui-page-header-title>
+<oui-header
+    class="cui-page__header"
```
marie-j, line 2: Necessary to keep the padding as bs4 utilities are not imported

PR #3382, opened 2020-08-05
Refactor: telephony architecture

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | none

## Description

⚠️ 1st commit is moving files around using git mv and replace paths
2nd commit is moving listing page as first architecture level : This is where the review should be focused


marie-j:
As discussed with @antleblanc, there might be a better approach to this and full architecture revamp can be handled as a later step

PR #3388, opened 2020-08-05
SMS: add breadcrumb

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | none

## Description

Add breadcrumb for SMS module

### DIFF
--- a/packages/manager/modules/sms/src/index.js
+++ b/packages/manager/modules/sms/src/index.js
@@ -14,7 +14,10 @@ angular
       $stateProvider
         .state('sms', {
           url: '/sms',
-          abstract: true,
+          redirectTo: 'sms.index',
+          resolve: {
+            breadcrumb: () => 'SMS',
+          },
         })
         .state('sms.index.**', {
           url: '',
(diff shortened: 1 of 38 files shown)


packages/manager/modules/sms/src/index.js
```diff
@@ -14,7 +16,10 @@ angular
       $stateProvider
         .state('sms', {
           url: '/sms',
-          abstract: true,
+          redirectTo: 'sms.index',
+          resolve: {
+            breadcrumb: () => PRODUCT_NAME,
```
frenautvh, line 21: don't you want to return $translate('sms_sms_breadcrumb') here ?

also i don't think it's necessary to create a constant for 'SMS' value given the module name is 'sms', a string literal would be completely fine imo

marie-j: This means that will have to load from another module or add a translation, I'm not sure it makes sense given the product name is static

ovh-cds:
CDS Report test-manager_2#14231.0 ✘  
* Stage 1
  * e2e -   
  * lint ✔ 
  * SemVer ✘

PR #3403, opened 2020-08-07
build: only provide esm target for private packages

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | 

## Description

There is no need in providing multiple targets for private monorepo packages. It slows down the build time.

### DIFF
--- a/packages/manager/modules/account-sidebar/rollup.config.js
+++ b/packages/manager/modules/account-sidebar/rollup.config.js
@@ -15,4 +15,4 @@ const config = rollupConfig(
   },
 );
 
-export default [config.es({})];
+export default [config.es()];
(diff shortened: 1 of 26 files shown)


packages/manager/modules/account-sidebar/rollup.config.js
```diff
@@ -15,19 +15,4 @@ const config = rollupConfig(
...
-        globals: {
-          angular: 'angular',
-        },
-      },
-    }),
-  );
-}
-
-export default outputs;
+export default [config.es({})];
```
antleblanc, line 18:
```suggestion
export default [config.es()];
```

PR #3418, opened 2020-08-11
Feat/emailpro standalone

marie-j:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/solarsteein`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | none
| License          | BSD 3-Clause

## Description

Clean up and create standalone app for emailpro/mxplan

PR #3431, opened 2020-08-13
Feat/exchange standalone app

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | none

## Requires 
- [x] https://github.com/ovh/manager/pull/3418/commits/ce88157173a8f9b55c85fd2700b53192e51b210c (#3418)


packages/manager/apps/exchange/src/index.html
```diff
@@ -0,0 +1,30 @@
+<!DOCTYPE html>
+<html>
+    <head>
+        <meta charset="utf-8" />
+        <title>Exchange App</title>
+        <style type="text/css">
+            .ng-cloak {
```
radireddy, line 7: consider moving this to css file

marie-j: I'm not sure it is possible. We need to check with @cbourgois but I think it was done that way because if not the styles are loaded too late

cbourgois: Yes, if we put this styles in a CSS loaded via webpack it will be loaded too late. It could be in external CSS (as recommended in https://docs.angularjs.org/api/ng/directive/ngCloak) but this CSS should be loaded as soon as possible.
Putting it in the html make it available when the sim is loaded.

PR #3446, opened 2020-08-17
ufrontend navbar fragment + refactoring

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | ufrontend
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | internal MANAGER-5357

## Description

Add navbar fragment, refactor ufrontend communication api


packages/manager/fragments/navbar/package.json
```diff
@@ -0,0 +1,53 @@
+{
+  "name": "@ovh-ux/manager-fragment-navbar",
```
marie-j, line 2: **thought**: A fragment generator could be provided to generate basic config in a next step

packages/manager/fragments/navbar/src/constants.js
```diff
@@ -0,0 +1,5 @@
+export const SECONDARY_UNIVERSES = ['partners'];
```
marie-j, line 1: I think this is no longer required as `Patners` universe no longer appears in navbar

packages/manager/fragments/navbar/src/hamburger-menu.jsx
```diff
@@ -0,0 +1,73 @@
...
+        className="oui-navbar-toggler oui-navbar-toggler_button"
+        onClick={() => setOpened(!opened)}
+        aria-expanded={opened}
+      >
+        <span className="oui-navbar-toggler__hamburger">
+          <span></span>
+          <span></span>
+          <span></span>
+          <span></span>
+        </span>
```
marie-j, line 33: Not yours but we should maybe ask CX to provide us with an icon for this

packages/manager/fragments/navbar/src/notifications.jsx
```diff
@@ -0,0 +1,41 @@
...
+      if (id === 'ovh.notifications.count') {
+        setNotificationsCount(count);
+      }
+    });
+  }, []);
+
+  return (
+    <button
+      role="button"
+      className={`oui-navbar-link oui-navbar-link_icon oui-navbar-link_tertiary ${style.navbarLink}`}
```
marie-j, line 21: I don't see `@ovh-ux/ui-kit` as a dependency so those classes are not of any use

frenautvh: You're right, it's a known issue we have, we don't want to bundle the whole ui-kit in the fragment but on the other hand we are not able to import only the needed css parts of the ui-kit components (using scoped imports).
Talked about it with @cbourgois, we choosed that we would temporarily use the ui-kit classes in the fragments, knowing that the ui-kit will be imported by applications and perform a refactoring later when the ui-kit provides more modularity. Do you have any opinion on it ?

marie-j: I understand and I agree that we shouldn't import the ui-kit yet in fragments to avoid importing things we don't need. But maybe the ui-kit could be set as a peer, just so the apps using the fragment knows they have to bring it in order to make the fragment work properly :)

packages/manager/fragments/navbar/src/constants.js
frenautvh: i removed the "partner" key in secondary_universes array but kept the array in case a secondary universe would need to be handled

packages/manager/fragments/navbar/src/notifications.jsx
frenautvh: added the peer dep

packages/manager/fragments/navbar/src/hamburger-menu.jsx
frenautvh: yep i can replace this one with two buttons and a oui-hambuger-icon but we would lose this marvelous animation ... https://github.com/ovh/ovh-ui-kit/blob/fe3fa0f5b7cb31001e88bf1fa54575c7e3f2f265/packages/components/navbar/src/less/navbar.less#L280

packages/manager/fragments/navbar/src/account.jsx
```diff
@@ -0,0 +1,49 @@
...
+  const [sidebarReady, setSidebarReady] = useState(false);
+  const { t } = useTranslation();
+  const firstName = capitalize(user.firstname);
+  const lastName = truncate(capitalize(user.name), {
+    length: 10,
+  });
+
+  useEffect(() => {
+    listen(({ id }) => {
+      if (id === 'ovh.account-sidebar.ready') {
```
radireddy, line 18: maybe move id to constants

packages/manager/fragments/navbar/src/account.jsx
```diff
@@ -0,0 +1,49 @@
...
+    <button
+      role="button"
+      className={`oui-navbar-link oui-navbar-link_icon oui-navbar-link_tertiary ${style.navbarLink}`}
+      aria-label={t('navbar_account')}
+      onClick={() => {
+        emit({
+          id: 'ovh.notifications.hide',
+        });
+        emit({
+          id: 'ovh.account-sidebar.toggle',
```
radireddy, line 34: maybe use an enum for id's and put it in constant file

ovh-cds:
CDS Report installation#14473.0 ✘  
* Stage 1
  * Install ✘

packages/manager/fragments/navbar/src/account.jsx
```diff
@@ -0,0 +1,50 @@
...
+  useEffect(() => {
+    listen(({ id }) => {
+      if (id === MESSAGES.accountSidebarReady) {
+        setSidebarReady(true);
+      }
+    });
+  }, []);
+
+  return (
+    <button
```
marie-j, line 26: No `type="button"` ?

PR #3447, opened 2020-08-17
fix(telephony.alias.configuration): prevent image from going out of tile

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | none

### DIFF
--- a/packages/manager/apps/telecom/src/app/telecom/telephony/alias/changeType/telecom-telephony-alias-changeType.html
+++ b/packages/manager/apps/telecom/src/app/telecom/telephony/alias/changeType/telecom-telephony-alias-changeType.html
@@ -110,6 +110,7 @@
                         class="transclude-image-container"
                     >
                         <img
+                            class="w-100"
                             src="assets/images/aliasFeature/schema-conference.svg"
                             alt="schema-conference.svg"
                         />


packages/manager/apps/telecom/src/app/telecom/telephony/alias/changeType/telecom-telephony-alias-changeType.html
```diff
@@ -110,6 +110,7 @@
                         class="transclude-image-container"
                     >
                         <img
+                            class="w-100"
```
antleblanc, line 113: I cannot reproduce the issue, tested on both with Firefox and Chromium.

Furthermore there are several `` tag in this following template. It only affect `schema-conference.svg`?

Is is linked to this rules: https://github.com/ovh/manager/blob/master/packages/manager/apps/telecom/src/app/telecom/telephony/telecom-telephony.less#L46?

marie-j: I thought it was a ng-repeat, I badly read the code, my bad 
I can reproduce on Chromium from `feat/solarsteein`, maybe I need to update my branch

PR #3471, opened 2020-08-19
fix(order): add order status tracking

varun257:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-4405

## Description

add a custom page tracking variable 

Signed-off-by: varun257 <varun257@gmail.com>

### DIFF
--- a/packages/components/ng-at-internet/src/constants.js
+++ b/packages/components/ng-at-internet/src/constants.js
@@ -49,6 +49,14 @@ export const AT_INTERNET_CUSTOM_VARS = {
     path: 'page.2', // OVH's AtInternet configuration
     format: '[%s]',
   },
+
+  /**
+   * Order status used
+   */
+  orderStatus: {
+    path: 'page.1',
+    format: '[%s]',
+  },
 };
 
 export default {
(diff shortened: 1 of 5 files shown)


packages/components/ng-at-internet/src/constants.js
```diff
@@ -49,6 +49,14 @@ export const AT_INTERNET_CUSTOM_VARS = {
     path: 'page.2', // OVH's AtInternet configuration
     format: '[%s]',
   },
+
+  /**
+   * Voucher code used
```
antleblanc, line 54: Looks like a copy/paste from the constant above.
Could you please update it to reflect the meaning of this constant?

In addition, maybe you can split your commit in two distinct commit.
The first one act more as a feature (adding a new constant)
and then the second one is fixing the order tracking component.

Is that make sense?

Jisay (approved):

antleblanc (approved):

antleblanc (approved):

PR #3479, opened 2020-08-19
feat: add plugin RetryChunkLoadPlugin

antleblanc:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | no

### :rocket: Features

9004b11  - feat: add plugin RetryChunkLoadPlugin

Covers following error: `ChunkLoadError`.

### :link: Relates

- https://github.com/mattlewis92/webpack-retry-chunk-load-plugin#readme

cc @jleveugle

### DIFF
--- a/packages/manager/tools/webpack-config/package.json
+++ b/packages/manager/tools/webpack-config/package.json
@@ -70,6 +70,7 @@
     "webpack-bundle-analyzer": "^2.13.1",
     "webpack-cli": "^3.3.9",
     "webpack-merge": "^4.2.2",
+    "webpack-retry-chunk-load-plugin": "^1.4.0",
     "webpackbar": "^3.2.0"
   },
   "devDependencies": {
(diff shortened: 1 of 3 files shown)


marie-j (approved):

Jisay (approved):

frenautvh:
i am wondering : did we already have ChunkLoadError issues reported? i'm don't see how it could append unless the infra is down, chunks are only static files.

Since the plugin does not seems to be widely used i'm not sure if it's worth adding it to our webpack build

antleblanc:
> i am wondering : did we already have ChunkLoadError issues reported? i'm don't see how it could append unless the infra is down, chunks are only static files.

Yes! Events are frequently reported from our monitoring system.

From the browser point of view, the infra is not seen as down.

The problem might happen (more ore less) because we are relying on a single page
application which is build at once and then we push the dist to the server.

Compared to previous releases, a large amount of chunks may changed and that's why
we have implemented some sort of mechanism to stay with previous chunks on the server.

> Since the plugin does not seems to be widely used i'm not sure if it's worth adding it to our webpack build

I'm wondering which inputs you're referring to.
npm download reports seems to argue positively (https://www.npmjs.com/package/webpack-retry-chunk-load-plugin ).

frenautvh:
i'm still doubtful that reloading chunks will avoid errors since the chunk src must not be the good one at first but let's test it and see how it turns out

frenautvh (approved):

JeremyDec (approved):

PR #3484, opened 2020-08-21
Hotfix 2020-08-21

marie-j:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #... 

## Description

DTRSD-18376, DTRSD-18231

### DIFF
--- a/packages/manager/modules/product-offers/src/workflows/product-offers-services-workflow.class.js
+++ b/packages/manager/modules/product-offers/src/workflows/product-offers-services-workflow.class.js
@@ -19,8 +19,9 @@ export default class ServicesWorkflow extends Workflow {
    * @param {Object} DetachService   Service to handle request to perform a
    * detach action, see /services API schema
    */
-  constructor($q, $translate, workflowOptions, detachService) {
+  constructor($q, $timeout, $translate, workflowOptions, detachService) {
     super($q, $translate, workflowOptions);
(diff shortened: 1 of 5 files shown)


packages/manager/modules/product-offers/src/workflows/product-offers-services-workflow.class.js
```diff
@@ -40,8 +41,10 @@ export default class ServicesWorkflow extends Workflow {
     this.pricings = this.computePricing(detachPlancode.prices);
 
     if (this.hasUniquePricing()) {
-      this.currentIndex += 1;
-      [this.pricing] = this.pricings;
+      this.$timeout(() => {
```
Jisay, line 44: I don't understand the use of `$timeout` here

JeremyDec: @Jisay the `$timeout` is necessary to increment correctly the currentIndex (exposed variable of `oui-stepper`), after the component is initialized

marie-j: @Jisay @JeremyDec I think it's ok as a workaround but we should try to find why this `$timeout` is needed and if we can avoid its usage

PR #3545, opened 2020-09-01
Feat/nasha breadcrumb

marie-j:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #... 


packages/manager/modules/nasha/src/index.js
```diff
@@ -3,47 +3,69 @@ import '@ovh-ux/manager-core';
...
+        .state('nasha.nasha-add.**', {
+          url: '/new',
+          lazyLoad,
+        })
+        .state('nasha.nasha-unavailable.**', {
+          url: '/unavailable',
           lazyLoad,
         });
+
+      $urlRouterProvider.when(/^\/paas\/nasha/, () => {
```
varun257, line 63: Is it possible to use `redirectTo` instead ?

marie-j: @varun257 A `redirectTo` will only work for root url (if customer target `/paas/nasha`). If I target `/paas/nasha/{serviceName}` the `redirectTo` won't apply except if for each nasha state we create a paas.nasha state that adds a `redirectTo`, it is easier to maintain the redirection that way imo

varun257 (approved):

packages/manager/modules/nasha/src/dashboard/routing.js
```diff
@@ -0,0 +1,66 @@
...
+      controllerAs: 'NashaOrderCompleteCtrl',
+      translations: {
+        value: ['order', '..'],
+        format: 'json',
+      },
+      params: {
+        orderUrl: null,
+      },
+      resolve: {
+        hideBreadcrumb: /* @ngInject */ () => true,
```
JeremyDec, line 37: Same thing for the two below
```suggestion
        hideBreadcrumb: () => true,
```

JeremyDec (approved):

marie-j:
Only updated with @JeremyDec suggestion

PR #3567, opened 2020-09-02
fix(billing.orders): filter null result from iceberg request

JeremyDec:
ref: DTRSD-20245

| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-20245

### DIFF
--- a/packages/manager/apps/dedicated/client/app/billing/orders/orders.routing.js
+++ b/packages/manager/apps/dedicated/client/app/billing/orders/orders.routing.js
@@ -21,7 +21,7 @@ export default /* @ngInject */ ($stateProvider) => {
           .sort('date', 'DESC')
           .limit(5000)
           .execute(null, true)
-          .$promise.then(({ data }) => data),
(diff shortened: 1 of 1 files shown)


packages/manager/apps/dedicated/client/app/billing/orders/orders.routing.js
```diff
@@ -21,7 +21,7 @@ export default /* @ngInject */ ($stateProvider) => {
           .sort('date', 'DESC')
           .limit(5000)
           .execute(null, true)
-          .$promise.then(({ data }) => data),
+          .$promise.then(({ data }) => data.filter((row) => row != null)),
```
marie-j, line 24: Why does some results are null ? Isn't it an API issue to return non valid values ?

JeremyDec: Issue from iceberg, the api route itself doesn't return any null value. Waiting their response, I will convert this PR in draft

PR #3628, opened 2020-09-16
Fix/manager 5290

mohammed-zahaf:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-5290

## Description

Manage translations according the API returned status

### :flags: Translations

- [x] TRANS-28451

### DIFF
--- a/packages/manager/modules/support/src/tickets/new-ticket/new-ticket.controller.js
+++ b/packages/manager/modules/support/src/tickets/new-ticket/new-ticket.controller.js
@@ -7,6 +7,7 @@ export default class SupportNewController {
     $q,
     $state,
     $window,
+    $translate,
     CORE_URLS,
     OvhApiMe,
     OvhApiSupport,
@@ -15,6 +16,7 @@ export default class SupportNewController {
     this.$q = $q;
     this.$state = $state;
     this.$window = $window;
+    this.$translate = $translate;
     this.CORE_URLS = CORE_URLS;
     this.OvhApiMe = OvhApiMe;
     this.OvhApiSupport = OvhApiSupport;
@@ -37,6 +39,7 @@ export default class SupportNewController {
       this.step = 'creation';
       this.issue = result.issue;
       this.service = result.service;
+      this.serviceType = result.serviceType;
       this.$window.scrollTo(0, 0);
     }
   }
@@ -66,6 +69,7 @@ export default class SupportNewController {
       .catch((error) => {
         this.error = {
           message: (error.data || { message: error.statusText }).message,
+          class: get(error, 'data.class'),
         };
         if (angular.isFunction(error.headers)) {
           this.error.queryId = error.headers('x-ovh-queryid');
@@ -85,5 +89,26 @@ export default class SupportNewController {
 
     this.step = 'issues';
     return this.$q.when();
+  }
+
+  isDedicatedServer() {
+    return get(this.serviceType, 'name') === 'dedicated_server';
+  }
+
+  getErrorMessage() {
+    if (this.isDedicatedServer() && this.error.class) {
+      if (this.error.class === 'Client::BadRequest::AlreadyOpened') {
+        return this.$translate.instant(
+          'ovhManagerSupport_new_creation_error_dedicated_server_already_opened',
+        );
+      }
+      if (this.error.class === 'Client::BadRequest::ServiceMandatory') {
+        return this.$translate.instant(
+          'ovhManagerSupport_new_creation_error_dedicated_server_service_mandatory',
+        );
+      }
+    }
+
+    return this.error.message;
   }
 }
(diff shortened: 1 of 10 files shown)


Jisay (approved):

packages/manager/modules/support/src/tickets/new-ticket/translations/Messages_fr_FR.json
```diff
@@ -120,6 +120,8 @@
   "ovhManagerSupport_new_creation_success": "Votre ticket numéro <a ui-sref=\"{{ state }}\">{{ id }}</a> a bien été pris en compte.",
   "ovhManagerSupport_new_creation_success_back": "Vous pouvez retrouver le fil de discussion de ce ticket depuis la <a ui-sref=\"{{ state }}\" ui-sref-opts=\"{ reload: true }\">liste des tickets</a>.",
   "ovhManagerSupport_new_creation_error": "Une erreur est survenue lors de la création de votre ticket.",
+  "ovhManagerSupport_new_creation_error_already_opened": "Un ticket d'intervention est déjà en cours pour ce serveur.",
```
frenautvh, line 123: how do you know that the ticket is about a dedicated server ?

frenautvh (changes requested):

packages/manager/modules/support/src/tickets/new-ticket/translations/Messages_fr_FR.json
mohammed-zahaf: Is about support-app

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.html
```diff
@@ -93,10 +93,16 @@
     <p>
         <strong data-translate="ovhManagerSupport_new_creation_error"></strong>
     </p>
-    <p>
-        <span data-ng-bind="$ctrl.error.message"></span>
-        <br />
-        <small data-ng-bind="$ctrl.error.queryId"></small>
+    <p data-ng-switch="$ctrl.error.class">
+        <span
+            data-ng-switch-when="Client::BadRequest::AlreadyOpened"
```
cbourgois, line 98: Not a big fan to do a switch directly in html based on API results.

cbourgois (changes requested):

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.html
mohammed-zahaf: Why ?

cbourgois: Because the way we fetch and manage the data/errors are not the html's responsibility.

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.controller.js
```diff
@@ -86,4 +88,18 @@ export default class SupportNewController {
     this.step = 'issues';
     return this.$q.when();
   }
+
+  isDedicatedServer() {
+    return get(this.serviceType, 'name') === 'dedicated_server';
+  }
+
+  getDedicatedServerErrorTranslationKey() {
```
cbourgois, line 96: Can we have something more generic ? and more consistent ? Here this function return sometimes a translation key, sometimes a message that doesn't need to be translated, I think we could use `$translate.instant` in it.

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.html
```diff
@@ -93,10 +93,13 @@
     <p>
         <strong data-translate="ovhManagerSupport_new_creation_error"></strong>
     </p>
-    <p>
+    <p data-ng-if="$ctrl.isDedicatedServer()">
```
cbourgois, line 96: Can we have something more generic? Tomorrow if we have to display a specific message for all products, do we plan to add multiple `p` with `ng-if` ?

packages/manager/modules/support/src/tickets/new-ticket/translations/Messages_fr_FR.json
mohammed-zahaf: Right

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.controller.js
mohammed-zahaf: i return the global message in case where the API status is not yet cover by our side or somethings goes bad.

packages/manager/modules/support/src/tickets/new-ticket/new-ticket.html
mohammed-zahaf: No there is a p for all other cases which display a global message

mohammed-zahaf: Like it was initialy

marie-j: @mohammed-zahaf there is for other cases for now

frenautvh (dismissed):

cbourgois (approved):

antleblanc (dismissed):

marie-j (approved):

marie-j (approved):

PR #3641, opened 2020-09-17
Telecom: avoid unhandled rejections

marie-j:
| Question         | Answer
| ---------------- | ---
| Branch?          | `release/2020-w38`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | yes
| Tickets          | MANAGER-7D and possibly more

## Description

Prevent TucToastError from returning a rejected promise as it is never catch and we don't expect this service to return a promise

### DIFF
--- a/packages/manager/apps/telecom/src/components/telecom/telephony/alias/members/telecom-telephony-alias-members.component.js
+++ b/packages/manager/apps/telecom/src/components/telecom/telephony/alias/members/telecom-telephony-alias-members.component.js
@@ -61,12 +61,9 @@ angular.module('managerApp').component('telecomTelephonyAliasMembers', {
       );
       return $translate.refresh().finally(() => {
         self.isInitialized = true;
-        return self
-          .refreshMembers()
-          .catch((err) => new TucToastError(err))
-          .finally(() => {
-            self.loaders.init = false;
-          });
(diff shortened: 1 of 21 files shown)


cbourgois (commented):
If there is no place where we catch the rejection it's ok.

marie-j:
> If there is no place where we catch the rejection it's ok.

@cbourgois Currently checking as this is used quite a log, if it is used I'll update the PR by updating the location where it needs to

frenautvh (approved):

packages/manager/apps/telecom/src/components/telecom/telephony/alias/members/telecom-telephony-alias-members.component.js
```diff
@@ -61,12 +61,9 @@ angular.module('managerApp').component('telecomTelephonyAliasMembers', {
       );
       return $translate.refresh().finally(() => {
         self.isInitialized = true;
-        return self
-          .refreshMembers()
-          .catch((err) => new TucToastError(err))
-          .finally(() => {
-            self.loaders.init = false;
-          });
+        return self.refreshMembers().finally(() => {
```
cbourgois, line 64: 👍

cbourgois (approved):

PR #3667, opened 2020-09-22
refactor(sharepoint): replace apiv7 with iceberg

radireddy:
closes #MANAGER-5634

| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-5634

## Description

replace apiv7 calls with iceberg


packages/manager/modules/sharepoint/src/sharepoint.service.js
```diff
@@ -623,32 +625,47 @@ export default class MicrosoftSharepointLicenseService {
...
+        )),
+      )
+      .then((services) => reduce(services, (flattened, other) => {
+          return flattened.concat(other);
+        }, []),
+      );
+  }
+
+  getExchangeServices(organizationName) {
+    return this.iceberg('/email/exchange/:organizationName/service')
```
marie-j, line 644: Nice job ! But what bothers me is that we've replaced one call with N+1 one, this might have quite some performance issue if there are quite a lot of services

radireddy: that's true

JeremyDec (dismissed):

antleblanc (approved):

packages/manager/modules/sharepoint/src/sharepoint.service.js
antleblanc: I do agree. Since it has been validated as is it could be part of the next release. We will keep an eye on that to make sure that performance are not impacted.

PR #3629, opened 2020-09-16
Feat/sms campaign

JeremyDec:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/sms-in-europe`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | #MANAGER-5201
| License          | BSD 3-Clause

## Description

Adding the new dashboard for sms service, and the campaign portal page.

PR #3692, opened 2020-09-25
Feat/sms campaign details

JeremyDec:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/sms-campaign-history`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-5201

## Description

Add batch details page


packages/manager/modules/sms/src/sms/batches/details/delete/routing.js
```diff
@@ -0,0 +1,30 @@
...
+      outgoing: /* @ngInject */ ($transition$) =>
+        $transition$.params().outgoing,
+      goBackToDetails: /* @ngInject */ ($state) => (options) =>
+        $state.go('sms.service.batches.details', {}, options),
+    },
+    layout: 'modal',
+    translations: {
+      value: ['.'],
+      format: 'json',
+    },
```
marie-j, line 28: Maybe replace by a 
```suggestion
 .run(/* @ngTranslationsInject:json ./translations */)
```

packages/manager/modules/sms/src/sms/batches/details/routing.js
```diff
@@ -0,0 +1,51 @@
...
+export default /* @ngInject */ ($stateProvider) => {
+  $stateProvider.state('sms.service.batches.details', {
+    url: '/:batchId',
+    views: {
+      'smsView@sms.service': {
+        component: component.name,
+      },
+    },
+    params: {
+      previousState: '^',
```
marie-j, line 14: Can't we use `$transition$.from()` or something like that instead ?

packages/manager/modules/sms/src/sms/batches/details/routing.js
```diff
@@ -0,0 +1,51 @@
...
+    },
+    resolve: {
+      batch: /* @ngInject */ ($transition$, batches) =>
+        batches.find(({ id }) => id === $transition$.params().batchId),
```
marie-j, line 18: batchId could be in its own resolve

packages/manager/modules/sms/src/sms/batches/details/routing.js
```diff
@@ -0,0 +1,51 @@
...
+        $state.go($transition$.params().previousState, {}, options),
+      goToDelete: /* @ngInject */ ($state) => (outgoing) =>
+        $state.go('sms.service.batches.details.delete', {
+          outgoing,
+        }),
+      outgoingSms: /* @ngInject */ (batch, getOutgoingSms) =>
+        getOutgoingSms(batch),
+    },
+    translations: {
+      value: ['.', '..'],
```
marie-j, line 47: To be standalone, the module shouldn't need a parent's translations

packages/manager/modules/sms/src/sms/batches/details/telecom-sms-batches-details.html
```diff
@@ -0,0 +1,106 @@
...
+            </span>
+            <span data-ng-if="$row.deliveryReceipt !== 1">
+                <span data-ng-if="$row.status.length >= 20">
+                    <span
+                        data-ng-bind="$row.status | limitTo: 20"
+                        data-oui-tooltip="{{ $row.status }}"
+                    >
+                    </span
+                    >…
+                </span>
```
marie-j, line 91: `class="text-ellipsis`

packages/manager/modules/sms/src/sms/batches/details/delete/routing.js
JeremyDec: See my comment on another PR : https://github.com/ovh/manager/pull/3629#discussion_r490807858 :)

packages/manager/modules/sms/src/sms/batches/details/routing.js
JeremyDec: Problem is that the details state has a child modal state (sms.service.batches.details.delete) which is called to open delete batch modal. If I use `$transition$.from()`, click on header back button may call the modal, instead of going back to parent page.

packages/manager/modules/sms/src/sms/batches/details/delete/routing.js
JeremyDec: Erratum: fix done

packages/manager/modules/sms/src/sms/batches/details/routing.js
JeremyDec: ok but I don't use it by itself

packages/manager/modules/sms/src/sms/batches/details/routing.js
marie-j: **suggestion**: Can't the header back button always redirect to the same parent page ? which is `sms.service.batches`

packages/manager/modules/sms/src/sms/batches/details/routing.js
Heiiji: this is seem to be an outdated part of code, can you re-check please ? :)

packages/manager/modules/sms/src/sms/batches/details/routing.js
marie-j: Ok but it would ease reuse in controller and other resolve if needed

packages/manager/modules/sms/src/sms/batches/details/delete/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,12 @@
...
+  "sms_batches_history_details_delete_title": "Confirmation",
+  "sms_batches_history_details_delete_question": "Êtes-vous sûr de vouloir supprimer le SMS affiché ci-dessous ?",
+  "sms_batches_history_details_delete_remove_ko": "Impossible de supprimer le SMS.<br/>{{error}}",
+  "sms_batches_history_details_delete_date": "Date",
+  "sms_batches_history_details_delete_sender": "Expéditeur",
+  "sms_batches_history_details_delete_receiver": "Destinataire",
+  "sms_batches_history_details_delete_destination": "Destination",
+  "sms_batches_history_details_delete_message": "Message",
+  "sms_batches_history_details_delete_error": "Une erreur est survenue lors de la suppression du SMS: {{error}}",
+  "sms_batches_history_details_delete_success": "Le SMS a été supprimé"
```
cbourgois, line 11: Don't find where theses translations are used.

packages/manager/modules/sms/src/sms/batches/pending/cancel/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,12 @@
...
+  "sms_cancel_title": "Confirmation",
+  "sms_cancel_details_name": "Nom",
+  "sms_cancel_details_date": "Date de création",
+  "sms_cancel_details_sender": "Expéditeur",
+  "sms_cancel_details_receivers": "Destinataires",
+  "sms_cancel_details_message": "Message",
+  "sms_cancel_question": "Êtes-vous sûr de vouloir annuler la campagne affichée ci-dessous ?",
+  "sms_removing": "Annulation en cours…",
+  "sms_cancel_error": "Impossible d‘annuler la campagne. {{error}}",
+  "sms_cancel_success": "La campagne a été annulée"
```
cbourgois, line 11: Don't find where theses translations are used.

packages/manager/modules/sms/src/sms/batches/details/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,7 @@
+{
+  "sms_batches_details_title": "Détail de la campagne",
+  "sms_batches_details_name": "Campagne: {{name}}",
+  "sms_batches_details_message": "Message délivré",
```
cbourgois, line 4: Don't find where this translation is used.

cbourgois (commented):
Some translations declared but never used in this PR (maybe they are already used and added because they are missing?).

packages/manager/modules/sms/src/sms/batches/details/delete/translations/Messages_fr_FR.json
cbourgois: `sms_batches_history_details_delete_error` & `sms_batches_history_details_delete_success`

packages/manager/modules/sms/src/sms/batches/details/translations/Messages_fr_FR.json
Heiiji: thanks for feedback, i've not found this one on any part of code. i've removed it on last commit :)

packages/manager/modules/sms/src/sms/batches/pending/cancel/translations/Messages_fr_FR.json
Heiiji: thanks for feedback, i've not found this one on any part of code. i've removed it on last commit :)

packages/manager/modules/sms/src/sms/batches/details/delete/translations/Messages_fr_FR.json
Heiiji: thanks for feedback, i've not found this one on any part of code. i've removed it on last commit :)

packages/manager/modules/sms/src/sms/batches/details/delete/routing.js
```diff
@@ -0,0 +1,26 @@
+import component from './telecom-sms-batches-history-details-delete.component';
+
+export default /* @ngInject */ ($stateProvider) => {
+  $stateProvider.state('sms.service.batches.details.delete', {
+    params: {
```
marie-j, line 5: No `url` ?

packages/manager/modules/sms/src/sms/batches/details/delete/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,10 @@
+{
+  "sms_batches_history_details_delete_title": "Confirmation",
+  "sms_batches_history_details_delete_question": "Êtes-vous sûr de vouloir supprimer le SMS affiché ci-dessous ?",
+  "sms_batches_history_details_delete_remove_ko": "Impossible de supprimer le SMS.<br/>{{error}}",
```
marie-j, line 4: Is it mandatory to have a new line for the `error` message

packages/manager/modules/sms/src/sms/batches/details/routing.js
JeremyDec: Why not, but here, the back button is set to go back to the page which has called the details page.
(It is not clear through this PR, but the details state can be called in two different points: through pending batches page or all batches page)

packages/manager/modules/sms/src/sms/batches/details/delete/routing.js
JeremyDec: Yes, I choose to not have this state accessible by reloading or with bookmark, as it is a modal to delete an item

packages/manager/modules/sms/src/sms/batches/details/delete/translations/Messages_fr_FR.json
JeremyDec: Copy paste is bad :)

frenautvh (approved):

packages/manager/modules/sms/src/sms/batches/details/delete/routing.js
marie-j: Imo, we're losing part of the benefits of having a state for a modal if we can't access it with a url

cbourgois: > Yes, I choose to not have this state accessible by reloading or with bookmark, as it is a modal to delete an item

The modal warn the customer about the deletion and ask a confirmation, so IMHO it's ok to be able to reach this modal directly. 
It's already the cases for some deletion modals in the repository : 
https://github.com/ovh/manager/blob/master/packages/manager/modules/pci/src/projects/project/instances/delete/delete.routing.js#L2-L3

JeremyDec: > The modal warn the customer about the deletion and ask a confirmation, so IMHO it's ok to be able to reach this modal directly.

In fact the modal state parameter `outgoing` is send by an action to be done by its parent state, not through a resolve. So it is strongly binded to it.
Adding a url to access to this modal would breaks the required `outgoing` param.

(See https://github.com/ovh/manager/pull/3692/files#diff-c54c173aa31a43fe8e6f5fc0fa9bf84eR102 + https://github.com/ovh/manager/pull/3692/files#diff-048f72c5252514d9d1d68957cdae4a18R40)

PR #3746, opened 2020-10-01
Fix/flatpickr locales

cbourgois:
| Question         | Answer
| ---------------- | ---
| Branch?          | `hotfixes-2020-10-01`
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix 

### DIFF
--- a/packages/manager/apps/public-cloud/src/app.module.js
+++ b/packages/manager/apps/public-cloud/src/app.module.js
@@ -92,6 +92,16 @@ angular
       );
     },
   )
+  .config(
+    /* @ngInject */ (ouiCalendarConfigurationProvider) => {
+      const lang = Environment.getUserLanguage();
+      return import(`flatpickr/dist/l10n/${lang}.js`)
+        .then((module) => {
+          ouiCalendarConfigurationProvider.setLocale(module.default[lang]);
+        })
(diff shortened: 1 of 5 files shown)


packages/manager/apps/public-cloud/src/app.module.js
```diff
@@ -92,6 +92,16 @@ angular
       );
     },
   )
+  .config(
+    /* @ngInject */ (ouiCalendarConfigurationProvider) => {
+      const lang = Environment.getUserLanguage();
+      return import(`flatpickr/dist/l10n/${lang}.js`)
+        .then((module) => {
+          ouiCalendarConfigurationProvider.setLocale(module.default[lang]);
```
marie-j, line 100: I wonder if this should belong to `@ovh-ux/manager-core` like other `oui` configurations (but as it is buit with rollup it will require to handle the lange import differently)

cbourgois: Yes it could move to `@ovh-ux/manager-core` but it means that if we don't use `flatpickr` (= don't use `oui-calendar`) in an application we still should import `flatpickr`.

marie-j (approved):

PR #3765, opened 2020-10-05
Fix/dedicated layout with bootstrap

Heiiji:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no

## Description

update 'offset' class on bootstrap 4 references to fox responssive result.

### DIFF
--- a/packages/manager/apps/dedicated/client/app/index.html
+++ b/packages/manager/apps/dedicated/client/app/index.html
@@ -75,9 +75,7 @@
                         data-sidebar-namespace="'account'"
                     ></div>
 
-                    <div
-                        class="col-md-9 col-md-offset-3 col-lg-10 col-lg-offset-2 px-0"
-                    >
+                    <div class="col-lg-10 offset-lg-2 px-0">
                         <!-- Skip content target -->
                         <div class="skiptarget">
                             <a id="maincontent" href>-</a>


packages/manager/apps/dedicated/client/app/index.html
```diff
@@ -75,9 +75,7 @@
                         data-sidebar-namespace="'account'"
                     ></div>
 
-                    <div
-                        class="col-md-9 col-md-offset-3 col-lg-10 col-lg-offset-2 px-0"
-                    >
+                    <div class="col-lg-10 offset-lg-2 px-0">
```
cbourgois, line 78: `md` not required ?

Heiiji: Nope, other CSS rule discard side-menu before `md` rule. :)

marie-j: I'm quite confused with this change as the same component and same configuration is applied on other universe (e.g: web https://github.com/ovh/manager/blob/master/packages/manager/apps/web/client/app/index.html#L68) and it doesn't seem to cause any issue or if it does maybe it should be applied on other universes

Heiiji: for apply on other universe, we need to import same version of bootstrap everywhere :/

PR #3802, opened 2020-10-08
[WIP] Server OS installation

Jisay:
| Question         | Answer
| ---------------- | ---
| Branch?          | refactor/server-os-install
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-5680

## Description

Refactor the section that install an OS from an OVH template on a dedicated server.

### DIFF
--- a/packages/manager/apps/dedicated/client/app/dedicated/server/install/ovh/components/raid/models/controller.js
+++ b/packages/manager/apps/dedicated/client/app/dedicated/server/install/ovh/components/raid/models/controller.js
@@ -1,0 +1,42 @@
+import get from 'lodash/get';
+
+import { HARDWARE_RAID_ENUM } from '../constants';
+
+export default class DedicatedServerRaidController {
+  constructor({ type, model, disks }) {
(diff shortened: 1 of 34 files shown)


packages/manager/apps/dedicated/client/app/dedicated/server/install/ovh/components/raid/models/controller.js
```diff
@@ -0,0 +1,42 @@
...
+
+  get displayName() {
+    return `${this.model} (${this.type})`;
+  }
+
+  getAvailableRaids() {
+    const numberOfDisk = get(this.disks, '[0].names').length;
+    const availableRaids = [];
+
+    if (numberOfDisk >= 8) {
```
frenautvh, line 20: not mandatory but you could also have this info in the enum, for example :
```
export const HARDWARE_RAID_ENUM = {
  raid0: {
    id: 'raid0',
    disks: 2,
  },
  ...
};
```

PR #3805, opened 2020-10-08
Fix/ldp

marie-j:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/ldp-migration`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-5750, MANAGER-5753, MANAGER-5754, MANAGER-5755

### DIFF
--- a/packages/manager/modules/dbaas-logs/package.json
+++ b/packages/manager/modules/dbaas-logs/package.json
@@ -12,6 +12,7 @@
   "author": "OVH SAS",
   "main": "./src/index.js",
   "dependencies": {
+    "@ovh-ux/manager-config": "^1.0.0",
     "lodash": "^4.17.20"
   },
   "devDependencies": {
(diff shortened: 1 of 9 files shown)


lizardK (approved):

packages/manager/modules/dbaas-logs/package.json
```diff
@@ -12,6 +12,7 @@
   "author": "OVH SAS",
   "main": "./src/index.js",
   "dependencies": {
+    "@ovh-ux/manager-config": "^1.0.0",
```
cbourgois, line 15: Should be in `peerDependencies` ?

marie-j: @cbourgois It's not an angular module so it's ok to add in it dependencies. Any reason you would rather add it as a peer instead ?

cbourgois: Ok, forgot that it's not an angular module. Thanks for reminding me ;)

cbourgois (approved):

PR #3984, opened 2020-11-05
fix(billing): remove pagination if empty list

Heiiji:
| Question         | Answer
| ---------------- | ---
| Branch?          |  `develop`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no

## Description

just small fix when nothing is listed -> no pagination needed

### DIFF
--- a/packages/manager/apps/dedicated/client/app/billing/payment/refunds/billing-refunds.html
+++ b/packages/manager/apps/dedicated/client/app/billing/payment/refunds/billing-refunds.html
@@ -216,6 +216,7 @@
         </tbody>
     </table>
     <div
+        data-ng-show="refunds.list.results.length > 0"
         data-pagination-server-side="refundTable"
         data-pagination-server-side-function="loadRefunds"
         data-pagination-server-side-paginated-stuff="refunds"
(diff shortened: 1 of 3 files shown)


packages/manager/apps/dedicated/client/app/billing/payment/refunds/billing-refunds.html
```diff
@@ -216,6 +216,7 @@ <h2 data-translate="refunds_title2"></h2>
         </tbody>
     </table>
     <div
+        data-ng-show="refunds.list.results.length > 0"
```
radireddy, line 219:
```suggestion
        data-ng-if="refunds.list.results.length > 0"
```

radireddy (approved):

packages/manager/apps/dedicated/client/app/billing/payment/refunds/billing-refunds.html
frenautvh: don't think ng-if will work here as pagination-server-side is still required to initialize and fetch data
ng-show seems appropriate

Heiiji: it break the display with this method :/

Jisay: yes the `div` must be in the dom in order to instanciate the directive on it. So `ngIf` won't work

Jisay (approved):

antleblanc (approved):

PR #3996, opened 2020-11-10
perf: lazy load angular-qr dependency

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | master
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #... 

## Description

angular-qr was imported for the whole application, now it's lazy loaded only in the module using it
remove unused dependency qrcode.js

un-gzipped gains : 50Kb on dedicated

### DIFF
--- a/packages/manager/apps/dedicated/client/app/account/user/user.routes.js
+++ b/packages/manager/apps/dedicated/client/app/account/user/user.routes.js
@@ -20,6 +20,10 @@ angular.module('UserAccount').config(
       },
       redirectTo: `${name}.method`,
       resolve: {
+        angularQr: /* @ngInject */ ($ocLazyLoad) =>
+          import('angular-qr').then((module) =>
+            $ocLazyLoad.inject(module.default || module),
+          ),
         schema: /* @ngInject */ (OvhApiMe) => OvhApiMe.v6().schema().$promise,
         supportLevel: /* @ngInject */ (OvhApiMe, schema) =>
           schema.models[API_MODEL_SUPPORT_LEVEL]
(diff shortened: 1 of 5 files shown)


packages/manager/apps/dedicated/client/app/account/user/user.routes.js
```diff
@@ -20,6 +20,10 @@ angular.module('UserAccount').config(
       },
       redirectTo: `${name}.method`,
       resolve: {
+        angularQr: /* @ngInject */ ($ocLazyLoad) =>
+          import('angular-qr').then((module) =>
```
radireddy, line 24: 👍 can this be imported in a lazy loaded module?

frenautvh: yes importing it in a lazy loaded module would have the same effect (or even better since the whole module would only be loaded when needed)
i didn't look at lazy load user module because it would take much more refactoring work, it's just a quick win to reduce the bundle size

marie-j, line 24: I don't see `angularQr` being declared as needed by the module like it used to be https://github.com/ovh/manager/pull/3996/files#diff-24e0955591a31a9b81232f4bfa927a3d31c3118725d09b1eee1b7d012646e1afL11
this has no impact ?

frenautvh: we used to declare 'ja.qr' because it was required by https://github.com/janantala/angular-qr
BUT
we are importing angular-qr which isn't the same repo, see : https://www.npmjs.com/package/angular-qr

qa.jr is not needed, also angular-qr is now injected dynamically

antleblanc (approved):

PR #4002, opened 2020-11-12
build(*): import d3 dependency if required

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | master
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | 

## Description

problem :
dedicated import d3 but it's not used in the app
cloud-universe-component import d3 but it's defined as **a peer dependency** in package.json

fix :
remove d3 imports from cloud-universe-component as it's a peer dep
import d3 in module pci
remove d3 import from dedicated as it's used only in cuc-consumption & cuc-monitoring directives (which are only used in module pci and cloud app)

### DIFF
--- a/packages/manager/modules/pci/src/index.js
+++ b/packages/manager/modules/pci/src/index.js
@@ -24,6 +24,7 @@ import '@ovh-ux/ui-kit/dist/css/oui.css';
 import 'ovh-ui-kit-bs/dist/css/oui-bs3.css';
 import 'font-awesome/css/font-awesome.css';
 import 'ovh-common-style/dist/ovh-common-style.css';
+import 'd3';
 
 import components from './components';
 import error from './error';
--- a/packages/manager/apps/dedicated/package.json
+++ b/packages/manager/apps/dedicated/package.json
@@ -114,7 +114,6 @@
     "clipboard": "^2.0.4",
     "components-jqueryui": "^1.12.1",
     "core-js": "^3.6.5",
-    "d3": "^5.12.0",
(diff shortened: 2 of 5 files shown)


packages/manager/modules/pci/src/index.js
```diff
@@ -24,6 +24,7 @@ import '@ovh-ux/ui-kit/dist/css/oui.css';
 import 'ovh-ui-kit-bs/dist/css/oui-bs3.css';
 import 'font-awesome/css/font-awesome.css';
 import 'ovh-common-style/dist/ovh-common-style.css';
+import 'd3';
```
radireddy, line 27: Very good initiative 👍 . Instead of importing it in PCI module, can you import only in module/route it is used and lazyload it?

antleblanc: I'm not sure that `d3` dependency is needed for following module `@ovh-ux/manager-pci`.

However the given module `@ovh-ux/ng-ovh-cloud-universe-components` has defined `d3` as a Peer Dependency meaning that it must be available on the parent application.

IMO, wherever `@ovh-ux/ng-ovh-cloud-universe-components` is installed, the parent application need to have `d3` from the dependency list.

frenautvh: i think you're right, some component from cloud-universe-components are only used once, maybe simply move those components we allow removing d3 from manager-dedicated build

closing this one, i'll reopen a new PR

PR #4006, opened 2020-11-12
build(*): dynamic import angular-websocket in dbaas module

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | master
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | 

## Description

dynamically import angular-websocket in dbaas module (as it's only used here)

### DIFF
--- a/packages/manager/modules/dbaas-logs/src/logs/logs.routing.js
+++ b/packages/manager/modules/dbaas-logs/src/logs/logs.routing.js
@@ -8,6 +8,12 @@ export default /* @ngInject */ ($stateProvider) => {
       format: 'json',
     },
     resolve: {
+      angularWebsocket: /* @ngInject */ ($ocLazyLoad) =>
+        import('script-loader!angular-websocket/dist/angular-websocket').then(
+          () => {
+            $ocLazyLoad.inject('angular-websocket');
+          },
+        ),
       me: /* @ngInject */ ($http) =>
         $http
           .get('/me')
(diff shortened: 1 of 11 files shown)


packages/manager/modules/dbaas-logs/src/logs/logs.routing.js
```diff
@@ -8,6 +8,12 @@ export default /* @ngInject */ ($stateProvider) => {
       format: 'json',
     },
     resolve: {
+      angularWebsocket: /* @ngInject */ ($ocLazyLoad) =>
+        import('script-loader!angular-websocket/dist/angular-websocket').then(
```
marie-j, line 12: I'm not sure we should use script-loader in modules or then we should add webpack as a requirement to bundle it

frenautvh: well it's already the case, this module contains script loader : 
https://github.com/ovh/manager/blob/bb6dc129c910ef248e285d0ef3fee9a54edd469e/packages/manager/apps/dbaas-logs/src/app.module.js#L1

marie-j: @frenautvh the input you've provided is from the `dbaas-logs` app and not module

frenautvh: oh my bad it's a fair point then, i guess since the module will be lazy loaded (see another PR)
i'll update this PR to keep angular websocket as a static import in this one

PR #4038, opened 2020-11-17
hotfix(web-email): not possible to create mx plan email account

radireddy:
ref #DTRSD-24203

| Question         | Answer
| ---------------- | ---
| Branch?          | master
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-24203

## Description

Issue: Not possible to create more than one mx plan email accounts. 
Fix: Remove account from new account options once the account is created. If not the same account is used again which breaks the API.

### DIFF
--- a/packages/manager/modules/emailpro/src/account/emailpro-account.controller.js
+++ b/packages/manager/modules/emailpro/src/account/emailpro-account.controller.js
@@ -1,4 +1,5 @@
 import angular from 'angular';
+
 import filter from 'lodash/filter';
 import find from 'lodash/find';
 import set from 'lodash/set';
@@ -254,6 +255,7 @@ export default /* @ngInject */ (
       primaryEmailAddress: newConfiguredAccount,
     })
       .then((account) => {
+        $scope.accountIds = $scope.accountIds.slice(1);
         set(account, 'completeDomain', {
           name: account.domain,
           displayName: punycode.toUnicode(account.domain),
@@ -264,12 +266,15 @@ export default /* @ngInject */ (
           account,
         );
       })
-      .catch((err) =>
+      .catch((err) => {
         $scope.setMessage(
           $translate.instant('emailpro_tab_ACCOUNTS_error_message'),
-          err,
-        ),
-      )
+          {
+            ...err,
+            type: err.type || 'error',
+          },
+        );
+      })
       .finally(() => {
         $scope.loadingNewConfiguredAccount = false;
       });


JeremyDec (approved):

packages/manager/modules/emailpro/src/account/emailpro-account.controller.js
```diff
@@ -254,6 +256,7 @@ export default /* @ngInject */ (
       primaryEmailAddress: newConfiguredAccount,
     })
       .then((account) => {
+        $scope.accountIds = drop($scope.accountIds, 1);
```
cbourgois, line 259: Maybe we can avoid to import `lodash/drop` ?

```suggestion
        $scope.accountIds = $scope.accountIds.slice(1);
```

packages/manager/modules/emailpro/src/account/emailpro-account.controller.js
```diff
@@ -264,12 +267,13 @@ export default /* @ngInject */ (
...
-      .catch((err) =>
+      .catch((err) => {
+        set(err, 'type', err.type || 'error');
         $scope.setMessage(
           $translate.instant('emailpro_tab_ACCOUNTS_error_message'),
           err,
-        ),
-      )
+        );
+      })
```
cbourgois, line 276:
```diff
      .catch((err) => {
-       set(err, 'type', err.type || 'error');
        $scope.setMessage(
          $translate.instant('emailpro_tab_ACCOUNTS_error_message'),	          
+         {
+            ..err,
+            type: err.type || 'error'
+         },
        );
      })
```

cbourgois (approved):

packages/manager/modules/emailpro/src/account/emailpro-account.controller.js
```diff
@@ -254,6 +255,7 @@ export default /* @ngInject */ (
       primaryEmailAddress: newConfiguredAccount,
     })
       .then((account) => {
+        $scope.accountIds = $scope.accountIds.slice(1);
```
marie-j, line 258: are we sure we always want to omit the first element ? shouldn't we check depending on a `domainName` or something else ?

antleblanc: @radireddy Do you have any information to provide regarding marie's comment?
Thank you!

antleblanc: ping @radireddy

radireddy: It is ok to drop the first one after it is used. Whenever mxplan email is created, the id is picked from the pool. Once the id is picked and used, it should not be picked again. which was the original issue.

antleblanc (approved):

PR #4051, opened 2020-11-18
perf: move out non-reused components

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #... 

## Description

Some modules in cloud-universe-components are used only once. Move them out to the corresponding app or module directly. This helps to avoid bundling unnecessary components.

(for example d3 has been removed from cloud-universe-components and won't be bundled anymore on manager dedicated app)

⚠️ billing/consumption section has been removed from cloud application as i suppose it's dead code, i would like some confirmation to ensure that i'm not missing something here

Note regarding adding breaking change or not to the commit :
I think it's debatable since cloud-universe-components is a private package, semver is basically ignored because at each build the last version will be used. Making the commit a breaking change would also major bump some apps & modules which i think is not what we want (and splitting in 2 commits seems difficult since files are moved and not created/deleted). Please share your point of view if you disagree thanks


packages/manager/modules/dbaas-logs/src/index.js
```diff
@@ -8,11 +8,13 @@ import '@ovh-ux/ng-ovh-doc-url';
 import { Environment } from '@ovh-ux/manager-config';
 
 import logs from './logs';
+import cuiDualList from './components/dual-list';
```
marie-j, line 11: I'm wondering if we can replace all those `cui` components by `ovh-ui-kit` ones as a next step

packages/manager/modules/pci/src/components/index.js
```diff
@@ -1,11 +1,21 @@
...
 import runabove from './runabove';
 import writeRightRequired from './writeRightRequired';
 
 const moduleName = 'ovhManagerPciComponents';
 
-angular.module(moduleName, [project, runabove, writeRightRequired]);
+angular.module(moduleName, [
+  cucConsumption,
+  cucCurrency,
+  cucFlavor,
```
marie-j, line 15: If I have a look this is only used for one function of one service across all manager, maybe we can just replace its usage

marie-j (commented):
Regarding commit history, it would be better to have a BREAKING CHANGE commit but I agree that it doesn't make sense for other modules than `cloud-universe-components`

packages/manager/modules/cloud-universe-components/src/index.js
```diff
@@ -1,30 +1,21 @@
...
-import cucMonitoring from './monitoring';
 import cucNavigation from './navigation';
 import cucOrderedHash from './orderedHash';
 import cucPoll from './poll';
 import cucPrice from './price';
 import cucRegion from './region';
 import cucProducts from './products';
-import cucSmoothScrollHere from './smoothScrollHere';
-import cucSpaceMeter from './space-meter';
 import cucSshKeyMin from './sshKeyMin';
```
radireddy, line 19: Looks like there is the same filter in the dedicated app.  We have moved it to dedicated universe components.

https://github.com/ovh/manager/blob/master/packages/manager/apps/dedicated/client/app/account/user/components/filters/sshkeyMin/sshkeyMin.js

packages/manager/modules/pci/src/components/index.js
frenautvh: i just deleted cucFlavor module and copied the "addOverQuotasInfos" where it's used (in kubernetes node-pool nodes add)

packages/manager/modules/cloud-universe-components/src/index.js
frenautvh: yes but it's also used in cloud app, how did you handle the case ?
(https://github.com/ovh/manager/blob/master/packages/manager/apps/cloud/client/app/cloud/project/compute/ssh/cloud-project-compute-ssh.html#L331)

radireddy: I moved it to dedicated universe componets. We need to revisit to consolidate this.

radireddy (dismissed):

marie-j (dismissed):

mohammed-zahaf (approved):

cbourgois (dismissed):

cbourgois:
Hey @frenautvh there is some conflicts (maybe related with #4306)

antleblanc (changes requested):
Well done!

Only noticed two translations files that can be removed.

packages/manager/modules/nasha/src/components/space-meter/translations/Messages_lt_LT.json
```diff
@@ -0,0 +1,19 @@
+{
```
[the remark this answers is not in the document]
frenautvh: thanks for spotting it, i deleted the two unused translation files

antleblanc (approved):

PR #4081, opened 2020-11-24
Replace API calls using user from environment

cbourgois:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/user-env`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix 

## Description

Replace following API calls 
* `/me`
* `/me/certificates` 
* `/me/supportLevel` 
* `/auth/details` 

 in 

* `@ovh-ux/manager-navbar`
* `@ovh-ux/manager-account-sidebar`
* `@ovh-ux/manager-core` 

using `Environment.getUser()` from `@ovh-ux/manager-config`

### DIFF
--- a/packages/manager/modules/account-sidebar/src/controller.js
+++ b/packages/manager/modules/account-sidebar/src/controller.js
@@ -1,20 +1,13 @@
+import { Environment } from '@ovh-ux/manager-config';
 import constants from './constants';
 
 export default class OvhManagerAccountSidebarCtrl {
   /* @ngInject */
-  constructor(
-    $q,
-    $rootScope,
-    $translate,
-    atInternet,
-    OvhApiMe,
-    RedirectionService,
-  ) {
+  constructor($q, $rootScope, $translate, atInternet, RedirectionService) {
     this.$q = $q;
     this.$rootScope = $rootScope;
     this.$translate = $translate;
     this.atInternet = atInternet;
-    this.OvhApiMe = OvhApiMe;
     this.RedirectionService = RedirectionService;
 
     this.$rootScope.$on('ovh::sidebar::toggle', () => {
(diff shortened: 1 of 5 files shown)


packages/manager/modules/account-sidebar/src/controller.js
```diff
@@ -28,15 +21,14 @@ export default class OvhManagerAccountSidebarCtrl {
...
-    return this.$q
-      .when(this.me ? this.me : this.OvhApiMe.v6().get().$promise)
-      .then((me) => {
-        this.me = me;
-        this.hasChatbot = constants.CHATBOT_SUBSIDIARIES.includes(
-          me.ovhSubsidiary,
-        );
-      })
-      .then(() => this.$translate.refresh())
+    if (!this.me) {
```
frenautvh, line 24: i believe you can remove all the ```if(!this.me)```test since you get user from environment now

cbourgois: As it's a component binding (https://github.com/ovh/manager/blob/master/packages/manager/modules/account-sidebar/src/component.js#L9), so I prefer to keep it.
If `me` is defined we use it, else we use the env user.

IMHO, I think we can refactor it later.

ganeshkumar1989 (approved):

marie-j (approved):

antleblanc (approved):
I've noticed that following module `@ovh-ux/manager-account-sidebar` is no more using services from `ovh-api-services` module.

It means that it can also be removed from the peerDependencies list, isn't it?

Do you prefer to handle it on a separate Pull Request?

antleblanc:
> I've noticed that following module `@ovh-ux/manager-account-sidebar` is no more using services from `ovh-api-services` module.
> 
> It means that it can also be removed from the peerDependencies list, isn't it?
> 
> Do you prefer to handle it on a separate Pull Request?

Seen together. Seems linked to #2303. Will be done later.

PR #4093, opened 2020-11-25
feat(dedicated.cloud-connect): display cloud-connect service for canada too

Steffy29:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #UXCT-350

## Description

Display Cloud Connect service into left menu into Canadian Manager too

### DIFF
--- a/packages/manager/modules/server-sidebar/src/sidebar.constants.js
+++ b/packages/manager/modules/server-sidebar/src/sidebar.constants.js
@@ -346,7 +346,6 @@ export const DEDICATED_NETWORK_CONFIG = {
     },
     {
       id: 'cloud_connect',
-      feature: 'cloud-connect',
       loadOnState: 'cloud-connect',
       types: [
         {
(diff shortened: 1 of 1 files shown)


packages/manager/modules/server-sidebar/src/sidebar.constants.js
```diff
@@ -346,20 +346,19 @@ export const DEDICATED_NETWORK_CONFIG = {
     },
     {
       id: 'cloud_connect',
-      feature: 'cloud-connect',
```
marie-j, line 349: I'm not sure we should remove the `feature` rule

Steffy29: If I let it, it doesn't work (I write same than vrack)

cbourgois (changes requested):
Don't remove the feature. If you do that, the Cloud Connect link will be displayed for all Regions (EU/CA/US).

packages/manager/modules/server-sidebar/src/sidebar.constants.js
cbourgois: Yes, we need to change feature rules to open it in CA. vrack doesn't use feature yet, because feature are the new way to enable/disable features based on region/sub

PR #4096, opened 2020-11-25
fix(dashboard.configuration): use current price from service

marie-j:
| Question         | Answer
| ---------------- | ---
| Branch?          |  `develop`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #... 

### DIFF
--- a/packages/manager/modules/vps/src/dashboard/vps-dashboard.routing.js
+++ b/packages/manager/modules/vps/src/dashboard/vps-dashboard.routing.js
@@ -88,14 +88,20 @@ export default /* @ngInject */ ($stateProvider) => {
         VpsService.getTaskInError(serviceName).then((tasks) =>
           head(filter(tasks, { type: 'migrate' })),
         ),
+      service: /* @ngInject */ ($http, serviceInfo) =>
+        $http
+          .get(`/services/${serviceInfo.serviceId}`)
+          .then(({ data }) => data),
 
       configurationTile: /* @ngInject */ (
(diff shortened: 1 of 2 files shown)


packages/manager/modules/vps/src/dashboard/vps-dashboard.routing.js
```diff
@@ -88,14 +88,20 @@ export default /* @ngInject */ ($stateProvider) => {
...
 
       configurationTile: /* @ngInject */ (
         availableUpgrades,
         catalog,
+        service,
         stateVps, // from apiv6
         vps, // from 2api
       ) => ({
         upgrades: VpsConfigurationTile.setVps(vps, stateVps.model)
+          .setCurrentService(service)
```
frenautvh, line 104: it's a better practice to have immutable service but since it's a fix and VpsConfigurationTile already has the mutable setVps method i guess this should be addressed in a refactoring PR

frenautvh (approved):

packages/manager/modules/vps/src/dashboard/vps-dashboard.routing.js
marie-j: @frenautvh I agree this is not a good practice, it was done that way because it was already the case. As the impact is quite huge, more time would be required for a refactor

JeremyDec (approved):

antleblanc (approved):

PR #4105, opened 2020-11-26
build: exclude node_modules from dynamicImport processing

marie-j:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master` for hotfix / `develop` for bug fixes 
| Bug fix?         | yes/no
| New feature?     | yes/no
| Breaking change? | yes/no
| Tickets          | Fix #... 

### DIFF
--- a/packages/manager/tools/component-rollup-config/src/index.ts
+++ b/packages/manager/tools/component-rollup-config/src/index.ts
@@ -20,9 +20,10 @@ import common from './plugins/common';
 
 const defaultName = path.basename(process.cwd());
 
-const mergeConfig = (config, customConfig) => mergeWith(config, customConfig, (obj, src) => (
-  Array.isArray(obj) && Array.isArray(src) ? src.concat(obj) : undefined
-));
+const mergeConfig = (config, customConfig) =>
+  mergeWith(config, customConfig, (obj, src) =>
+    Array.isArray(obj) && Array.isArray(src) ? src.concat(obj) : undefined,
+  );
 
 const getLanguages = (pluginsOpts) => {
   if (isString(process.env.LANGUAGES)) {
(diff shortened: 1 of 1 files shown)


frenautvh (approved):

frenautvh:
i'm curious does it solve some issue ?
also does it mean that we are not allowed to import('my-dep').then(...) ?

antleblanc:
Could you please elaborate if the Pull Request have to target `develop` or `master` branch?

In addition to Florian's comment, I'm also curious about some description regarding this changes.

**nitpick:** `.ts` files are not formatted (yet) via our current ESLint configuration.
This leads to an extra diff that make the original diff more complicated to read.

marie-j:
> i'm curious does it solve some issue ?
> also does it mean that we are not allowed to import('my-dep').then(...) ?

Yes this solves an issue using `@stencil`. Using this, only the things that are imported and used will be processed. 
For example let's say I have two chunk of code `a` and `b` in a dependency, if I only import `a`, only `a` will be processed whereas before `a` and `b` were processed with no reason. It works the same as for babel plugins

frenautvh:
> > i'm curious does it solve some issue ?
> > also does it mean that we are not allowed to import('my-dep').then(...) ?
> 
> Yes this solves an issue using `@stencil`. Using this, only the things that are imported and used will be processed.
> For example let's say I have two chunk of code `a` and `b` in a dependency, if I only import `a`, only `a` will be processed whereas before `a` and `b` were processed with no reason. It works the same as for babel plugins

do you mean that if i dynamically import a dependency and this dependency dynamically import some other dependencies (from node_modules), then only the first imported dependency will be processed ?
if so i think you're right since node_modules deps are supposed to be already bundled

Jisay (approved):

cbourgois (approved):

marie-j:
> > > i'm curious does it solve some issue ?
> > > also does it mean that we are not allowed to import('my-dep').then(...) ?
> > 
> > 
> > Yes this solves an issue using `@stencil`. Using this, only the things that are imported and used will be processed.
> > For example let's say I have two chunk of code `a` and `b` in a dependency, if I only import `a`, only `a` will be processed whereas before `a` and `b` were processed with no reason. It works the same as for babel plugins
> 
> do you mean that if i dynamically import a dependency and this dependency dynamically import some other dependencies (from node_modules), then only the first imported dependency will be processed ?
> if so i think you're right since node_modules deps are supposed to be already bundled

It depends how the dependency is handled, if it's already bundled then the dynamic import have already been replaced , if not it will be processed only if used in the sources

PR #4124, opened 2020-12-02
Configure 'app' for feature flipping

ganeshkumar1989:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-5737

## Description

Set the 'app' for all the main applications, to be used for feature flipping

**Related:**
features-rules: 19
feature-availability: 43

### DIFF
--- a/packages/manager/apps/hub/src/app.module.js
+++ b/packages/manager/apps/hub/src/app.module.js
@@ -12,6 +12,7 @@ import ngUiRouterBreadcrumb from '@ovh-ux/ng-ui-router-breadcrumb';
 import { isString, get, has } from 'lodash-es';
 
 import '@ovh-ux/ui-kit';
+import ngOvhFeatureFlipping from '@ovh-ux/ng-ovh-feature-flipping';
 import ovhManagerAccountSidebar from '@ovh-ux/manager-account-sidebar';
 import ovhManagerCore from '@ovh-ux/manager-core';
 import ovhManagerHub from '@ovh-ux/manager-hub';
@@ -44,6 +45,7 @@ angular
       dashboard,
       errorPage,
       'ngAnimate',
+      ngOvhFeatureFlipping,
       ngOvhSsoAuthModalPlugin,
       ngOvhUiRouterLineProgress,
       ngUiRouterBreadcrumb,
@@ -63,6 +65,11 @@ angular
   .controller('HubController', controller)
   .config(
     /* @ngInject */ ($locationProvider) => $locationProvider.hashPrefix(''),
+  )
+  .config(
+    /* @ngInject */ (ovhFeatureFlippingProvider) => {
+      ovhFeatureFlippingProvider.setApplicationName('hub');
+    },
   )
   .config(routing)
   .run(
--- a/packages/manager/apps/hub/package.json
+++ b/packages/manager/apps/hub/package.json
@@ -35,6 +35,7 @@
     "@ovh-ux/ng-at-internet": "^5.3.0",
     "@ovh-ux/ng-at-internet-ui-router-plugin": "^3.1.0",
     "@ovh-ux/ng-ovh-api-wrappers": "^4.0.8",
+    "@ovh-ux/ng-ovh-feature-flipping": "^1.0.1",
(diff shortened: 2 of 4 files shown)


Jisay (dismissed):

cbourgois (dismissed):

packages/manager/apps/hub/src/app.module.js
```diff
@@ -64,6 +66,11 @@ angular
   .config(
     /* @ngInject */ ($locationProvider) => $locationProvider.hashPrefix(''),
   )
+  .config(
+    /* @ngInject */ (ovhFeatureFlippingProvider) => {
+      ovhFeatureFlippingProvider.setApplicationName('hub');
```
antleblanc, line 71: **question:** Does the application's name should reflect the folder name (here `hub`) instead of `dashboard` for instance?

Same could be applied to following list:
- `dedicated` => `bare-metal-cloud`
- `web` => `web-cloud`

Or it doesn't have sense?

My 2cts, `hub` sounds more like internal codename for a given `application`.

antleblanc:
Regarding code changes, it looks like a new addition to these applications 
so I would prefer rely on a `feat` instead of a `bug` as commit type.


Do you agree?

packages/manager/apps/hub/src/app.module.js
cbourgois: `dedicated` => `bare-metal-cloud` && `hosted-private-cloud` ? and `cloud` is also `bare-metal-cloud`.

IMHO we should use internal codename (`hub`, `dedicated` etc) as it's for "internal" purpose. 
It's an application name, but the concept behind is to have a `uniqid` for each application, no matter where  it's hosted or how it's named for customers (knowing that these two parameters can change).

marie-j (dismissed):

packages/manager/apps/hub/src/app.module.js
ganeshkumar1989: Hello @antleblanc . The app name need not match the folder name. Hence, we can name the apps the way you have mentioned. But we will need to decide on the app name that we would like to use. For example, we have `bare-metal-cloud` and `hosted-private-cloud` under dedicated app. Please let me know your thoughts :)

antleblanc: As the `name` can evolved based on how we want to expose it from a customer point of view, we can stick with folder name as an identifier to the application name as @cbourgois mentioned.

ganeshkumar1989:
> Regarding code changes, it looks like a new addition to these applications
> so I would prefer rely on a `feat` instead of a `bug` as commit type.
> 
> Do you agree?

This change does not produce any change in the behavior of the apps as such. I feel this is more of a feature change for the feature flipping capability, that the Manager apps (where the change is made in this PR). I am not sure whether we can this a 'feat'... (We now just pass the app name to the feature-flipping API call). Please let me know what you think @antleblanc
Comments from the team are welcome :)

antleblanc:
> This change does not produce any change in the behavior of the apps as such. I feel this is more of a feature change for the feature flipping capability, that the Manager apps (where the change is made in this PR).

I'm wondering why you prefer mark it as a bugfix by using `fix` as a commit type.

Is the application is buggy at some point? Are you fixing something here?

In my opinion it looks like you want the application to enable a new mechanism (which is the feature-flipping stuff) that's
why it's sound (to me) more like a enhancement (at least `feat` commit type)

ganeshkumar1989:
> > This change does not produce any change in the behavior of the apps as such. I feel this is more of a feature change for the feature flipping capability, that the Manager apps (where the change is made in this PR).
> 
> I'm wondering why you prefer mark it as a bugfix by using `fix` as a commit type.
> 
> Does the application is buggy at some point? Are you fixing something here?
> 
> In my opinion it looks like you want the application to enable a new mechanism (which is the feature-flipping stuff) that's
> why it's sound (to me) more like a enhancement (at least `feat` commit type)

Got it. This is done @antleblanc . Thanks for your inputs

JeremyDec (approved):

antleblanc (approved):

PR #4133, opened 2020-12-03
fix: Wrong SPF for CA

mohammed-zahaf:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | DTRSD-25117

## Description

Aims of this fix is to set the SPF value to CA or COM value.

### DIFF
--- a/packages/manager/apps/web/client/app/domain/zone/record/domain-zone-record.controller.js
+++ b/packages/manager/apps/web/client/app/domain/zone/record/domain-zone-record.controller.js
@@ -8,6 +8,7 @@ import isEmpty from 'lodash/isEmpty';
 import isString from 'lodash/isString';
 import trim from 'lodash/trim';
 import set from 'lodash/set';
+import { Environment } from '@ovh-ux/manager-config';
 
 angular.module('App').controller(
   'DomainZoneRecordCtrl',
@@ -700,7 +701,7 @@ angular.module('App').controller(
 
     useSpfOvh() {
       this.model.target = {};
-      this.model.target.include = 'mx.ovh.com';
+      this.model.target.include = DomainZoneRecordAddCtrl.getSPFIncludeName();
       this.model.target.all = '~all';
       this.setTargetValue('spf');
       this.checkIfRecordCanBeAdd();
@@ -767,6 +768,10 @@ angular.module('App').controller(
         .finally(() => {
           this.$scope.resetAction();
         });
+    }
+
+    static getSPFIncludeName() {
+      return `mx.ovh.${Environment.getRegion() === 'CA' ? 'ca' : 'com'}`;
     }
   },
 );
--- a/packages/manager/apps/web/client/app/domain/zone/record/SPF/domain-zone-record-SPF.html
+++ b/packages/manager/apps/web/client/app/domain/zone/record/SPF/domain-zone-record-SPF.html
@@ -4,7 +4,7 @@
         data-translate-values="{ t0: ctrl.domain.displayName }"
     ></p>
     <code
-        data-ng-bind="ctrl.domain.displayName + '. IN TXT &quot;v=spf1 include:mx.ovh.com ~all&quot;'"
+        data-ng-bind="ctrl.domain.displayName + '. IN TXT &quot;v=spf1 include:' + ctrl.constructor.getSPFIncludeName() + ' ~all&quot;'"
     ></code>
     <div class="mt-3">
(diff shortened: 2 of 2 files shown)


Heiiji (approved):

packages/manager/apps/web/client/app/domain/zone/record/domain-zone-record.controller.js
```diff
@@ -8,6 +8,8 @@ import isEmpty from 'lodash/isEmpty';
 import isString from 'lodash/isString';
 import trim from 'lodash/trim';
 import set from 'lodash/set';
+import { Environment } from '@ovh-ux/manager-config';
+import { CANADA_REGION } from '../../../../../../../modules/config/src/environment/environment.constants';
```
cbourgois, line 12: Why do you need to import `environment.constants` like this ?
If you really need it, constants should be exported by `@ovh-ux/manager-config` ?

Here `import ... from '@ovh-ux/manager-config';` and `import ... from '../../../../../../../modules/config/src/` target the same package.

packages/manager/apps/web/client/app/domain/zone/record/domain-zone-record.controller.js
```diff
@@ -700,7 +702,9 @@ angular.module('App').controller(
 
     useSpfOvh() {
       this.model.target = {};
-      this.model.target.include = 'mx.ovh.com';
+      this.model.target.include = `mx.ovh.${
+        Environment.getRegion() === CANADA_REGION ? 'ca' : 'com'
+      }`;
```
cbourgois, line 707:
```suggestion
      this.model.target.include = Environment.getRegion() === 'CA' ? 'mx.ovh.ca' : 'mx.ovh.com';
```

As it seems to be an hotfix, I think we can get rid of refactor everything to use a constant and use `'CA'`.
The refacto should be done in another PR.

packages/manager/modules/config/src/environment/environment.constants.js
```diff
@@ -1,5 +1,6 @@
 export const ALLOWED_REGIONS = ['EU', 'CA', 'US'];
 export const DEFAULT_REGION = 'EU';
+export const CANADA_REGION = 'CA';
```
cbourgois, line 3: The constant `CANADA_REGION` is exported but never used in `@ovh-ux/manager-config` package, and it's never exported in the dist.

We don't have `EUROPE_REGION` or `UNITED_STATES_REGION`. If we add them, we should at least export them in the dist (as `REGIONS:{EU, CA, US}` , and refactor every place in our code where `EU` / `CA` / `US` strings are used.

cbourgois (changes requested):

packages/manager/apps/web/client/app/domain/zone/record/domain-zone-record.controller.js
mohammed-zahaf: I had try create a static method `isCanadaRegion()`.
When i try to call it like so `Environment.isCanadaRegion()` i got error.
Now i understand the mistake the `Environment` is already an instance object not a Class.

So i don't need anymore the `CANADA_REGION` import.

packages/manager/apps/web/client/app/domain/zone/record/domain-zone-record.controller.js
```diff
@@ -768,5 +769,13 @@ angular.module('App').controller(
...
         });
     }
+
+    static getSPFIncludeName() {
+      return `mx.ovh.${
+        Environment.constructor.isCanadaRegion(Environment.getRegion())
+          ? 'ca'
+          : 'com'
+      }`;
+    }
```
frenautvh, line 779:
```js
return `mx.ovh.${Environment.getRegion() === 'ca' ? 'ca' : 'com'}`
```
just do that, no need to change Environement

mohammed-zahaf: Ok changed

packages/manager/apps/web/client/app/domain/zone/record/domain-zone-record.controller.js
mohammed-zahaf: Ok got it.

frenautvh (approved):

antleblanc (approved):

PR #3993, opened 2020-11-10
Application manifest

cbourgois:
| Question         | Answer
| ---------------- | ---
| Branch?          | `develop`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | yes
| Tickets          | Fix MANAGER-5859
| License          | BSD 3-Clause

## Description

Add `@ovh-ux/manager-manifest-generator`: 

```sh
$ yarn run manager-manifest-generator --help
Usage: manager-manifest-generator [options] <basePath>

Options:
  -V, --version                output the version number
  --application <application>  application name (default: "application")
  -f, --file <filename>        manifest file name (default: "manifest.json")
  --verbose                    output extra debugging
  -h, --help                   display help for command
```

- Add manifest generation in `hub` application

## Related 
- [ ] https://github.com/ovh/manager/pull/4005 — `url-resolver`

## TODO
- [ ] add manifest generation in all apps (if the review is ok )

PR #4156, opened 2020-12-08
Feat/apps navigation

cbourgois:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix MANAGER-5847

## Description

* Add `applicationURLs` in `Environment` (require `/configuration#7`)
* add `url-builder` in `@ovh-ux/ufrontend`

Exemple on how to build an url to an external application: 

```js 
import { buildURL } from '@ovh-ux/ufrontend/url-builder';

const url = buildURL('hub', '#/catalog', { q: 'domaine' });
console.log(url);
// https://www.ovh.com/manager/#/catalog?q=domaine
```

Replace #3993

### DIFF
--- a/packages/manager/modules/config/src/environment/environment.service.js
+++ b/packages/manager/modules/config/src/environment/environment.service.js
@@ -11,6 +11,7 @@ export default class EnvironmentService {
     this.userLocale = findAvailableLocale(detectUserLocale(), this.region);
     this.version = null;
     this.user = {};
+    this.applicationURLs = {};
   }
 
   setRegion(region = DEFAULT_REGION) {
@@ -56,5 +57,17 @@ export default class EnvironmentService {
 
   getVersion() {
     return this.version;
+  }
+
+  setApplicationURLs(applicationURLs) {
+    this.applicationURLs = applicationURLs;
+  }
+
+  getApplicationURLs() {
+    return this.applicationURLs;
+  }
+
+  getApplicationURL(id) {
+    return this.applicationURLs[id];
   }
 }
--- a/packages/manager/modules/ufrontend/README.md
+++ b/packages/manager/modules/ufrontend/README.md
@@ -114,3 +114,43 @@ emit(
   },
 );
 ```
+
+### URL Builder
+
+URL builder is here to help us to build links between applications.
+When you want to build an URL for an outside application, you will use `url-builder` to build the URL.
+
+#### Examples
+
+Build the URL that redirect to `#/catalog` state from the `hub` application with query parameter (`expand`)
+
+```js
+import { buildURL } from '@ovh-ux/ufrontend/url-builder';
+
+const url = buildURL('hub', '#/catalog', { expand: true });
+// use `url`;
+```
+
+Build multiples routes using an array of `{application, url, query}`
+
+```js
+import { buildURLs } from '@ovh-ux/ufrontend/url-builder';
+
+const [dashboard, catalog] = buildURLs([
+  { application: 'hub', url: '#/', query: { expand: true } },
+  { application: 'hub', url: '#/catalog' },
+]);
+// use `dashboard` and `catalog` URLs
+```
+
+Build multiples routes using an object of `{application, url, query}`
+
+```js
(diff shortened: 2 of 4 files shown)


packages/manager/modules/config/src/environment/environment.service.js
```diff
@@ -11,6 +11,7 @@ export default class EnvironmentService {
     this.userLocale = findAvailableLocale(detectUserLocale(), this.region);
     this.version = null;
     this.user = {};
+    this.baseUrls = {};
```
frenautvh, line 14: maybe ```applicationURLs``` would describe the attribute better, what do you think?

frenautvh (changes requested):
i think we need to talk more about it, i'm not sure that we should base routing on state identifiers because it does make sense only in the context of some routers (ui router for example)

take a look at https://reactrouter.com/web/guides/quick-start the routing is resolved using URLs, the concept of state id is not necessary. So in your example you would do '#/goto/app/dashboard' and perform an un-necessary redirection in the react app and it would just be simpler to just do appURL + /app/dashboard

in my opinion the inter-app routing should only rely on crafting URLs, there is no concept of state

packages/manager/modules/config/src/environment/environment.service.js
cbourgois: Make sense, it's applications bases urls.

cbourgois:
> i think we need to talk more about it, i'm not sure that we should base routing on state identifiers because it does make sense only in the context of some routers (ui router for example)
> 
> take a look at https://reactrouter.com/web/guides/quick-start the routing is resolved using URLs, the concept of state id is not necessary. So in your example you would do '#/goto/app/dashboard' and perform an un-necessary redirection in the react app and it would just be simpler to just do appURL + /app/dashboard
> 
> in my opinion the inter-app routing should only rely on crafting URLs, there is no concept of state

In fact two way to think about it : 
- state doesn't change and url will change (using goto)
- state will change but urls will remain the same (using baseUrl + urlPart as constant)

So we have to choose one way (both are good but we need to keep one ;) )

If we use urlPart, `buildUrl` will be simplier : 

```js
const buildUrl = (app, url) => {
  return `${Environment.getBaseUrl(app)}${url}`;
};
```

cbourgois:
Ok, `state name` could change and `URL` too, so we will use `URL`.
If URL changes the application that handle should provide a redirection (a customer who has a bookmark on old URL should be redirected to the new one)

Build URL should be simpler : 
```js
const buildUrl = (application, url) => `${Environment.getApplicationURL(application)}${url}`
```

packages/manager/modules/ufrontend/url-builder.js
```diff
@@ -0,0 +1,36 @@
+import { Environment } from '@ovh-ux/manager-config';
+
+const buildQueryString = (data) =>
+  Object.keys(data)
+    .map((key) => `${encodeURIComponent(key)}=${encodeURIComponent(data[key])}`)
+    .join('&');
+
+export const buildUrl = (application, url, query) => {
```
frenautvh, line 8: it's really nitpicking but buildURL would be more consistent (encodeURIComponent for example)

frenautvh (dismissed):

packages/manager/modules/ufrontend/README.md
```diff
@@ -114,3 +114,43 @@ emit(
...
 ```
+
+### URL Builder
+
+URL builder is here to help us to build links between applications.
+When you want to build an URL for an outside application, you will use `url-builder` to build the URL.
+
+#### Examples
+
+Build the url that redirect to `#/catalog` state from the `hub` application with query parameter (`expand`)
```
antleblanc, line 125:
```suggestion
Build the URL that redirect to `#/catalog` state from the `hub` application with query parameter (`expand`)
```

antleblanc (approved):

marie-j (approved):

Jisay (dismissed):

antleblanc (approved):

PR #4159, opened 2020-12-09
fix(alias): sort repayment by date on init

JeremyDec:
ref: DTRSD-25868

| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-25868

### DIFF
--- a/packages/manager/apps/telecom/src/app/telecom/telephony/alias/special/repayments/repayments.controller.js
+++ b/packages/manager/apps/telecom/src/app/telecom/telephony/alias/special/repayments/repayments.controller.js
@@ -32,7 +32,15 @@ export default class TelecomTelephonyAliasSpecialRepaymentsCtrl {
       })
       .$promise.then((repayments) => {
         this.allRepayments = repayments;
-        this.repayments = repayments.slice(0, LIMIT);
+        this.repayments = repayments
+          .slice(0, LIMIT)
+          .sort(({ creationDatetime: dateA }, { creationDatetime: dateB }) => {
+            if (new Date(dateA) < new Date(dateB)) {
+              return 1;
+            }
+
+            return new Date(dateA) > new Date(dateB) ? -1 : 0;
+          });
 
         let otherRepayments = [];
         [this.repaymentsInfos.feesToPay, otherRepayments] = partition(


packages/manager/apps/telecom/src/app/telecom/telephony/alias/special/repayments/repayments.controller.js
```diff
@@ -32,7 +32,15 @@ export default class TelecomTelephonyAliasSpecialRepaymentsCtrl {
       })
       .$promise.then((repayments) => {
         this.allRepayments = repayments;
-        this.repayments = repayments.slice(0, LIMIT);
+        this.repayments = repayments
+          .slice(0, LIMIT)
+          .sort(({ creationDatetime: dateA }, { creationDatetime: dateB }) => {
```
frenautvh, line 37: is it working? don't you have to use moment(date).isAfter()?

JeremyDec: yes, the dates have the following format 'YYYY-MM-DDTHH:MM:SS'

JeremyDec: In fact no you're right, some comparison won't work that way for all cases. It could have been possible with 'YYYY-MM_DD' format

JeremyDec: Sorry for confusion, in the manager, the format is like 'DD-MM-YYYY', but API returns 'YYYY-MM-DD'. So that's why it works.
But I can still secure the test

Heiiji (approved):

packages/manager/apps/telecom/src/app/telecom/telephony/alias/special/repayments/repayments.controller.js
antleblanc: I'm wondering if the sorting couldn't be instead apply directly into the 2API endpoint.
Is that make sense for you?

Jisay (approved):

packages/manager/apps/telecom/src/app/telecom/telephony/alias/special/repayments/repayments.controller.js
```diff
@@ -32,7 +32,15 @@ export default class TelecomTelephonyAliasSpecialRepaymentsCtrl {
...
+            if (new Date(dateA) < new Date(dateB)) {
```
radireddy, line 38: are you sure this works? when I run new Date() == new Date() or new Date() > new Date() or new Date() < new Date(), I always get false.

JeremyDec: Yes it is working, as explained to @frenautvh , the format 'YYYY-MM-DDTHH:MM:SS' make it working fine. Certainly date comparison is relying on this format

packages/manager/apps/telecom/src/app/telecom/telephony/alias/special/repayments/repayments.controller.js
JeremyDec: If we consider that these dates should always be sorted this way, yes it could be directly done in the 2api endpoint

packages/manager/apps/telecom/src/app/telecom/telephony/alias/special/repayments/repayments.controller.js
radireddy: new Date('2020-12-24T10:30:00') === new Date('2020-12-24T10:30:00')
The above statement returns false which is not true.

JeremyDec: You're right, and here's why:

`new Date() === new Date()` won't work as `new Date()` creates object. Object comparison, even if they have the same values, will always be falsy.

But instead, Date comparison with `<|>|<=|>=` (also with this format of date), will work, because they use .valueOf() under the hood.

JeremyDec:
Closing PR has this logic is moved to 2API endpoint

PR #3992, opened 2020-11-10
Feat/bm commitment

marie-j:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | yes
| Tickets          | MANAGER-5701
| License          | BSD 3-Clause

## Description

- [x] https://github.com/ovh/manager/pull/3858
- [x] https://github.com/ovh/manager/pull/3977

## Related
- [x] 2API billing-autorenew-services#46
- [x] 2API sws-dedicated-server#13

PR #4189, opened 2020-12-17
Feat/bm resiliate

JeremyDec:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/bm-commitment`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-2379

## Description

Add new resiliate action for commitment baremetal servers.

Translation request: 
- [x]  TRANS-31203


packages/manager/apps/dedicated/client/app/billing/autoRenew/cancel-resiliation/cancel-resiliation.routing.js
```diff
@@ -0,0 +1,62 @@
...
+        if (engagement && hasEndRuleStrategies) {
+          return setReactivateEngagementStrategy();
+        }
+
+        service.cancelResiliation();
+        return BillingAutoRenew.updateService(service);
+      },
+      engagement: /* @ngInject */ (Server, service) =>
+        (service.canHaveEngagement()
+          ? Server.getSelected(service.serviceId)
```
marie-j, line 28: I wonder if this condition has changed and other services can have an engagement

packages/manager/apps/dedicated/client/app/billing/autoRenew/resiliation/resiliation.routing.js
```diff
@@ -0,0 +1,38 @@
+import { BillingService as Service } from '@ovh-ux/manager-models';
+
+export default /* @ngInject */ ($stateProvider) => {
+  $stateProvider.state('app.account.billing.autorenew.resiliation', {
+    url: '/resiliation?serviceId&serviceName',
+    component: 'ovhManagerBillingResiliation',
+    translations: { value: ['.'], format: 'json' },
```
marie-j, line 7: Maybe we could use an annotation instead
```suggestion
.run(/* @ngTranslationsInject:json ./translations */
```

packages/manager/apps/dedicated/client/app/dedicated/server/dashboard/resiliation/resiliation.routing.js
```diff
@@ -0,0 +1,53 @@
...
+        goBack(successMessage),
+      serviceId: /* @ngInject */ (serviceInfos) => serviceInfos.serviceId,
+      serviceName: /* @ngInject */ ($transition$) =>
+        $transition$.params().productId,
+      service: /* @ngInject */ ($http, serviceId) =>
+        $http
+          .get(`/services/${serviceId}`)
+          .then(({ data }) => new Service(data)),
+      validate: /* @ngInject */ (BillingService, serviceId) => (strategy) =>
+        BillingService.putEndRuleStrategy(serviceId, strategy),
```
marie-j, line 50: If the API calls are always the same and always will be, maybe the component should handle it instead. If not we'll need to duplicate every time we want to use the resiliation component

packages/manager/modules/billing/src/components/resiliation/resiliation.controller.js
```diff
@@ -0,0 +1,26 @@
+export default class BillingResiliationController {
+  /* @ngInject */
+  constructor($translate, $timeout) {
+    this.$translate = $translate;
+    this.$timeout = $timeout;
```
marie-j, line 5:
```suggestion

```
doesn't seem used

packages/manager/modules/billing/src/components/services-actions/services-actions.controller.js
```diff
@@ -46,7 +49,10 @@ export default class ServicesActionsCtrl {
         this.renewLink = `${this.CORE_MANAGER_URLS.telecom}sms/${this.service.serviceId}/options/recredit`;
         break;
       default:
-        this.resiliateLink = `${this.autorenewLink}/delete?serviceId=${this.service.serviceId}${serviceTypeParam}`;
+        this.resiliateLink = this.service.engagementDetails
+          ? (this.getResiliationLink && this.getResiliationLink()) ||
```
marie-j, line 53: Maybe we could set a `resiliationLink` variable to ease maintenant in reading this condition

packages/manager/modules/billing/src/components/services-actions/services-actions.html
```diff
@@ -100,7 +100,12 @@
         data-ng-if="(!$ctrl.service.shouldDeleteAtExpiration() || !$ctrl.service.isResiliated()) && !$ctrl.service.hasDebt() && !$ctrl.service.hasPendingResiliation()"
     >
         <oui-action-menu-item
-            ng-if="$ctrl.service.canBeResiliated($ctrl.user.nichandle)"
+            ng-if="$ctrl.service.canBeResiliated($ctrl.user.nichandle) && !$ctrl.service.engagementDetails && !$ctrl.service.canBeEngaged
+            || (
+              $ctrl.service.engagementDetails
+              && $ctrl.service.engagementDetails.endRule.strategy === 'REACTIVATE_ENGAGEMENT'
+              && $ctrl.service.engagementDetails.endRule.possibleStrategies.length > 0
```
marie-j, line 107: `service` has a method `isEngaged` that we could reuse. Also we could maybe abstract those conditions by having method in the class used. For instance `canReacrivateEngagement()` or `canResiliateEngagement`

packages/manager/modules/billing/src/components/utils/billing.service.js
```diff
@@ -36,9 +36,21 @@ export default class BillingService {
       .then((commitments) => commitments.map((c) => new Commitment(c)));
   }
 
+  getEngagementDetails(serviceId) {
+    return this.$http
+      .get(`/services/${serviceId}/billing/engagement`)
+      .then(({ data }) => data);
+  }
+
```
marie-j, line 44: I think there are info about engagement in the `/services` API maybe we might not need another call

packages/manager/modules/billing/src/components/services-actions/services-actions.html
JeremyDec: Here `service` is not a instance of `Service` class, but is one of `BillingService` class.
So this won't be able to be used. But method could be done in the component controller, which is not very good.
What do you think ?

marie-j: Is there any issue adding those to the `BillingService` class then ?

JeremyDec: Hmm I find it confusing that `BillingService`class handle these information as they are not part of what is returned by `/serviceInfos`.

marie-j: Why add it here then https://github.com/ovh/manager/pull/4189/files#diff-eb74d56185f345b0fdc694c3333879f25a6725b2958587f4da1e77543edb3e43R76 ? 😅 
Here it works fine because we don't destructure the properties in the constructor, if it's confusing then we should find another way to pass this information than through the created instance

JeremyDec: TBH, it is more a "copy" of what you've done here: https://github.com/ovh/manager/pull/3992/files#diff-eb74d56185f345b0fdc694c3333879f25a6725b2958587f4da1e77543edb3e43R63-R64. 

I agree this is not the best though.

marie-j:
Rebase due to updates on base branch

PR #4181, opened 2020-12-16
refactor(link): use buildURL instead of redirectionService

cbourgois:
| Question         | Answer
| ---------------- | ---
| Branch?          | `develop`
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix 
| License          | BSD 3-Clause

## Description

Use `buildURL` from `@ovh-ux/ufrontend/url-builder` instead of `RedirectionService`.

`RedirectionService` store URL hash behind identifier (`autorenew`, `contactManagement`...) but URL hash are already identifiers.

I keep `RedirectionService` for now for external URLs (guides/help/tasks).

PR #4193, opened 2020-12-21
Refactor/core urls

cbourgois:
| Question         | Answer
| ---------------- | ---
| Branch?          | `refactor/redirection-links`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | yes
| Tickets          | Fix 

## Description

Remove `CORE_MANAGER_URLS` and `CORE_REDIRECT_URLS` constants and replace usages by `buildURL` 

Related to #4181


packages/manager/apps/dedicated/client/app/components/services/services.constants.js
```diff
@@ -1,138 +1,140 @@
-angular.module('App').constant('SERVICES_TARGET_URLS', {
+export const SERVICES_TARGET_URLS = {
```
marie-j, line 1: Depending on the use we might be able to replace those with the configuration exposed in a 2APU

marie-j (approved):
🧹

packages/manager/apps/dedicated/client/app/components/services/services.constants.js
cbourgois: `SERVICES_TARGET_URLS` is only used in `packages/manager/apps/dedicated/client/app/components/services/services.service.js` and we dont update the key, but yes, the association between api routes/universe should be updated.

In theses PR I don't define specific app and use existing and prodded apps only (`dedicated`/`public-cloud`/`web`...)

packages/manager/modules/core/src/manager-core.constants.js
```diff
@@ -1,83 +1,3 @@
...
-    'https://www.ovh.com/manager/dedicated/index.html#/useraccount/emails',
-  userSubscriptions:
-    'https://www.ovh.com/manager/dedicated/index.html#/useraccount/subscriptions',
-  userSSH: 'https://www.ovh.com/manager/dedicated/index.html#/useraccount/ssh',
-  userAdvanced:
-    'https://www.ovh.com/manager/dedicated/index.html#/useraccount/advanced',
-  contacts:
-    'https://www.ovh.com/manager/dedicated/index.html#/useraccount/contacts?tab=SERVICES&serviceName={serviceName}',
-  horizon:
-    'https://horizon.cloud.ovh.net/openstackdashboard?username={username}',
```
antleblanc, line 45: I'm wondering if removing this particular URL still work as expected!

cbourgois: Thanks to report it! 
This one is not used. There is some other `REDIRECT_URLS` constants (with `horizon` inside) and it's done in another PR: https://github.com/ovh/manager/pull/4197/files#diff-206c4f156a8a305eec45c2b281664dc31fa64174e988507b1baac8b57c3e6038R1-R9

antleblanc (approved):

PR #4227, opened 2021-01-03
feat(vps-advice-upsell): add upsell advice for vps

varun257:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/upsell-crossell`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-6072

## Description

add vps upsell crossell advice to vps
add component dependency in vps and dedicated


packages/manager/modules/vps/src/dashboard/vps-dashboard.routing.js
```diff
@@ -118,6 +118,9 @@ export default /* @ngInject */ ($stateProvider) => {
           upgradeType,
         }),
 
+      adviceGroup: /* @ngInject */ (serviceName, VpsService) =>
```
radireddy, line 121: don't load advices in resolve, it blocks page load. We don't want to block page load while loading advice. Advice must be loaded lazily. Use a separate advice component and load devices in a controller. Refer to https://github.com/ovh/manager/pull/4218/files

radireddy (changes requested):

packages/manager/modules/vps/src/dashboard/advices/controller.js
```diff
@@ -0,0 +1,31 @@
...
+        this.adviceGroups = adviceGroups;
+      })
+      .finally(() => {
+        this.loading = false;
+      });
+  }
+
+  loadAdvices() {
+    return this.$http
+      .get(`/advices/vps/${this.serviceName}`, {
```
cbourgois, line 23: Maybe we could have a common `ovh-api-advices` (or another name) that take two parameters : `serviceType` and `serviceName` and call the API `/advices/${this.serviceType}/${this.serviceName}` and use `ovh-advices` to display results ?

frenautvh (approved):

radireddy (approved):

packages/manager/modules/vps/src/dashboard/advices/controller.js
radireddy: I have moved logic of fetching all advices to module. All service has to do is to use the component and pass service name and service type.

PR #4284, opened 2021-01-14
fix: use right label for primary action

antleblanc:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-6233

### :bug: Bug Fixes

db325f0 - fix: use right label for primary action

### DIFF
--- a/packages/manager/modules/cda/src/ip/delete/cda-ip-delete.html
+++ b/packages/manager/modules/cda/src/ip/delete/cda-ip-delete.html
@@ -5,7 +5,7 @@
 >
     <oui-modal
         heading="{{ ::'cda_ip_delete_title' | translate }}"
-        primary-label="{{ ::'delete' | translate }}"
+        primary-label="{{ ::'cda_common_delete' | translate }}"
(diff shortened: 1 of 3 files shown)


lizardK (approved):

packages/manager/modules/cda/src/ip/delete/cda-ip-delete.html
```diff
@@ -5,7 +5,7 @@
 >
     <oui-modal
         heading="{{ ::'cda_ip_delete_title' | translate }}"
-        primary-label="{{ ::'delete' | translate }}"
```
antleblanc, line 8: Following translation key is not available into the `ovhManagerCdaIp` module.
Here as a hotfix, I'm using a translation available from the same given module.

Maybe we can consider to rename this translation key in order to avoid using some `common` translations.
But this can be done in a separate Pull Request if needed.

marie-j (approved):

PR #4299, opened 2021-01-18
fix(server): show bandwidth plans irrespective of capacity

radireddy:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-6230

## Description

Manager filters bandwidth plans to remove plans with current bandwidth capacity. This is wrong because there can be different bandwidth plans with the same bandwidth capacity. For example, there are two plans, guaranteed and ultimate, with 1Gbps capacity. Bandwidth plans can not be filtered only based on current capacity. As of now, there is no API to know the current bandwidth plan to remove it from the plan's list. There is a change request on PU to enable this. For now, its better to show all plans.

Required for up-sell cross-sell

### DIFF
--- a/packages/manager/apps/dedicated/client/app/dedicated/server/server.service.js
+++ b/packages/manager/apps/dedicated/client/app/dedicated/server/server.service.js
@@ -1,4 +1,3 @@
-import assign from 'lodash/assign';
 import camelCase from 'lodash/camelCase';
 import compact from 'lodash/compact';
 import filter from 'lodash/filter';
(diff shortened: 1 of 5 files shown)


ganeshkumar1989 (dismissed):

packages/manager/apps/dedicated/client/app/dedicated/server/server.service.js
```diff
@@ -1505,10 +1505,7 @@ export default class ServerF {
           head(filter(plan.productName.split('-'), (ele) => /^\d+$/.test(ele))),
         );
         assign(plan, { bandwidth });
-
-        if (bandwidth !== existingBandwidth) {
-          return plan;
-        }
+        return plan;
```
marie-j, line 1508:
```suggestion
        return {
          ...plan,
          bandwidth
        }
```
that way we can  remove call to `_.assign`

antleblanc (approved):

PR #4327, opened 2021-01-22
perf: avoid unecessary ssoAuth login call

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | 

## Description

login is already done in manager-core here : https://github.com/ovh/manager/blob/master/packages/manager/modules/core/src/index.js#L148

avoids making unecessary calls to /me


packages/manager/apps/web/client/app/app.js
```diff
@@ -440,21 +440,6 @@ angular
       );
     },
   ])
-  .run([
```
frenautvh, line 443: i think this part was buggy, calling login() multiple time for each URLS is wrong and useless

antleblanc (approved):

PR #4355, opened 2021-01-27
feat(hub): fix cloud connect issues

varun257:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-22063

## Description

add translations and region information

### DIFF
--- a/packages/manager/apps/hub/src/dashboard/products-list/cloud-connect.routing.js
+++ b/packages/manager/apps/hub/src/dashboard/products-list/cloud-connect.routing.js
@@ -1,0 +1,53 @@
+import { compact, get, map, set } from 'lodash-es';
+
+import { ListLayoutHelper } from '@ovh-ux/manager-ng-layout-helpers';
+import { component, resolves } from './config';
+import { POP_MAP } from './products-list.constants';
+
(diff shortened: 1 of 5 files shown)


packages/manager/apps/hub/src/dashboard/products-list/cloud-connect.routing.js
```diff
@@ -0,0 +1,53 @@
+import { compact, get, map, set } from 'lodash-es';
+
+import { ListLayoutHelper } from '@ovh-ux/manager-ng-layout-helpers';
+import { component, resolves } from './config';
+import { POP_MAP } from './products-list.constants';
+
+export default /* @ngInject */ ($stateProvider) => {
+  $stateProvider.state('app.dashboard.ovh_cloud_connect', {
```
marie-j, line 8: I'm not sure we should add a state for this as with solarsteein we will need to duplicate this configuration, maybe instead we should find a way to fetch/provide a configuration depending on the product that we can pass (and avoid duplicating states) because any change we will have to make to this configuration we will have to replicate it elsewhere

PR #4383, opened 2021-02-03
fix(overthebox.remote): update poller when tab switching before result

Steffy29:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #UXCT-369

## Description

When tab switching before result for the poller call, not display error message which is not necessary here.

### DIFF
--- a/packages/manager/modules/overthebox/src/overthebox/remote/overTheBox-remote.controller.js
+++ b/packages/manager/modules/overthebox/src/overthebox/remote/overTheBox-remote.controller.js
@@ -173,8 +173,7 @@ export default class OverTheBoxRemoteCtrl {
    * @param {Date} when Date to compare
    * @return {Boolean}
    */
-  // eslint-disable-next-line class-methods-use-this
-  isInFuture(when) {
+  static isInFuture(when) {
     return !when || moment(when).isAfter(new Date());
   }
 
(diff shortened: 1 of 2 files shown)


antleblanc (approved):

packages/manager/modules/overthebox/src/overthebox/remote/overTheBox-remote.controller.js
```diff
@@ -86,7 +86,9 @@ export default class OverTheBoxRemoteCtrl {
           (remotes) => {
             this.remotes = remotes;
           },
-          this.TucToastError,
+          () => {
```
antleblanc, line 89: This part looks like a `fix` rather than a `feat`.

packages/manager/modules/overthebox/src/overthebox/remote/overTheBox-remote.controller.js
```diff
@@ -171,8 +173,7 @@ export default class OverTheBoxRemoteCtrl {
    * @param {Date} when Date to compare
    * @return {Boolean}
    */
-  // eslint-disable-next-line class-methods-use-this
-  isInFuture(when) {
+  static isInFuture(when) {
```
antleblanc, line 176: Regarding the rest of the diff, it looks like a `refactor` because class methods
has been just turned into static ones.

antleblanc:
History has been updated.

marie-j (approved):

PR #4388, opened 2021-02-04
Feature/poc vue3

zakariasahmane:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | yes

## Description
Reproduce hub in Vue3. This is a draft PR, there are still things to do, just opened for early feedback.
Still missing some translations, few components and some api configuration.


packages/manager/apps/hub-next/package.json
```diff
@@ -0,0 +1,47 @@
...
+  "private": true,
+  "scripts": {
+    "build": "vue-cli-service build",
+    "lint": "vue-cli-service lint",
+    "serve": "vue-cli-service serve",
+    "test:unit": "vue-cli-service test:unit"
+  },
+  "dependencies": {
+    "@ovh-ux/manager-config": "^3.0.1",
+    "@ovh-ux/manager-core": "^12.0.0",
```
marie-j, line 13: `@ovh-ux/manager-core` requires a peerDep to `angular: ^1.7.x`, why do we need it here ?

packages/manager/apps/hub-next/package.json
```diff
@@ -0,0 +1,47 @@
...
+    "@ovh-ux/manager-webpack-dev-server": "^2.8.1",
+    "@ovh-ux/ui-kit": "^4.4.3",
```
marie-j, line 15: Need the package for styles ?

packages/manager/apps/hub-next/src/i18n.js
```diff
@@ -0,0 +1,44 @@
+import { nextTick } from 'vue';
+
+/* eslint-disable no-param-reassign */
+export function setI18nLanguage(i18n, locale) {
+  if (i18n.mode === 'legacy') {
+    i18n.global.locale = locale;
+  } else {
+    i18n.global.locale.value = locale;
+  }
```
marie-j, line 9: Is this some i18n library recommandation ?

packages/manager/apps/hub-next/src/main.ts
```diff
@@ -0,0 +1,37 @@
...
+
+// css
+import 'bootstrap/dist/css/bootstrap.min.css';
+import '@ovh-ux/ui-kit/dist/css/oui.css';
+
+import App from './App.vue';
+import router from './router';
+import store from './store';
+
+fetchConfiguration('hub-next').then(async () => {
```
marie-j, line 15: Not sure we'll be able to have the correct app returned with this, but not sure about conflicts if we register `hub` twice

packages/manager/apps/hub-next/package.json
zakariasahmane: We don't, i imported it to use "bootstrapApplication" but we don't need it anymore, i'll remove it, thanks !

packages/manager/apps/hub-next/package.json
zakariasahmane: Yes !

packages/manager/apps/hub-next/src/views/Activity.vue
```diff
@@ -0,0 +1,294 @@
...
+              : this.t('hub_support_account_management'),
+          },
+        ],
+        ticket.subject,
+        this.t(`hub_support_state_${ticket.state}`),
+        [
+          {
+            tag: 'a',
+            attrs: {
+              href: `https://www.ovh.com/manager/dedicated/#/support/tickets/${ticket.ticketId}`,
```
marie-j, line 112: We might need to handle the different regions here

packages/manager/apps/hub-next/src/views/Activity.vue
```diff
@@ -0,0 +1,294 @@
...
+          { tag: 'div', value: this.t(`manager_hub_products_${billingService.serviceType}`) },
+        ],
+      ]);
+    },
+  },
+});
+</script>
+
+<style lang="scss" scoped>
+@import '@ovh-ux/manager-hub/src/variables.scss';
```
marie-j, line 132: Not declared as a dependency

packages/manager/apps/hub-next/src/i18n.js
zakariasahmane: Yes, just picked up the code they're recommending from their docs :  https://vue-i18n.intlify.dev/guide/advanced/lazy.html

packages/manager/apps/hub-next/src/App.vue
```diff
@@ -0,0 +1,24 @@
+<template>
+  <div class="container-fluid hub-main-view_container">
+    <div id="nav">
+      <router-link to="/">Home</router-link> |
+      <router-link to="/product-details">About</router-link>
```
marie-j, line 5: Some products have specific needs due to the way the API is built and need unfortunately some custom route declaration

zakariasahmane: If you're refering to the product-details route yep, the work to get the custom API is done under ProductsList.vue . 
I made this to be able to go to the route when i had static data to be able to test how the route is behaving but this whole section should definetly be deleted later and replaced with a proper breadcrumbs.

packages/manager/apps/hub-next/src/main.ts
cbourgois: you could call it with `fetchConfiguration('hub')`. This application stands for the `hub` application.

packages/manager/apps/hub-next/src/i18n.js
cbourgois: > If you specified `legacy: true` option in `createI18n`, return `legacy`, else `composition`
https://vue-i18n.intlify.dev/api/general.html#mode

> In Vue I18n v9 and later, the API offered by Vue I18n v8.x is called Legacy API mode.

It could impact the way you access the VueI18n instance (example: https://vue-i18n.intlify.dev/api/general.html#global)

Here, we are using composition mode, so I think we could remove the condition.

packages/manager/apps/hub-next/src/main.ts
```diff
@@ -0,0 +1,44 @@
...
+  attach(Environment.getUserLanguage());
+
+  const i18n = createI18n({
+    locale,
+    fallbackLocale,
+    datetimeFormats,
+  });
+
+  router.beforeEach(async (to, from, next) => {
+    const translationFolders = to.meta.relatedTranslations;
```
Jisay, line 33: If I understand well, here you use the Router to load translations files. Should it be possible to use the same way as we do now in manager? I mean loading asynchronously translations with their corresponding components? Like so, translations are not linked to a route but to a component.

zakariasahmane: Theoretically, it should be possible though i never tried it. I'll try it and see how it goes. It could be better tbh
I'll keep u updated

Jisay: Cool! 
In fact, this implementation is the way we did before. It's not a bad solution but, load the translations asynchronously allowed us to re-use our components easily.

PR #4396, opened 2021-02-05
fix(pack.migration): update list for services to delete

Steffy29:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #UXCT-372

## Description

Update list for services to delete with the correct number of services to delete

### DIFF
--- a/packages/manager/apps/telecom/src/app/telecom/pack/migration/offers/pack-migration-offers.controller.js
+++ b/packages/manager/apps/telecom/src/app/telecom/pack/migration/offers/pack-migration-offers.controller.js
@@ -84,19 +84,13 @@ export default class TelecomPackMigrationOffersCtrl {
     };
 
     // Retrieve option values (included and added)
-    const options = [];
-
-    Object.entries(selectedOffer.options).map(([key, option]) => {
-      if (option.included > 0 || option.choosedValue > 0) {
-        options.push({
-          quantity:
-            option.included +
-            (option.choosedValue > 0 ? option.choosedValue : 0),
-          name: option.name,
-        });
-      }
-      return key;
-    });
+    const options = Object.entries(selectedOffer.options)
+      .filter(([, option]) => option.included > 0 || option.choosedValue > 0)
+      .map(([, option]) => ({
+        quantity:
+          option.included + (option.choosedValue > 0 ? option.choosedValue : 0),
+        name: option.name,
+      }));
 
     if (options.length > 0) {
       params.options = options;
@@ -107,7 +101,6 @@ export default class TelecomPackMigrationOffersCtrl {
       .$promise.then((result) => {
         selectedOffer.subServicesToDelete = result;
         this.TucPackMigrationProcess.selectOffer(selectedOffer);
-
         return result;
       })
       .catch((error) => {
--- a/packages/manager/apps/telecom/src/app/telecom/pack/migration/offers/pack-migration-offers.controller.js
+++ b/packages/manager/apps/telecom/src/app/telecom/pack/migration/offers/pack-migration-offers.controller.js
@@ -84,9 +84,9 @@ export default class TelecomPackMigrationOffersCtrl {
     };
 
     // Retrieve option values (included and added)
-    const options = Object.entries(selectedOffer.options)
-      .filter(([, option]) => option.included > 0 || option.choosedValue > 0)
-      .map(([, option]) => ({
+    const options = Object.values(selectedOffer.options)
+      .filter((option) => option.included > 0 || option.choosedValue > 0)
+      .map((option) => ({
         quantity:
           option.included + (option.choosedValue > 0 ? option.choosedValue : 0),
         name: option.name,
--- a/packages/manager/apps/telecom/src/app/telecom/pack/migration/offers/pack-migration-offers.controller.js
+++ b/packages/manager/apps/telecom/src/app/telecom/pack/migration/offers/pack-migration-offers.controller.js
@@ -7,9 +7,16 @@ import { PROMO_DISPLAY } from '../pack-migration.constant';
 
 export default class TelecomPackMigrationOffersCtrl {
   /* @ngInject */
-  constructor($q, $translate, TucPackMigrationProcess, TucToast) {
+  constructor(
+    $q,
+    $translate,
+    OvhApiPackXdsl,
+    TucPackMigrationProcess,
+    TucToast,
+  ) {
     this.$q = $q;
     this.$translate = $translate;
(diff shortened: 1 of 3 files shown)


packages/manager/apps/telecom/src/app/telecom/pack/migration/offers/pack-migration-offers.controller.js
```diff
@@ -70,8 +77,50 @@ export default class TelecomPackMigrationOffersCtrl {
     );
   }
 
+  static isChosen(option) {
```
frenautvh, line 80: simply use ```option.choosedValue > 0``` where it's needed, not need to create a static method for that

packages/manager/apps/telecom/src/app/telecom/pack/migration/service-delete/pack-migration-service-delete.controller.js
```diff
@@ -11,6 +13,16 @@ export default class TelecomPackMigrationServiceDeleteCtrl {
...
+    this.process.selectedOffer.subServicesToDelete.forEach((subService) => {
+      set(
+        subService,
+        'services',
+        map(subService.services, (service, index, originalArray) => ({
+          name: service,
+          selected: originalArray.length === subService.numberToDelete,
+        })),
+      );
+    });
```
cbourgois, line 25: IMO we should avoid using `lodash/set` even if it's done like that in the codebase. Maybe you could use a `map` for that.
And I'm don't get what the `map` is supposed to do 🤔

Steffy29: I did this to transform the services array like ['service1', 'service2',...] to array like [{name: 'service1', selected: false}, {name:'service2', selected: false}...] that it could be display into the view.

packages/manager/apps/telecom/src/app/telecom/pack/migration/offers/pack-migration-offers.controller.js
```diff
@@ -71,7 +78,45 @@ export default class TelecomPackMigrationOffersCtrl {
...
+    const params = {
+      offerName: selectedOffer.offerName,
+    };
+
+    // Retrieve option values (included and added)
+    const options = [];
+
+    Object.entries(selectedOffer.options).map(([key, option]) => {
+      if (option.included > 0 || option.choosedValue > 0) {
+        options.push({
```
marie-j, line 91: map + if and push = filter , no ?

packages/manager/apps/telecom/src/app/telecom/pack/migration/offers/pack-migration-offers.controller.js
```diff
@@ -71,7 +78,38 @@ export default class TelecomPackMigrationOffersCtrl {
...
 
   selectOffer(offer) {
-    this.TucPackMigrationProcess.selectOffer(offer);
+    const selectedOffer = offer;
+    const params = {
+      offerName: selectedOffer.offerName,
+    };
+
+    // Retrieve option values (included and added)
+    const options = Object.entries(selectedOffer.options)
```
JeremyDec, line 87: If no use of key for each entry:
```suggestion
    const options = Object.values(selectedOffer.options)
```

packages/manager/apps/telecom/src/app/telecom/pack/migration/offers/pack-migration-offers.controller.js
```diff
@@ -71,7 +78,38 @@ export default class TelecomPackMigrationOffersCtrl {
   }
 
   selectOffer(offer) {
```
JeremyDec, line 80: The method seems to delete services, so the name should be changed

packages/manager/apps/telecom/src/app/telecom/pack/migration/service-delete/pack-migration-service-delete.controller.js
JeremyDec: As @cbourgois suggested, this code could be simplified to use at least two map instructions.
```suggestion
    this.process.selectedOffer.subServiceToDelete = this.process.selectedOffer.subServicesToDelete.map((subService) => ({
    ...subService,
    services: services.map(() => // etc...)
    }));
```

Also the logic of the `map`on `subService.services` is quite disturbing. You're setting a `selected` boolean, which will be true if the subService has all its services to be deleted ?

packages/manager/apps/telecom/src/app/telecom/pack/migration/offers/pack-migration-offers.controller.js
Steffy29: No, this method is called when you click on the "Choose this offer" button. We then need to update the list of services to be deleted from the additional options selected in this same page. So the name is correct.

packages/manager/apps/telecom/src/app/telecom/pack/migration/service-delete/pack-migration-service-delete.controller.js
Steffy29: this `selected` boolean is for the checkbox displayed into the view of services to delete

cbourgois: Not sure, but if you have something like : 

```javascript
subServicesToDelete = [
  {
    services: ["a", "b", "c", "d"],
    numberToDelete: 3,
    // ... props
  },
  {
    services: ["x", "y", "z"],
    numberToDelete: 3,
    // ... props
  }
]
```

The expected result is : 
- `a/b/c/d` service with `selected` set to `false`
- `x/y/z` service with `selected` set to `true`  

Can you confirm ?

cbourgois: The only thing I want to know is that we really mark service as `selected` if the parent `numberToDelete` is equal to the number of services in `services`. Strange rule, I don't get it.

Steffy29: Yes it's that

cbourgois: Ok, let the customer choose which services or select all because all should be deleted. Thanks for the infos.

Heiiji (approved):

antleblanc (approved):

PR #4509, opened 2021-03-08
Refactor: clean unused translations

ovh-ux-cds:


packages/manager/apps/telecom/src/app/telecom/pack/slots/emailPro/add/translations/Messages_de_DE.json
```diff
@@ -1,7 +1,6 @@
 {
-  "add_an_account": "Ajouter un compte xDSL E-mail",
```
antleblanc, line 2: Following key `add_an_account` is removed because it's not present in the source language file, isn't it?

I mean: `packages/manager/apps/telecom/src/app/telecom/pack/slots/emailPro/add/translations/Messages_fr_FR.json`

antleblanc (changes requested):
I would recommend to use a commit type `fix` instead of `refactor` for this use case.

Could you please consider using a signed commit (if possible)?

Thanks for this nice addition!

packages/manager/apps/telecom/src/app/telecom/pack/slots/emailPro/add/translations/Messages_de_DE.json
cbourgois: @antleblanc, @marie-j

The tool compare source language translations keys with target language translations keys in the same directory.

`add_an_account` isn't present in `fr_FR` file, so it will be deleted from this target file.

marie-j:
I'm closing this in order to address reviews (PR auto-generated)

PR #4505, opened 2021-03-05
Feat/ufrontend env apps

cbourgois:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/ufrontend-environment`
| Bug fix?         | yes
| New feature?     | yes
| Breaking change? | yes
| Tickets          | Fix MANAGER-6500
| License          | BSD 3-Clause

### Environment

The `Environment` is no more a Singleton. A new `Environment` is created, hydrated and provided by the `fetchConfiguration`.
Impacts are the following 
* the `ovhManagerCore` module needs the Environment to be defined dynamically, so the AngularJS  module is no more declared by default.
* the ufrontend package, replace the `config` object by the `environment` to inject the `environment` in `application` and `fragments` dynamically.

### Modules

- `@ovh-ux/manager-config` and `@ovh-ux/manager-core` bumped 

### Applications

All applications are up to date and use the `registerApplication` from `@ovh-ux/ufrontend` (EnvExecutor behind), attach/detach the preloader and inject the `environment` in the application. 
Applications are compatible with fragment integration.

Apps
- `@ovh-ux/manager-hub-app`
- `@ovh-ux/manager-dedicated`
- `@ovh-ux/manager-public-cloud`
- `@ovh-ux/manager-web`
- `@ovh-ux/manager-telecom`
- `@ovh-ux/sign-up-app`

Apps using fragment
- `@ovh-ux/manager-freefax-app`
- `fragment/navbar`

Standalone apps
- `@ovh-ux/manager-carrier-sip-app`
- `@ovh-ux/manager-cda-app`
- `@ovh-ux/manager-cloud-connect-app`
- `@ovh-ux/manager-dbaas-logs-app`
- `@ovh-ux/manager-enterprise-cloud-database-app`
- `@ovh-ux/manager-exchange-app`
- `@ovh-ux/manager-iplb-app`
- `@ovh-ux/manager-metrics-app`
- `@ovh-ux/manager-nasha-app`
- `@ovh-ux/manager-office-app`
- `@ovh-ux/order-tracking-app`
- `@ovh-ux/manager-overthebox-app`
- `@ovh-ux/manager-pci-app`
- `@ovh-ux/manager-sharepoint-app`
- `@ovh-ux/manager-sms-app`
- `@ovh-ux/manager-support-app`
- `@ovh-ux/manager-telecom-dashboard-app`
- `@ovh-ux/manager-telecom-task-app`
- `@ovh-ux/manager-veeam-cloud-connect-app`
[...]

PR #4528, opened 2021-03-12
feat(ng-at-internet): set US custom vars

Jisay:
Signed-off-by: Jisay <jean-christophe.alleman@corp.ovh.com>

| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-6432

## Description

Add `setRegion` in provider in order to differenciate custom vars from EU/CA and US.

### DIFF
--- a/packages/components/ng-at-internet/src/constants.js
+++ b/packages/components/ng-at-internet/src/constants.js
@@ -6,7 +6,10 @@ export const AT_INTERNET_CUSTOM_VARS = {
    * Country code of the customer making the order
    */
   countryCode: {
-    path: 'site.1',
+    path: {
+      default: 'site.1',
+      US: 'site.12',
+    },
     format: '[%s]',
   },
 
@@ -14,7 +17,10 @@ export const AT_INTERNET_CUSTOM_VARS = {
    * Currency code of the customer
    */
   currencyCode: {
-    path: 'site.16',
+    path: {
+      default: 'site.16',
+      US: 'site.15',
+    },
     format: '[%s]',
   },
 
@@ -23,7 +29,10 @@ export const AT_INTERNET_CUSTOM_VARS = {
    */
 
   referrerSite: {
-    path: 'site.8', // OVH's AtInternet configuration
+    path: {
+      default: 'site.8', // OVH's AtInternet configuration
+      US: 'site.11',
+    },
     format: '[%s]',
   },
 
@@ -31,12 +40,18 @@ export const AT_INTERNET_CUSTOM_VARS = {
    * Event
    */
   event: {
-    path: 'site.20',
+    path: {
+      default: 'site.20',
+      US: 'site.6',
+    },
     format: '[%s]',
   },
 
   siteName: {
-    path: 'site.13',
+    path: {
+      default: 'site.13',
+      US: 'site.14',
+    },
     format: '[%s]',
   },
 
@@ -44,7 +59,9 @@ export const AT_INTERNET_CUSTOM_VARS = {
    * Project Id
    */
   projectId: {
-    path: 'page.1', // OVH's AtInternet configuration
+    path: {
+      default: 'page.1', // OVH's AtInternet configuration
+    },
     format: '[%s]',
   },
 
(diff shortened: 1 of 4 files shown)


mohammed-zahaf (approved):

JeremyDec (dismissed):

packages/manager/modules/at-internet-configuration/src/index.js
```diff
@@ -28,6 +28,7 @@ angular
     ) => {
       atInternetProvider.setEnabled(trackingEnabled);
       atInternetProvider.setDebug(!trackingEnabled);
+      atInternetProvider.setRegion(Environment.getRegion());
```
cbourgois, line 31: This package has no dependency on `@ovh-ux/manager-config`.
And we should avoid to add a dependency on `Environment`, as https://github.com/ovh/manager/pull/4505 will remove it as a singleton.

I think if we really need the region we should have a provider to configure it.

cbourgois (changes requested):

packages/manager/modules/at-internet-configuration/src/index.js
Jisay: > This package has no dependency on @ovh-ux/manager-config.

So we are lucky :-) as it's also used at: https://github.com/ovh/manager/blob/9bba0f475e2b99ac7d9746a3e0981de95e4f8971/packages/manager/modules/at-internet-configuration/src/index.js#L78

> I think if we really need the region we should have a provider to configure it.

Ok, how to get the correct region? All apps will require to configure the `at-internet-configuration` module, I'm right?

cbourgois: Ok I updated it to use coreConfig instead from `manager-core` (https://github.com/ovh/manager/blob/feat%2Fufrontend-environment/packages/manager/modules/at-internet-configuration/src/index.js#L81).

`coreConfig` is already in the codebase, so it's already ok to use `coreConfig.getRegion()` instead of `Environment.getRegion()`

Jisay: Done. `manager-config` has been replaced by coreConfig

cbourgois (approved):

packages/components/ng-at-internet/src/constants.js
```diff
@@ -23,44 +29,59 @@ export const AT_INTERNET_CUSTOM_VARS = {
...
     format: '[%s]',
   },
 
   /**
    * Project Id
    */
   projectId: {
-    path: 'page.1', // OVH's AtInternet configuration
+    path: {
+      default: 'page.1', // OVH's AtInternet configuration
```
marie-j, line 63: This doesn't exist for US or it is the same ?

packages/components/ng-at-internet/src/provider.js
```diff
@@ -185,10 +199,14 @@ export default /* @ngInject */ function(AT_INTERNET_CUSTOM_VARS) {
       }
     }
 
+    function getCustomVarConfigPath(conf) {
+      return get(conf.path, config.region, get(conf.path, 'default'));
```
marie-j, line 203: **nitpick**
```suggestion
      return conf.path[config.region] || conf.path.default;
```

marie-j (dismissed):

packages/manager/modules/at-internet-configuration/src/index.js
```diff
@@ -4,8 +4,6 @@ import '@ovh-ux/manager-core';
 import '@ovh-ux/ng-at-internet';
 import '@ovh-ux/ng-at-internet-ui-router-plugin';
 
-import { Environment } from '@ovh-ux/manager-config';
```
antleblanc, line 7: Shouldn't be removed from `packages/manager/modules/at-internet-configuration/package.json` also?

packages/components/ng-at-internet/src/constants.js
Jisay: Yep, not specified in MANAGER-6432

JeremyDec (approved):

marie-j (approved):

antleblanc:
Rebasing on `develop` in order to rewrite git commit history (got 2 approvals).

PR #4595, opened 2021-03-22
fix(tracking): update defaults initialization

cbourgois:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix MANAGER-5814,

## Description

Avoid race condition between API call and routing initialization to send first track page event.
Use defaultPromise instead of setting defaults when the promise is resolved.

### DIFF
--- a/packages/manager/modules/at-internet-configuration/src/index.js
+++ b/packages/manager/modules/at-internet-configuration/src/index.js
@@ -77,18 +77,19 @@ angular
         ...atInternetConfiguration.getConfig(Environment.getRegion()),
         ...(referrerSite ? { referrerSite } : {}),
       };
-      $http
-        .get('/me')
-        .then(({ data: me }) => me)
-        .catch(() => {})
-        .then((me) => {
-          atInternet.setDefaults({
+
+      atInternet.setDefaultsPromise(
+        $http
+          .get('/me')
+          .then(({ data: me }) => me)
+          .catch(() => {})
+          .then((me) => ({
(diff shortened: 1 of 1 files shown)


packages/manager/modules/at-internet-configuration/src/index.js
```diff
@@ -77,18 +77,19 @@ angular
...
-        .then(({ data: me }) => me)
-        .catch(() => {})
-        .then((me) => {
-          atInternet.setDefaults({
+
+      atInternet.setDefaultsPromise(
+        $http
+          .get('/me')
+          .then(({ data: me }) => me)
+          .catch(() => {})
```
frenautvh, line 85: won't it fail if the catch is before ? it'll return an empty object then me.country will break in the next "then" block ?

cbourgois: If the code fails, the catch returns an empty object, and it won't break, only return `undefined` values.

```js
const me = {};

console.log({
  countryCode: me.country,
  currencyCode: me.currency && me.currency.code,
  visitorId: me.customerCode,
});
```

```js
{countryCode: undefined, currencyCode: undefined, visitorId: undefined}
```

frenautvh (approved):

JeremyDec (approved):

antleblanc (approved):

marie-j (approved):

PR #4702, opened 2021-04-07
fix(telecom.phone): set a height for tiles item

fchabrolle:
Signed-off-by: Fabien Chabrolle <fabien.chabrolle@corp.ovh.com>

| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-6663

fix display of phone tiles

### DIFF
--- a/packages/manager/apps/telecom/src/app/telecom/telephony/line/phone/order/choice/choice.less
+++ b/packages/manager/apps/telecom/src/app/telecom/telephony/line/phone/order/choice/choice.less
@@ -3,6 +3,7 @@
 .telecom-telephony-line-phone-order {
   .phone-item {
     margin: 8px 0;
+    height: 300px;
 
     .oui-radio_thumbnail-light {
       height: 300px;


packages/manager/apps/telecom/src/app/telecom/telephony/line/phone/order/choice/choice.less
```diff
@@ -3,6 +3,7 @@
 .telecom-telephony-line-phone-order {
   .phone-item {
     margin: 8px 0;
+    height: 300px;
```
radireddy, line 6: why 300px? why not 100%? let it occupy the height it needs.

fchabrolle: 100% doesn't fix the issue some tiles don't have the same content, so they're smaller.

Heiiji (approved):

packages/manager/apps/telecom/src/app/telecom/telephony/line/phone/order/choice/choice.less
antleblanc: Probably `min-height` could be safer here but the best option we could figure out is that replacing actual grid by using flex.
Could be handle into a separate Pull Request.

antleblanc (approved):
Thank you @fchabrolle for your contribution.

PR #4747, opened 2021-04-12
feat(pci.components.project.images-list): display newest os imgs first

mohammed-zahaf:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-6627

## Description

Display newest OS images first into the select box.
Currently on the Manager, the sort is ascending and made into OVH-API-SERVICES based on `name` attribute.
The sort based on name attribute it not relevant, so we decide that sort will be drove by API side.

For this case i copy and adapt the related OVH-API-SERVICE code directly into Manager.

### DIFF
--- a/packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
+++ b/packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
@@ -3,12 +3,26 @@ import groupBy from 'lodash/groupBy';
 import map from 'lodash/map';
 import pick from 'lodash/pick';
 import omit from 'lodash/omit';
+import snakeCase from 'lodash/snakeCase';
 
 import Image from './images.class';
+import { IMAGES_REGEX } from './images.constants';
+
+function getDistribution(name, type) {
+  const os = IMAGES_REGEX[type];
+  if (os) {
+    const distribution = os.find((distrib) => distrib.regex.test(name));
+
+    return distribution?.name || ''.concat(type, '_other');
+  }
+
+  return 'unknown';
+}
 
 export default class ImagesList {
   /* @ngInject */
-  constructor(OvhApiCloudProjectImage, OvhApiCloudProjectSnapshot) {
+  constructor($http, OvhApiCloudProjectImage, OvhApiCloudProjectSnapshot) {
+    this.$http = $http;
     this.OvhApiCloudProjectImage = OvhApiCloudProjectImage;
     this.OvhApiCloudProjectSnapshot = OvhApiCloudProjectSnapshot;
   }
@@ -38,9 +52,18 @@ export default class ImagesList {
   }
 
   getImages(serviceName) {
-    return this.OvhApiCloudProjectImage.v6()
-      .query({ serviceName })
-      .$promise.then((images) => ImagesList.groupByName(images));
+    return this.$http
+      .get(`/cloud/project/${serviceName}/image`)
+      .then(({ data: images }) =>
+        images.map((image) => {
+          return {
+            ...image,
+            nameGeneric: snakeCase(image.name),
+            distribution: getDistribution(image.name, image.type),
+          };
+        }),
+      )
+      .then((images) => ImagesList.groupByName(images));
   }
 
   getSnapshots(serviceName) {
--- a/packages/manager/modules/pci/src/components/project/images-list/images.constants.js
+++ b/packages/manager/modules/pci/src/components/project/images-list/images.constants.js
@@ -25,6 +25,49 @@ import ASSET_OS_WINDOWS_OTHER from './assets/os/windows_other.png';
 import ASSET_OS_WINDOWS_SERVER_2012 from './assets/os/windows_server_2012.png';
 import ASSET_OS_WINDOWS_SERVER_2016 from './assets/os/windows_server_2016.png';
 
+export const IMAGES_REGEX = {
+  linux: [
+    {
+      name: 'ubuntu',
+      regex: /^Ubuntu/i,
+    },
+    {
+      name: 'freebsd',
+      regex: /^FreeBSD/i,
+    },
+    {
+      name: 'coreos',
+      regex: /^CoreOS/i,
+    },
+    {
+      name: 'debian',
+      regex: /^Debian/i,
+    },
+    {
+      name: 'centos',
+      regex: /^Cent[\s-]?OS/i,
+    },
+    {
+      name: 'fedora',
+      regex: /^Fedora/i,
+    },
+    {
+      name: 'dokku',
+      regex: /^Dokku/i,
+    },
+  ],
+  windows: [
+    {
+      name: 'windows_server_2012',
+      regex: /^Win[a-zA-Z\s-]+2012/i,
+    },
+    {
+      name: 'windows_server_2016',
+      regex: /^Win[a-zA-Z\s-]+2016/i,
+    },
+  ],
+};
+
 export const ACTIVE_STATUS = ['active'];
 
 export const APPLICATION_TAG = 'application';
@@ -135,4 +178,5 @@ export default {
   APPLICATION_TAG,
   APPLICATION_LIST,
   IMAGE_ASSETS,
+  IMAGES_REGEX,
 };


packages/manager/modules/pci/src/components/project/images-list/images-list.controller.js
```diff
@@ -42,11 +43,13 @@ export default class ImagesListController {
   }
 
   getImages() {
-    return this.PciProjectImages.getImages(this.serviceName).then((images) => {
-      this.images = images;
-      this.updateImages(images);
-      return images;
-    });
+    return this.PciProjectImages.getImages(this.serviceName)
+      .then((images) => orderBy(images, ['name'], ['desc']))
```
frenautvh, line 47: you PR says 'newest' but you are sorting by name

cbourgois: You can't sort by `name` to sort by `date`.

cbourgois (changes requested):

packages/manager/modules/pci/src/components/project/images-list/images-list.controller.js
mohammed-zahaf: The aims is to display the new OS images first.
Eg for Fedora: `['Fedora 33', 'Fedora 32']`

mohammed-zahaf: is't an issue @cbourgois ?

cbourgois: This will work: 

```js
_.orderBy(
    [
        {
            name: 'Fedora 33'
        },
        {
            name: 'Fedora 34'
        },
        {
            name: 'Fedora 32'
        }
    ],
    ['name'], 
    ['desc'],
)
```

will output 

```
Fedora 34 
Fedora 33 
Fedora 32
```

But not this:

```js
_.orderBy(
    [
        {
            name: 'Fedora 1'
        },
        {
            name: 'Fedora 10'
        },
        {
            name: 'Fedora 9'
        }
    ],
    ['name'], 
    ['desc'],
)
```

will output

```
Fedora 9
Fedora 10
Fedora 1
```

So yes it's an issue.

mohammed-zahaf:
Need a new field from API to be able to order properly by OS versions.

mohammed-zahaf:
Now the ordering images is droved by PCI team.

packages/manager/modules/pci/src/components/project/images-list/images-list.controller.js
mohammed-zahaf: Outdate.

packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
```diff
@@ -3,12 +3,71 @@ import groupBy from 'lodash/groupBy';
 import map from 'lodash/map';
 import pick from 'lodash/pick';
 import omit from 'lodash/omit';
+import snakeCase from 'lodash/snakeCase';
 
 import Image from './images.class';
 
+function getDistribution(name, type) {
+  const regex = {
+    linux: [
```
marie-j, line 12: Not sure it is a good idea to map distributions this way. Each time a new distribution is added or updated we might have to update and maintain regexps

mohammed-zahaf: It was made like this into OVH-API-SERVICES.
I just import the function like it.

@marie-j do you have a suggestion ?

marie-j: @mohammed-zahaf Just because it was done that way doesn't mean it should be repeated 😄 
What is the need here ?

mohammed-zahaf: No, I let it to preserve the behavior.
Currently, from returned images list we don't have the `distribution` attribute.

So this function was made to deduce this attribute from the `name` attribute.
But sure it was to be provided via an attribute.

mohammed-zahaf: Sample:
`
creationDate: "2020-02-06T15:32:47Z"
flavorType: null
id: "blblablabla"
minDisk: 0
minRam: 0
name: "Debian 9 - IaaS hardened"
planCode: null
region: "GRA5"
size: 0.70184326171875
status: "active"
tags: []
type: "linux"
user: "root"
visibility: "private"
`

mohammed-zahaf: So @marie-j as i think, the API don't give us this info.
I asked the PCI team they said it will be planned in the future.

So we can let it like this until we have this attribute.

packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
```diff
@@ -38,9 +97,17 @@ export default class ImagesList {
...
 
   getImages(serviceName) {
-    return this.OvhApiCloudProjectImage.v6()
-      .query({ serviceName })
-      .$promise.then((images) => ImagesList.groupByName(images));
+    return this.$http
+      .get(`/cloud/project/${serviceName}/image`)
+      .then(({ data: images }) => {
+        images.forEach((image) => {
+          Object.assign(image, {
```
frenautvh, line 104: instead of Object.assign it's preferred to use
```js
return {
  ...image,
  nameGeneric: ...,
  distribution: ...,
};
```

mohammed-zahaf: I do like this, i will recreate the whole images list.
In my case i need just to add two new attribute ( `nameGeneric` and `distribution`).

mohammed-zahaf: Maybe like this:
```suggestion
          image.nameGeneric = '...'
          image.distribution = '...'
```

Like this we avoid the Object.assign, is't ok @frenautvh ?

packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
```diff
@@ -3,12 +3,71 @@ import groupBy from 'lodash/groupBy';
 import map from 'lodash/map';
 import pick from 'lodash/pick';
 import omit from 'lodash/omit';
+import snakeCase from 'lodash/snakeCase';
 
 import Image from './images.class';
 
+function getDistribution(name, type) {
+  const regex = {
```
radireddy, line 11: Please move this regex definition to a constant file. Now its getting initialized every time getDistribution method is called.

packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
```diff
@@ -3,12 +3,71 @@ import groupBy from 'lodash/groupBy';
...
+    for (let i = 0, l = regex[type].length; i < l; i += 1) {
+      if (regex[type][i].regex.test(name)) {
+        return regex[type][i].name;
+      }
+    }
+
+    return ''.concat(type, '_other');
+  }
+
+  return 'unknown';
```
radireddy, line 64: is it required to translate this? is it shown on UI?

packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
mohammed-zahaf: Fixed

packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
mohammed-zahaf: No, it's not required.

packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
```diff
@@ -3,12 +3,29 @@ import groupBy from 'lodash/groupBy';
...
 import Image from './images.class';
+import { IMAGES_REGEX } from './images.constants';
+
+function getDistribution(name, type) {
+  if (IMAGES_REGEX[type]) {
+    for (let i = 0, l = IMAGES_REGEX[type].length; i < l; i += 1) {
+      if (IMAGES_REGEX[type][i].regex.test(name)) {
+        return IMAGES_REGEX[type][i].name;
+      }
+    }
```
marie-j, line 17: Can't we use a find here instead ?

packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
marie-j: This looks achievable by a `map` 
Also as @frenautvh mentionned, spread operator is preffered over `Object.assign` as it prevents updating the value and creates a new object instead

packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
mohammed-zahaf: Fixed

radireddy (dismissed):

packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
```diff
@@ -3,12 +3,29 @@ import groupBy from 'lodash/groupBy';
...
 import omit from 'lodash/omit';
+import snakeCase from 'lodash/snakeCase';
 
 import Image from './images.class';
+import { IMAGES_REGEX } from './images.constants';
+
+function getDistribution(name, type) {
+  const os = IMAGES_REGEX[type];
+  if (os) {
+    const distribution = os.find((distributionId, index) => {
```
ganeshkumar1989, line 14:
```suggestion
    const distribution = os.find((distribution) => distribution.regex.test(name))
```

ganeshkumar1989 (changes requested):

packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
```diff
@@ -3,12 +3,28 @@ import groupBy from 'lodash/groupBy';
...
+import snakeCase from 'lodash/snakeCase';
 
 import Image from './images.class';
+import { IMAGES_REGEX } from './images.constants';
+
+function getDistribution(name, type) {
+  const os = IMAGES_REGEX[type];
+  if (os) {
+    const distribution = os.find(
+      (distrib, index) => index < os.length && distrib.regex.test(name),
```
ganeshkumar1989, line 15: why do we need the check `index < os.length`? all elements of the array will definitely will have the index less than the length :)

ganeshkumar1989 (changes requested):

ganeshkumar1989 (approved):

packages/manager/modules/pci/src/components/project/images-list/images-list.service.js
```diff
@@ -38,9 +52,18 @@ export default class ImagesList {
...
 
   getImages(serviceName) {
-    return this.OvhApiCloudProjectImage.v6()
-      .query({ serviceName })
-      .$promise.then((images) => ImagesList.groupByName(images));
+    return this.$http
+      .get(`/cloud/project/${serviceName}/image`)
+      .then(({ data: images }) =>
+        images.map((image) => {
+          return {
```
marie-j, line 59: Nitpick 
```suggestion
        images.map((image) => ({
```

marie-j (approved):

mohammed-zahaf:
Hello @antleblanc, this feature can be part of release.
No trad required, no internal endpoint to prod before.

Thanks.

PR #4762, opened 2021-04-14
feat(cloud.project): handle quantity unit in billing display

christophe-rannou:
Signed-off-by: Christophe Rannou <christophe.rannou@ovhcloud.com>

| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | 

In the billing control part of the Public Cloud control panel when displaying hourly usage the unit of the actual billed quantity is not handled leading to inconsistent values.

### DIFF
--- a/packages/manager/modules/pci/src/components/project/billing/resource-usage/controller.js
+++ b/packages/manager/modules/pci/src/components/project/billing/resource-usage/controller.js
@@ -3,4 +3,14 @@ export default class BillingResourceUsageCtrl {
   constructor(CucRegionService) {
     this.CucRegionService = CucRegionService;
   }
+
+  static hourlyQuantity(quantity) {
+    switch (quantity.unit) {
+      case 'Minute':
+        return (quantity.value / 60).toFixed(2);
+      case 'Hour':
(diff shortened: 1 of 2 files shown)


packages/manager/modules/pci/src/components/project/billing/resource-usage/controller.js
```diff
@@ -3,4 +3,14 @@ export default class BillingResourceUsageCtrl {
   constructor(CucRegionService) {
     this.CucRegionService = CucRegionService;
   }
+
+  static hourlyQuantity(quantity) {
+    switch (quantity.unit) {
```
antleblanc, line 8: **question:** Does the value can be returned by the API instead?

antleblanc: I discussed with @christophe-rannou and for a short term period, that's the only patch that can be applied.
So to improve readability we will integrate this patch.

antleblanc (approved):

PR #4763, opened 2021-04-14
fix(ip): ip firewall rules are not shown

radireddy:

| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | DTRSD-34405

## Description

Firewall rules are not visible if the rules contain duplicate records. This creates confusion for customers and they are trying to create them again.

### DIFF
--- a/packages/manager/apps/dedicated/client/app/ip/ip/firewall/ip-ip-firewall.controller.js
+++ b/packages/manager/apps/dedicated/client/app/ip/ip/firewall/ip-ip-firewall.controller.js
@@ -12,6 +12,8 @@ angular
       goToDashboard,
       $location,
       $route,
+      $stateParams,
+      $http,
     ) => {
       $scope.selectedBlock = null;
       $scope.selectedIp = null;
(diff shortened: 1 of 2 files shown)


JeremyDec (approved):

antleblanc (dismissed):

cbourgois (dismissed):

packages/manager/apps/dedicated/client/app/ip/ip/firewall/ip-ip-firewall.controller.js
```diff
@@ -40,69 +42,77 @@ angular
...
-            )
-            .finally(() => {
-              $scope.rulesLoading = false;
-            });
+      $scope.$on('ips.firewall.informations.reload', () => {
+        reloadRules();
+      });
+
+      $scope.loadRules = function loadRules(rulesCount, offset) {
+        if (!$scope.selectedIp && !$scope.selectedBlock) {
```
JeremyDec, line 108: Is `$scope.selectedBlock` test useful here ? You could set `$scope.selectedBlock = data.ip;` in all cases no ?

radireddy: Thanks for the review. No, I can not set it to data.ip in all cases. There are two cases.
First: user comes from the action menu. In this case, both selectedIp and selectedIpBlock are provided.
Second: User enters URL directly. In this case, only IP is available, need to fetch the ipBlock based on the IP.

JeremyDec (approved):

antleblanc (approved):

marie-j (approved):

JeremyDec (approved):

packages/manager/apps/dedicated/client/app/ip/ip/firewall/ip-ip-firewall.controller.js
```diff
@@ -40,69 +42,81 @@ angular
...
-          )
-            .then(
-              (rules) => {
-                $scope.rules = rules;
-                let options;
+              if (
+                $scope.rules &&
+                $scope.rules.list &&
+                $scope.rules.list.results &&
+                $scope.rules.list.results.length
```
mohammed-zahaf, line 63: i think we can use lodash to avoid the long path:
```suggestion
                get($scope.rules, 'list.results.length') > 0
```

What do you think @radireddy ?

antleblanc: Alternative: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining

Not mandatory since it has been already validated. Can be handled into a separate Pull Request.

PR #4798, opened 2021-04-20
feat(project): add pci project advices

radireddy:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | yes
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-6677

## Description

Added PCI project advices

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/project.html
+++ b/packages/manager/modules/pci/src/projects/project/project.html
@@ -84,22 +84,35 @@
                                         </a>
                                     </div>
                                 </div>
-                                <h2
-                                    data-translate="pci_projects_project_documentation"
-                                ></h2>
-                                <ul>
-                                    <li
-                                        data-ng-repeat="link in $ctrl.links track by $index"
-                                    >
-                                        <a
-                                            data-ng-href="{{ ::link.href }}"
-                                            target="_blank"
-                                            rel="noopener"
-                                            >{{ ::link.translation | translate
-                                            }}</a
+                                <div class="row">
+                                    <div class="col-sm-6 mb-4">
+                                        <h2
+                                            data-translate="pci_projects_project_documentation"
+                                        ></h2>
+                                        <ul>
+                                            <li
+                                                data-ng-repeat="link in $ctrl.links track by $index"
+                                            >
+                                                <a
+                                                    data-ng-href="{{ ::link.href }}"
+                                                    target="_blank"
+                                                    rel="noopener"
+                                                    data-translate="{{ ::link.translation }}"
+                                                ></a>
+                                            </li>
+                                        </ul>
+                                    </div>
+                                    <div class="col-sm-6 mb-4">
+                                        <!-- Advices -->
+                                        <ovh-advices
+                                            url="/cloud/:serviceName"
+                                            url-params="{
+                                                serviceName: $ctrl.projectId,
+                                            }"
                                         >
-                                    </li>
-                                </ul>
+                                        </ovh-advices>
+                                    </div>
+                                </div>
                             </div>
                         </div>
                     </div>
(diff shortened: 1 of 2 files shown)


packages/manager/modules/pci/src/projects/project/project.html
```diff
@@ -84,22 +84,40 @@
...
                                 </div>
-                                <h2
-                                    data-translate="pci_projects_project_documentation"
-                                ></h2>
-                                <ul>
-                                    <li
-                                        data-ng-repeat="link in $ctrl.links track by $index"
+                                <div class="row">
+                                    <div
+                                        class="col-sm-6 col-lg-6 col-xl-6 mb-4"
```
JeremyDec, line 89: No `col-md-6` ? Also, you need to specify for all the sizes ?

radireddy: I think it will be the same as col-lg-6. Correct me if I am wrong, if the lower-end classes are not provided, it will consider the next higher class. For example, if only col-xl-6 is provided, bootstrap will consider 6 column layout for the rest of the lower-end classes. In this case, providing col-xl-6 is sufficient.

JeremyDec: Yes you're right, so no need `col-sm/lg-6` ? :)

JeremyDec: We are using bootstrap4 in this module (https://github.com/ovh/manager/blob/master/packages/manager/modules/pci/package.json#L67).

If I check the documentation, I see the following:

"_Grid breakpoints are based on minimum width media queries, meaning they apply to that one breakpoint and all those above it (e.g., .col-sm-4 applies to small, medium, large, and extra large devices, but not the first xs breakpoint)._"

So, it should be the contrary no ? (lower classes applies to higher sizes)

frenautvh (approved):

packages/manager/modules/pci/src/projects/project/project.html
radireddy: Yes, you are right, it's exactly the opposite of what I said :). In this case, just having col-sm-4 is sufficient.

antleblanc: You mean `col-sm-6` (not `col-sm-4`) is required? Isn't it?

antleblanc, line 89:
```suggestion
                                        class="col-sm-6 mb-4"
```

packages/manager/modules/pci/src/projects/project/project.html
```diff
@@ -84,22 +84,40 @@
...
+                                                    target="_blank"
+                                                    rel="noopener"
+                                                    >{{ ::link.translation |
+                                                    translate }}</a
+                                                >
+                                            </li>
+                                        </ul>
+                                    </div>
+                                    <div
+                                        class="col-sm-6 col-lg-6 col-xl-6 mb-4"
```
antleblanc, line 109:
```suggestion
                                        class="col-sm-6 mb-4"
```

packages/manager/modules/pci/src/projects/project/project.html
```diff
@@ -84,22 +84,40 @@
...
+                                        ></h2>
+                                        <ul>
+                                            <li
+                                                data-ng-repeat="link in $ctrl.links track by $index"
+                                            >
+                                                <a
+                                                    data-ng-href="{{ ::link.href }}"
+                                                    target="_blank"
+                                                    rel="noopener"
+                                                    >{{ ::link.translation |
```
antleblanc, line 102: Not directly brought by this Pull Request but maybe we can consider replacing `{{` syntax and prefer using `translate` directive instead.

antleblanc (changes requested):

mohammed-zahaf (approved):

marie-j (approved):

antleblanc (approved):

PR #4811, opened 2021-04-22
fix(monitoring): update monitoring value after put

Heiiji:
| Question         | Answer
| ---------------- | ---
| Branch?          |`develop` 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-6406

## Description

need to update this boolean to display the right value otherwhere. No need process time so no need polling.

### DIFF
--- a/packages/manager/apps/dedicated/client/app/dedicated/server/details/server.controller.js
+++ b/packages/manager/apps/dedicated/client/app/dedicated/server/details/server.controller.js
@@ -220,6 +220,14 @@ export default class ServerCtrl {
       this.loadServer();
     });
 
+    this.$scope.$on(
+      'dedicated.informations.update.monitoring',
+      (_, monitored) => {
+        this.server.monitored = monitored;
+        this.loadServer();
+      },
+    );
+
     this.$scope.$on('$destroy', () => {
       this.Polling.addKilledScope();
       this.Poller.kill({
--- a/packages/manager/apps/dedicated/client/app/dedicated/server/details/server.service.js
+++ b/packages/manager/apps/dedicated/client/app/dedicated/server/details/server.service.js
@@ -261,8 +261,19 @@ export default class ServerF {
       data: {
         monitoring,
       },
-      broadcast: 'dedicated.informations.reload',
-    });
+    }).then(() =>
+      this.OvhHttp.get('/dedicated/server/{serviceName}', {
+        rootPath: 'apiv6',
+        urlParams: {
+          serviceName,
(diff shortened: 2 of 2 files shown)


packages/manager/apps/dedicated/client/app/dedicated/server/details/server.controller.js
```diff
@@ -220,6 +220,11 @@ export default class ServerCtrl {
       this.loadServer();
     });
 
+    this.$scope.$on('dedicated.informations.update.monitoring', () => {
+      this.server.monitored = !this.server.monitored;
```
frenautvh, line 224: how are you sure that the monitoring value is toggled ?
it's possible to call updateMonitoring method with the same value, in that case monitoring won't change

i think it would be safer to pass the monitoring value as an argument of this event so you can write
```js
this.$scope.$on('dedicated.informations.update.monitoring', (isMonitored) => {
// ...
```

packages/manager/apps/dedicated/client/app/dedicated/server/details/server.service.js
```diff
@@ -261,8 +261,19 @@ export default class ServerF {
...
-      broadcast: 'dedicated.informations.reload',
-    });
+    }).then(() =>
+      this.OvhHttp.get('/dedicated/server/{serviceName}', {
+        rootPath: 'apiv6',
+        urlParams: {
+          serviceName,
+        },
+      }).then((data) => {
+        return this.$rootScope.$broadcast(
```
marie-j, line 271: Can't we reload the state or somethign instead of broadcasting through `$rootScope` ? Or at least move this callback to the call so we can  maybe use  `$scope.$emit` instead

Heiiji: i understand this point but with actual tructure i have not find solution to do this differently with my actual capability :'(

Heiiji: if is a blocker i can cancel this PR and this fix became a subject for a bigger rework :/

radireddy: We don't need to call the API again or refresh the route. I made the same fix in PR https://github.com/ovh/manager/pull/4835. Please take a look at it.

radireddy:
Same change made as part of PR https://github.com/ovh/manager/pull/4835

PR #4822, opened 2021-04-26
Feat/manager 6681: Add autoscaling to Node Pools

mohammed-zahaf:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-6681

## Description

This feature will allow customers to scale their node-poll 

### :flags: Translations

- [x] TRANS-35267


packages/manager/modules/pci/src/projects/project/kubernetes/add/add.controller.js
```diff
@@ -60,12 +61,16 @@ export default class {
...
-      flavorName: this.cluster.nodePool.flavor.name,
-      antiAffinity: this.cluster.nodePool.antiAffinity,
-      monthlyBilled: this.cluster.nodePool.monthlyBilling,
+      desiredNodes: nodePool.nodeCount,
+      flavorName: nodePool.flavor.name,
+      antiAffinity: nodePool.antiAffinity,
+      monthlyBilled: nodePool.monthlyBilling,
+      autoscale: nodePool.autoscaling.autoscale,
+      minNodes: nodePool.autoscaling.nodes.lowest.value,
+      maxNodes: nodePool.autoscaling.nodes.highest.value,
```
marie-j, line 73: Wouldn't it be easier to use a model that would abstract all the nesting ?

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.html
```diff
@@ -0,0 +1,100 @@
...
+            data-label="{{:: 'kube_common_node_pool_autoscaling_autoscale' | translate}}"
+            data-size="xl"
+        >
+            <oui-switch
+                model="$ctrl.model.autoscale"
+                on-change="$ctrl.onAutoscaleChanged({status: modelValue})"
+            >
+            </oui-switch>
+            <strong
+                data-translate="{{'kube_common_node_pool_autoscaling_toggle_' + $ctrl.model.autoscale | translate }}"
```
marie-j, line 54:
```suggestion
                data-translate="{{'kube_common_node_pool_autoscaling_toggle_' + $ctrl.model.autoscale  }}"
```

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.html
```diff
@@ -0,0 +1,100 @@
...
+            data-size="xl"
+        >
+            <oui-numeric
+                id="autoscaling-node-pool-lowest-count"
+                disabled="!$ctrl.model.autoscale"
+                name="autoscaling-node-pool-lowest-count"
+                min="$ctrl.model.nodes.lowest.min"
+                model="$ctrl.model.nodes.lowest.value"
+                max="$ctrl.model.nodes.desired.value"
+                on-change="!$ctrl.isEditMode ? $ctrl.nodeCount = modelValue : null"
```
marie-j, line 75: Maybe logic could be extracted to a function

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,5 @@
+{
+  "kubernetes_add_node_pool_autoscaling_description": "Vous pouvez choisir de mettre à l'échelle automatiquement chacun de vous pool de nœuds. Cette fonctionnalité est parfois appelée \"cluster autoscaler\" et vient compléter le horizontal pod autoscaler de Kubernetes.",
```
marie-j, line 2:
```suggestion
  "kubernetes_add_node_pool_autoscaling_description": "Vous pouvez choisir de mettre à l'échelle automatiquement chacun de vos pool de nœuds. Cette fonctionnalité est parfois appelée \"cluster autoscaler\" et vient compléter le horizontal pod autoscaler de Kubernetes.",
```

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/adjust/adjust.component.js
```diff
@@ -0,0 +1,19 @@
...
+const component = {
+  bindings: {
+    autoscaling: '<',
+    goBack: '<',
+    kubeId: '<',
+    nodePool: '<',
+    nodePoolId: '<',
+    nodePoolName: '<',
+    projectId: '<',
+    sendKubeTrack: '<',
```
marie-j, line 13:
```suggestion
    sendKubeTracker: '<',
```

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/adjust/adjust.controller.js
```diff
@@ -0,0 +1,66 @@
...
+    this.$translate = $translate;
+    this.Kubernetes = Kubernetes;
+    this.DELETE_CONFIRMATION_INPUT = DELETE_CONFIRMATION_INPUT;
+  }
+
+  $onInit() {
+    const { autoscale } = this.nodePool;
+    const { lowest, desired, highest } = this.autoscaling.nodes;
+
+    this.isAjusting = false;
```
marie-j, line 18:
```suggestion
    this.isAdjusting = false;
```

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/adjust/adjust.routing.js
```diff
@@ -0,0 +1,31 @@
...
+      views: {
+        modal: {
+          component: 'pciProjectKubernetesNodePoolsAdjustComponent',
+        },
+      },
+      layout: 'modal',
+      translations: {
+        value: ['.'],
+        format: 'json',
+      },
```
marie-j, line 15:
```suggestion
```
can we use a 
```suggestion
  .run(/* @ngTranslationsInject:json ./translations */)
```

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/node-pool.html
```diff
@@ -73,6 +73,26 @@
                 data-ng-bind="$row.availableNodes+'/'+$row.desiredNodes"
             ></span>
         </oui-datagrid-column>
+        <oui-datagrid-column
+            data-title="'Autoscaling'"
```
marie-j, line 77: Not translated ?

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/translations/Messages_fr_FR.json
```diff
@@ -4,12 +4,15 @@
...
   "kube_node_pool_delete": "Supprimer le pool",
   "kube_node_pool_add": "Ajouter un pool de nœuds",
   "kube_node_pool_refreh": "Recharger le tableau",
   "kube_node_pool_title": "Pools de nœuds",
   "kube_node_pool_anti_affinity": "Anti-affinité",
   "kube_add_billing_type_description": "Vous pourrez passer de la facturation horaire au forfait mensuel pour chaque nœud, dans la vue édition du pool de nœud.",
   "kube_node_pool_enabled": "Activé",
-  "kube_node_pool_disabled": "Désactivé"
+  "kube_node_pool_disabled": "Désactivé",
+  "kube_node_pool_autoscale_true": "Active",
```
marie-j, line 16:
```suggestion
  "kube_node_pool_autoscale_true": "Actif",
```
no ?

packages/manager/modules/pci/src/projects/project/kubernetes/service.js
```diff
@@ -110,11 +110,11 @@ export default class Kubernetes {
     );
   }
 
-  resizeNodePool(projectId, kubeId, nodePoolId, desiredNodes) {
+  resizeNodePool(projectId, kubeId, nodePoolId, kubeNodePool) {
     return this.$http.put(
       `/cloud/project/${projectId}/kube/${kubeId}/nodepool/${nodePoolId}`,
       {
-        desiredNodes,
+        ...kubeNodePool,
```
marie-j, line 117: Destructure doesn't seem needed

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/adjust/adjust.component.js
mohammed-zahaf: Thanks for suggestion.
This function was developed in another feature.

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/node-pool.html
mohammed-zahaf: No, it's a globish word.

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.html
mohammed-zahaf: Fixed

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.html
mohammed-zahaf: Changed.

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/translations/Messages_fr_FR.json
mohammed-zahaf: Fixed

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/adjust/adjust.controller.js
mohammed-zahaf: Fixed by using scaling instead of ajusting.

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/adjust/adjust.routing.js
mohammed-zahaf: Fixed

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/translations/Messages_fr_FR.json
mohammed-zahaf: Fixed

packages/manager/modules/pci/src/projects/project/kubernetes/service.js
mohammed-zahaf: Fixed

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.constants.js
```diff
@@ -0,0 +1,7 @@
+export const AUTOSCALING_LINK = {
+  DEFAULT: 'https://docs.ovh.com/gb/en/kubernetes/using-cluster-autoscaler/',
```
JeremyDec, line 2: Why not store this guide link here: https://github.com/ovh/manager/commits/master/packages/manager/modules/core/src/manager-core.constants.js

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
```diff
@@ -0,0 +1,142 @@
...
+    );
+
+    AutoscalingController.triggerHandlerFunction(
+      this.onHighestPoolValueChanged,
+      max,
+    );
+  }
+
+  static triggerHandlerFunction(fn, data) {
+    if (isFunction(fn)) fn(data);
```
JeremyDec, line 127:
```suggestion
    if (typeof fn === 'function') {
      fn(data)
    };
```
Or, to keep one line you could also, and avoid with no brackets: (nitpick)
```suggestion
    typeof fn === 'function' && fn(data);
```

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,11 @@
+{
+  "kubernetes_node_pool_autoscaling_description": "Vous pouvez choisir de mettre à l'échelle automatiquement chacun de vous pool de nœuds. Cette fonctionnalité est parfois appelée \"cluster autoscaler\" et vient compléter le horizontal pod autoscaler de Kubernetes.",
```
JeremyDec:
```suggestion
  "kubernetes_node_pool_autoscaling_description": "Vous pouvez choisir de mettre à l'échelle automatiquement chacun de vos pool de nœuds. Cette fonctionnalité est parfois appelée \"cluster autoscaler\" et vient compléter le horizontal pod autoscaler de Kubernetes.",
```

packages/manager/modules/pci/src/projects/project/kubernetes/add/add.controller.js
mohammed-zahaf: Iam not sure we can make a model an just send it like it.
for example my `autoscaling` model can't be just composed of `minNodes`, `maxNodes` ...

So for me no.

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.controller.js
```diff
@@ -0,0 +1,59 @@
...
+
+  $onInit() {
+    const { autoscale } = this.nodePool;
+    const { lowest, desired, highest } = this.autoscaling.nodes;
+
+    this.isScaling = false;
+    this.autoscaling.autoscale = autoscale;
+    lowest.value = this.nodePool.minNodes;
+    desired.value = this.nodePool.desiredNodes;
+    highest.value = this.nodePool.maxNodes;
```
JeremyDec, line 21: Why not directly declare `lowest/desired/highest` to `this.nodePool` corresponding values ?

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.controller.js
```diff
@@ -0,0 +1,59 @@
...
+            message: get(error, 'data.message'),
+          }),
+          'error',
+        ),
+      );
+  }
+
+  onNodePoolScaleModalCancel() {
+    this.sendKubeTrack('details::nodepools::scale::cancel');
+    this.goBack();
```
JeremyDec, line 57:
```suggestion
    return this.goBack();
```

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.routing.js
```diff
@@ -0,0 +1,31 @@
...
+      views: {
+        modal: {
+          component: 'pciProjectKubernetesNodePoolsScaleComponent',
+        },
+      },
+      layout: 'modal',
+      translations: {
+        value: ['.'],
+        format: 'json',
+      },
```
JeremyDec: You're already loading translations in the module
```suggestion
```

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.routing.js
```diff
@@ -0,0 +1,31 @@
...
+      params: {
+        nodePoolId: null,
+        nodePoolName: null,
+      },
+      resolve: {
+        goBack: /* @ngInject */ (goToNodePools) => goToNodePools,
+        nodePool: /* @ngInject */ (Kubernetes, kubeId, nodePoolId, projectId) =>
+          Kubernetes.getNodePool(projectId, kubeId, nodePoolId),
+        nodePoolId: /* @ngInject */ ($stateParams) => $stateParams.nodePoolId,
```
JeremyDec, line 20: `$stateParams` is obsolete, use `$transition$.params()` instead please

packages/manager/modules/pci/src/projects/project/kubernetes/translations/Messages_fr_FR.json
```diff
@@ -9,7 +9,8 @@
   "kube_common_save": "Valider",
   "kube_common_cancel": "Annuler",
   "kube_common_delete": "Supprimer",
-  "kube_common_node_pool_title": "Configurez votre pool de nœuds",
+  "kube_common_node_pool_title": "Configurez votre type de nœuds",
```
JeremyDec, line 12: Is this change legit ? You're keeping `pool de noeuds` just below

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.constants.js
mohammed-zahaf: The current link is specific for the Kube to scale nodePool.
I think is right location.

What do you think @JeremyDec ?

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
mohammed-zahaf: Thanks for suggestion.
Initially i was wrote it like the second suggestion, but finally i preferred to use Lodash function (elegant).

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/translations/Messages_fr_FR.json
mohammed-zahaf: Already fixed.

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.controller.js
mohammed-zahaf: It's a generic model, to be reusable on some other kube components.

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.routing.js
mohammed-zahaf: Already fixed.

packages/manager/modules/pci/src/projects/project/kubernetes/translations/Messages_fr_FR.json
mohammed-zahaf: Yes it's normal.
Because i made some refactoring on different stepper section.
So the initial step can't be stay with same title.

Done with UX.

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
JeremyDec: I agree `isFunction` lodash method is more elegant. But in fact you're importing a dependency here, and so even if it small you're adding weight to your code. And IMO it is no useful as it is just for cosmetic details

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.constants.js
JeremyDec: Hmm yes it is a specific part of the documentation. IMO it annoys me to have multiple places for documentation link. But also I guess it could make the core constants heavier to store all specific links. So you can let this as is

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.routing.js
mohammed-zahaf: Copy past from another module.
Thanks for your attention.

Fixed.

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.controller.js
mohammed-zahaf: Fixed

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
mohammed-zahaf: Right, fixed.

packages/manager/modules/pci/src/projects/project/kubernetes/add/add.controller.js
marie-j: I was thinking of a model for `NodePool` that way we could just nodePool.maxNodes and it would abstract all the nesting

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.component.js
```diff
@@ -0,0 +1,18 @@
+import controller from './autoscaling.controller';
+import template from './autoscaling.html';
+
+export default {
+  bindings: {
+    isEditMode: '<',
+    displayDesiredNodes: '<',
+    displayNodePoolSizing: '<',
+    model: '<',
```
marie-j, line 9: Not a two way binding ?

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
```diff
@@ -0,0 +1,140 @@
...
+
+export default class AutoscalingController {
+  /* @ngInject */
+  constructor(coreConfig) {
+    this.ovhSubsidiary = coreConfig.getUser().ovhSubsidiary;
+    this.autoscalingInfoLink = this.getAutoscalingInfoLink();
+  }
+
+  $onInit() {
+    this.setLowestHighestLimit();
```
marie-j, line 12: Having trouble understanding as we have both lowest and highest in the name 😅 Maybe we could just name it `setLimits`

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
```diff
@@ -0,0 +1,140 @@
...
+    );
+
+    AutoscalingController.triggerHandlerFunction(
+      this.onHighestPoolValueChanged,
+      max,
+    );
+  }
+
+  static triggerHandlerFunction(fn, data) {
+    if (fn instanceof Function) fn(data);
```
marie-j, line 125: I think if the function binding is set to `&?` this is not needed

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,11 @@
+{
+  "kubernetes_node_pool_autoscaling_description": "Vous pouvez choisir de mettre à l'échelle automatiquement chacun de vos pool de nœuds. Cette fonctionnalité est parfois appelée \"cluster autoscaler\" et vient compléter le horizontal pod autoscaler de Kubernetes.",
+  "kubernetes_node_pool_autoscaling_description_link": "Plus d'informations.",
+  "kubernetes_node_pool_autoscaling_autoscale": "Autoscaling",
+  "kubernetes_node_pool_autoscaling_autoscale_toggle_true": "Autoscaling Actif",
+  "kubernetes_node_pool_autoscaling_autoscale_toggle_false": "Autoscaling Inactif",
+  "kubernetes_node_pool_autoscaling_lowest_nodes_size": "Taille Minimale du pool",
+  "kubernetes_node_pool_autoscaling_desired_nodes_size": "Nombre de nœuds dans le pool",
+  "kubernetes_node_pool_autoscaling_highest_nodes_size": "Taille Maximale du pool",
```
marie-j, line 9:
```suggestion
  "kubernetes_node_pool_autoscaling_highest_nodes_size": "Taille maximale du pool",
```

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/index.js
```diff
@@ -3,18 +3,18 @@ import '@uirouter/angularjs';
 import 'oclazyload';
 
 const moduleName =
-  'ovhManagerPciProjectKubernetesDetailsNodesBulkDeleteLazyloading';
+  'ovhManagerPciProjectKubernetesDetailsNodePoolsScaleLazyloading';
 
 angular.module(moduleName, ['ui.router', 'oc.lazyLoad']).config(
   /* @ngInject */ ($stateProvider) => {
     $stateProvider.state(
-      'pci.projects.project.kubernetes.details.nodepools.details.nodes.bulk-delete.**',
```
marie-j, line 11: We completely remove the feature ?

packages/manager/modules/pci/src/projects/project/kubernetes/kubernetes.constants.js
```diff
@@ -1,7 +1,15 @@
 export const VERSION_ENUM_KEY = "models['cloud.kube.VersionEnum'].enum";
 export const ANTI_AFFINITY_MAX_NODES = 5;
+export const SCALE_DEFAULT_VALUES = {
+  LOWEST_MIN_VALUE: 0,
+  LOWEST_VALUE: 3,
+  DESIRED_VALUE: 3,
+  HIGHEST_VALUE: 100,
+  HIGHEST_MAX_VALUE: 100,
```
marie-j, line 8: Can't be exposed by API ?

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
```diff
@@ -0,0 +1,140 @@
...
+  getAutoscalingInfoLink() {
+    return AUTOSCALING_LINK[this.ovhSubsidiary] || AUTOSCALING_LINK.DEFAULT;
+  }
+
+  resetScaleValuesToDefault() {
+    const { lowest, highest } = this.model.nodes;
+    lowest.min = SCALE_DEFAULT_VALUES.LOWEST_MIN_VALUE;
+    lowest.value = SCALE_DEFAULT_VALUES.LOWEST_MIN_VALUE;
+
+    highest.value = SCALE_DEFAULT_VALUES.HIGHEST_VALUE;
```
radireddy, line 30: above you have set both lowest values to LOWEST_MIN_VALUE. It's different for the highest. Is it ok?

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
```diff
@@ -0,0 +1,140 @@
...
+      this.setLowestHighestLimit();
+    }
+
+    AutoscalingController.triggerHandlerFunction(
+      this.onAutoscalePoolChanged,
+      autoscale,
+    );
+  }
+
+  onLowestValueChanged(min) {
```
radireddy, line 53:
```suggestion
  onLowestValueChange(min) {
```

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
```diff
@@ -0,0 +1,140 @@
...
+      highest.value,
+    );
+
+    AutoscalingController.triggerHandlerFunction(
+      this.onLowestPoolValueChanged,
+      min,
+    );
+  }
+
+  onDesiredValueChanged(need) {
```
radireddy, line 75:
```suggestion
  onDesiredValueChange(need) {
```

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/node-pool.class.js
```diff
@@ -18,4 +18,8 @@ export default class NodePool {
   isError() {
     return this.status === 'ERROR';
   }
+
+  getMinMaxNodes() {
```
radireddy, line 22: Can you rename the function name to better align with returned value like getMinMaxNodesLabel?

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.module.js
```diff
@@ -0,0 +1,14 @@
...
+
+import component from './scale.component';
+import routing from './scale.routing';
+import autoscaling from '../../../components/autoscaling';
+
+const moduleName = 'ovhManagerPciProjectKubernetesNodePoolsScale';
+angular
+  .module(moduleName, [autoscaling])
+  .config(routing)
+  .run(/* @ngTranslationsInject:json ./translations */)
```
radireddy, line 11: no lazy loading the module?

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.routing.js
```diff
@@ -0,0 +1,28 @@
+export default /* @ngInject */ ($stateProvider) => {
+  $stateProvider.state(
+    'pci.projects.project.kubernetes.details.nodepools.scale',
+    {
+      url: '/scale?nodePoolId',
```
radireddy, line 5: if you need a nodePoolId than why can't you have this as a child route of nodepools.nodepool. By doing this you can avoid nodePoolId in query params.

packages/manager/modules/pci/src/projects/project/kubernetes/add/add.controller.js
mohammed-zahaf: yes i understand @marie-j.
my `autoscaling` model is deeply composed.

You can see it here:
https://github.com/ovh/manager/blob/607cde76019fdfe2ff07913df3fa6ed2f625d95c/packages/manager/modules/pci/src/projects/project/kubernetes/kubernetes.routing.js#L57

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
mohammed-zahaf: Yes nice 👍.

packages/manager/modules/pci/src/projects/project/kubernetes/kubernetes.constants.js
mohammed-zahaf: No, unfortunately there aren't default values returned from API.

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/node-pool.class.js
mohammed-zahaf: Yes right.
Fixed.

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
mohammed-zahaf: Fixed.

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
mohammed-zahaf: Fixed.

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.module.js
mohammed-zahaf: It was done into:
`packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/index.js`

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
mohammed-zahaf: Yes it's normal.
If costumer disable the scale option the lowest should be set to 0 but the max to default which is 100 as max.

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.routing.js
mohammed-zahaf: I made it like for the delete component to be consistent.

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/index.js
mohammed-zahaf: Yes, it can be done via the new scale component.

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/translations/Messages_fr_FR.json
mohammed-zahaf: Fixed.

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
mohammed-zahaf: the `&?` is efficient if we passed or not a callback function via the component attribute.
In my case i implement the on-change handlers into the component. So if parent passed a callback fn it will override the implemented one into the child component, so just one will be triggered not both.
For this case i implement this function, if parent passed a callback i checked if it's a function and trigger it.

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.component.js
mohammed-zahaf: No, i don't need two way binding.

packages/manager/modules/pci/src/projects/project/kubernetes/add/add.controller.js
radireddy: I agree with marie. Maybe we can create a nodepool and autoscaling class and initiate it with all this info and pass it to service/api. The autoscaling class can be a composite class in nodepool class. You can move all the complex logic of creating an autoscaling object into these classes. This way it looks clean, easy to understand, and can scale well with further enhancements. Just my thoughts :)

mohammed-zahaf: Why not for a classes.
I will create a Jira task to make refactor cross Kube modules.

Thanks for suggestion.

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
marie-j: @mohammed-zahaf I'm having trouble understanding, doesn't `&?`  binding already catch the case if function is undefined ? Maybe we don't need to check if the binding is a function instead of calling it

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/index.js
marie-j: I wonder if we should add a redirection in this case then

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.html
```diff
@@ -0,0 +1,110 @@
...
+                max="$ctrl.model.nodes.desired.max"
+                on-change="$ctrl.onDesiredValueChange({value: modelValue})"
+                required
+            >
+            </oui-numeric>
+        </oui-field>
+
+        <oui-message
+            data-type="warning"
+            data-ng-if="$ctrl.model.nodes.desired.value < 3"
```
marie-j, line 85: `3` could be stored as a constant maybe

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/index.js
mohammed-zahaf: I am not sure to get your point @marie-j.

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.html
mohammed-zahaf: Fixed.

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
mohammed-zahaf: It do it perfectly, but.
Case to consider:
   Case1: we have this: `onLowestPoolValueChange: '&?',`
   if we dont pass a parameter `data-on-lowest-pool-value-change` the onLowestPoolValueChange will be trigger from 
   autoscaling component.
   Case2: we have this: `onLowestPoolValueChange: '&?',`
    if we passed a parameter `data-on-lowest-pool-value-change` the onLowestPoolValueChange will be triggered only 
    from parent not from autoscaling component.
    
You can try it from your side.

packages/manager/modules/pci/src/projects/project/kubernetes/add/add.controller.js
mohammed-zahaf: We already have a NodePool class which give us the nodePool info (scale, minNodes, maxNodes, ...).
In my case i created a model to manage 3 Numeric values (lowest, desired and highest).
All the logic for the autosacling is implemented into the AutoscalingController.

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/index.js
marie-j: Well if I'm a customer and I access to the `/delete` url directly I will no longer be able to so I wonder if it is relevant to add a redirection in that case or not

mohammed-zahaf: I am not sure it's relevant.
1 - The logic is to navigate from the UI to do things.
2 - To delete nodes our customer must use OVH API note using a saved manager URL.
3 - If it do it via the URL at least he should save one or many urls for eache nodes he had.
4 - If we do i special page to cover the redirection we must planned a new global feature to cover this case. (may be we have this @marie-j ?)

radireddy (dismissed):

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.html
```diff
@@ -0,0 +1,110 @@
...
+        <oui-message
+            data-type="warning"
+            data-ng-if="$ctrl.model.nodes.desired.value < $ctrl.LOWEST_VALUE"
+        >
+            <p data-translate="kubernetes_node_pool_autoscaling_desired_nodes_warning"></p>
+        </oui-message>
+    </div>
+
+    <!--Autoscaling highest config-->
+    <div class="mb-4" data-ng-show="$ctrl.model.autoscale">
```
marie-j, line 92:
```suggestion
    <div class="mb-4" data-ng-if="$ctrl.model.autoscale">
```

marie-j (approved):

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.html
mohammed-zahaf: Thanks for this suggestion.
Initially i used `ng-if` directive but the UI produce an uncomfortable display (take time to build the full elements).
To have a smoothly display i used  `data-ng-show` to avoid to rebuild elements.

Heiiji (approved):

mohammed-zahaf:
Hello @antleblanc , this PR is ready to merge.
Trad: OK
API: OK

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
```diff
@@ -0,0 +1,195 @@
...
+
+  $onInit() {
+    this.model = this.nodePool.autoscaling;
+    const { lowest, desired, highest } = this.model.nodes;
+
+    this.init(true);
+    this.setAutoscalingValidity(lowest.value, desired.value, highest.value);
+    this.$scope.$watch('$ctrl.nodePool.antiAffinity', () =>
+      this.onAntiAffinityChange(),
+    );
```
marie-j, line 24: Can't we use something else ? Like a `onChange` binding ?

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.routing.js
```diff
@@ -0,0 +1,42 @@
...
+      views: {
+        modal: {
+          component: 'pciProjectKubernetesNodePoolsScaleComponent',
+        },
+      },
+      layout: 'modal',
+      params: {
+        nodePoolId: null,
+        nodePoolName: null,
```
marie-j, line 14: Can't we fetch the name from the `nodePoolId` ?

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
mohammed-zahaf: No Marie, the `antiAffinity` is implemented into another step component.
I need it here to follow the changes when customer change the `antiAffinity` in another step.

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.routing.js
mohammed-zahaf: I am not sure to understand.
Name and Id are passed from parent state.

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
marie-j: Is it inside a `oui-step-form` component, in that case maybe we can check when we enter the step that needs the value no ?

mohammed-zahaf: Yes, is't inside a `oui-step-form`.
No i can't, because the autoscaling is a few step before the antiAffinity step.

If customer set the antiaffinity we must set the autosacling limits.
I hope it's more clear @marie-j.

marie-j: I'm having trouble understanding why using a `onChange` binding that updates `$ctrl.model.autoscaling` is not possible even if the step is before (also I'm not sure about the experience here, shouldn't antiaffinity and autoscaling be in another order or in the same step ?

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.routing.js
marie-j: So if we want to access the url directly with the id we have to pass the name as well ? 
Also this means that if we need something else like description, we would pass it as well through parameters ? Not sure it is the best way to achieve things that to pass every info through params, id should be sufficient to get all the required info

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
mohammed-zahaf: Because the switch component is inside a dedicated step component and the autoscaling has it dedicated component.
The both are not connected one to each other. To be able to be follow the changes i used $watch.

About the step, Maxime said it should stay as we already QC it.

packages/manager/modules/pci/src/projects/project/kubernetes/details/node-pool/scale/scale.routing.js
mohammed-zahaf: Thanks for the remark @marie-j.
It's a mistake from my side, i missed to remove the `nodePoolName` param (no needed)

packages/manager/modules/pci/src/projects/project/kubernetes/components/autoscaling/autoscaling.controller.js
marie-j: Simplified example of what I had in mind: 
```
//template.html
<oui-step-form>
 <anti-affinity-component model="$ctrl.antiAffinityModel"></anti-affinity-component>
</oui-step-form>
<oui-step-form>
 <component
  on-change="$ctrl.onChange()>
 </component>
</oui-step-form>

//controller.js
onChange() {
 this.antiAffinityModel= someValue 
}

//antiaffinity.component.js
export default {
 bindings: {
  antiAffinityModel: '<'
 }
}
```

mohammed-zahaf: I have read more about $watch, and it's a not an angularjs antipattern.
I think we can let it @marie-j.

antleblanc (dismissed):

JeremyDec (dismissed):

radireddy (dismissed):

mohammed-zahaf:
Hello @antleblanc and @marie-j.
The QC is done, i pulled the translations.
We can merged it, thanks.

antleblanc (approved):

PR #4833, opened 2021-04-27
Refactor/product offers

JeremyDec:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-6790

## Description

Some refactor work made on product-offers, following CDN Shared recently merged feature

Remove use of `coreConfig.getUserLocale()` as it is not used
Remove user passed as component bindings in web app

### DIFF
--- a/packages/manager/modules/product-offers/src/workflows/product-offers-order-workflow.class.js
+++ b/packages/manager/modules/product-offers/src/workflows/product-offers-order-workflow.class.js
@@ -12,7 +12,6 @@ import { CATALOG_ITEM_TYPE_NAMES } from './product-offers-workflow.constants';
  */
 export default class OrderWorkflow extends Workflow {
   /**
-   * @param {string} locale          User locale
    * @param {Object} $translate      AngularJS provider
    * @param {Object} workflowOptions Specific options for this workflow,
    *  must contains the following values:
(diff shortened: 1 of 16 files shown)


antleblanc (approved):

packages/manager/modules/product-offers/src/workflows/product-offers-order-workflow.class.js
```diff
@@ -68,16 +69,8 @@ export default class OrderWorkflow extends Workflow {
...
-    locale,
-    $q,
-    $timeout,
-    $translate,
-    workflowOptions,
-    WucOrderCartService,
-  ) {
-    super(locale, $q, $translate, workflowOptions);
-    this.$timeout = $timeout;
+  constructor($q, $timeout, $translate, workflowOptions, WucOrderCartService) {
```
marie-j, line 72: As `locale` was the first parameter and even if uses were updates, this looks like a `BREAKING CHANGE`

JeremyDec: IMO, it would be breaking if the updates were not embedded in the same commit.
If you cherry-pick the related commit, the app is no more functioning, so it isn't breaking then, what do you think ?

marie-j: @JeremyDec Imo we should have two commits : one with the breaking change and the other one fixing usage (even though package is private, if it was published it would be best for other users to know the new version is breaking)

JeremyDec: @marie-j this is where I disagree. I don't think it is a good practice to force a breaking change here. 
The changes and its fixes, are all located in the module, and won't change the usage.

radireddy (approved):

packages/manager/modules/product-offers/src/workflows/product-offers-order-workflow.class.js
marie-j: Could the class used here be used outside of the module ? If the answer is yes then it's breaking, if not then it's fine

JeremyDec: Yes the workflow class are used only inside the module

marie-j (approved):

PR #4866, opened 2021-04-30
fix: prevent to reach clickable items when element is hidden

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | 

## Description

In the manager when clicking next to the scrollbar, some account sidebar links can be triggered despite the sidebar being hidden. Also visible when inspecting dom elements.
This PR makes links & button unclickable when the sidebar is hidden.

### DIFF
--- a/packages/manager/modules/account-sidebar/src/index.less
+++ b/packages/manager/modules/account-sidebar/src/index.less
@@ -17,6 +17,10 @@
 
   .manager-hub-user-panel {
     width: 0;
+    a,
+    button {
+      pointer-events: none;
+    }
   }
 
   .sidebar-show() {
@@ -27,6 +31,10 @@
 
     .manager-hub-user-panel {
       width: @ovh-sidebar-width;
+      a,
+      button {
+        pointer-events: auto;
+      }
     }
   }
 


Heiiji (approved):

marie-j (approved):

antleblanc:
Your approach solve this issue you mentioned but I'm wondering if we can fix
this bug by taking a look at what's wrong in term of elements positioning.

If we hover over the account sidebar `.ovh-account-sidebar-wrapper` element
from the DevTools, we are able to see that its not completely moved out of the
viewport.

screenshot-1

Issue that causing this element to be in the viewport is the `padding` style
applied to following element `.manager-hub-user-panel`.

Another approach could be to first, remove this padding and then applying
it only if the account sidebar is toggled (visible in the viewport).

Here are my through:

```css
.manager-hub-user-panel {
    padding: 0;
}

[aria-expanded="true"] .manager-hub-user-panel {
    padding: 3rem 2rem 2rem;
}
```

Please let me know what you think about this approach.
Thank you

frenautvh:
@antleblanc 
I first had the idea of removing the padding and wrapping the content in a div with some margin, it fixes the problem but the animation when closing the sidebar disappear. I couldn't find why since the transition seems to be on the width attribute.

I agree that fixing the padding is the way to go, but since removing it caused some animation issues i provided this quick fix as an alternative. I still think we should merge this one since it solves an issue, then we'll had time to think about how to solve the padding root cause

JeremyDec (approved):

antleblanc:
> @antleblanc
> I first had the idea of removing the padding and wrapping the content in a div with some margin, it fixes the problem but the animation when closing the sidebar disappear. I couldn't find why since the transition seems to be on the width attribute.
> 
> I agree that fixing the padding is the way to go, but since removing it caused some animation issues i provided this quick fix as an alternative. I still think we should merge this one since it solves an issue, then we'll had time to think about how to solve the padding root cause

I agree!

We could apply a stronger solution to address the positioning issue after this one which fit the needs.

As you said we can also optimize the animation. We shouldn't animate `width` property since it triggers layout/paint/composite on each keyframes.

PR #4896, opened 2021-05-05
fix(sms): remaining chars estimate

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | master
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | internal DTRSD-32868

## Description

fix remaining chars in sms composition not correctly displayed

### DIFF
--- a/packages/manager/modules/sms/src/sms/components/sms-composition/sms-composition.controller.js
+++ b/packages/manager/modules/sms/src/sms/components/sms-composition/sms-composition.controller.js
@@ -3,7 +3,8 @@ import { SMS_COMPOSE } from '../../sms/compose/telecom-sms-sms-compose.constant'
 
 export default class SmsCompositionController {
   /* @ngInject */
-  constructor(TucSmsMediator) {
+  constructor($timeout, TucSmsMediator) {
+    this.$timeout = $timeout;
     this.TucSmsMediator = TucSmsMediator;
   }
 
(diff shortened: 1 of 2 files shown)


packages/manager/modules/sms/src/sms/components/sms-composition/sms-composition.html
```diff
@@ -80,8 +80,7 @@ <h2 class="heading_underline" data-translate="sms_message_title"></h2>
                 <oui-checkbox
                     name="noCommercialClause"
                     id="noCommercialClause"
-                    model="$ctrl.model.noStopClause"
-                    on-change="$ctrl.getMessageDetails(modelValue)"
```
JeremyDec, line 84: No more calculation made after checkbox changed ?

packages/manager/modules/sms/src/sms/components/sms-composition/sms-composition.controller.js
```diff
@@ -12,16 +13,20 @@ export default class SmsCompositionController {
...
-      SmsCompositionController.isShortNumber(this.model.sender)
-        ? true
-        : this.model.noStopClause;
-    const suffix = noStopValue !== undefined ? noStopValue : isShort;
-    this.model.messageDetails = this.TucSmsMediator.getSmsInfoText(
-      this.model.message,
-      !suffix,
-    );
+    // timeout is here because the onChange event of the oui-textarea
+    // is triggered before the model is updated
```
marie-j, line 17: Isn't the new model the value passed as parameter ?

marie-j (approved):

packages/manager/modules/sms/src/sms/components/sms-composition/sms-composition.html
frenautvh: from what i understand only the noStopClause does change the calculation, am i missing something ?

JeremyDec: Oh I didn't saw you change from `noStopClause` to `noCommercialClause`, my bad, it's ok then

JeremyDec (approved):

antleblanc (approved):

packages/manager/modules/sms/src/sms/components/sms-composition/sms-composition.controller.js
frenautvh: yes you're right, the issue here is that this method is also used as a callback for the on-change method on some radio-buttons in the template, so changing the method signature (textarea value as parameter) will conflict with the behavior of the radio buttons that are passing their model value as parameter

PR #4977, opened 2021-05-18
fix(security): display popover template for maxPasswordAge field

antleblanc:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | DTRSD-37373

## Description

Following patch is inspired by the way that popover are currently displayed into this section.
See: https://github.com/ovh/manager/blob/c7074416bc31eb2af8ba15cdd80dd16271a4b253/packages/manager/modules/exchange/src/security/security.html#L47-L54

We could imagine as a long-term solution to rely on following component:
- https://ovh.github.io/ovh-ui-kit/?path=/story/design-system-components-field-webcomponent--label-popover

### :bug: Bug Fixes

c707441 - fix(security): display popover template for maxPasswordAge field

### DIFF
--- a/packages/manager/modules/exchange/src/security/security.html
+++ b/packages/manager/modules/exchange/src/security/security.html
@@ -169,7 +169,10 @@
             </oui-field>
         </div>
 
-        <h2 class="oui-heading_underline">Sécurité des comptes</h2>
+        <h2
+            class="oui-heading_underline"
+            data-translate="exchange_ACTION_configure_services_label_security_account"
+        ></h2>
 
         <!-- Complexity Enabled -->
         <oui-checkbox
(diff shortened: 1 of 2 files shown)


packages/manager/modules/exchange/src/security/security.html
```diff
@@ -169,7 +169,10 @@
             </oui-field>
         </div>
 
-        <h2 class="oui-heading_underline">Sécurité des comptes</h2>
```
antleblanc, line 172: Moved content into its own translation file.

frenautvh (approved):

cbourgois (approved):

PR #4986, opened 2021-05-20
Feature/hub partial loading

zakariasahmane:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | no
| Breaking change? | yes
| Tickets | Fix MANAGER-6678

## Description

Splitting the calls of the hub to reduce the time the preloader is shown before the first drawing of the page. 
Instead of waiting for all the API to finish we show skeletons for the API's that finished loading, so that the user can start interacting with what's already loaded.

### DIFF
--- a/packages/manager/apps/hub/src/dashboard/dashboard.html
+++ b/packages/manager/apps/hub/src/dashboard/dashboard.html
@@ -1,8 +1,6 @@
 <div data-ui-view>
     <div class="hub-dashboard-content">
-        <hub-dashboard-welcome
-            data-name="$ctrl.me.firstname"
-        ></hub-dashboard-welcome>
+        <hub-dashboard-welcome></hub-dashboard-welcome>
         <div
             class="d-flex flex-wrap w-100 minw-0 align-items-center justify-content-between"
         >
@@ -12,7 +10,6 @@
             ></ovh-manager-banner>
             <ovh-manager-hub-carousel
                 class="w-100"
-                items="$ctrl.notifications"
                 tracking-prefix="{{ $ctrl.trackingPrefix }}"
             ></ovh-manager-hub-carousel>
         </div>
@@ -46,7 +43,6 @@
             <ovh-manager-hub-support
                 class="col-md-8 mb-3 mb-md-4 order-2 order-md-3"
                 me="$ctrl.me"
-                tickets="$ctrl.tickets"
                 tracking-prefix="{{ $ctrl.trackingPrefix }}"
             >
             </ovh-manager-hub-support>
@@ -59,14 +55,11 @@
             >
             </hub-order-tracking>
         </div>
-        <div data-ng-if="$ctrl.products.length" class="hub-dashboard-product">
+        <div class="hub-dashboard-product">
             <h2 data-translate="manager_hub_dashboard_services"></h2>
             <hub-products
-                products="$ctrl.products"
-                on-product-select="$ctrl.goToProductPage(productType)"
+                products-promise="$ctrl.products"
                 tracking-prefix="{{ $ctrl.trackingPrefix }}"
-                on-expand="$ctrl.expandProducts(expand)"
-                expand="$ctrl.expand"
             ></hub-products>
         </div>
     </div>
--- a/packages/manager/apps/hub/src/dashboard/routing.js
+++ b/packages/manager/apps/hub/src/dashboard/routing.js
@@ -1,5 +1,22 @@
 import { filter, get, groupBy, map, reverse, sortBy } from 'lodash-es';
 
+const getProducts = (services, order, catalog) => {
+  return get(services, 'data.count') === 0 && !order
+    ? groupBy(
+        filter(catalog.data, ({ highlight }) => highlight),
+        'universe',
+      )
+    : reverse(
+        sortBy(
+          map(services.data?.data, (service, productType) => ({
+            ...service,
+            productType,
+          })),
+          'count',
+        ),
+      );
+};
+
 export default /* @ngInject */ ($stateProvider, $urlRouterProvider) => {
   $stateProvider.state('app.dashboard', {
     url: '/?expand',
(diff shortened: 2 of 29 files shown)


packages/manager/modules/hub/src/components/products/controller.js
```diff
@@ -4,17 +4,52 @@ import {
...
     this.DISPLAYED_PRODUCTS_NUMBER = DISPLAYED_PRODUCTS_NUMBER;
+    this.$state = $state;
+    this.atInternet = atInternet;
+    this.loading = true;
+  }
+
+  $onInit() {
+    this.skeletonServices = Array.from({ length: 6 });
+    this.skeletonProducts = Array.from({ length: 4 });
+    this.products
```
frenautvh, line 19: it looks unclear to use a promise for "products" component binding, it makes more sense to expect a list of products

packages/manager/modules/hub/src/components/catalog-items/controller.js
```diff
@@ -1,4 +1,18 @@
 export default class {
+  constructor() {
+    this.loading = true;
+  }
+
+  $onInit() {
+    this.items
```
frenautvh, line 7: i expect items to be a list of items, maybe use a method like fetchItems instead ? to make it clear that it's a promise

packages/manager/modules/hub/src/components/billing-summary/controller.js
```diff
@@ -8,46 +8,57 @@ export default class ManagerHubBillingSummaryCtrl {
...
+      .catch(() => {
+        this.billsError = true;
+      })
+      .finally(() => {
+        this.loading = false;
+      });
+  }
+
+  loadBills() {
+    return this.$q.when(this.fetchBills()).then(({ data }) => {
```
frenautvh, line 43: i don't think $q.when is necessary here

packages/manager/modules/hub/src/components/billing-summary/controller.js
```diff
@@ -8,46 +8,57 @@ export default class ManagerHubBillingSummaryCtrl {
...
+        data.total,
+        get(data, 'currency.code'),
+      );
+      this.buildPeriodFilter(data.period);
+      return this.bills;
+    });
+  }
+
+  loadDebt() {
+    return this.$q.when(this.fetchDebt()).then(({ data }) => {
```
frenautvh, line 55: i don't think $q.when is necessary here

frenautvh (changes requested):

packages/manager/modules/hub/src/components/products/controller.js
zakariasahmane: I agree. 

Though this is the result of using the `resolvePolicy : NOWAIT` in the caller component. 

I think it would be tricky to pass the products directly to the component bindings because the caller has no controller where we can handle the loading states, but i can actually rename the binding so that products would be "products-promise" or something like that to make it clear that it's not the products, what do you think?

frenautvh: yes something like ```productsPromise``` would be better suited :)

packages/manager/apps/hub/src/dashboard/routing.js
```diff
@@ -57,10 +73,12 @@ export default /* @ngInject */ ($stateProvider, $urlRouterProvider) => {
       breadcrumb: /* @ngInject */ ($translate) =>
         $translate.instant('manager_hub_dashboard'),
     },
-    componentProvider: /* @ngInject */ (order, services) =>
-      get(services, 'data.count') === 0 && !order
-        ? 'hubOrderDashboard'
-        : 'hubDashboard',
+    resolvePolicy: {
+      async: 'NOWAIT',
```
Jisay, line 77: nice :-)

packages/manager/apps/hub/src/dashboard/dashboard.html
```diff
@@ -12,7 +10,6 @@
             ></ovh-manager-banner>
             <ovh-manager-hub-carousel
                 class="w-100"
-                items="$ctrl.notifications"
```
cbourgois, line 15: We should be sure of that because if you remove the items property here, you will fetch `notifications` API to display the carousel, but `notifications` API is also called to initialize notifications badges in navbar + notifications sidebar.

cbourgois (approved):
Could be great to rebase and squash some commit and add a `ref: MANAGER-6678` in commit body.

packages/manager/apps/hub/src/dashboard/dashboard.html
```diff
@@ -1,8 +1,6 @@
 <div data-ui-view>
     <div class="hub-dashboard-content">
-        <hub-dashboard-welcome
-            data-name="$ctrl.me.firstname"
-        ></hub-dashboard-welcome>
+        <hub-dashboard-welcome data-me="$ctrl.me"></hub-dashboard-welcome>
```
marie-j, line 3: binding doesn't seem needed anymore

cbourgois (dismissed):

Jisay (approved):

frenautvh (dismissed):

packages/manager/modules/hub/src/components/billing-summary/controller.js
```diff
@@ -8,46 +8,57 @@ export default class ManagerHubBillingSummaryCtrl {
...
+        this.loading = false;
+      });
+  }
+
+  loadBills() {
+    return this.fetchBills().then(({ data }) => {
+      this.bills = data;
+      this.formattedBillingPrice = this.getFormattedPrice(
+        data.total,
+        get(data, 'currency.code'),
```
marie-j, line 47: nitpick: we could
```suggestion
        data?.currency?.code,
```

packages/manager/modules/hub/src/components/carousel/carousel.controller.js
```diff
@@ -1,8 +1,27 @@
...
+  constructor($q, $http, $translate, atInternet) {
     this.atInternet = atInternet;
     this.index = 0;
+    this.$http = $http;
+    this.$q = $q;
+    this.$translate = $translate;
+  }
+
+  $onInit() {
+    if (!this.items || this.items.length === 0) {
```
marie-j, line 14: why check the items length ? if there is no notification why refetch them ?

zakariasahmane: The idea was to leave the items binding if we want to call the component with items other than notifications( or if the caller already has notifications items). If the caller doesn't specify anything, we fetch notifications by default

Maybe we could change the condition or document it, after reading your comment, now i'm thinking that the caller could want to pass an empty items object, maybe bind a boolean instead? like "fetchNotifications" in the component to replace the condition?

packages/manager/modules/hub/src/components/billing-summary/controller.js
zakariasahmane: We could, i'll check if rollup is configured to accept this, if it is, i'll change it everywhere. :)

packages/manager/modules/hub/src/components/carousel/carousel.controller.js
marie-j: @zakariasahmane it depends on the usage but maybe we could just check `!this.items` and create a Breaking change on the component

packages/manager/modules/hub/src/components/billing-summary/controller.js
marie-j: @zakariasahmane did you figure out if rollup handles it ?

zakariasahmane: Hi @marie-j !

Sorry i didn't give feedback ! Actually it looks like it's handled ( i tested it for this module and everything works just fine ), but instead of removing it just here, it will be the subject of another PR to remove lodash-es when not necessary, there are a lot of packages that could benefit from removing it.

marie-j (approved):

marie-j:
Hello @zakariasahmane 
Could you please add the ticket reference to your commit ?
Thanks

antleblanc (approved):

PR #4989, opened 2021-05-21
feat(bm-server-components): add general info component

radireddy:
ref: MANAGER-6930

| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-6930

## Description

Added below baremetal reusable components
1. Baremetal server general info tile


packages/manager/modules/bm-server-components/src/general-info-tile/index.js
```diff
@@ -0,0 +1,22 @@
+import angular from 'angular';
```
JeremyDec, line 1: You should have a `@ovh-ux/ui-kit` import here, and pass it as dep of the `angular.module(...)`

packages/manager/modules/bm-server-components/src/general-info-tile/template.html
```diff
@@ -0,0 +1,146 @@
...
+            data-ng-if="!$ctrl.server.isExpired"
+            aria-label="{{:: 'menu_action_label' | translate:{ t0: ('server_configuration_boot' | translate) } }}"
+            data-compact
+            data-placement="end"
+        >
+            <oui-action-menu-item
+                data-on-click="$ctrl.editNetboot('choice')"
+                data-disabled="$ctrl.server.isOsInstallationAllowed() ||
+                    $ctrl.server.state === 'HACKED' ||
+                    $ctrl.server.state === 'HACKED_BLOCKED' ||
```
JeremyDec, line 112: **suggestion**
It could be interesting to have this logic into `server` class

packages/manager/modules/bm-server-components/src/datacenter/template.html
```diff
@@ -0,0 +1 @@
+<span data-ng-bind="$ctrl.getFormatedName()"> </span>
```
marie-j, line 1:
```suggestion
<span data-ng-bind="$ctrl.getFormatedName()"></span>
```

packages/manager/modules/bm-server-components/src/general-info-tile/controller.js
```diff
@@ -0,0 +1,31 @@
...
+
+export default class BmServerGeneralInfoTileController {
+  /* @ngInject */
+  constructor($translate, coreConfig) {
+    this.$translate = $translate;
+    this.region = coreConfig.getRegion();
+  }
+
+  showNameActions() {
+    return this.region !== 'US' && isFunction(this.onEditName);
```
marie-j, line 11: Check on region could be done using Feature Flipping maybe ?

packages/manager/modules/bm-server-components/src/general-info-tile/index.js
```diff
@@ -0,0 +1,26 @@
...
+
+import '@ovh-ux/ui-kit';
+import 'angular-translate';
+import '@ovh-ux/ng-translate-async-loader';
+import '@ovh-ux/ng-at-internet';
+import ovhManagerCore from '@ovh-ux/manager-core';
+
+import Datacenter from '../datacenter';
```
marie-j, line 9:
```suggestion
import datacenter from '../datacenter';
```

packages/manager/modules/bm-server-components/src/general-info-tile/template.html
```diff
@@ -0,0 +1,145 @@
...
+            </oui-action-menu-item>
+        </oui-action-menu>
+    </oui-tile-definition>
+    <!-- commercial range -->
+    <oui-tile-definition
+        term="{{:: 'server_configuration_description' | translate }}"
+    >
+        <oui-tile-description>
+            <span
+                data-ng-bind="$ctrl.server.commercialRange !== 'housing' ? $ctrl.server.description || '-' : ('server_configuration_description_housing' | translate)"
```
marie-j, line 29: Maybe the check on commercialRange could be extracted from the template

marie-j (commented):
Constant file may be removed if empty

packages/manager/modules/bm-server-components/src/datacenter/index.js
```diff
@@ -0,0 +1,13 @@
+import angular from 'angular';
+import 'angular-translate';
+
```
JeremyDec, line 3:
```suggestion
import '@ovh-ux/ng-translate-async-loader';
```

packages/manager/modules/bm-server-components/src/general-info-tile/template.html
```diff
@@ -0,0 +1,145 @@
...
+            >
+                <span
+                    data-translate="server_configuration_installation_button"
+                ></span>
+            </oui-action-menu-item>
+        </oui-action-menu>
+    </oui-tile-definition>
+    <!-- boot -->
+    <oui-tile-definition
+        data-ng-if="$ctrl.server.commercialRange !== 'housing'"
```
ganeshkumar1989, line 86: Probably the commercialRange Types can be moved to a constant file

ganeshkumar1989 (approved):

packages/manager/modules/bm-server-components/src/general-info-tile/controller.js
radireddy: This was the old code. I will do this change in a separate ticket.

varun257:
`Breaking commit` is missing while adding new module

frenautvh:
in my opinion you should also refactor dedicated manager to use this component in the same PR

we should be able to perform some QA by checking that the infos are still correctly displayed in the manager, also we should be able to keep the git history of those files

by copying you are introducing code duplication, we are loosing the history of changes and also we'll have to backport fixes or changes from dedicated manager to your component

frenautvh (changes requested):
i'm requesting changes since i think the manager dedicated should be refactored in the same pull request (for being able to QA your PR and also keep git history)

radireddy:
> in my opinion you should also refactor dedicated manager to use this component in the same PR
> 
> we should be able to perform some QA by checking that the infos are still correctly displayed in the manager, also we should be able to keep the git history of those files
> 
> by copying you are introducing code duplication, we are loosing the history of changes and also we'll have to backport fixes or changes from dedicated manager to your component

@frenautvh This was the plan decided by the team. The plan was to redevelop these components from scratch using latest coding guidelines. Use them in both dedicated and Nutanix. No git move and making changes to existing code.

PR #4994, opened 2021-05-25
Fix/sass tilde import

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | internal MANAGER-6999

## Description

See : https://webpack.js.org/loaders/sass-loader/#resolving-import-at-rules

using `~` in sass import rules is deprecated

### DIFF
--- a/packages/manager/apps/telecom/src/app/app-scss.scss
+++ b/packages/manager/apps/telecom/src/app/app-scss.scss
@@ -1,4 +1,4 @@
-@import 'bootstrap4/scss/functions';
-@import 'bootstrap4/scss/variables';
-@import 'bootstrap4/scss/mixins';
-@import 'bootstrap4/scss/_utilities.scss';
+@import 'bootstrap4/scss/_functions';
+@import 'bootstrap4/scss/_variables';
+@import 'bootstrap4/scss/_mixins';
+@import 'bootstrap4/scss/_utilities';
--- a/packages/manager/apps/telecom/src/app/app-scss.scss
+++ b/packages/manager/apps/telecom/src/app/app-scss.scss
@@ -1,4 +1,4 @@
-@import '~bootstrap4/scss/functions';
-@import '~bootstrap4/scss/variables';
-@import '~bootstrap4/scss/mixins';
-@import '~bootstrap4/scss/_utilities.scss';
+@import 'bootstrap4/scss/functions';
+@import 'bootstrap4/scss/variables';
+@import 'bootstrap4/scss/mixins';
+@import 'bootstrap4/scss/_utilities.scss';
--- a/packages/manager/modules/pci/src/projects/project/workflow/index.scss
+++ b/packages/manager/modules/pci/src/projects/project/workflow/index.scss
@@ -1,9 +1,9 @@
 ovh-manager-kubernetes-list {
-  @import '~bootstrap4/scss/_functions';
-  @import '~bootstrap4/scss/_variables';
-  @import '~bootstrap4/scss/_mixins';
-  @import '~bootstrap4/scss/_utilities';
-  @import '~bootstrap4/scss/_grid';
+  @import 'bootstrap4/scss/_functions';
+  @import 'bootstrap4/scss/_variables';
+  @import 'bootstrap4/scss/_mixins';
+  @import 'bootstrap4/scss/_utilities';
+  @import 'bootstrap4/scss/_grid';
 }
 
 .no-label .oui-radio__label-container {
--- a/packages/manager/tools/component-rollup-config/src/index.ts
+++ b/packages/manager/tools/component-rollup-config/src/index.ts
@@ -2,6 +2,7 @@ import { get, isString, mergeWith } from 'lodash';
 import babel from '@rollup/plugin-babel';
 import camelcase from 'camelcase';
 import commonjs from '@rollup/plugin-commonjs';
+import fs from 'fs';
 import html from 'rollup-plugin-html';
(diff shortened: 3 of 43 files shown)


packages/manager/tools/component-rollup-config/src/index.ts
```diff
@@ -52,6 +61,19 @@ const generateConfig = (opts, pluginsOpts) =>
         sass({
           insert: true,
           output: false,
+          options: {
+            importer(url) {
```
frenautvh, line 65: in order to handle sass imports correctly in rollup i had to manually resolve url using node resolver
(see: https://rollupjs.org/guide/en/#why-isnt-node-resolve-a-built-in-feature)

packages/manager/apps/telecom/src/app/app-scss.scss
```diff
@@ -1,4 +1,4 @@
-@import '~bootstrap4/scss/functions';
```
frenautvh, line 1: those paths were wrong

Jisay (approved):

zakariasahmane (dismissed):

cbourgois (approved):

antleblanc:
@frenautvh Looks like there is a mismatch between versions for both `webpack`
and `sass-loader` from here (according to your description).

Here we are still using `webpack@4`, so we should rely on following
documentation: https://v4.webpack.js.org/loaders/sass-loader/#resolving-import-at-rules

As you can see, it also reports the warning about `~` usage but if we pay
attention to the `sass-loader` changelog, this warning only appears since
`sass-loader@11.0.0`.
See: https://github.com/webpack-contrib/sass-loader/releases/tag/v11.0.0

According to the `yarn.lock` file, here we are still using `sass-loader@8.0.2`
so I'm wondering if following changes make sense or if some addition
dependencies requires to be upgraded too.

Please let me know your thought about it.
Thanks

frenautvh:
> @frenautvh Looks like there is a mismatch between versions for both `webpack`
> and `sass-loader` from here (according to your description).
> 
> Here we are still using `webpack@4`, so we should rely on following
> documentation: https://v4.webpack.js.org/loaders/sass-loader/#resolving-import-at-rules
> 
> As you can see, it also reports the warning about `~` usage but if we pay
> attention to the `sass-loader` changelog, this warning only appears since
> `sass-loader@11.0.0`.
> See: https://github.com/webpack-contrib/sass-loader/releases/tag/v11.0.0
> 
> According to the `yarn.lock` file, here we are still using `sass-loader@8.0.2`
> so I'm wondering if following changes make sense or if some addition
> dependencies requires to be upgraded too.
> 
> Please let me know your thought about it.
> Thanks

I think it would make easier to upgrade later or switch to another bundler by not using the `~` syntax.
From sass-loader webpack point of view i see that removing it has no impact,

antleblanc:
As you are updating 70%+ Sass imports, are you considering, for consistency,
to "import" Sass files with (or without) `.scss` extension?

I'm still having trouble to understand why we are applying changes "encouraged"
by `sass-loader@11.0.0` whereas we keep using `sass-loader@8.0.2`.

According to Sass language documentation:
> The Sass team discourages the continued use of the @import rule..

see: https://sass-lang.com/documentation/at-rules/import

frenautvh:
> As you are updating 70%+ Sass imports, are you considering, for consistency,
> to "import" Sass files with (or without) `.scss` extension?

Yes we could use the same syntax everywhere ; it depends if we explicitely want to specify that we are importing sass or not. Apart from that it doesn't make any difference since both syntax are valid.

> I'm still having trouble to understand why we are applying changes "encouraged"
> by `sass-loader@11.0.0` whereas we keep using `sass-loader@8.0.2`.

The '~' import syntax is supported by sass-loader from webpack but let's say we are using another bundler instead of webpack. Then we may have trouble importing existing sass styles since we are using the '~' syntax introduced by sass-loader. It's better to make our sass code not dependant on sass-loader.

packages/manager/modules/pci/src/projects/project/workflow/index.scss
```diff
@@ -1,9 +1,9 @@
 ovh-manager-kubernetes-list {
-  @import '~bootstrap4/scss/_functions';
-  @import '~bootstrap4/scss/_variables';
-  @import '~bootstrap4/scss/_mixins';
-  @import '~bootstrap4/scss/_utilities';
-  @import '~bootstrap4/scss/_grid';
+  @import 'bootstrap4/scss/_functions';
```
marie-j, line 2: It looks like we are importing bootstrap4 a lot, maybe we can import it at a parent level (out of scope, just a thought)

marie-j (dismissed):

packages/manager/modules/pci/src/projects/project/workflow/index.scss
frenautvh: Yes we might have something to do here, i also found that scoping bootstrap might look like it's a best practice but in reality it makes the css dist a LOTS heavier : each bootstrap's class definition is duplicated for the new scope.

JeremyDec (dismissed):

Heiiji (dismissed):

frenautvh:
rebased on develop 0a30f6da7109b85aa33bb8c2adfdf47831eaae94

marie-j (approved):

JeremyDec (approved):

Heiiji (approved):

PR #5000, opened 2021-05-25
Feature/Remove typescript

zakariasahmane:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no

## Description

Remove Typescript.

### :zap: Performance

6b5b460 - perf: rename *.ts files into *.js

### :construction_worker: Build

c784131 - build(dev-deps): add missing eslint-formatter-pretty


cbourgois (dismissed):

antleblanc:
Congratulations, you reached Pull Request 5000 :tada: :)

According to your first commit message, I'm wondering if `feat` really make
sense here.

From the given packages impacted here, I don't see any feature addition.

- @ovh-ux/component-rollup-config
- @ovh-ux/manager-webpack-config
- @ovh-ux/manager-webpack-dev-server

Same remark with the scope `ts` which provide non-needed information in my point
of view.

Looks like you are only upgrading some devDependencies for the packages listed
above.

According to `yarn why typescript` command, it also reports following dependency:
```sh
...
=> Found "testcafe#typescript@3.8.3"
```

As you are replacing typescript here from `typescript@3.7.3` to `typescript@4.2.4`,
you have to make sure that it keeps working correctly with `testcafe`.
For such, see with @fchabrolle @blaryjp.

To be honest, for the three packages listed above, there are no benefit to use
TypeScript syntax. Additional trade off, it brings an extra compilation phase
(around 9sec (3sec per package)) which slow down the build.

I would recommend to switch back from `*.ts` to `*.js` for theses packages since
it don't bring any additional values.

zakariasahmane:
> Congratulations, you reached Pull Request 5000 🎉 :)
> 
> According to your first commit message, I'm wondering if `feat` really make
> sense here.
> 
> From the given packages impacted here, I don't see any feature addition.
> 
>     * @ovh-ux/component-rollup-config
> 
>     * @ovh-ux/manager-webpack-config
> 
>     * @ovh-ux/manager-webpack-dev-server
> 
> 
> Same remark with the scope `ts` which provide non-needed information in my point
> of view.
> 
> Looks like you are only upgrading some devDependencies for the packages listed
> above.
> 
> According to `yarn why typescript` command, it also reports following dependency:
> 
> ```shell
> ...
> => Found "testcafe#typescript@3.8.3"
> ```
> 
> As you are replacing typescript here from `typescript@3.7.3` to `typescript@4.2.4`,
> you have to make sure that it keeps working correctly with `testcafe`.
> For such, see with @fchabrolle @blaryjp.
> 
> To be honest, for the three packages listed above, there are no benefit to use
> TypeScript syntax. Additional trade off, it brings an extra compilation phase
> (around 9sec (3sec per package)) which slow down the build.
> 
> I would recommend to switch back from `*.ts` to `*.js` for theses packages since
> it don't bring any additional values.

Hi @antleblanc ! Thanks for your feedback ! 

I agree about the branch and commit naming, it's actually a bit off you're right.

As to why we're bumping the ts version is not for these packages directly. It's because we have noticed that if you add any vue repo generated with some of the latest `vue-cli` versions, the other repos such as the hub for instance won't build anymore because there will be a conflicting version of typescript. 
The main reason this happens, is because `vue-cli` also uses `webpack-dev-server`, but in later versions. And the version it uses actually writes types for this package in `typescript@4.1.0` at least and the latest version of `webpack-dev-server` uses the `typescript@4.2.4` version. 
This makes it, that the `tsc` used in the monorepo to build our sites blocks the build because he's unable to understand the` *.d.ts` files that came with the `vue-cli` newly generated project. 
This wasn't an issue a couple months ago but it is now so i thought that since these three packages won't break from 3.9 to 4.2 it doesn't cost anything to do it. 

But if it actually is costly, maybe we can totally get rid of ts in these packages you're right. 
We can also maybe explore the possibility of using Vite instead of vue-cli in newly generated Vue projects. But that only solves the vue-cli issue, and if any other package in the future uses conflicting versions like this with something we already use, we will encouter the same issue.

antleblanc (dismissed):
Could you please update commit type according to previous comments?

In addition, Did you received inputs regarding TestCafé usage based on this new version of TypeScript? (if required)

Or do we plan, for some packages under following workspace `packages/manager/tools` to move from `*.ts` to `*.js`?

Thank you!

zakariasahmane:
> Could you please update commit type according to previous comments?
> 
> In addition, Did you received inputs regarding TestCafé usage based on this new version of TypeScript? (if required)
> 
> Or do we plan, for some packages under following workspace `packages/manager/tools` to move from `*.ts` to `*.js`?
> 
> Thank you!

Yes sure, i'll update the commit. 

And for testcafé, actually i'm rewriting everything from .ts to .js right now. Should be better

packages/manager/tools/component-rollup-config/package.json
```diff
@@ -62,14 +50,12 @@
     "slash": "^2.0.0"
   },
   "devDependencies": {
-    "@typescript-eslint/eslint-plugin": "^2.11.0",
-    "@typescript-eslint/parser": "^2.11.0",
+    "@ovh-ux/component-rollup-config": "^8.0.0 || ^9.0.0",
```
frenautvh, line 53: dev dependency of itself ?

yarn.lock
```diff
@@ -20888,10 +20888,10 @@ typescript@^3.3.3:
   resolved "https://registry.yarnpkg.com/typescript/-/typescript-3.8.3.tgz#409eb8544ea0335711205869ec458ab109ee1061"
   integrity sha512-MYlEfn5VrLNsgudQTVJeNaQFUAI7DkhnOjdpAp4T+ku1TfQClewlbSuTVHiA+8skNBgaf02TL/kLOvig4y3G8w==
 
-typescript@^3.7.3:
-  version "3.7.3"
-  resolved "https://registry.yarnpkg.com/typescript/-/typescript-3.7.3.tgz#b36840668a16458a7025b9eabfad11b66ab85c69"
-  integrity sha512-Mcr/Qk7hXqFBXMN7p7Lusj1ktCBydylfQM/FZCk5glCNQJrCUKPkMHdo9R0MTFWsC/4kPFvDS0fDPvukfCkFsw==
+typescript@^4.2.4:
```
frenautvh, line 20891: yarn why typescript ?

packages/manager/tools/component-rollup-config/package.json
zakariasahmane: Nice, should be removed, a lousy copy/paste i forgot !

yarn.lock
zakariasahmane: The yarn lock file should be updated, there's only ts left in testcafe and no need for it in 4.2.4, i'll update the yarn lock, thanks ! :)

cbourgois (approved):

Jisay (approved):

packages/manager/tools/webpack-dev-server/src/config.js
```diff
@@ -30,7 +30,7 @@ export = (env) => {
     proxy.unshift(serverProxy.aapi);
   }
   if (env.dev) {
-    proxy.unshift(...env.dev.map((config) => serverProxy.dev(config)));
+    proxy.unshift(...env.dev.map((c) => serverProxy.dev(c)));
```
marie-j, line 33: Issue with leaving `config` as the name of parameter ?

antleblanc (changes requested):

packages/manager/tools/webpack-dev-server/src/config.js
zakariasahmane: Yep, it gives the error "config is already declared in the upper scope. 

Although, now that i re-read the code, i think i can rename it "config" and remove the upper-scope object exported to export it directly ! Thanks !

cbourgois (approved):

Jisay (approved):

antleblanc:
Rebasing on `develop` branch.

antleblanc:
Rebased successfully on `develop` branch.

marie-j (approved):

PR #5021, opened 2021-06-01
Feat/bm components ipmi

radireddy:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/baremetal-server-components` 
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-6937

## Description

Added ipmi tab, ipmi restart and ipmi order components.


packages/manager/modules/bm-server-components/src/ipmi/controller.js
```diff
@@ -0,0 +1,690 @@
...
+      pingDone: false,
+
+      buttonStart: false,
+      navigationLoading: false,
+      navigationReady: null,
+      javaLoading: false,
+      javaReady: false,
+      sshLoading: false,
+      kvmhtmlLoading: false,
+      solSshKeyLoading: false,
```
frenautvh, line 62: are those infos really needed ? i'm not sure it's used

packages/manager/modules/bm-server-components/src/ipmi/controller.js
```diff
@@ -0,0 +1,690 @@
...
+  }
+
+  startIpmiTestStatus() {
+    this.disable.testActive = true;
+    this.disable.testIpmi = true;
+  }
+
+  startIpmiPollPing(task) {
+    this.startIpmiTestStatus();
+    this.setHttpState(false, true, false);
```
frenautvh, line 176: i think having to call this function like that is kind of bad practice :
```js
this.setHttpState(false, true, false)
```

false, true, false doesn't means a lot when you are reading the code, it's kind of difficult to maintain

It looks like http state could simply be an enumeration, something like :
```js
this.httpState = this.STATE_LOADING;
this.httpState = this.STATE_ERROR;
// etc
```

if a state can only have one value at a time they is no need to have a separate variable for each state value

packages/manager/modules/bm-server-components/src/ipmi/order-kvm/template.html
```diff
@@ -0,0 +1,72 @@
...
+                <span>{{ option.duration | wucDuration:'longDate' }}</span>
+                <strong
+                    class="text-danger float-right"
+                    data-ng-bind-html="option.prices | ducPrice:$ctrl.user.ovhSubsidiary"
+                >
+                </strong>
+                <!-- <ovh-manager-catalog-price
+                    data-price="offer.price"
+                    data-user="$ctrl.user"
+                > -->
```
frenautvh, line 34: to remove

packages/manager/modules/bm-server-components/src/ipmi/order-kvm/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,12 @@
+{
+  "bm_server_ipmi_order_kvm": "Commander un KVM",
+  "bm_server_kvm_order_step1_title": "Choisissez la durée",
+  "bm_server_kvm_order_step2_title": "Commande d'un KVM sur le serveur <em>{{serverName}}</em>",
+  "bm_server_kvm_order_step2_duration": "Période :",
+  "bm_server_kvm_order_step2_details": "Détails de la commande :",
+  "bm_server_kvm_order_step2_price": "Prix :",
+  "bm_server_kvm_order_finish_success": "La commande numéro <a href='{{orderUrl}}' target='_blank'>{{orderId}}</a> a bien été prise en compte.",
+  "bm_server_kvm_order_error": "Une erreur est survenue lors de la commande de KVM: {{message}}",
+  "bm_server_common_confirm": "Confirmez",
```
frenautvh, line 10:
```suggestion
  "bm_server_common_confirm": "Confirmer",
```

packages/manager/modules/bm-server-components/src/ipmi/restart/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,8 @@
+{
+  "bm_server_ipmi_restart_title": "Redémarrage de l'IPMI",
+  "bm_server_ipmi_restart_information": "Voulez-vous vraiment redémarrer le module IPMI ?",
+  "bm_server_ipmi_restart_error": "Une erreur est survenue lors de la demande de redémarrage du module IPMI: {{ message }}",
+  "bm_server_ipmi_restart_loading": "Le module IPMI est en cours de redémarrage.",
+  "bm_server_common_confirm": "Confirmez",
```
frenautvh, line 6:
```suggestion
  "bm_server_common_confirm": "Confirmer",
```

packages/manager/modules/bm-server-components/src/ipmi/controller.js
frenautvh: oh i see that this code already exists so just ignore my comments regarding code review since it's not in the scope of your PR

frenautvh:
if you are refactoring existing code in a new component it should be great if you could also remove the code from dedicated in order to avoid having the code duplicated in two places in the monorepo

packages/manager/modules/bm-server-components/src/ipmi/controller.js
radireddy: yes all these are used.

packages/manager/modules/bm-server-components/src/ipmi/controller.js
radireddy: Its a valid comment. I am rewriting the IPMI feature as a reusable component that can be used in dedicated server and nutanix. Its better to remove the legacy code and write new code as per our latest guidelines. I will fix this comment.

radireddy:
> if you are refactoring existing code in a new component it should be great if you could also remove the code from dedicated in order to avoid having the code duplicated in two places in the monorepo
> 
> by doing that you'll also be able to send your refactoring to QA

Yes I am rewriting the IPMI feature as a reusable component that can be used in dedicated server and nutanix. But replacing the code in dedicated is not planned as part of this task. Myself along with both the Cyrils has come up with this plan. First step is to rewrite the code as reusable components. Second step is to use it in Nutanix and third step is to use them in Dedicated server.

radireddy:
opening a new PR for this.

PR #5015, opened 2021-05-28
fix(*): remove less tilde imports

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | internal MANAGER-7041
| License          | BSD 3-Clause

## Description

remove @ovh-ux/rollup-plugin-less-tilde-importer usage (~ import syntax in less files)as the dependency is deprecated
also the benefits is to make our stack simpler and more respectful of standards

PR #5048, opened 2021-06-02
feat: add manager components module

ganeshkumar1989:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? |no
| Tickets          | Fix #MANAGER-7074

## Description

Add Manager Components module to house all reusable UI components, to be used, across the manager. Currently, the resource-selector component has been added.

### DIFF
--- a/packages/manager/modules/manager-components/package.json
+++ b/packages/manager/modules/manager-components/package.json
@@ -1,0 +1,23 @@
+{
+  "name": "@ovh-ux/manager-components",
+  "version": "0.0.0",
+  "private": true,
+  "description": "A set of reusable UI components, to be used across the OVH Manager",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/ovh/manager.git",
+    "directory": "packages/manager/modules/manager-components"
+  },
+  "license": "BSD-3-Clause",
+  "author": "OVH SAS",
+  "sideEffects": false,
+  "main": "./src/index.js",
+  "dependencies": {
+    "bootstrap4": "twbs/bootstrap#v4.0.0",
+    "lodash": "^4.17.15"
+  },
+  "peerDependencies": {
+    "@ovh-ux/ui-kit": "^4.4.3",
+    "angular": "^1.7.5"
+  }
+}
--- a/packages/manager/modules/manager-components/src/index.js
+++ b/packages/manager/modules/manager-components/src/index.js
@@ -1,0 +1,3 @@
+import resourceSelector from './resource-selector';
+
+export default resourceSelector;
--- a/packages/manager/modules/manager-components/src/resource-selector/README.md
+++ b/packages/manager/modules/manager-components/src/resource-selector/README.md
@@ -1,0 +1,71 @@
+# ovh-manager-resource-selector
+
+The `ovh-manager-resource-selector` component provides a quick way to create a horizontal tile, with the price and additional controls included. This is usually used, but not limited to, for resource selection and hence the name.
+
+## Installation
+
+```js
+import 'angular';
+import resourceSelector from '@ovh-ux/manager-components';
+
+angular.module('myModule', [resourceSelector]);
+```
+
+## Attributes
+
+| Attribute         | Type            | Binding | One-time binding | Values                    | Default    | Description
+| ----              | ----            | ----    | ----             | ----                      | ----       | ----
+| `price`            | string          | @       | no              | n/a                       | n/a        | The price to be shown
+| `price-suffix`      | string          | @?      | no              | n/a                       | n/a        | This is a suffix that gets added to the price (in the next line)
+| `title`       | string          | @?      | no              | n/a  | n/a  | The title for the resource selector tile
+
+## Basic Usage
+Below is the simplest form of using this component, where all the attributes are supplied
+
+```html
+<ovh-manager-resource-selector
+    price="100 €"
+    price-suffix="/month"
+    title="My title"
+>
+</ovh-manager-resource-selector>
+```
+
+## Transclude price
+Sometimes, you might need to format the price differently, or might need to customise it further. In this case, the `ovh-manager-resource-selector` component supports a named transclude `resourceSelectorPrice`. When this named transclude is used, it overrides the `price` and the `price-suffix` attributes, which are no longer required. Below is an example usage
+
+```html
+<ovh-manager-resource-selector
+    title="My title"
+>
+    <resource-selector-price>
+        <p>$100/Month/Resource</p>
+        <p>Only</p>
+    </resource-selector-price>
+</ovh-manager-resource-selector>
+```
+
+## Default Transclude
+The component also supports a default transclude. This content gets transcluded to the right of the price slot.
+
+```html
+<ovh-manager-resource-selector
+    title="My title"
+>
+    <resource-selector-price>
+        <p>$100/Month/Resource</p>
+        <p>Only</p>
+    </resource-selector-price>
+    <div>
+        <p>Default transclude content here</p>
+    </div>
+</ovh-manager-resource-selector>
+```
+
+## Contributing
+
+Always feel free to help out! Whether it's [filing bugs and feature requests](https://github.com/ovh/manager/issues/new) or working on some of the [open issues](https://github.com/ovh/manager/issues), our [contributing guide](https://github.com/ovh/manager/blob/master/CONTRIBUTING.md) will help get you started.
+
+## License
+
+[BSD-3-Clause](LICENSE) © OVH SAS
--- a/packages/manager/modules/manager-components/src/resource-selector/component.js
+++ b/packages/manager/modules/manager-components/src/resource-selector/component.js
@@ -1,0 +1,13 @@
+import template from './template.html';
+
+export default {
+  bindings: {
+    price: '<?',
+    priceSuffix: '@?',
+    title: '@',
+  },
+  template,
+  transclude: {
+    price: '?resourceSelectorPrice',
+  },
+};
--- a/packages/manager/modules/manager-components/src/resource-selector/index.js
+++ b/packages/manager/modules/manager-components/src/resource-selector/index.js
@@ -1,0 +1,13 @@
+import angular from 'angular';
+import '@ovh-ux/ui-kit';
+
+import component from './component';
+import './index.scss';
+
+const moduleName = 'ovhManagerComponentsResourceSelector';
+
+angular
+  .module(moduleName, ['oui'])
+  .component('managerResourceSelector', component);
+
+export default moduleName;
--- a/packages/manager/modules/manager-components/src/resource-selector/index.scss
+++ b/packages/manager/modules/manager-components/src/resource-selector/index.scss
@@ -1,0 +1,6 @@
+ovh-manager-resource-selector {
+  @import 'bootstrap4/scss/_functions.scss';
+  @import 'bootstrap4/scss/_variables.scss';
+  @import 'bootstrap4/scss/_mixins.scss';
+  @import 'bootstrap4/scss/_utilities.scss';
+}
--- a/packages/manager/modules/manager-components/src/resource-selector/template.html
+++ b/packages/manager/modules/manager-components/src/resource-selector/template.html
@@ -1,0 +1,8 @@
+<div class="oui-tile d-flex m-2">
+    <h6 class="w-50" data-ng-bind="$ctrl.title"></h6>
+    <div class="text-right ml-auto mr-3" ng-transclude="price">
+        <p class="oui-heading_6 m-0" data-ng-bind="$ctrl.price"></p>
+        <span data-ng-bind=":: $ctrl.priceSuffix"></span>
+    </div>
+    <div class="oui-tile__body" ng-transclude></div>
+</div>
--- a/packages/manager/modules/manager-components/README.md
+++ b/packages/manager/modules/manager-components/README.md
@@ -1,0 +1,25 @@
+# manager-components
+
+## Install
+
+```sh
+yarn add @ovh-ux/manager-components
+```
+
+## Usage
+This module supports importing only the required component, instead of importing the entire module at 1 shot. For example, to import a component named `my-component`, below is the code. Refer to the individual component's documentation, to find out exactly how to use the respective components.
+
+```js
+import 'angular';
+import myComponent from '@ovh-ux/manager-components';
+
+angular.module('myApp', [myComponent]);
+```
+
+## Contributing
+
+Always feel free to help out! Whether it's [filing bugs and feature requests](https://github.com/ovh/manager/issues/new) or working on some of the [open issues](https://github.com/ovh/manager/issues), our [contributing guide](https://github.com/ovh/manager/blob/master/CONTRIBUTING.md) will help get you started.
+
+## License
+
+[BSD-3-Clause](LICENSE) © OVH SAS


packages/manager/modules/manager-components/package.json
```diff
@@ -0,0 +1,23 @@
+{
+  "name": "@ovh-ux/manager-manager-components",
```
radireddy, line 2:
```suggestion
  "name": "@ovh-ux/manager-components",
```

packages/manager/modules/manager-components/src/resource-selector/component.js
```diff
@@ -0,0 +1,14 @@
+import template from './template.html';
+
+export default {
+  bindings: {
+    currencySymbol: '@?',
```
radireddy, line 5: is this not available in price? price will have price text which has currency symbol. We need to avoid taking currencySymbol as input. If required you can parse price and get it.

packages/manager/modules/manager-components/src/resource-selector/index.js
```diff
@@ -0,0 +1,13 @@
+import angular from 'angular';
+import '@ovh-ux/ui-kit';
+
+import component from './component';
+import './index.scss';
+
+const moduleName = 'ovhManagerManagerComponentsResourceSelector';
```
radireddy, line 7:
```suggestion
const moduleName = 'ovhManagerComponentsResourceSelector';
```

packages/manager/modules/manager-components/src/index.js
```diff
@@ -0,0 +1,9 @@
+import angular from 'angular';
+
+import resourceSelector from './resource-selector';
+
+const moduleName = 'ovhManagerManagerComponents';
```
radireddy, line 5:
```suggestion
const moduleName = 'ovhManagerComponents';
```

packages/manager/modules/manager-components/src/resource-selector/index.js
```diff
@@ -0,0 +1,13 @@
...
+
+angular
+  .module(moduleName, ['oui'])
+  .component('ovhManagerResourceSelector', component);
```
radireddy, line 11: can we avoid ovh? just rename it to managerResourceSelector. This will avoid impact incase company name is changed.

packages/manager/modules/manager-components/package.json
```diff
@@ -0,0 +1,23 @@
+{
```
radireddy, line 1: can you fill README.md file. Add example on how to use the components.

packages/manager/modules/manager-components/src/index.js
```diff
@@ -0,0 +1,9 @@
...
+
+angular.module(moduleName, [resourceSelector]);
+
+export default moduleName;
```
radireddy, line 9: This will export all components in this module irrespective of them being used or not. Lets find a way to support only components needed like the way date-fn does.

packages/manager/modules/manager-components/src/resource-selector/component.js
```diff
@@ -0,0 +1,14 @@
...
+    price: '<?',
```
radireddy, line 6: accept a price object instead of price and currencySymbol

packages/manager/modules/manager-components/src/resource-selector/template.html
```diff
@@ -0,0 +1,11 @@
+<div class="oui-tile d-flex m-2">
+    <h6 class="w-50" data-ng-bind="$ctrl.title"></h6>
+    <div class="text-right ml-auto mr-3" ng-transclude="price">
+        <h6 class="m-0">
+            <span data-ng-bind="$ctrl.price"></span>
```
radireddy, line 5: does this info shown if resourceSelectorPrice is provided?

radireddy (changes requested):

packages/manager/modules/manager-components/src/index.scss
```diff
@@ -0,0 +1,5 @@
+@import '~bootstrap4/scss/_functions.scss';
```
marie-j, line 1: I wonder if this should be scoped to the component (in case we use this in an app that still use bootstrap 3

packages/manager/modules/manager-components/src/resource-selector/template.html
```diff
@@ -0,0 +1,11 @@
+<div class="oui-tile d-flex m-2">
+    <h6 class="w-50" data-ng-bind="$ctrl.title"></h6>
+    <div class="text-right ml-auto mr-3" ng-transclude="price">
+        <h6 class="m-0">
```
marie-j, line 4: Isn't the `h6` overridden with the transclude here ? 
Also maybe we should just use the style as we don't know if other headings will be provided across the same view

packages/manager/modules/manager-components/package.json
ganeshkumar1989: The name of the module is `manager-components`. The prefix `@ovh-ux/manager-` gets added automatically

ganeshkumar1989: Changed

packages/manager/modules/manager-components/src/resource-selector/component.js
ganeshkumar1989: This has been removed. The user will have to pass a string for the price, that includes the currency as well

packages/manager/modules/manager-components/src/resource-selector/index.js
ganeshkumar1989: Changed

packages/manager/modules/manager-components/src/index.js
ganeshkumar1989: Changed

packages/manager/modules/manager-components/src/resource-selector/index.js
ganeshkumar1989: Done

packages/manager/modules/manager-components/package.json
ganeshkumar1989: Done

packages/manager/modules/manager-components/src/index.js
ganeshkumar1989: Yeah... Probably, we can have it as a separate story...

packages/manager/modules/manager-components/src/resource-selector/component.js
ganeshkumar1989: The price has to be a string that also contains the currency. The currencySymbol has been removed

packages/manager/modules/manager-components/src/resource-selector/template.html
ganeshkumar1989: No... The user can override it using the `resourceSelectorPrice`. The user can directly give the price string. But `resourceSelectorPrice` takes higher precedence

packages/manager/modules/manager-components/src/index.scss
ganeshkumar1989: Done

packages/manager/modules/manager-components/src/resource-selector/template.html
ganeshkumar1989: Yeah... The user can override it using the resourceSelectorPrice. The user can directly give the price string or transclude the required content using `resourceSelectorPrice`. `resourceSelectorPrice` takes higher precedence

packages/manager/modules/manager-components/src/resource-selector/component.js
radireddy: ok then is it possible to use string binding?

radireddy (approved):

radireddy (dismissed):

marie-j:
Hello @ganeshkumar1989 
Could you amend https://github.com/ovh/manager/pull/5048/commits/f677061d0fe62b8249d002a0ac649c9594688833 so BREAKING CHANGE mention appear last ? (If not, the ticket reference will appear in Changelog) 

Thanks

ganeshkumar1989:
> Hello @ganeshkumar1989
> Could you amend [f677061](https://github.com/ovh/manager/commit/f677061d0fe62b8249d002a0ac649c9594688833) so BREAKING CHANGE mention appear last ? (If not, the ticket reference will appear in Changelog)
> 
> Thanks

This is done @marie-j . Thanks

packages/manager/modules/manager-components/src/index.js
```diff
@@ -0,0 +1,2 @@
+/* eslint-disable import/prefer-default-export */
```
marie-j, line 1: No other solution than to disable eslint ?

radireddy (dismissed):

packages/manager/modules/manager-components/src/index.js
JeremyDec: You could do: 
```
import { default as managerResourceSelector } from './resource-selector`
export default managerResourceSelector;
```

packages/manager/modules/manager-components/src/resource-selector/index.scss
```diff
@@ -0,0 +1,14 @@
+ovh-manager-resource-selector {
+  @import '~bootstrap4/scss/_functions.scss';
+  @import '~bootstrap4/scss/_variables.scss';
+  @import '~bootstrap4/scss/_mixins.scss';
+  @import '~bootstrap4/scss/_utilities.scss';
+  @import '~bootstrap4/scss/_grid.scss';
```
JeremyDec, line 6: Keep in mind that there's a work to remove less tilde imports (see: https://github.com/ovh/manager/pull/5015).

packages/manager/modules/manager-components/package.json
```diff
@@ -0,0 +1,23 @@
...
+  "license": "BSD-3-Clause",
+  "author": "OVH SAS",
+  "main": "./src/index.js",
+  "dependencies": {
+    "bootstrap4": "twbs/bootstrap#v4.0.0",
+    "lodash": "^4.17.15"
+  },
+  "peerDependencies": {
+    "@ovh-ux/ui-kit": "^4.4.3",
+    "@uirouter/angularjs": "^1.0.23",
```
JeremyDec, line 20: I don't see this dependency used by your module, you can remove this then

packages/manager/modules/manager-components/src/resource-selector/template.html
```diff
@@ -0,0 +1,8 @@
+<div class="oui-tile d-flex m-2">
+    <h6 class="w-50" data-ng-bind="$ctrl.title"></h6>
+    <div class="text-right ml-auto mr-3" ng-transclude="price">
+        <p class="oui-heading_6 m-0" data-ng-bind="$ctrl.price"></p>
+        <span data-ng-bind=":: $ctrl.priceSuffix"></span>
+    </div>
+    <div class="oui-tile__body" ng-transclude></div>
```
JeremyDec, line 7: **question**:
No unwanted effect to have transclude slot unnamed aside to one slot with name ?

packages/manager/modules/manager-components/src/index.js
ganeshkumar1989: It seems to be the same. I will have to add `/* eslint-disable import/no-named-default */` if I go with this suggestion :)

packages/manager/modules/manager-components/src/resource-selector/index.scss
ganeshkumar1989: Yeah... But I think the scss imports have to be handled separately...
#5015 handles only `.less` imports

packages/manager/modules/manager-components/package.json
ganeshkumar1989: Done. Thanks :)

packages/manager/modules/manager-components/src/resource-selector/template.html
ganeshkumar1989: No side-effects... Its tested :)

packages/manager/modules/manager-components/src/index.js
ganeshkumar1989: I will have to add `/* eslint-disable import/no-named-default */` if I use the above suggestion... I think it is more or less the same....

packages/manager/modules/manager-components/src/resource-selector/index.scss
JeremyDec: my bad, yes it does not concern scss imports

Heiiji (approved):

packages/manager/modules/manager-components/src/index.js
mohammed-zahaf: @ganeshkumar1989, to avoid to add an ESLINT exception, i think we can just import it like this:
```suggestion
export managerResourceSelector from './resource-selector';
```

For info, when you import something that is exported with `default` keyword, we can rename it like we would.
I hope it will helps.

ganeshkumar1989: This does not seem to work either... I get an error "Declaration or statement expected"...

packages/manager/modules/manager-components/src/resource-selector/index.scss
cbourgois: Hello, the PR concerning `~` on sass import was merged : https://github.com/ovh/manager/pull/4994

packages/manager/modules/manager-components/src/index.js
cbourgois:
```js
import { default as managerResourceSelector } from './resource-selector` 
```
will cause eslint `import/no-named-default` error, because you don't have to destructure the `default` when importing.


something similar to @JeremyDec 's proposal should works : 
```js
import resourceSelector from './resource-selector` 
export default resourceSelector;
```

packages/manager/modules/manager-components/src/resource-selector/index.scss
```diff
@@ -0,0 +1,14 @@
...
+
+  .oui-tile__body {
+    .oui-field {
```
cbourgois, line 9: Why should we need custom css ?

cbourgois (changes requested):
The objective of this module is to expose some components globally for the manager. How to avoid to import everything if we only need one component in it ?

packages/manager/modules/manager-components/src/index.js
ganeshkumar1989: This is done now. Thanks a lot for the suggestions/help

packages/manager/modules/manager-components/src/resource-selector/index.scss
ganeshkumar1989: This has been removed. Thanks

packages/manager/modules/manager-components/src/resource-selector/index.scss
ganeshkumar1989: This has been corrected. Thanks

ganeshkumar1989:
> The objective of this module is to expose some components globally for the manager. How to avoid to import everything if we only need one component in it ?

After a discussion, it was agreed that we will not have a dist generated for this module... The importing app will generate the dist for the imported part.

mohammed-zahaf (approved):

packages/manager/modules/manager-components/package.json
cbourgois: I don't see where is the documentation ?

varun257 (approved):

marie-j (dismissed):

marie-j (approved):

cbourgois (approved):
The component's README is a good thing! 👍

ganeshkumar1989:
> The component's README is a good thing! 👍

Thanks @cbourgois  :)

marie-j (approved):

packages/manager/modules/manager-components/package.json
```diff
@@ -0,0 +1,23 @@
...
+    "url": "git+https://github.com/ovh/manager.git",
+    "directory": "packages/manager/modules/manager-components"
+  },
+  "license": "BSD-3-Clause",
+  "author": "OVH SAS",
+  "sideEffects": false,
+  "main": "./src/index.js",
+  "dependencies": {
+    "bootstrap4": "twbs/bootstrap#v4.0.0",
+    "lodash": "^4.17.15"
```
antleblanc, line 17: Are you sure that `lodash` is required?

packages/manager/modules/manager-components/src/resource-selector/README.md
```diff
@@ -0,0 +1,71 @@
...
+import resourceSelector from '@ovh-ux/manager-components';
+
+angular.module('myModule', [resourceSelector]);
+```
+
+## Attributes
+
+| Attribute         | Type            | Binding | One-time binding | Values                    | Default    | Description
+| ----              | ----            | ----    | ----             | ----                      | ----       | ----
+| `price`            | string          | @       | no              | n/a                       | n/a        | The price to be shown
```
antleblanc, line 18: Binding differ from component definition file.
see: https://github.com/ovh/manager/pull/5048/files#diff-d2a23a5eefe28602b3290df7da36607b30ff0c42e1c5c5de14c2816015cb1768R5

antleblanc (commented):
Sorry to come late on the review process but it seems that it's remains some errors.
Please let me know what do you think.
Thank you!

Issue #4099, opened 2020-11-26
Rollup and optional chaining


marie-j:
**Describe the bug**
When using optional chaining in a module bundled with rollup, an error is thrown `(plugin commonjs) SyntaxError: Unexpected token `

**Additional context**
Looks like this issue was fixed in version 2.11.0 of rollup (https://github.com/rollup/rollup/blob/master/CHANGELOG.md#pull-requests-51)

antleblanc:
Hi @marie-j!

Yup! We will bump `rollup` on our shared rollup configuration package located here:

- https://github.com/ovh/manager/tree/master/packages/manager/tools/component-rollup-config

**Currently**:

```sh
$ yarn why rollup | grep Found
=> Found "rollup@2.0.2"
```

**Pending branch**:

- https://github.com/ovh/manager/compare/develop...bugfix/upgrade-rollup-config?expand=1#diff-055d85d1d80602832dc33d851fc197cc1d6d30381fd10d79448f19012d02f13eR57

antleblanc:
**package.json**:
```json
{
  "name": "foo",
  "version": "1.0.0",
  "main": "index.js",
  "license": "MIT",
  "devDependencies": {
    "@ovh-ux/component-rollup-config": "^9.0.1"
  },
  "scripts": {
    "build": "rollup -c"
  }
}
```

**rollup.config.js**:
```js
import rollupConfig from '@ovh-ux/component-rollup-config';

const config = rollupConfig({
  input: './src/index.js',
});

const outputs = [
  config.es({
    output: {
      sourcemap: false,
    },
  }),
];

export default outputs;
```

**src/index.js**:

```js
const adventurer = {
  name: 'Alice',
  cat: {
    name: 'Dinah'
  }
};

const dogName = adventurer.dog?.name;
console.log(dogName);
// expected output: undefined

console.log(adventurer.someNonExistentMethod?.());
// expected output: undefined
```

**dist/esm/index.js**:

```js
$ cat dist/esm/index.js 
function ___$insertStyle(css) {
  if (!css) {
    return;
  }
  if (typeof window === 'undefined') {
    return;
  }

  var style = document.createElement('style');

  style.setAttribute('type', 'text/css');
  style.innerHTML = css;
  document.head.appendChild(style);
  return css;
}

var _adventurer$dog, _adventurer$someNonEx;

var adventurer = {
  name: 'Alice',
  cat: {
    name: 'Dinah'
  }
};
[...]

PR #5062, opened 2021-06-03
fix(service): check if billing pricing value exists

antleblanc:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | DTRSD-38569, DTRSD-39838

### :bug: Bug Fixes

161d60a - fix(service): check if billing pricing value exists

### DIFF
--- a/packages/manager/modules/models/service/Service.class.js
+++ b/packages/manager/modules/models/service/Service.class.js
@@ -103,7 +103,7 @@ export default class Service {
   isEngaged() {
     return (
       !!this.billing.engagement ||
-      !!this.billing.pricing.engagementConfiguration
+      !!(this.billing.pricing && this.billing.pricing.engagementConfiguration)
     );
   }
 }


JeremyDec (approved):

packages/manager/modules/models/service/Service.class.js
```diff
@@ -103,7 +103,7 @@ export default class Service {
   isEngaged() {
     return (
       !!this.billing.engagement ||
-      !!this.billing.pricing.engagementConfiguration
+      !!(this.billing.pricing && this.billing.pricing.engagementConfiguration)
```
frenautvh, line 106:
```suggestion
      !!(this.billing.pricing?.engagementConfiguration)
```

antleblanc: You're right! But currently we are facing an issue with Rollup config that's why, as an Hotfix, I've picked this solution.

marie-j (approved):

packages/manager/modules/models/service/Service.class.js
marie-j: Related to https://github.com/ovh/manager/issues/4099

antleblanc: #5064 solves this issue.
Still need some more investigation. We kill keep you posted.
Thank you!

PR #5065, opened 2021-06-04
fix(form): reorder legal field value and update wording

Heiiji:
| Question         | Answer
| ---------------- | ---
| Branch?          |`develop`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-5342

## Description

It is some rewording and a custom order on legals list. 

## Related

- [x] TRANS-35640

### DIFF
--- a/packages/manager/modules/sign-up/src/form/activity/translations/Messages_fr_FR.json
+++ b/packages/manager/modules/sign-up/src/form/activity/translations/Messages_fr_FR.json
@@ -9,12 +9,13 @@
   "sign_up_activity_field_vat_personalcorporation_more": "Numéro d'identification d'entreprise ({{ vatLabel }})",
   "sign_up_activity_field_organisation_association": "Nom de l'association",
   "sign_up_activity_field_organisation_corporation": "Nom de l'entreprise",
-  "sign_up_activity_field_organisation_administration": "Nom de l'administration",
+  "sign_up_activity_field_organisation_administration": "Raison sociale",
   "sign_up_activity_field_organisation_personalcorporation": "Nom de l'entreprise individuelle",
-  "sign_up_activity_field_organisation": "Organisation",
+  "sign_up_activity_field_organisation": "Raison sociale",
   "sign_up_activity_field_corporationType": "Type d'entreprise",
   "sign_up_activity_field_nationalIdentificationNumber": "Numéro d'identification national",
   "sign_up_activity_field_companyNationalIdentificationNumber": "Numéro d'identification national",
-  "sign_up_activity_field_companyNationalIdentificationNumber_fr": "Numéro d'identification national (SIREN)",
+  "sign_up_activity_field_companyNationalIdentificationNumber_fr": "Numéro d'identification national (SIRET)",
+  "sign_up_activity_field_companyNationalIdentificationNumber_fr_tooltip": "Le numéro SIRET est nécessaire afin d’améliorer le processus de facturation électronique mise en place par l’état (Chorus)",
   "sign_up_activity_field_italian_sdi": "Code du destinataire pour la facturation électronique (SDI)"
 }
--- a/packages/manager/modules/sign-up/src/form/form.controller.js
+++ b/packages/manager/modules/sign-up/src/form/form.controller.js
@@ -80,6 +80,14 @@ export default class SignUpFormCtrl {
       this.model,
     );
 
+    const legalformOrder = {
+      corporation: 0,
+      individual: 1,
+      administration: 2,
+      association: 3,
+      other: 4,
+    };
+
     return this.signUp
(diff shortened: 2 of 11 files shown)


packages/manager/modules/sign-up/src/form/form.controller.js
```diff
@@ -102,6 +102,31 @@ export default class SignUpFormCtrl {
...
+                  prev.splice(1, 1, cur);
+                  break;
+                default:
+                  prev.push(cur);
+              }
+              return prev;
+            },
+            [null, null, null, null],
+          )
+          .filter((el) => el);
```
radireddy, line 128: If I am correct, the output after reduce will be
["corporation", "individual", "administration", "association"]
Is this what you are trying to get?

Heiiji: ["corporation", "individual", "administration", "association", "other"] for Fr, but for US we dant have "administration" for exemple and i'm trying to make a résilient order if API add other.

frenautvh: i think it should be simpler, maybe take a look here : https://stackoverflow.com/questions/16221197/sort-javascript-array-in-a-pre-defined-order

ganeshkumar1989: What I see from the code is that you are trying to sort the values in a particular order. Is that all required here?

Heiiji: yes we want order them : ["corporation", "individual", "administration", "association", "other"]
but i want to make it resilient and if API change, this is working. For exemple remove Organisation and add "gaming" for exemple. 

It's useless maybe, so i can change method for the @frenautvh ref, the first thread awnser crawl the array much time (each indexOf + ref array) where i crawl array 2 time. So i'm looking for second one (but it need to be a predictible array)

frenautvh: @Heiiji i think it doesn't matter if the first thread answer is slower that your solution, the array is so small that there is no need to seek performance, you shoud favor code readability in this case

frenautvh (dismissed):

JeremyDec (approved):

packages/manager/modules/sign-up/src/form/activity/translations/Messages_fr_FR.json
```diff
@@ -9,12 +9,13 @@
...
+  "sign_up_activity_field_organisation_administration": "Raison sociale",
   "sign_up_activity_field_organisation_personalcorporation": "Nom de l'entreprise individuelle",
-  "sign_up_activity_field_organisation": "Organisation",
+  "sign_up_activity_field_organisation": "Raison sociale",
   "sign_up_activity_field_corporationType": "Type d'entreprise",
   "sign_up_activity_field_nationalIdentificationNumber": "Numéro d'identification national",
   "sign_up_activity_field_companyNationalIdentificationNumber": "Numéro d'identification national",
-  "sign_up_activity_field_companyNationalIdentificationNumber_fr": "Numéro d'identification national (SIREN)",
+  "sign_up_activity_field_companyNationalIdentificationNumber_fr": "Numéro d'identification national (SIRET)",
+  "sign_up_activity_field_companyNationalIdentificationNumber_fr_tooltip": "Le numéro SIRET est nécessaire afin d’améliorer le processus de facturation électronique mise en place par l’état (Chorus)",
```
radireddy, line 19: is it only applicable for FR? what about other subsidiaries?

packages/manager/modules/sign-up/src/form/form.controller.js
```diff
@@ -80,6 +80,14 @@ export default class SignUpFormCtrl {
       this.model,
     );
 
+    const legalformOrder = {
```
marie-j, line 83: Can't the order be handled by the API ?

packages/manager/modules/sign-up/src/form/activity/translations/Messages_fr_FR.json
Heiiji: "SIRET" is present only with this key yes, for translation a request is pending :)

marie-j: I think what @radireddy meant is that you will display the popover for FR but what about other subsidiaries that has a `companyNationalIdentificationNumber`

packages/manager/modules/sign-up/src/form/form.controller.js
antleblanc: We can request this kind of enhancement.
Since this ticket already received a validation I will merge it in order to be able to deploy it.
Thank you!

antleblanc (approved):

PR #5121, opened 2021-06-15
refactor(dedicated): move billing section into module

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | internal MANAGER-7162

## Description

create "billing" module, first step to the final goal of making dedicated's billing section standalone


packages/manager/modules/billing/package.json
```diff
@@ -0,0 +1,49 @@
+{
+  "name": "@ovh-ux/manager-billing",
+  "version": "0.0.0-alpha",
```
cbourgois, line 3: I think we could use a `.prerelease` file for that
https://github.com/ovh/manager/blob/develop/scripts/common/repository.js#L243

Because I think the release script won't release an alpha.

cbourgois (dismissed):

packages/manager/modules/billing/package.json
frenautvh: thanks, just added it here :
https://github.com/ovh/manager/pull/5121/files#diff-98c4c0a9eda678e34b5c6f1aa0521862a24274272e0a2d3c6b90274c492ac7c1R1

packages/manager/apps/dedicated/package.json
```diff
@@ -25,6 +25,7 @@
     "@ovh-ux/manager-at-internet-configuration": "^0.0.0 || ^1.0.0",
     "@ovh-ux/manager-banner": "^1.1.3",
     "@ovh-ux/manager-beta-preference": "^1.0.0",
+    "@ovh-ux/manager-billing": "^0.0.0-alpha || ^1.0.0",
```
marie-j, line 28: Shouldn't it be something like 
```suggestion
    "@ovh-ux/manager-billing": "^0.0.0 || ^0.1.0-alpha",
```
?

cbourgois (dismissed):

cbourgois (approved):

Jisay (approved):

PR #5126, opened 2021-06-16
Feat/vps migrate page modal

varun257:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-7062, #MANAGER-7061 

## Description

add migration details for new and current plan
add migration confirmation modal


packages/manager/modules/vps/package.json
```diff
@@ -21,6 +21,7 @@
   "peerDependencies": {
     "@ovh-ux/manager-advices": "^0.0.0 || ^1.0.0",
     "@ovh-ux/manager-billing": "^2.0.0",
+    "@ovh-ux/manager-catalog-price": "^1.0.0",
```
marie-j, line 24: Not needed to add it in both apps `dedicated` and `vps` ?

packages/manager/modules/vps/src/dashboard/migrate/index.scss
```diff
@@ -0,0 +1,6 @@
+oui-select-picker {
+  .oui-select-picker__body ul li {
+    justify-content: space-between;
+    display: flex;
```
marie-j, line 4: This looks achievable with some bootstrap4 utilities

packages/manager/modules/vps/src/dashboard/migrate/migrate-confirm/migrate-confirm.controller.js
```diff
@@ -0,0 +1,40 @@
...
+  /* @ngInject */
+  constructor($translate, VpsMigrationService) {
+    this.$translate = $translate;
+    this.MIGRATE_CONFIRMATION_INPUT_PATTERN = MIGRATE_CONFIRMATION_INPUT_PATTERN;
+    this.vpsMigrationService = VpsMigrationService;
+  }
+
+  cancel() {
+    this.goBackToMigrate();
+  }
```
marie-j, line 14: Why not directly use `$ctrl.goBackToMigrate` ?

packages/manager/modules/vps/src/dashboard/migrate/migrate.constants.js
```diff
@@ -0,0 +1,6 @@
+export const MIGRATE_FAQ_LINK =
+  'https://www.ovhcloud.com/lang/vps/vps-offer-migration/';
```
marie-j, line 2:
```suggestion
  'https://www.ovhcloud.com/lang/vps/vps-offer-migration/';
```
/lang ? Shouldn't we replace for each subsidiary ?

packages/manager/modules/vps/src/dashboard/migrate/migrate.controller.js
```diff
@@ -0,0 +1,42 @@
...
+  constructor() {
+    this.MIGRATE_FAQ_LINK = MIGRATE_FAQ_LINK;
+  }
+
+  $onInit() {
+    set(
+      this.newPlan,
+      'configuration',
+      this.parseRangeConfiguration(this.newPlan.planCode),
+    );
```
marie-j, line 16: Maybe we should recreate the property instead of updating an existing binding

packages/manager/modules/vps/src/dashboard/migrate/migrate.controller.js
```diff
@@ -0,0 +1,42 @@
...
+  }
+
+  /* eslint-disable-next-line class-methods-use-this */
```
marie-j, line 19: Prefer using a `static` method instead of disabling eslint

packages/manager/modules/vps/src/dashboard/migrate/migrate.controller.js
```diff
@@ -0,0 +1,42 @@
...
+
+    return {
+      cpu: { cores },
+      memory: { size: memory },
+      storage: { disks: [{ capacity: storage }] },
+    };
+  }
+
+  getRenewablePrice() {
+    return get(this.newPlan, 'pricings').find(({ capacities }) =>
```
marie-j, line 38:
```suggestion
    return get(this.newPlan, 'pricings', []).find(({ capacities }) =>
```
if not get is not useful here you could simply this.newPlan?.pricings

packages/manager/modules/vps/src/dashboard/migrate/migrate.html
```diff
@@ -0,0 +1,110 @@
...
+            <span data-translate="vps_dashboard_migrate_faq_title"></span>
+            <a
+                data-ng-href="{{:: $ctrl.MIGRATE_FAQ_LINK }}"
+                target="_blank"
+                rel="noopener"
+            >
+                <span
+                    data-translate="vps_dashboard_migrate_faq_link_text"
+                ></span>
+                <span class="fa fa-external-link" aria-hidden="true"></span>
```
marie-j, line 51: Prefer using icons from ui-kit

packages/manager/modules/vps/src/dashboard/migrate/migrate.service.js
```diff
@@ -0,0 +1,11 @@
+export default class VpsMigrationService {
+  /* @ngInject */
+  constructor($q, $http) {
+    this.$q = $q;
```
marie-j, line 4:
```suggestion
```

packages/manager/modules/vps/src/dashboard/migrate/migrate.service.js
```diff
@@ -0,0 +1,11 @@
...
+    this.$http = $http;
+  }
+
+  migrateVps(serviceName) {
+    return this.$http.post(`/vps/${serviceName}/migration2016`);
```
marie-j, line 9:
```suggestion
    return this.$http.post(`/vps/${serviceName}/migration2016`)
    .then(({data}) => data)
```
If not I think having a service is a bit overkill here if we don't do anything apart from one call

marie-j (changes requested):

packages/manager/modules/vps/src/dashboard/migrate/migrate.module.js
```diff
@@ -0,0 +1,19 @@
+import angular from 'angular';
+import '@uirouter/angularjs';
+import '@ovh-ux/ui-kit';
+
+import routing from './migrate.routing';
+import component from './migrate.component';
+import confirm from './migrate-confirm';
+import vpsMigrationService from './migrate.service';
+
```
JeremyDec, line 9: This should have `import ovhManagerCatalogPrice from '@ovh-ux/manager-catalog-price'` + inject the dep into `angular.module`

packages/manager/modules/vps/src/dashboard/migrate/migrate.controller.js
```diff
@@ -0,0 +1,42 @@
...
+    const [cores, memory, storage] = this.extractConfigurationFromPlanCode(
+      rangeFullName,
+    );
+
+    return {
+      cpu: { cores },
+      memory: { size: memory },
+      storage: { disks: [{ capacity: storage }] },
+    };
+  }
```
JeremyDec, line 35: This logic is a duplicate of this: 
- https://github.com/ovh/manager/blob/master/packages/manager/modules/vps/src/upscale/upscale.controller.js#L352
- https://github.com/ovh/manager/blob/master/packages/manager/modules/vps/src/upscale/upscale.controller.js#L366

Maybe an helper service could be build for the two use cases?
Also, `/* eslint-disable [...]` should be avoid

packages/manager/modules/vps/src/dashboard/migrate/migrate.controller.js
```diff
@@ -0,0 +1,42 @@
...
+
+  getRenewablePrice() {
```
JeremyDec, line 37: This logic could be replaced by https://github.com/ovh/manager/blob/master/packages/manager/modules/product-offers/src/services/product-offers.service.js#L15

packages/manager/modules/vps/src/dashboard/migrate/translations/Messages_fr_FR.json
```diff
@@ -0,0 +1,12 @@
+{
+  "vps_dashboard_migrate_configuration_vcore": "{{count}} vCores",
+  "vps_dashboard_migrate_configuration_label_ram": "Mémoire vive :",
+  "vps_dashboard_migrate_configuration_label_core": "Processeur :",
+  "vps_dashboard_migrate_configuration_label_disk": "Stockage :",
+  "vps_dashboard_migrate_title": "Migration de votre VPS",
+  "vps_dashboard_migrate_faq_title": "Pour plus de détails, ",
+  "vps_dashboard_migrate_faq_link_text": "veuillez consulter la FAQ dédiée.",
```
JeremyDec, line 8: The whole sentence should be the link for simplicity

JeremyDec (changes requested):

packages/manager/modules/vps/src/dashboard/migrate/index.scss
varun257: Yes but that means i have to add a class to every `li`, so using the css selector instead

packages/manager/modules/vps/src/dashboard/migrate/migrate-confirm/migrate-confirm.controller.js
varun257: i added this with `click` tracking in mind

packages/manager/modules/vps/package.json
varun257: Has to be listed in project dependency to be used inside module.

packages/manager/modules/vps/src/dashboard/migrate/translations/Messages_fr_FR.json
varun257: Will check with PU for this change.

packages/manager/modules/vps/package.json
marie-j: Yes but here it is declared as a `peer`, we might need to add it as a dependency to apps using this module (like https://github.com/ovh/manager/blob/master/packages/manager/apps/vps/package.json for instance)

packages/manager/modules/vps/src/dashboard/migrate/migrate.html
```diff
@@ -0,0 +1,113 @@
...
+                            data-translate="vps_dashboard_migrate_configuration_vcore"
+                            data-translate-values="{ count: $ctrl.stateVps.vcore }"
+                        ></span>
+                    </li>
+                    <li>
+                        <span
+                            data-translate="vps_dashboard_migrate_configuration_label_ram"
+                        ></span>
+                        <span
+                            data-ng-bind=":: ($ctrl.stateVps.model.memory / 1024) | cucBytes:0:false:'GB'"
```
marie-j, line 26: Maybe we can avoid doing some math in the template and extract the `$ctrl.stateVps.model.memory / 1024`

packages/manager/modules/vps/src/dashboard/migrate/migrate.html
```diff
@@ -0,0 +1,113 @@
...
+                        <span
+                            data-ng-bind=":: ($ctrl.newPlan.configuration.memory.size) | cucBytes:0:false:'GB'"
+                        ></span>
+                    </li>
+                    <li>
+                        <span
+                            data-translate="vps_dashboard_migrate_configuration_label_disk"
+                        ></span>
+                        <span
+                            data-ng-bind=":: $ctrl.newPlan.configuration.storage.disks[0].capacity | cucBytes:0:false:'GB'"
```
marie-j, line 85: What happens if there are no disks ?

packages/manager/modules/vps/src/dashboard/migrate/index.scss
```diff
@@ -0,0 +1,6 @@
+oui-select-picker {
```
radireddy, line 1: can you scope this only to required module or select picker instance? this will be applied to all select pickers, across Manager, once this is loaded.

packages/manager/modules/vps/src/dashboard/migrate/migrate-confirm/migrate-confirm.html
```diff
@@ -0,0 +1,33 @@
...
+        >
+            <input
+                class="oui-input"
+                type="text"
+                id="deleteConfirmation"
+                name="deleteConfirmation"
+                data-ng-model="$ctrl.deleteConfirmation"
+                data-ng-pattern="$ctrl.MIGRATE_CONFIRMATION_INPUT_PATTERN"
+                required
+                autofocus
```
radireddy, line 29: 👍

packages/manager/modules/vps/src/dashboard/migrate/migrate.constants.js
```diff
@@ -0,0 +1,26 @@
...
+  IT: 'https://www.ovhcloud.com/it/vps/vps-offer-migration/',
+  MA: 'https://www.ovhcloud.com/fr-ma/vps/vps-offer-migration/',
+  NL: 'https://www.ovhcloud.com/nl/vps/vps-offer-migration/',
+  PL: 'https://www.ovhcloud.com/pl/vps/vps-offer-migration/',
+  PT: 'https://www.ovhcloud.com/pt/vps/vps-offer-migration/',
+  QC: 'https://www.ovhcloud.com/fr-ca/vps/vps-offer-migration/',
+  SG: 'https://www.ovhcloud.com/en-sg/vps/vps-offer-migration/',
+  SN: 'https://www.ovhcloud.com/fr-sn/vps/vps-offer-migration/',
+  TN: 'https://www.ovhcloud.com/fr-tn/vps/vps-offer-migration/',
+  WE: 'https://www.ovhcloud.com/en/vps/vps-offer-migration/',
```
radireddy, line 20: You can remove this. DEFAULT will cover this.

packages/manager/modules/vps/src/dashboard/migrate/migrate.controller.js
```diff
@@ -0,0 +1,39 @@
...
+      this.newPlan?.pricings,
+      pricingConstants.PRICING_CAPACITIES.RENEW,
+    ).price;
+  }
+
+  getStorageCapacity() {
+    return this.newPlan.configuration?.storage?.disks[0]?.capacity || 0;
+  }
+
+  getMemory(memory) {
```
radireddy, line 36: rename this to formatMemory or something more relevant/meaningful.

packages/manager/modules/vps/src/dashboard/migrate/migrate.html
```diff
@@ -0,0 +1,113 @@
+<oui-tile-definition
+    data-term="{{:: 'vps_dashboard_migrate_title'| translate}}"
+>
+    <oui-tile-description>
+        <oui-select-picker
```
radireddy, line 5: why use a select picker when user has nothing to select? this will have impact on accessibility. Consider creating a box that looks like select picker using select picker styles.

packages/manager/modules/vps/src/dashboard/migrate/migrate.html
```diff
@@ -0,0 +1,113 @@
...
+                <span
+                    data-translate="vps_dashboard_migrate_faq_link_text"
+                ></span>
+                <span
+                    class="oui-icon oui-icon-external-link"
+                    aria-hidden="true"
+                ></span>
+            </a>
+        </p>
+        <oui-select-picker
```
radireddy, line 57: same as above

packages/manager/modules/vps/src/dashboard/migrate/migrate.html
```diff
@@ -0,0 +1,113 @@
+<oui-tile-definition
```
radireddy, line 1: <oui-tile-definition> without oui-tile?

packages/manager/modules/vps/src/import/helper.service.js
```diff
@@ -0,0 +1,20 @@
...
+    const [cores, memory, storage] = planCode.match(/\d+/g);
+    return [parseInt(cores, 10), parseInt(memory, 10), parseInt(storage, 10)];
+  }
+
+  parseRangeConfiguration(rangeFullName) {
+    const [
+      cores,
+      memory,
+      storage,
+    ] = this.constructor.extractConfigurationFromPlanCode(rangeFullName);
```
radireddy, line 12:
```suggestion
    ] = VpsHelperService.extractConfigurationFromPlanCode(rangeFullName);
```

packages/manager/modules/vps/src/import/helper.service.js
```diff
@@ -0,0 +1,20 @@
+export default class VpsHelperService {
```
radireddy, line 1: This should be a util not a service. You don't need to create a object to call these util methods. AngularJs will create a instance of service when used.

radireddy (changes requested):

packages/manager/modules/vps/src/import/helper.service.js
varun257: These functions are dependent on each other so i went with service and not `utils`.

packages/manager/modules/vps/src/dashboard/migrate/migrate.html
varun257: I am passing `0`  if there are no disks now.

packages/manager/modules/vps/src/dashboard/migrate/migrate.html
varun257: I am retaining `<select-picker>` here and added a `model` to bind. Since there is only one element it is always selected. And i can avoid writing class and customising styles

packages/manager/modules/vps/src/dashboard/migrate/migrate.html
varun257: Has benn removed

packages/manager/modules/vps/src/import/helper.service.js
varun257: Prefer using constructor as i can bypass eslint warning about not using this inside class method.

radireddy (approved):

Heiiji (approved):

ganeshkumar1989 (approved):

ganeshkumar1989 (approved):

PR #5169, opened 2021-06-24
fix(redirection): avoid redirection loop during lazyloading

cbourgois:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix MANAGER-7204,

## Description

Using `redirectTo` and using future states, this triggers the lazy loading mechanism which will, once the state is loaded, execute a retry on the transition (https://github.com/ui-router/core/blob/master/src/hooks/lazyLoad.ts#L32-L67 )

If we have some API calls in the `redirectTo`, calls are repeated.
We can't use resolvables on the state that use `redirectTo` because `resolves` are executed once the state is entered, which is not the case.

So this is a sort of hack : create an isolated state (without children), and in his resolves we use `$state.go`.

### DIFF
--- a/packages/manager/apps/public-cloud/src/index.routes.js
+++ b/packages/manager/apps/public-cloud/src/index.routes.js
@@ -5,53 +5,59 @@ import { DEFAULT_PROJECT_KEY } from './index.constants';
 export default /* @ngInject */ ($stateProvider, $urlRouterProvider) => {
   $stateProvider.state('app', {
     url: '/?onboarding',
-    redirectTo: (trans) => {
-      const $q = trans.injector().get('$q');
-      const publicCloud = trans.injector().get('publicCloud');
-      return $q
-        .all([
-          publicCloud.getDefaultProject(),
-          publicCloud.getServices([
-            {
-              field: 'route.path',
-              comparator: 'eq',
-              reference: '/cloud/project/{serviceName}',
-            },
-            {
-              field: 'billing.lifecycle.current.state',
-              comparator: 'eq',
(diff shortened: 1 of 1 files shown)


Jisay (approved):

antleblanc (dismissed):

frenautvh (approved):

packages/manager/apps/public-cloud/src/index.routes.js
```diff
@@ -5,56 +5,62 @@ import { DEFAULT_PROJECT_KEY } from './index.constants';
...
-          };
-        });
-    },
+    redirectTo: 'app.redirect',
     resolve: {
       rootState: () => 'app',
     },
   });
 
+  $stateProvider.state('app.redirect', {
```
marie-j, line 14: Maybe we could add a comment following this PR description so we avoid removing this in some near future thinking it is non sense

JeremyDec (dismissed):

packages/manager/apps/public-cloud/src/index.routes.js
marie-j: @cbourgois do you plan on documenting this change ?

cbourgois: Done in a dedicated commit. 
Could be squashed.

frenautvh (approved):

antleblanc (approved):

marie-j (approved):

PR #5181, opened 2021-06-28
Feature/add typescript to manager config & ufrontend

zakariasahmane:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-7185

## Description

The goal of this PR is to rewrite manager-config and ufrontend in Typescript in order to provide safe typings for different apps/libs that are going to use them.


Jisay (dismissed):
Don't understand everything but good job! :-)

packages/manager/modules/config/.eslintrc.js
```diff
@@ -0,0 +1,22 @@
...
+        '@typescript-eslint/explicit-module-boundary-types': 'off',
+        '@typescript-eslint/no-empty-function': 'off',
+        'lines-between-class-members': 'off',
+        'import/extensions': 'off',
+        'no-use-before-define': 'off',
+        'no-shadow': 'off',
+        '@typescript-eslint/no-shadow': ['error'],
+      },
+    },
+  ],
```
marie-j, line 21: Is this a standard configuration ?

packages/manager/modules/config/package.json
```diff
@@ -13,14 +13,17 @@
     "directory": "packages/manager/modules/config"
   },
   "license": "BSD-3-Clause",
-  "main": "./dist/esm/index.js",
+  "sideEffects": false,
+  "main": "dist/index.js",
+  "module": "dist/index.js",
```
marie-j, line 18: Need to define both ?

packages/manager/modules/config/package.json
```diff
@@ -31,6 +34,9 @@
     "whatwg-fetch": "^3.5.0"
   },
   "devDependencies": {
-    "@ovh-ux/component-rollup-config": "^9.0.0 || ^10.0.0"
+    "@typescript-eslint/eslint-plugin": "^4.27.0",
+    "@typescript-eslint/parser": "^4.27.0",
+    "eslint": "^7.28.0",
+    "typescript": "^4.3.2"
```
marie-j, line 40: ❤️

packages/manager/modules/config/src/types/ovhLanguages.ts
```diff
@@ -0,0 +1,45 @@
...
+  ES = 'ES',
+  PL = 'PL',
+  PT = 'PT',
+  IT = 'IT',
+  DE = 'DE',
+}
+export type LangId =
+  | 'cs'
+  | 'fi'
+  | 'lt'
```
marie-j, line 21:
```suggestion

```
I think those can be removed now

packages/manager/modules/config/.eslintrc.js
zakariasahmane: @marie-j Actually, no. But these are rules that were either in conflict with the current eslint config in the manager or some eslint config we had that didn't make sense in typescript. So i configured it to either be remplaced or not trigger in typescript files.

For instance, i disabled 'no-shadow' of the eslint to activate the no-shadow rule of the ts instead as it didn't understand everything. So the rule is actually still active, it's just suited for ts.

I disabled 'import/extensions' because the current eslint config in the manager wanted me to have extensions for the imports since it couldn't infer ".ts" files. 

The 'no-use-before-define' is a rule that is activated by eslint, but it doesn't make sense in typescript files, it reports incorrect errors so i disabled it. Actually, now that i think about it, there is an equivalent rule in TS '@typescript-eslint/no-use-before-define' that should report the right errors. I'll add this one to throw an error to be as close as possible with our eslint config.

I disabled the rule '@typescript-eslint/no-empty-function' because we return an empty function in the code and didn't want to change the behavior for now as the goal is just to introduce typescript.

The rule '@typescript-eslint/explicit-module-boundary-types' is there to force you to write the return type of your functions etc. But in the most cases, TS can infer return types correctly. I can actually put this one on 'warning' to recall us to check for the return type. 

And finally, the rule 'lines-between-class-members' was activated by our current eslint, but since now  we're implemeting interfaces in our classes, i thought that there were too many spaces. I can reactivate this one since it's the only one i disabled on personal preference !  

If you have other workarounds for some of these rules i'll take them into account, or if you want me to reactivate some of them i can too!

packages/manager/modules/config/package.json
zakariasahmane: @marie-j I put them both because of what i understood, module is for packages that can handle ES modules. If a package that can use ES Module see the module entry, it will use it instead of main for better use of the package. But for packages who don't understand ES Modules or are written in CommonJS etc, the need the main entry to be able to stuff like `require('myModule')`. 

To be honest, i don't know if i need to define them both, i just thought of main as a fallback if the package couldn't understand the module entry. 

The first answer in this SO topic actually gives a lot of info about it : https://stackoverflow.com/questions/42708484/what-is-the-module-package-json-field-for
But i couldn't find anywhere if i should use it alone or couple it with main too.

packages/manager/modules/config/.eslintrc.js
marie-j: @zakariasahmane Thank you for this verbose explanation :) I was just curious if this was some standard eslint config for TS or just adapted to our current context 
About `lines-between-class-members` if it is consistent with the rest of our codebase I don't mind keeping it

cbourgois (dismissed):

frenautvh (dismissed):

frenautvh (dismissed):

Jisay (dismissed):

Jisay (dismissed):

cbourgois (approved):

cbourgois (approved):

PR #5204, opened 2021-07-02
Refactor/sso jquery

JeremyDec:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-4997

## Description

Re-opening of previous @FredericEspiau' PR https://github.com/ovh/manager/pull/2852 to get rid of jQuery dependency on `ng-ovh-sso-auth` and `ng-ovh-sso-auth-modal-plugin`. (Rebase on `develop` since then)

### DIFF
--- a/packages/components/ng-ovh-sso-auth-modal-plugin/package.json
+++ b/packages/components/ng-ovh-sso-auth-modal-plugin/package.json
@@ -39,9 +39,6 @@
     "start:dev": "lerna exec --stream --scope='@ovh-ux/ng-ovh-sso-auth-modal-plugin' --include-dependencies -- yarn run dev",
     "start:watch": "lerna exec --stream --parallel --scope='@ovh-ux/ng-ovh-sso-auth-modal-plugin' --include-dependencies -- yarn run dev:watch"
   },
-  "dependencies": {
-    "lodash": "~4.17.11"
-  },
   "devDependencies": {
     "@ovh-ux/component-rollup-config": "^9.0.0 || ^10.0.0"
   },
--- a/packages/components/ng-ovh-sso-auth-modal-plugin/package.json
+++ b/packages/components/ng-ovh-sso-auth-modal-plugin/package.json
@@ -47,7 +47,6 @@
     "@ovh-ux/ng-translate-async-loader": "^2.1.2",
     "angular": "^1.3.x",
     "angular-translate": "^2.18.1",
-    "angular-ui-bootstrap": "1.3.3",
-    "jquery": "^2.1.3"
+    "angular-ui-bootstrap": "1.3.3"
   }
 }
--- a/packages/components/ng-ovh-sso-auth-modal-plugin/src/controller.js
+++ b/packages/components/ng-ovh-sso-auth-modal-plugin/src/controller.js
@@ -1,5 +1,3 @@
-import $ from 'jquery';
-
 /**
  * @ngdoc controller
  * @name ng-ovh-sso-auth-modal-plugin.SsoAuthModalController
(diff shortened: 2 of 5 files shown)


frenautvh (approved):

packages/components/ng-ovh-sso-auth-modal-plugin/package.json
```diff
@@ -47,7 +47,9 @@
     "@ovh-ux/ng-translate-async-loader": "^2.1.2",
     "angular": "^1.3.x",
     "angular-translate": "^2.18.1",
-    "angular-ui-bootstrap": "1.3.3",
-    "jquery": "^2.1.3"
+    "angular-ui-bootstrap": "1.3.3"
+  },
+  "optionalDependencies": {
```
cbourgois, line 52: I think `optionalDependencies` are dependencies that are not required to run, but used in your code. 
https://docs.npmjs.com/cli/v6/configuring-npm/package-json#optionaldependencies

Here `whatwg-fetch` is not imported, so not used; it's a _recommended_ dependencies.

If you use `optionnal` you should support the case when dependency is not installed (`npm install --no-optional` or `yarn install --ignore-optional`)

packages/components/ng-ovh-sso-auth-modal-plugin/src/index.js
```diff
@@ -27,9 +26,10 @@ angular
     $provide.decorator(
       'ssoAuthentication',
       ($delegate, ssoAuthModalPluginFct) => {
-        set($delegate, 'handleSwitchSession', () =>
-          ssoAuthModalPluginFct.handleSwitchSession($delegate),
-        );
+        // eslint-disable-next-line no-param-reassign
```
FredericEspiau, line 29: wtf I like the previous way better 🤔

packages/components/ng-ovh-sso-auth-modal-plugin/package.json
FredericEspiau: Don't know about that, the guy who opened the first PR seems to know what he is doing: https://github.com/ovh/manager/pull/2852/files#diff-6800f225d7d13b98b4781559c10c478d3f4fe3bd2c4ade35ca643f2aa9a8ea6fR51

FredericEspiau:
I can't approve, there must be a bug with Github

packages/components/ng-ovh-sso-auth-modal-plugin/package.json
JeremyDec: Seen together, dependency is removed, thanks.

packages/components/ng-ovh-sso-auth-modal-plugin/src/controller.js
```diff
@@ -41,13 +39,18 @@ export default /* @ngInject */ function SsoAuthModalController(
 
   function getUser() {
     const deferredObj = $q.defer();
-    $.ajax(self.data.userConfig)
-      .done((data) => {
-        deferredObj.resolve(data);
+
+    fetch(self.data.userConfig.url, {
```
radireddy, line 43: do you think fetch is supported by all browsers? or do we need a polyfill? IE does not support it https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#browser_compatibility

JeremyDec: Hello @radireddy, seen with @cbourgois, the polyfill, if required should be handled by the app which use this module, not by the module itself. So it will use the fetch, regardless of the targeted browser

FredericEspiau: Which means that the target should know it must handle fetch. I didn't check, is it specified anywhere?

cbourgois (dismissed):

frenautvh (approved):

PR #5227, opened 2021-07-07
feat(dashboard): update list for new options

varun257:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/high-perf-storage`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-7046

## Description

Activate/Deactivate switch to be added
Add "Add a user to a container" option to high perf objects alone

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/storages/containers/containers.controller.js
+++ b/packages/manager/modules/pci/src/projects/project/storages/containers/containers.controller.js
@@ -9,38 +9,37 @@ export default class PciStoragesContainersController {
 
   $onInit() {
     this.loadMessages();
+    this.publicToggleLoading = false;
   }
 
-  onPublicToggle(data) {
-    this.publicToggleLoading = true;
+  onPublicToggle(container) {
+    this.loadingContainer = container.id;
     return this.$http
-      .put(`/cloud/project/${this.projectId}/storage/${data.id}`, {
-        containerType: data.public ? 'private' : 'public',
+      .put(`/cloud/project/${this.projectId}/storage/${container.id}`, {
+        containerType: container.state ? 'private' : 'public',
       })
       .then(() =>
         this.CucCloudMessage.success(
           this.$translate.instant(
-            data.public
+            container.state
               ? 'pci_projects_project_storages_containers_toggle_private_succeed'
               : 'pci_projects_project_storages_containers_toggle_public_succeed',
-            { name: data.name },
+            { name: container.name },
           ),
           'pci.projects.project.storages.containers.container',
         ),
       )
-      .catch((err) =>
+      .catch((error) =>
         this.CucCloudMessage.error(
           this.$translate.instant(
             'pci_projects_project_storages_containers_toggle_fail',
-            {
-              message: err.message || err.data?.message,
-            },
+            { message: error.data?.message },
           ),
           'pci.projects.project.storages.containers.container',
         ),
       )
       .finally(() => {
-        this.publicToggleLoading = false;
+        this.loadingContainer = null;
       });
   }
 
--- a/packages/manager/modules/pci/src/projects/project/storages/containers/containers.html
+++ b/packages/manager/modules/pci/src/projects/project/storages/containers/containers.html
@@ -22,7 +22,7 @@
 
         <oui-datagrid data-rows="$ctrl.containers">
             <oui-datagrid-column
-                data-title="'pci_projects_project_storages_containers_name_label' | translate"
+                data-title=":: 'pci_projects_project_storages_containers_name_label' | translate"
                 data-property="name"
                 data-type="string"
                 data-searchable
@@ -36,24 +36,25 @@
                 ></a>
             </oui-datagrid-column>
             <oui-datagrid-column
-                data-title="'pci_projects_project_storages_containers_region_label' | translate"
+                data-title=":: 'pci_projects_project_storages_containers_region_label' | translate"
                 data-property="region"
                 data-type="string"
                 data-sortable
                 data-filterable
             ></oui-datagrid-column>
             <oui-datagrid-column
-                data-title="'pci_projects_project_storages_containers_containerType_label' | translate"
+                data-title=":: 'pci_projects_project_storages_containers_containerType_label' | translate"
                 data-property="containerType"
                 data-sortable
                 data-type="string"
             >
                 <span
+                    data-ng-if="$row.containerType"
                     data-ng-bind="'pci_projects_project_storages_containers_containerType_' + $row.containerType | translate"
                 ></span>
             </oui-datagrid-column>
             <oui-datagrid-column
-                data-title="'pci_projects_project_storages_containers_offer_label' | translate"
+                data-title=":: 'pci_projects_project_storages_containers_offer_label' | translate"
                 data-property="offer"
                 data-type="string"
                 data-sortable
@@ -61,7 +62,7 @@
             >
             </oui-datagrid-column>
             <oui-datagrid-column
-                data-title="'pci_projects_project_storages_containers_storedObjects_label' | translate"
+                data-title=":: 'pci_projects_project_storages_containers_storedObjects_label' | translate"
                 data-property="storedObjects"
                 data-type="number"
                 data-sortable
@@ -69,12 +70,16 @@
             >
             </oui-datagrid-column>
             <oui-datagrid-column
-                data-title="'pci_projects_project_storages_containers_storedBytes_label' | translate"
+                data-title=":: 'pci_projects_project_storages_containers_storedBytes_label' | translate"
                 data-property="storedBytes"
                 data-type="number"
                 data-sortable
             >
-                <span data-ng-bind="$value | RAUnitsBits"></span>
+                <span
+                    data-ng-if="$value > 0"
+                    data-ng-bind="$value | RAUnitsBits"
+                ></span>
+                <span data-ng-if="!$value">-</span>
             </oui-datagrid-column>
 
             <oui-action-menu data-compact data-placement="end">
@@ -99,15 +104,16 @@
                 </oui-action-menu-item>
             </oui-action-menu>
             <oui-datagrid-column
-                data-title="'pci_projects_project_storages_containers_public_label' | translate"
-                data-property="public"
+                data-title=":: 'pci_projects_project_storages_containers_public_label' | translate"
+                data-property="state"
                 data-sortable
                 data-filterable
             >
                 <oui-switch
                     model="$value"
-                    disabled="$ctrl.publicToggleLoading"
+                    disabled="$row.id === $ctrl.loadingContainer"
                     on-change="$ctrl.onPublicToggle($row)"
+                    data-ng-if="!$row.isHighPerfStorage"
                 >
                 </oui-switch>
             </oui-datagrid-column>
--- a/packages/manager/modules/pci/src/projects/project/storages/containers/containers.service.js
+++ b/packages/manager/modules/pci/src/projects/project/storages/containers/containers.service.js
@@ -20,6 +20,7 @@ import {
   OBJECT_CONTAINER_TYPE_STATIC,
   OBJECT_CONTAINER_TYPE_PUBLIC,
   OBJECT_TYPE_SEALED,
+  OPENIO_DEFAULT_REGION,
   STORAGE_GATEWAY,
   X_AUTH_TOKEN,
   X_CONTAINER_HEADERS_REGEX,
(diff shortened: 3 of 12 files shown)


packages/manager/modules/pci/src/projects/project/storages/containers/containers.controller.js
```diff
@@ -1,14 +1,44 @@
 export default class PciStoragesContainersController {
   /* @ngInject */
-  constructor($translate, CucCloudMessage) {
+  constructor($translate, $http, CucCloudMessage) {
     this.$translate = $translate;
+    this.$http = $http;
     this.CucCloudMessage = CucCloudMessage;
+    this.publicToggleLoading = false;
```
ganeshkumar1989, line 7: May be move this to `$onInit` to have a cleaner code :)

packages/manager/modules/pci/src/projects/project/storages/containers/containers.controller.js
```diff
@@ -1,14 +1,44 @@
...
   }
 
   $onInit() {
     this.loadMessages();
   }
 
+  onPublicToggle(data) {
```
ganeshkumar1989, line 14:
```suggestion
  onPublicToggle(container) {
```

packages/manager/modules/pci/src/projects/project/storages/containers/containers.html
```diff
@@ -82,6 +91,19 @@
...
             </oui-action-menu>
+            <oui-datagrid-column
+                data-title="'pci_projects_project_storages_containers_public_label' | translate"
+                data-property="public"
+                data-sortable
+                data-filterable
+            >
+                <oui-switch
+                    model="$value"
+                    disabled="$ctrl.publicToggleLoading"
```
ganeshkumar1989, line 102: Are we sure that we will have to disable this toggle for all the rows, when one of the rows is being toggled?

packages/manager/modules/pci/src/projects/project/storages/containers/containers.controller.js
```diff
@@ -1,12 +1,44 @@
...
+              ? 'pci_projects_project_storages_containers_toggle_private_succeed'
+              : 'pci_projects_project_storages_containers_toggle_public_succeed',
+            { name: container.name },
+          ),
+          'pci.projects.project.storages.containers.container',
+        ),
+      )
+      .catch(() =>
+        this.CucCloudMessage.error(
+          'yay',
```
radireddy, line 35: what is 'yay'?

packages/manager/modules/pci/src/projects/project/storages/containers/containers.service.js
```diff
@@ -132,20 +135,34 @@ export default class PciStoragesContainersService {
...
+        ),
       );
   }
 
-  getContainer(projectId, containerId) {
-    return this.OvhApiCloudProjectStorage.v6()
-      .get({
+  getContainer(projectId, containerId, isHighPerfStorage) {
+    if (isHighPerfStorage) {
+      this.promise = this.$http
```
radireddy, line 153: do you need to store the promise in Object? if not use can use a constant within the method

packages/manager/modules/pci/src/projects/project/storages/containers/containers.html
```diff
@@ -65,7 +75,11 @@
                 data-type="number"
                 data-sortable
             >
-                <span data-ng-bind="$value | RAUnitsBits"></span>
+                <span
+                    data-ng-if="$value > 0"
+                    data-ng-bind="$value | RAUnitsBits"
+                ></span>
+                <span data-ng-if="!$value > 0">-</span>
```
JeremyDec, line 82: **nitpick**
Ok it works, but more simple to do
```suggestion
                <span data-ng-if="$value === 0">-</span>
```

packages/manager/modules/pci/src/projects/project/storages/containers/translations/Messages_fr_FR.json
```diff
@@ -1,8 +1,15 @@
 {
   "pci_projects_project_storages_containers_archive_title": "Cloud Archive",
   "pci_projects_project_storages_containers_object_title": "Object Storage",
+  "pci_projects_project_storages_containers_object_description": "Hébergez votre catalogue de données dans notre solution Object Storage. Vous pouvez administrer ces données depuis votre application grâce aux API S3. Content !!!",
```
JeremyDec, line 4:
```suggestion
  "pci_projects_project_storages_containers_object_description": "Hébergez votre catalogue de données dans notre solution Object Storage. Vous pouvez administrer ces données depuis votre application grâce aux API S3.",
```

JeremyDec (changes requested):

radireddy (approved):

packages/manager/modules/pci/src/projects/project/storages/objects/object/object.routing.js
```diff
@@ -1,16 +1,23 @@
 export default /* @ngInject */ ($stateProvider) => {
   $stateProvider.state('pci.projects.project.storages.objects.object', {
-    url: '/{containerId}',
+    url: '/{containerId}/{isHighPerfStorage}',
```
JeremyDec, line 3: Do we expect to have `https://www.ovh.com/manager/public-cloud/xxx/true` in the url path ? Should it be transformed as `/{containerId}?isHighPerfStorage` ?

packages/manager/modules/pci/src/projects/project/storages/objects/object/delete/delete.routing.js
```diff
@@ -1,21 +1,28 @@
 export default /* @ngInject */ ($stateProvider) => {
   $stateProvider.state('pci.projects.project.storages.objects.delete', {
-    url: '/delete?containerId',
+    url: '/delete?containerId?isHighPerfStorage',
```
JeremyDec, line 3:
```suggestion
    url: '/delete?containerId&isHighPerfStorage',
```

JeremyDec (changes requested):
Hello @varun257, I have trouble to know where does come the `isHighPerfStorage` flag. Must this be set by the front ? Isn't it a value return on container details by API ?

packages/manager/modules/pci/src/projects/project/storages/objects/object/delete/delete.routing.js
marie-j: Not sure this belongs in the url. If customer wants a direct access to this they shouldn't be bothering if their container is `highPerfStorage` or not

packages/manager/modules/pci/src/projects/project/storages/containers/containers.controller.js
```diff
@@ -1,12 +1,47 @@
...
     this.CucCloudMessage = CucCloudMessage;
   }
 
   $onInit() {
     this.loadMessages();
+    this.publicToggleLoading = false;
+  }
+
+  onPublicToggle(container) {
+    set(container, 'publicToggleLoading', true);
```
marie-j, line 17: We shouldn't update the parameter value. Please create a copy or handle it differently

packages/manager/modules/pci/src/projects/project/storages/containers/containers.controller.js
```diff
@@ -1,12 +1,47 @@
...
+    return this.$http
+      .put(`/cloud/project/${this.projectId}/storage/${container.id}`, {
+        containerType: container.publicState ? 'private' : 'public',
```
marie-j, line 20: If there is a publicState then the container is `private` ? Not really intuitive 
Also maybe it could be handled by the model

packages/manager/modules/pci/src/projects/project/storages/containers/containers.html
```diff
@@ -48,9 +49,18 @@
                 data-type="string"
             >
                 <span
+                    data-ng-if="$row.containerType"
                     data-ng-bind="'pci_projects_project_storages_containers_containerType_' + $row.containerType | translate"
                 ></span>
             </oui-datagrid-column>
+            <oui-datagrid-column
+                data-title="'pci_projects_project_storages_containers_offer_label' | translate"
```
marie-j, line 57:
```suggestion
                data-title="::'pci_projects_project_storages_containers_offer_label' | translate"
```

packages/manager/modules/pci/src/projects/project/storages/containers/containers.html
```diff
@@ -65,7 +75,11 @@
                 data-type="number"
                 data-sortable
             >
-                <span data-ng-bind="$value | RAUnitsBits"></span>
+                <span
+                    data-ng-if="$value > 0"
+                    data-ng-bind="$value | RAUnitsBits"
+                ></span>
+                <span data-ng-if="$value === 0">-</span>
```
marie-j, line 82:
```suggestion
                <span data-ng-if="!$value">-</span>
```
Not enough ?

packages/manager/modules/pci/src/projects/project/storages/containers/containers.service.js
```diff
@@ -58,7 +59,9 @@ export default class PciStoragesContainersService {
           endpoints,
           (result, endpoint) => ({
             ...result,
-            [endpoint.region.toLowerCase()]: endpoint.url,
+            [(
+              endpoint.region || OPENIO_DEFAULT_REGION
```
marie-j, line 63:
```suggestion
              endpoint.region
```
Already handled by the model, no ?

packages/manager/modules/pci/src/projects/project/storages/objects/objects.routing.js
```diff
@@ -25,17 +25,20 @@ export default /* @ngInject */ ($stateProvider) => {
       viewContainer: /* @ngInject */ ($state, projectId) => (container) =>
         $state.go('pci.projects.project.storages.objects.object', {
           projectId,
-          containerId: container.id,
+          containerId: container.id || container.name,
+          isHighPerfStorage: container.isHighPerfStorage || false,
```
marie-j, line 29: Is it necessary to specify `|| false` ?

packages/manager/modules/pci/src/projects/project/storages/objects/objects.routing.js
```diff
@@ -25,17 +25,20 @@ export default /* @ngInject */ ($stateProvider) => {
       viewContainer: /* @ngInject */ ($state, projectId) => (container) =>
         $state.go('pci.projects.project.storages.objects.object', {
           projectId,
-          containerId: container.id,
+          containerId: container.id || container.name,
```
marie-j, line 28: I'm not sure it is a good idea to start either having `id` or `name`. Especially it will require to make some adaptation to find id depending on name in the used component and it starts to get tricky

marie-j (changes requested):

packages/manager/modules/pci/src/projects/project/storages/objects/object/delete/delete.routing.js
varun257: Yes that would be the normal case. But right now to get high performance storage, we have to pass a static region to the API to get the details of this. I dont see any other way of doing this.
Also to combine both swift and open IO containers in the same list page, we are fetching results(openIO) based on static region and passing `isHighPerfStorage` as a parameter from the 2API.

packages/manager/modules/pci/src/projects/project/storages/objects/objects.routing.js
varun257: Well this is a bit strange to see i agree but with swift they had id and name and faced some difficulties.
There is no ID(openIO) either. Containers have no ID, they only have a unique name. For swift, we used to compute an ID using name and region (base64) but this was dirty and led to issue. We do not want to reproduce this with OpenIO.

varun257:
> Hello @varun257, I have trouble to know where does come the `isHighPerfStorage` flag. Must this be set by the front ? Isn't it a value return on container details by API ?

Hi @JeremyDec , This is returned from the 2API, but since i wanted to make the route independent, i had to pass whether the storage is swift or openIO. hence using a flag in route `isHighPerfStorage` since we have to send static `region` in API to get openIO storage

packages/manager/modules/pci/src/projects/project/storages/objects/objects.routing.js
marie-j: Ok so maybe we should abstract with those kind of containers that the id is the name (with a model for instance)

ganeshkumar1989 (approved):

PR #5231, opened 2021-07-07
refactor(region): move region to common module

marie-j:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/netapp`
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-7320

## Description

Move region service from `cloud-universe-components` to `manager-components` to avoid having to import `cuc` when a region selector is needed (in Dedicated for instance) 

## Related 
- [x] https://github.com/ovh/manager/pull/5232


frenautvh (dismissed):

marie-j:
Rebased to target `develop` and amend commit with ticket reference

packages/manager/modules/manager-components/src/region/service.js
```diff
@@ -1,6 +1,7 @@
 import trim from 'lodash/trim';
+import { COUNTRIES } from './constants';
 
-export default class CucRegionService {
+export default class ovhManagerRegionService {
```
radireddy, line 4: 👍  great initiative. I have created a reusable region formater/directive (https://github.com/ovh/manager/pull/5234). Do you think we can still keep it and may be refactor it to use this region service?

packages/manager/modules/manager-components/src/region/service.js
```diff
@@ -23,28 +24,30 @@ export default class CucRegionService {
 
   getAllTranslatedMacroRegion() {
```
radireddy, line 25: There is a requirement to show region with city name, country name and datacenter name like "Strasbourg (SBG1) France". Is it supported?

marie-j: @radireddy Not sure I've just kept it as is to not break existing usage. I guess if we have other need we can extend the service to add needed formatters/functions

packages/manager/modules/manager-components/src/region/service.js
marie-j: @radireddy Yes I suppose we better reuse existing code than duplicating this kind of logic (purpose here is to move so we can use it in `feat/netapp` instead of reimplementing a region logic)

ganeshkumar1989 (approved):
Neat!!! Much needed and very well done. Thanks :)
Just another thought... Can we also move https://github.com/ovh/manager/tree/master/packages/manager/modules/pci/src/components/project/regions-list into this module, since it is used in a lot of places? Can be a separate PR though :)

marie-j:
> Neat!!! Much needed and very well done. Thanks :)
> Just another thought... Can we also move https://github.com/ovh/manager/tree/master/packages/manager/modules/pci/src/components/project/regions-list into this module, since it is used in a lot of places? Can be a separate PR though :)

Sure we can imagine centralizing region selection/handling to one place (but in a separate PR 😅 )

marie-j:
Rebase to fix conflicts + patch notebooks

ganeshkumar1989 (dismissed):

antleblanc (approved):

PR #5262, opened 2021-07-15
fix(exchange): remove version test on get display value

JeremyDec:
ref: DTRSD-43026

| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-43062

## Description

Remove test to get type display text for exchange account

### DIFF
--- a/packages/manager/modules/exchange/src/account/accountTypes.service.js
+++ b/packages/manager/modules/exchange/src/account/accountTypes.service.js
@@ -38,10 +38,6 @@ export default class ExchangeAccountTypes {
 
     switch (formattedAccountType) {
       case this.TYPES.BASIC:
-        if (!this.CAN_DO.BASIC()) {
-          throw new Error("Current service doesn't allow Basic account types");
-        }
-
         return this.exchangeServiceInfrastructure.isDedicatedCluster()
           ? this.$translate.instant(
               'exchange_accounts_types_dedicatedCluster_BASIC',


radireddy (approved):

packages/manager/modules/exchange/src/account/accountTypes.service.js
```diff
@@ -38,10 +38,6 @@ export default class ExchangeAccountTypes {
 
     switch (formattedAccountType) {
       case this.TYPES.BASIC:
-        if (!this.CAN_DO.BASIC()) {
```
frenautvh, line 41: can't you just update the condition here :
https://github.com/ovh/manager/blob/409d42926d36d13ef21924d308d5d78dc385ad4f/packages/manager/modules/exchange/src/account/accountTypes.service.js#L21
instead of removing the check ?

JeremyDec: Why not, but the question is: is this legit to make the test here ? As the method is here to get the value to display, regarding the type, why do we need to make this test to just display `Basic` text ?

frenautvh: sure i was wondering if it's legit to remove the test but in case of displaying the value i think you are right it can be removed

frenautvh (approved):

PR #5264, opened 2021-07-16
feat(exchange.disclaimer): not editable editor

varun257:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-41448

## Description

Remove tabindex attribute from bootstrap modal as it clashes with ckeditor modal

### DIFF
--- a/packages/manager/modules/exchange/src/disclaimer/update/disclaimer-update.controller.js
+++ b/packages/manager/modules/exchange/src/disclaimer/update/disclaimer-update.controller.js
@@ -4,8 +4,16 @@ import isEmpty from 'lodash/isEmpty';
 
 export default class ExchangeUpdateDisclaimerCtrl {
   /* @ngInject */
-  constructor($scope, wucExchange, navigation, messaging, $translate) {
+  constructor(
+    $document,
+    $scope,
+    wucExchange,
+    navigation,
+    messaging,
+    $translate,
+  ) {
     this.services = {
+      $document,
       $scope,
       wucExchange,
       navigation,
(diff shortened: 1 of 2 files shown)


varun257 (commented):
There are 2 modals here, bootstrap modal and a ckeditor modal created fetched dynamically.
The tabindex attribute of the ckeditor is clashing with the bootstrap modal tabindex
Hence i am removing this manually after modal is loaded.

packages/manager/modules/exchange/src/disclaimer/update/disclaimer-update.controller.js
```diff
@@ -25,6 +25,9 @@ export default class ExchangeUpdateDisclaimerCtrl {
   loadOptions() {
     this.loadingData = true;
 
+    // Remove tabindex from modal as it clashes with ckeditor modal
+    document.getElementsByClassName('modal')[0].removeAttribute('tabindex');
```
frenautvh, line 29: it's kind of a dirty fix, .modal class refer to which element, the ckeditor one or the bootstrap one ?
maybe you can scope the query by className a little more narrower, i didn't check but something like :
```js
document.getElementsByClassName('.ckeditor-modal > .modal'); // just an example
```
also please use angularJS api to query elements : https://docs.angularjs.org/api/ng/service/$document

frenautvh (changes requested):
see comments

packages/manager/modules/exchange/src/disclaimer/update/disclaimer-update.controller.js
varun257: I cannot find a better fix for this. I cannot add class as the bootstrap modal is outside the component.
I have used angularJS api to query the modal

varun257: @frenautvh is this okay ?

frenautvh (approved):

ganeshkumar1989 (approved):

PR #5272, opened 2021-07-19
Feat/uxct 413 add hec fec

Steffy29:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #UXCT-414

## Description

Toggle button for content sharing doesn't work because the modal could not open (problem with initialization of the service)


Steffy29:
Not the correct PR

PR #5331, opened 2021-07-28
fix(translations): run translations at the right moment

zakariasahmane:
| Question         | Answer
| ---------------- | ---
| Branch?          |  master
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-41612

## Description

When we clicked on the add config pop to open the modal, the modal open with the keys of the translations instead of translations the first time. If you close and reopen the modal, they are okay. 

The goal of this PR is to load translations for the first time. 

Please refer to the ticket for more infos.

### DIFF
--- a/packages/manager/modules/cloud-connect/src/details/overview/add-pop-configuration/add-pop-configuration.module.js
+++ b/packages/manager/modules/cloud-connect/src/details/overview/add-pop-configuration/add-pop-configuration.module.js
@@ -9,6 +9,14 @@ angular
   .module(moduleName, [])
   .config(routing)
   .component('cloudConnectDetailsAddPopConfiguration', component)
-  .run(/* @ngTranslationsInject:json ./translations */);
+  .run(/* @ngTranslationsInject:json ./translations */)
+  .run(
+    /* @ngInject */ ($translate, $transitions) => {
+      $transitions.onBefore(
+        { to: 'cloud-connect.details.overview.add-pop' },
+        () => $translate.refresh(),
+      );
+    },
+  );
 
 export default moduleName;


packages/manager/modules/cloud-connect/src/details/overview/add-pop-configuration/index.js
```diff
@@ -29,6 +29,7 @@ angular
         },
       });
     },
-  );
+  )
+  .run(/* @ngTranslationsInject:json ./translations */);
```
cbourgois, line 33: If we do that it means that we lazy load only the component and not all things related to it.

marie-j: Maybe what's needed is a `$translate.refresh` before entering the states

zakariasahmane: I'll try that

zakariasahmane: Yes but i think the refresh might be done in the upper state, i actually added it and it works now

marie-j: @cbourgois as it is asynchronous, component might be loaded and translations still being fetched right ? 
Maybe we can prevent this error more globally with scoping to the root state ?

cbourgois: But some translations (that are in the same translations file) are OK, some not.
I think it's because the modal is displayed before the translations's injections is completed, and 
- when using `translate` directive, it's OK, 
- when using `translation` filter in one time expression (`::`), it doesn't works.

I try to understand adding the following code in the template:

```html
<p data-translate="cloud_connect_pop_configure_title"></p>
<p>{{:: 'cloud_connect_pop_configure_title' | translate }}</p>
<p>{{ 'cloud_connect_pop_configure_title' | translate }}</p>
```

it displays

```html
Ajout d’une configuration PoP
cloud_connect_pop_configure_title
Ajout d’une configuration PoP
```

So I try to remove one time expression to have

```html
<oui-modal
        data-heading="{{ 'cloud_connect_pop_configure_title' | translate }}"
```

instead of

```html
<oui-modal
        data-heading="{{:: 'cloud_connect_pop_configure_title' | translate }}"
```
Cf https://github.com/ovh/manager/blob/master/packages/manager/modules/cloud-connect/src/details/overview/add-pop-configuration/template.html#L8

But it doesn't work because the first time the modal is rendered I think translation are not injected, and as oui-modal#heading is a one time binding (cf https://ovh.github.io/ovh-ui-kit/?path=/story/design-system-components-modal-introduction--page), it won't be updated when translations are injected.

Adding a `$translate.refresh()` could help to fix the bug, but it doesn't fix the root cause.

radireddy (approved):

packages/manager/modules/cloud-connect/src/details/overview/add-pop-configuration/index.js
cbourgois: We will fix the root cause in another PR. This one fixes this issue, so no blocker.

cbourgois (approved):

PR #5395, opened 2021-08-10
Feat/netapp configuration

marie-j:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/netapp`
| Bug fix?         |no
| New feature?     | yes
| Breaking change? |no
| Tickets          | MANAGER-7145


packages/manager/modules/manager-components/src/region/component.js
```diff
@@ -0,0 +1,11 @@
+import controller from './controller';
+import template from './template.html';
+
+export default {
+  bindings: {
+    regions: '<',
+    model: '=',
```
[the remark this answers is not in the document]
marie-j: This would oblige all consumers of this component to manually update the model on their side. It could be added later on if it is needed but in this case (netapp) it is not.
I agree that we should pay attention to the bindings use but I'm curious how using a callback is better here, because from first sight it just seems to add complexity in the use (declare a function, update the model, having to pass parameters correctly etc), let me know your thoughts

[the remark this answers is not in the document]
JeremyDec: Yes the two-way binding is more simpler to use, but it has its downsides, like adding watchers to the global scope to look for changes, and so may lower performances, even more if it is used many times in the whole application.

Having a callback, indeed bring code complexity, but may limit this effect as it won't be triggered at each scope value update.

JeremyDec (approved):

antleblanc (approved):

packages/manager/modules/billing-components/src/components/commitment-duration/CommitmentDuration.class.js
```diff
@@ -0,0 +1,31 @@
...
+      );
+    }
+  }
+
+  get monthlyDuration() {
+    return moment.duration(this.duration).asMonths();
+  }
+
+  get monthlyPrice() {
+    return this.commitment.pricing.monthlyPrice.text;
```
ganeshkumar1989, line 29: If we doubt its availability
```suggestion
    return this.commitment.pricing?.monthlyPrice?.text;
```

packages/manager/modules/catalog-price/src/catalog-price.html
```diff
@@ -12,6 +12,12 @@
                         data-translate="{{:: 'order_catalog_price_interval_' + $ctrl.interval}}"
                     ></span>
                 </span>
+                <span data-ng-if="$ctrl.intervalUnit">
+                    &#47;
+                    <span
+                        data-translate="{{:: 'order_catalog_price_interval_' + $ctrl.intervalUnit }}"
+                    ></span>
+                </span>
```
ganeshkumar1989, line 20: It is confusing as to why we need both `interval` and `intervalUnit`. Seems like both result in the same text being rendered... Can't we remove one of them?

packages/manager/modules/netapp/src/order/translations/Messages_fr_CA.json
```diff
@@ -0,0 +1,3 @@
+{
+  "world": "World!"
```
ganeshkumar1989, line 2: Probably, can be removed?

packages/manager/modules/catalog-price/src/catalog-price.html
marie-j: @ganeshkumar1989 No because `interval` specifies the duration for the given price so let's say I set it to month it will divide the price per 12

packages/manager/modules/billing-components/src/components/commitment-duration/CommitmentDuration.class.js
marie-j: In fact this should have been a `git mv`

ganeshkumar1989 (approved):

marie-j:
Rebase to fix conflicts

PR #5405, opened 2021-08-11
fix(translations): refresh translation after transition

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-7433

## Description

Ensure translations are refreshed after state change

### DIFF
--- a/packages/manager/apps/dedicated/client/app/app.module.js
+++ b/packages/manager/apps/dedicated/client/app/app.module.js
@@ -430,11 +430,6 @@ export default (containerEl, environment) => {
         });
       },
     )
-    .run(
-      /* @ngInject */ ($translate, $transitions) => {
-        $transitions.onBefore({ to: 'app.**' }, () => $translate.refresh());
-      },
-    )
     .config(
       /* @ngInject */ (ovhFeatureFlippingProvider) => {
         ovhFeatureFlippingProvider.setApplicationName(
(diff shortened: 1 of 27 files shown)


packages/components/ng-translate-async-loader/src/index.js
```diff
@@ -4,6 +4,15 @@ import factory from './factory';
 
 const moduleName = 'ngTranslateAsyncLoader';
 
-angular.module(moduleName, []).factory('asyncLoader', factory);
+angular
+  .module(moduleName, [])
+  .factory('asyncLoader', factory)
+  .run(
```
antleblanc, line 10: Both `$transitions` and `$translate` come from `@uirouter/angularjs` and `angular-translate` which are not defined as peerDependencies of this module.

In some way, it looks like a Breaking Change but out of curiosity, shouldn't be done inside `@ovh-ux/manager-core`?

frenautvh: i think you're right it would make sense to add this in core, i'm going to update this PR.

radireddy (approved):

marie-j (approved):

Jisay (approved):

cbourgois (approved):

PR #5600, opened 2021-09-22
fix(pci.project.quota): set issue type constants to id list

JeremyDec:
ref: DTRSD-22073

| Question         | Answer
| ---------------- | ---
| Branch?          | `develop` for bug fixes 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #... DTRSD-22073

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/quota/increase-request/increase-request.controller.js
+++ b/packages/manager/modules/pci/src/projects/project/quota/increase-request/increase-request.controller.js
@@ -1,7 +1,7 @@
 import find from 'lodash/find';
 import get from 'lodash/get';
 import isNil from 'lodash/isNil';
-import { ISSUE_TYPE_ID } from './increase.constants';
+import { ISSUE_TYPE_IDS } from './increase.constants';
 
 export default class PciProjectQuotaIncreaseController {
   /* @ngInject */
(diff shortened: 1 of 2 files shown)


packages/manager/modules/pci/src/projects/project/quota/increase-request/increase-request.controller.js
```diff
@@ -66,7 +66,7 @@ export default class PciProjectQuotaIncreaseController {
 
     return this.OvhApiSupport.v6()
       .createTickets({
-        issueTypeId: ISSUE_TYPE_ID,
+        issueTypeId: this.issueType.id,
```
radireddy, line 69: do you think the issueType will never be null? there is always a issue for ISSUE_QUOTA_RANK

JeremyDec: I assume it will never be, as it has the same value than the previous ISSUE_TYPE_ID constant, which stored potential ID defining the Quota issue type. So if API returns issue type with no id, sure it will be broken here, but the problem would be on the API side. (Why would they send an issue type without identifier?)

packages/manager/modules/pci/src/projects/project/quota/increase-request/increase.constants.js
```diff
@@ -1,9 +1,9 @@
 export const ISSUE_CATEGORY = 'assistance';
 export const ISSUE_SERVICE_TYPE = 'cloud_project';
-export const ISSUE_TYPE_ID = 33381;
+export const ISSUE_TYPE_IDS = [33381, 33382, 43227, 44486, 44800, 44801];
```
JeremyDec, line 3: For the moment, no other way than storing hardcoded values

antleblanc (approved):

marie-j (approved):

varun257 (approved):

PR #5603, opened 2021-09-23
feat(hosting): hide the CDN order button if needed

y4nnL:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix MANAGER-6751

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

The user may not have enough privileges to order a CDN, therefore we hide the order button in this case

### DIFF
--- a/packages/manager/apps/web/client/app/hosting/general-informations/GENERAL_INFORMATIONS.controller.js
+++ b/packages/manager/apps/web/client/app/hosting/general-informations/GENERAL_INFORMATIONS.controller.js
@@ -30,6 +30,7 @@ export default class HostingGeneralInformationsCtrl {
     HostingRuntimes,
     hostingSSLCertificate,
     OvhApiScreenshot,
+    user,
   ) {
     this.$q = $q;
     this.$scope = $scope;
@@ -52,6 +53,7 @@ export default class HostingGeneralInformationsCtrl {
     this.HostingRuntimes = HostingRuntimes;
     this.hostingSSLCertificate = hostingSSLCertificate;
     this.OvhApiScreenshot = OvhApiScreenshot;
+    this.user = user;
   }
 
   $onInit() {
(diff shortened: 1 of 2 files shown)


packages/manager/apps/web/client/app/hosting/general-informations/GENERAL_INFORMATIONS.controller.js
```diff
@@ -188,6 +190,34 @@ export default class HostingGeneralInformationsCtrl {
...
+      flushCdnState,
+      hosting: { hasCdn, offer },
+    } = this.$scope;
+
+    const isFlushCdnStateOk = flushCdnState === 'ok';
+    const isOfferStart10M = offer === 'START_10_M';
+
+    const canOrderCdn =
+      (isFlushCdnStateOk && !hasCdn && !isOfferStart10M) ||
+      (hasCdn && isFlushCdnStateOk) ||
```
JeremyDec, line 204: This condition can be removed as the following OR will return true regardless of `isFlushCdnStateOk`

packages/manager/apps/web/client/app/hosting/general-informations/GENERAL_INFORMATIONS.controller.js
```diff
@@ -188,6 +190,34 @@ export default class HostingGeneralInformationsCtrl {
...
+    }
+
+    const {
+      hosting: {
+        serviceInfos: { contactAdmin, contactBilling },
+      },
+    } = this.$scope;
+    const { nichandle } = this.user;
+
+    return [contactAdmin, contactBilling].includes(nichandle);
```
JeremyDec, line 218: May this logic be in separated method? This way you call this new method and combine its result with the previous return.
And then finish this method by `return canOrderCdn || this.hasRightsToOrderCdn()`

packages/manager/apps/web/client/app/hosting/general-informations/GENERAL_INFORMATIONS.html
```diff
@@ -407,7 +407,7 @@
                         <oui-action-menu
                             data-compact
                             data-placement="end"
-                            data-ng-if="flushCdnState === 'ok' && !hosting.hasCdn && hosting.offer!== 'START_10_M' || hosting.hasCdn && flushCdnState === 'ok' || hosting.hasCdn"
+                            data-ng-if="$ctrl.canOrderCdn()"
```
JeremyDec, line 410: Nice catch to use method here :)

packages/manager/apps/web/client/app/hosting/general-informations/GENERAL_INFORMATIONS.controller.js
y4nnL: :+1: but with a `&&` operator

packages/manager/apps/web/client/app/hosting/general-informations/GENERAL_INFORMATIONS.controller.js
y4nnL: IMO we could push it a little bit further and simply write
`hasCdn || (isFlushCdnStateOk && !isOfferStart10M)`

But frankly I didn't wanted to rework anything about this business rule as it often leads to potential side effects (and because that's a first for me :stuck_out_tongue:)

packages/manager/apps/web/client/app/hosting/general-informations/GENERAL_INFORMATIONS.controller.js
JeremyDec: Yes I meant with `&&` ^^

frenautvh (approved):

PR #5611, opened 2021-09-23
fix: fixes the display of the SSL mode

Dimfacion:
PUD-911

Signed-off-by: Gael Leblan <gael.leblan@ovhcloud.com>

| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/PUD-748` 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #PUD-911 

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
fixes the display of the SSL mode

### DIFF
--- a/packages/manager/modules/pci/src/components/project/storages/databases/database.class.js
+++ b/packages/manager/modules/pci/src/components/project/storages/databases/database.class.js
@@ -1,6 +1,7 @@
 import find from 'lodash/find';
 
 import Base from './base.class';
+import { SSL_MODE_REQUIRED, SSL_MODE_NA } from './databases.constants';
 
 export default class Database extends Base {
   constructor({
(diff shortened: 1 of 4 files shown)


packages/manager/modules/pci/src/components/project/storages/databases/database.class.js
```diff
@@ -89,6 +90,16 @@ export default class Database extends Base {
     return this.currentEngine;
   }
 
+  getSSLModeKey() {
```
radireddy, line 93: This method might be called multiple times by the digest loop. Because of this the method includes will iterate the array multiple times. Is it possible to do this check once at instance level or only when something is changed? if not its ok with me. Its might improve negligible performance.

Dimfacion: It's a good point. Fixed :)

antleblanc (approved):

PR #5664, opened 2021-10-01
feat: add nutanix node interventions page

radireddy:
ref: MANAGER-7508

| Question         | Answer
| ---------------- | ---
| Branch?          | feat/nutanix
| Bug fix?         | yes/no
| New feature?     | yes/no
| Breaking change? | yes/no
| Tickets          | Fix #MANAGER-7508

- [ ] Commits are signed-off
- [ ] Lint has passed locally

## Description

Add Nutanix node interventions page

### DIFF
--- a/packages/manager/modules/nutanix/src/dashboard/nodes/node/interventions/module.js
+++ b/packages/manager/modules/nutanix/src/dashboard/nodes/node/interventions/module.js
@@ -1,0 +1,17 @@
+import angular from 'angular';
+import '@uirouter/angularjs';
+
+import { serverInterventions } from '@ovh-ux/manager-bm-server-components';
+
+import routing from './routing';
+import component from './component';
+
+const moduleName = 'ovhManagerNutanixNodeInterventions';
+
+angular
+  .module(moduleName, ['ui.router', serverInterventions])
+  .config(routing)
+  .component('nutanixNodeInterventions', component)
+  .run(/* @ngTranslationsInject:json ./translations */);
+
+export default moduleName;
--- a/packages/manager/modules/nutanix/src/dashboard/nodes/node/interventions/template.html
+++ b/packages/manager/modules/nutanix/src/dashboard/nodes/node/interventions/template.html
@@ -1,0 +1,4 @@
+<server-interventions
+    service-name="{{ $ctrl.node.name }}"
+    on-error="$ctrl.handleError(error)"
+></server-interventions>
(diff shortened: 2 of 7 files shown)


packages/manager/modules/nutanix/src/dashboard/nodes/node/interventions/module.js
```diff
@@ -0,0 +1,17 @@
...
+
+import { serverInterventions } from '@ovh-ux/manager-bm-server-components';
+
+import routing from './routing';
+import component from './component';
+
+const moduleName = 'ovhManagerNutanixNodeInterventions';
+
+angular
+  .module(moduleName, ['ui.router', serverInterventions])
```
antleblanc, line 12: **question:** Since you are using `breadcrumb` into the routing part, shouldn't `ngUiRouterBreadcrumb` be injected as a module dependency here?

ganeshkumar1989 (approved):

packages/manager/modules/nutanix/src/dashboard/nodes/node/interventions/template.html
```diff
@@ -0,0 +1,4 @@
+<server-interventions
+    service-name="{{ $ctrl.node.name }}"
+    on-error="$ctrl.handleError(error)"
+></server-interventions>
```
marie-j, line 4: I'm wondering if we could use this component directly in routing and avoid a view + controller in this case

packages/manager/modules/nutanix/src/dashboard/nodes/node/interventions/module.js
radireddy: I need clarification here. I don't see breadcrumb used in all modules. Its mainly used in the parent module. I have used it in the parent module of Nutanix. Since these modules (nutanix child modules) can not be used in isolation, I don't see the need for it. But if it is something a guideline or practice, I can add it.

packages/manager/modules/nutanix/src/dashboard/nodes/node/interventions/template.html
radireddy: I prefer to use routes only for pure resolver methods/variables. Controller for other kind of methods. If I directly use the component inside the route, I need to define the handleError method inside the route which I would like to avoid. Also having a controller provides flexibility to add other features easily like tracking etc.

mohammed-zahaf (approved):

PR #5668, opened 2021-10-04
Feat/shell routing plugin

frenautvh:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-7750

move shell routing into a plugin


packages/components/ovh-shell/package.json
```diff
@@ -11,17 +11,20 @@
...
-  "exports": {
-    ".": "./dist/index.js",
-    "./client": "./dist/client/index.js",
-    "./plugin/i18n": "./dist/plugin/i18n/index.js"
-  },
   "main": "./dist/index.js",
   "scripts": {
-    "build": "tsc"
+    "build": "tsc",
+    "dev": "tsc -w"
```
mohammed-zahaf, line 17: i try to found `-w` option into the typescript compiler documentation but nothing.
is't sure this option is available ?

I found this `tsc --watch`

antleblanc: It's a shorthand

```sh
$ npx tsc --help
Version 3.8.3
Syntax:   tsc [options] [file...]

Examples: tsc hello.ts
          tsc --outFile file.js file.ts
          tsc @args.txt
          tsc --build tsconfig.json

Options:
 -h, --help                                         Print this message.
 -w, --watch                                        Watch input files.
```

;)

packages/components/ovh-shell/src/client/index.ts
```diff
@@ -1,8 +1,10 @@
 import ShellClient from './shell-client';
 import exposeApi from './api';
 
-export default function useShellApi() {
+export function useShellApi() {
```
Jisay, line 4: Any reason of not using arrow functions?

Jisay (approved):

packages/components/ovh-shell/src/client/index.ts
zakariasahmane: I don't think it changes anything here, the function relies on its own context so using arrow functions doesn't have any added value

zakariasahmane (approved):

packages/components/ovh-shell/src/client/index.ts
frenautvh: yes i think it's a matter of preferences, i believe that using named regular function might facilitates debugging as you get the function name in the debugger but that's pretty much it

PR #5679, opened 2021-10-04
fix(dedicated.server.ipmi): add with-geo-location parameter

mohammed-zahaf:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-7424

- [ X ] Try to keep pull requests small so they can be easily reviewed.
- [ X ] Commits are signed-off
- [ ] Lint has passed locally

## Description
add `withGeoLocation` parameter


frenautvh (approved):

packages/manager/apps/dedicated/client/app/dedicated/server/ipmi/dedicated-server-ipmi.controller.js
```diff
@@ -549,13 +549,18 @@ angular
 
       // ------------Start KVM URL------------
       $scope.getIpmiKvmUrl = function getIpmiKvmUrl() {
+        const withGeolocation =
+          !includes(['HIL_1', 'VIN_1'], $scope.server.datacenter) &&
+          coreConfig.isRegion('US');
```
marie-j, line 554: Maybe the `coreConfig.isRegion` could be replaced with feature flipping

mohammed-zahaf: Thanks Marie, goal here is to pass `withGeolocation` parameter (true/false) when we start an IPMI session not to activate/deactivated a feature.

marie-j: Adding geolocation could be seen as a feature in some sense and here it seems that it's a feature only available for US region so it could make sense to use feature flipping imo (we could want to add geolocation in some future for other regions let's say)

mohammed-zahaf: The KVM console don't accept a connection without a trusted IP address.
So we have to pass `withGeolocation: true` to pass the IP addess.

varun257 (approved):

radireddy (approved):

PR #5908, opened 2021-11-08
Feat/manager shell callback

frenautvh:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-7942

## Description

Refactor ovh-shell component to add callback / listener capabilities between shell client & shell

### DIFF
--- a/packages/components/ovh-shell/features/client/shell-client.feature
+++ b/packages/components/ovh-shell/features/client/shell-client.feature
@@ -6,3 +6,8 @@ Scenario: Plugin method invokation
   When I invoke a method from my plugin
   Then The invokation should be resolved with the method result
 
+Scenario: Plugin event listener
+  Given My shell and shell client are configured with a direct message bus
+  And I add an event listener on the shell client
+  When The shell emits an event
+  Then The event emitted from the shell should be received by the client
--- a/packages/components/ovh-shell/src/client/api.ts
+++ b/packages/components/ovh-shell/src/client/api.ts
@@ -1,19 +1,29 @@
 import ShellClient from './shell-client';
 
-export default function exposeApi(shell: ShellClient) {
+export default function exposeApi(shellClient: ShellClient) {
   return {
     i18n: {
       getLocale: () =>
-        shell.invokePluginMethod({
+        shellClient.invokePluginMethod({
           plugin: 'i18n',
           method: 'getLocale',
+        }),
+      onLocaleChange: (callback: CallableFunction) =>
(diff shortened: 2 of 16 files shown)


packages/components/ovh-shell/features/client/shell-client.feature
```diff
@@ -6,3 +6,7 @@ Scenario: Plugin method invokation
   When I invoke a method from my plugin
   Then The invokation should be resolved with the method result
 
+Scenario: Plugin event listener
+  When My shell and shell client are configured with a direct message bus
```
zakariasahmane, line 10: `When` should be used when an action happens. This line is more related to configuration/prerequisites, so it should be `Given`. 

Also, to respect the Gherkin pattern, we should always have `Given, When, Then` and they are nearly the same as the AAA pattern ( Arrange, Act, Assert ). 

In this case, it should be more like : 

`Given My shell and shell client are configured with a direct message bus`
`And I add an event listener on the shell client`
`When The shell emits an event`
`Then The event emitted from the shell should be received by the client`

packages/components/ovh-shell/features/client/shell-client.feature
```diff
@@ -6,3 +6,8 @@ Scenario: Plugin method invokation
   When I invoke a method from my plugin
   Then The invokation should be resolved with the method result
 
+Scenario: Plugin event listener
+  Given My shell and shell client are configured with a direct message bus
+  And I add an event listener on the shell client
+  When Then shell emits an event
```
Jisay, line 12:
```suggestion
  When Shell emits an event
```
?

packages/components/ovh-shell/src/client/api.ts
```diff
@@ -1,19 +1,29 @@
...
   return {
     i18n: {
       getLocale: () =>
-        shell.invokePluginMethod({
+        shellClient.invokePluginMethod({
           plugin: 'i18n',
           method: 'getLocale',
         }),
+      onLocaleChange: (callback: CallableFunction) =>
+        shellClient.addEventListener('locale-change', callback),
```
Jisay, line 12: Maybe it should be a good idea to prefix the event name with the name of the plugin?
Something like `i18n:locale-change`?
I searched for best practice and I found this article: https://gomakethings.com/custom-event-naming-conventions-in-vanilla-js/

cbourgois (approved):

Jisay (approved):

zakariasahmane (approved):
Nice !

PR #5934, opened 2021-11-09
feat(baremetal.commitment): fix issue with upfront price

varun257:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-56164

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

move logic to main controller to execute on click of next


marie-j (changes requested):
This will break logic for Netapp order funnel as component is used directly and not using commitment page.

packages/manager/modules/billing-components/src/components/commitment/commitment.controller.js
```diff
@@ -70,9 +71,9 @@ export default class {
   }
 
   onDurationChange(duration) {
-    this.pricingModes = this.availableEngagements.filter(
-      (commitment) => commitment.durationInMonths === duration.monthlyDuration,
-    );
+    this.pricingModes = this.availableEngagements.filter((commitment) => {
+      return commitment.durationInMonths === duration.monthlyDuration;
+    });
```
JeremyDec, line 76: Why not keep the current form?

varun257:
> This will break logic for Netapp order funnel as component is used directly and not using commitment page.

Oh okay i did not see that. I have to execute the function on load then and not `onInit`

PR #5937, opened 2021-11-09
Feat/manager shell i18n plugin

Jisay:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/manager-shell-callback`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-7858

## Description

Add i18n plugin to ovh-shell.

### DIFF
--- a/packages/components/ovh-shell/features/plugin/i18n/i18n.feature
+++ b/packages/components/ovh-shell/features/plugin/i18n/i18n.feature
@@ -1,0 +1,16 @@
+Feature: i18n Plugin
+
+Scenario: Retrieving locale when nothing has changed
+  Given I have a i18n plugin instanciated
+  When I try to get the locale
+  Then I should get the default locale
+
+Scenario: Setting the locale
+  Given I have a i18n plugin instanciated
+  When I try to change the locale
+  Then I should retrieve the setted locale
+
+Scenario: Getting the list of available locales
+  Given I have a i18n plugin instanciated
+  When I try to get the list of available locales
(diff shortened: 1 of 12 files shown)


packages/components/ovh-shell/features/plugin/i18n/i18n.feature
```diff
@@ -0,0 +1,16 @@
...
+  Given I have a i18n plugin instanciated
+  When I call getLocale method
+  Then I should get the default locale
+
+Scenario: setLocale method
+  Given I have a i18n plugin instanciated
+  When I call setLocale method
+  Then I should have the locale updated
+
+Scenario: getAvailableLocales method
```
zakariasahmane, line 13: In the `Scenario`, you shouldn't test a "method", but a scenario as written. 

As an example, for the first block of code, the scenario would be "Retrieving a locale" or something like this. The scenarios shouldn't be tied to how the code is written so that they're futurproof and we can refactor the code safely without breaking the feature.

So the first one would be : 

`Scenario: Retrieving locale`
`Given I have a i18n plugin instanciated`
`When I try to get the locale`
`Then I should get the default locale of my plugin`

And let the code in your test do the details that are related to the code.

Jisay: Thanks for your feedback, I have updated the tests

frenautvh (approved):

zakariasahmane (approved):

PR #5965, opened 2021-11-16
Fix/url html injection

JeremyDec:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-8092

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Block HTML tag potentially present in any url params (route path or url search query).

(See DTRSD-11881 internal ticket for more details)

### DIFF
--- a/packages/manager/modules/core/src/index.js
+++ b/packages/manager/modules/core/src/index.js
@@ -28,6 +28,7 @@ import redirectionService from './redirection/redirection.service';
 import managerCoreUrlBuilder from './url-builder';
 
 import { registerConfigModule } from './config';
+import sanitizeUrl from './sanitize-url/sanitize-url';
 
 import { URLS } from './manager-core.constants';
 
@@ -270,7 +271,8 @@ export const registerCoreModule = (environment) => {
       /* @ngInject */ ($transitions, $translate) => {
         $transitions.onBefore({}, () => $translate.refresh());
       },
-    );
+    )
+    .run(sanitizeUrl);
 
   return moduleName;
 };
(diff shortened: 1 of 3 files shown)


antleblanc (commented):
**question:** Can this logic be delegated to the `@ovh-ux/manager-core` module instead? (if make sense).

My main concern is to avoid code duplication.

Thanks!

packages/manager/modules/core/src/secure-url/index.js
```diff
@@ -0,0 +1,8 @@
+import angular from 'angular';
+import secureUrl from './secure-url.config';
+
+const moduleName = 'coreSecureUrl';
+
+angular.module(moduleName, []).run(secureUrl);
```
antleblanc, line 6: **question:** As `$transitions` provider is used, shouldn't `@uirouter/angularjs` module be injected as a module dependency?

packages/manager/modules/core/src/index.js
```diff
@@ -28,6 +28,7 @@ import redirectionService from './redirection/redirection.service';
 import managerCoreUrlBuilder from './url-builder';
 
 import { registerConfigModule } from './config';
+import secureUrl from './secure-url/secure-url';
```
marie-j, line 31: suggestion: 
```suggestion
import sanitizeUrl from './secure-url/secure-url';
```
is more explicit about what the module does

marie-j (dismissed):

packages/manager/modules/core/src/index.js
JeremyDec: changed

antleblanc (dismissed):

varun257 (dismissed):

JeremyDec:
rebase on `develop` since last approvals

marie-j (approved):

antleblanc (approved):

PR #5977, opened 2021-11-17
Feat/manager shell auth plugin

frenautvh:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-7886

## Description

Add authentication plugin in manager shell

### DIFF
--- a/packages/components/ovh-shell/__tests__/plugin/auth/auth.spec.ts
+++ b/packages/components/ovh-shell/__tests__/plugin/auth/auth.spec.ts
@@ -1,0 +1,46 @@
+import { loadFeature, defineFeature } from 'jest-cucumber';
+import auth from '../../../src/plugin/auth';
+
+const feature = loadFeature('../../../features/plugin/auth/auth.feature', {
+  loadRelativePath: true,
+});
+
+defineFeature(feature, (test) => {
+  let authPlugin;
+  const loginFn = jest.fn();
+  const logoutFn = jest.fn();
+
+  test('Log in', ({ given, when, then }) => {
+    given('An authentication plugin', () => {
+      authPlugin = auth({
+        goToLogin: loginFn,
+        goToLogout: logoutFn,
+      });
+    });
+
+    when('I try to log in', () => {
+      authPlugin.login();
+    });
+
+    then('I should go to the login page', () => {
+      expect(loginFn).toHaveBeenCalled();
+    });
+  });
+
+  test('Log out', ({ given, when, then }) => {
+    given('An authentication plugin', () => {
+      authPlugin = auth({
+        goToLogin: loginFn,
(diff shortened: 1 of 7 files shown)


packages/manager/modules/config/src/index.ts
```diff
@@ -20,6 +20,15 @@ export { LANGUAGES } from './locale/locale.constants';
 
 export const isTopLevelApplication = () => window.top === window.self;
 
+export const getAuthenticationURL = () => {
```
antleblanc, line 23: **thought:** This logic looks quite similar to https://github.com/ovh/manager/blob/master/packages/components/ovh-reket/src/redirections.js

**question:** Is there any plan to add some refactor around it?

Jisay: Could be a good point to share this logic at the same place

Jisay (approved):

cbourgois (approved):

packages/components/ovh-shell/__tests__/plugin/auth/auth.spec.ts
```diff
@@ -0,0 +1,102 @@
...
+    });
+
+    when('I try to log out', () => {
+      authPlugin.logout();
+    });
+
+    then(
+      'I should get redirected to the authentication page with logout parameters',
+      () => {
+        expect(window.location.assign).toHaveBeenCalledWith(
```
zakariasahmane, line 94: This is not a reliable test. It tests too much the implementation details. We shouldn't test if `window.location.assign`is called with certain parameters, and we don't care that if it's called or not. 

Because it can change. 

But i know that DOM methods can be painful to test, but to get around that, we should split the code. 
We know that `window.location.assign`, will always call `addParameters`, so we should test the output of that method instead. So that we know that that method, is tested in a **unitary** way. 
And since you have logic in `login`/`logout`  that construct `params` based on `authURL` and `loginParameters`, you should extract that logic to test it separately, for example in a `constructAuthParameters(login/logout, LoginParams)` method.
And since these methods will be tested unitarily, you won't have to test the whole `login`/`logout` method. Which can actually be better tested with `e2e`tests. 

Or another solution, is to wrap your `window.location.assign` in another method, for example `redirect` and test that `redirect` is called with the parameters you want. But this solution should be your last resort if there is no another solution. So the above solution is better to have more reliable unit tests.

zakariasahmane (changes requested):

packages/manager/modules/config/src/index.ts
frenautvh: good point i'll see what i can do to reuse it

Jisay (approved):

zakariasahmane (approved):

PR #6002, opened 2021-11-22
Feat/manager shell ux plugin

Jisay:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/manager-shell`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | yes
| Tickets          | MANAGER-8052

## Description

Add ux plugin to ovh-shell and app-shell

### DIFF
--- a/packages/components/ovh-shell/src/plugin/ux/components/component.ts
+++ b/packages/components/ovh-shell/src/plugin/ux/components/component.ts
@@ -1,0 +1,43 @@
+export interface IUxComponentOptions {
+  visible?: boolean;
+}
+
+export interface IUxComponent {
+  toggleVisibility: CallableFunction;
+  show: CallableFunction;
+  hide: CallableFunction;
+  getVisibility: CallableFunction;
+}
+
+const uxComponent = (options: IUxComponentOptions = {}): IUxComponent => {
+  let visible = options.visible || false;
+
+  const setVisibility = (visibility: boolean) => {
+    visible = visibility;
+  };
+
+  const toggleVisibility = () => {
+    setVisibility(!visible);
+  };
+
+  const show = () => {
+    setVisibility(true);
+  };
+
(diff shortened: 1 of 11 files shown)


cbourgois (approved):

packages/components/ovh-shell/src/plugin/ux/components/component.ts
```diff
@@ -0,0 +1,43 @@
+export interface IUxComponentOptions {
+  visible?: boolean;
+}
+
+export interface IUxComponent {
+  toggleVisibility: CallableFunction;
+  show: CallableFunction;
+  hide: CallableFunction;
+  getVisibility: CallableFunction;
```
zakariasahmane, line 9: These function have a specific signature and should be defined rather than using `CallableFunction` to define them. 
We use `CallableFunction` when we don't know the signature of the function, like in the PluginManager.

packages/components/ovh-shell/src/plugin/ux/ux.ts
```diff
@@ -0,0 +1,73 @@
...
+}
+
+export interface IShellUx {
+  registerSidebar: CallableFunction;
+  isSidebarVisible: CallableFunction;
+  toggleSidebarVisibility: CallableFunction;
+  showSidebar: CallableFunction;
+  hideSidebar: CallableFunction;
+  registerNavbar: CallableFunction;
+  getNavbar: CallableFunction;
```
zakariasahmane, line 15: Same as previous comment for CallableFunction

zakariasahmane:
I also see in the `core/ux/index.js` that we are importing `lodash-es` to use `set`. Since we are using modern Javascript, i think we could avoid using lodash and doing it in plain Javascript.

zakariasahmane (changes requested):

PR #6019, opened 2021-11-24
Feat/manager shell config

frenautvh:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | yes
| Tickets          | MANAGER-8163

## Description

fetch /configuration in manager-shell

### DIFF
--- a/packages/components/ovh-shell/src/client/index.ts
+++ b/packages/components/ovh-shell/src/client/index.ts
@@ -4,13 +4,13 @@ import { isTopLevelApplication } from '@ovh-ux/manager-config';
 import ShellClient from './shell-client';
 import StandaloneShellClient from './standalone-shell-client';
 import IFrameMessageBus from '../message-bus/iframe';
-import exposeApi from './api';
 
 interface ApplicationConfiguration {
   universe: string;
   url: string;
-  standalone?: boolean;
+  useShell?: boolean;
   shellPath?: string;
+  publicURL?: string;
 }
 
 function fetchApplications(): Promise<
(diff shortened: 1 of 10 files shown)


Jisay (approved):

packages/components/ovh-shell/src/client/index.ts
```diff
@@ -24,26 +26,45 @@ function fetchApplications(): Promise<
...
+  apps: Record<string, ApplicationConfiguration>,
+): Promise<StandaloneApplicationResult> {
+  return new Promise<StandaloneApplicationResult>((resolve) => {
+    Object.entries(apps).forEach(([, appConfig]) => {
+      const urlWithoutHash = new URL(window.location.href);
+      urlWithoutHash.hash = '';
+      if (!appConfig.standalone && urlWithoutHash.href === appConfig.url) {
+        const redirection = new URL(window.location.href);
+        redirection.pathname = appConfig.shellPath;
+        resolve({ isStandalone: false, redirectionURL: redirection.href });
```
zakariasahmane, line 44: So in the end we're naming it `isStandalone` instead of `useShell` ?

packages/components/ovh-shell/src/client/standalone-shell-client.ts
```diff
@@ -0,0 +1,34 @@
...
+  init(): Promise<ShellClient> {
+    return initShell(this.shell).then((shellApi) => {
+      // @TODO set universe here when 2API configuration is patched
+      // const environment = shellApi.getEnvironment();
+      // environment.setUniverse();
+      shellApi.registerPlugin(
+        'i18n',
+        plugin.i18n(this.shell, shellApi.getEnvironment()),
+      );
+      return this;
```
zakariasahmane, line 31: Is this `return` necessary ? 

Since this is a class, you already know the `this` before calling it, otherwise, you wouldn't be able to call the init method. 
The method just registers a `Plugin` so it should just be a `Promise<void>`, no ?

packages/components/ovh-shell/src/shell/index.ts
```diff
@@ -1,22 +1,39 @@
...
+interface IShell {
+  getEnvironment: () => Environment;
+  emitEvent: (eventId: string, data: unknown) => void;
+  registerPlugin: (
+    pluginId: string,
+    pluginApi: Record<string, CallableFunction>,
+  ) => void;
+  setPluginAvailability: (pluginId: string, availability: boolean) => void;
+  setIframeMessageBus: (iframe: HTMLIFrameElement) => void;
+  i18n: () => unknown;
```
zakariasahmane, line 16: I think this is of type `Record<string, CallableFunction>`. 

It comes from `shell.getPluginManager().plugins.i18n.instance` , `i18n` being of type `PluginEntry` which has the `instance` object in it.

zakariasahmane (changes requested):

Jisay (approved):

Jisay (approved):

PR #6048, opened 2021-11-29
feat(reket): export redirection functions

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | 

## Description

For manager shell authentication plugin we need to have access to login & logout functions.
In order to avoid code duplication, export of those functions have been added.

Also removed the default export to avoid "mixing import" warning when building with cjs target


packages/components/ovh-reket/src/index.js
```diff
@@ -2,7 +2,13 @@ import { Reket } from '@ovhcloud/reket-core';
...
 
 import { DEFAULT_REQUEST_TYPES } from './constants';
-import { redirectToLoginPage, redirectToLogoutPage } from './redirections';
+import {
+  redirectToLoginPage as toLogin,
+  redirectToLogoutPage as toLogout,
+} from './redirections';
+
+export const redirectToLoginPage = toLogin;
+export const redirectToLogoutPage = toLogout;
```
zakariasahmane, line 11: `import { redirectToLoginPage, redirectToLogoutPage } from './redirection`
`export redirectToLoginPage`
`export redirectToLogoutPage`

Doesn't work? you are forced to work with aliases?

zakariasahmane (dismissed):

packages/components/ovh-reket/src/index.js
frenautvh:
```js
export redirectToLoginPage; // Unexpected token (missing const)
```
```js
export const redirectToLoginPage; // Identifier 'redirectToLoginPage' has already been declared
```
it's kind of ugly but i'm not sure if there are other ways around that

Jisay: Something like:
```
export * from './redirections';
``` ?

Jisay (approved):

PR #6098, opened 2021-12-07
feat(hub): use i18n shell plugin

Jisay:
Signed-off-by: Jisay <jean-christophe.alleman@corp.ovh.com>

| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/manager-shell`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #... 

## Description

Use i18n shell plugin in `@ovh-ux/manager-hub-app` package.


packages/manager/apps/hub/src/shell.js
```diff
@@ -0,0 +1,19 @@
+let shellClient;
+
+export const setShellClient = (client) => {
+  shellClient = client;
+
+  // set callbacks
+  // set on locale change
+  shellClient.i18n.onLocaleChange(() => {
+    window.location.reload();
```
Jisay, line 9: Don't know if we need to reload `window.top`?

packages/manager/modules/core/src/index.js
```diff
@@ -158,8 +158,12 @@ export const registerCoreModule = (environment) => {
     .run(
       /* @ngInject */ ($rootScope, coreConfig) => {
         $rootScope.$on('lang.onChange', (event, { lang }) => {
-          coreConfig.setUserLocale(lang);
-          window.location.reload();
+          if (shellClient) {
```
frenautvh, line 161: instead of passing a shellClient instance to core module i think it's cleaner to make the onChange callback function configurable (if not set do the default reload behavior, otherwise call the callback)

that way manager-core is not tied to ovh-shell

packages/manager/apps/hub/src/shell.js
frenautvh: do we need to reload the app-shell to see changes ? if so we need to reload window.top too, if it's dynamically updated in the react app-shell then it's ok

packages/manager/modules/core/src/index.js
Jisay: This done

packages/components/ovh-shell/src/plugin/i18n/index.ts
```diff
@@ -3,11 +3,7 @@ import { KeyPairName } from '@ovh-ux/manager-config/types/locale';
 
 import Shell from '../../shell/shell';
 
-export interface I18nPlugin {
```
frenautvh, line 6: why removing the plugin interface definition with an abstract Record<string, CallableFunction> ?

packages/manager/modules/core/src/index.js
```diff
@@ -158,8 +165,12 @@ export const registerCoreModule = (environment) => {
     .run(
       /* @ngInject */ ($rootScope, coreConfig) => {
         $rootScope.$on('lang.onChange', (event, { lang }) => {
-          coreConfig.setUserLocale(lang);
-          window.location.reload();
+          if (onLocaleChange) {
+            onLocaleChange(lang);
```
frenautvh, line 169: don't you want to call
```js
coreConfig.setUserLocale(lang);
```
in both cases ?

packages/manager/apps/hub/src/shell.js
Jisay: For now, I think we need to reload top window as in the app-shell, we don't manage i18n refresh.

packages/manager/modules/core/src/index.js
Jisay: I think you have right (even it's working without >_<)

packages/components/ovh-shell/src/plugin/i18n/index.ts
zakariasahmane: Because it caused problems when integrating elsewhere, ideally, we should have Plugin definitions, with a `Plugin` interface, and other child interfaces that inherit from it. 

We should work on all Plugin definitions like we said, to have a more generic class etc

packages/manager/modules/core/src/index.js
Jisay: In fact it's not necessary :-)

As `onLocaleChange` callback does the same as `coreConfig.setUserLocale`. It sets the user locale of `environment` object through i18n plugin : https://github.com/ovh/manager/pull/6098/files#diff-17953c7ed48e08a390075d0a4ab263796472f2d1e6c2a1d167a7eef2fb72b85aR54-R59

frenautvh (approved):

zakariasahmane (approved):

PR #6251, opened 2022-01-06
fix(netapp): use cucBytes to get right value and unit

mohammed-zahaf:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-7916

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
Use bytes filter to convert storage capacity to right unit value

### DIFF
--- a/packages/manager/modules/netapp/package.json
+++ b/packages/manager/modules/netapp/package.json
@@ -25,6 +25,7 @@
     "@ovh-ux/manager-core": "^11.0.0",
     "@ovh-ux/manager-models": "^1.11.1",
     "@ovh-ux/manager-ng-layout-helpers": "^2.0.0",
+    "@ovh-ux/manager-filters": "^0.1.0 || ^1.0.0",
     "@ovh-ux/ng-at-internet": "^5.8.0",
     "@ovh-ux/ng-at-internet-ui-router-plugin": "^3.2.0",
     "@ovh-ux/ng-ovh-feature-flipping": "^1.0.4",
--- a/packages/manager/modules/netapp/src/dashboard/template.html
+++ b/packages/manager/modules/netapp/src/dashboard/template.html
@@ -59,7 +59,7 @@
                     </oui-tile-definition>
                     <oui-tile-definition
                         term="{{:: 'netapp_dashboard_information_quota' | translate }}"
-                        description="{{ 'netapp_dashboard_information_quota_formatted' | translate:{quota: $ctrl.storage.quota } }}"
+                        description="{{ $ctrl.storage.quota | bytes:'GB':false:0 }}"
                     >
                     </oui-tile-definition>
                 </oui-tile>
--- a/packages/manager/apps/netapp/package.json
+++ b/packages/manager/apps/netapp/package.json
@@ -30,6 +30,7 @@
     "@ovh-ux/manager-netapp": "^0.0.0 || ^1.0.0",
     "@ovh-ux/manager-ng-layout-helpers": "^2.0.0",
     "@ovh-ux/manager-preloader": "^1.1.0",
+    "@ovh-ux/manager-filters": "^0.1.0 || ^1.0.0",
     "@ovh-ux/ng-at-internet": "^5.8.0",
     "@ovh-ux/ng-at-internet-ui-router-plugin": "^3.2.0",
     "@ovh-ux/ng-ovh-api-wrappers": "^4.0.10",
--- a/packages/manager/modules/netapp/src/dashboard/module.js
+++ b/packages/manager/modules/netapp/src/dashboard/module.js
@@ -1,5 +1,6 @@
 import angular from 'angular';
 import '@ovh-ux/manager-core';
+import '@ovh-ux/manager-filters';
 import '@uirouter/angularjs';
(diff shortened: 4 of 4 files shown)


packages/manager/modules/netapp/src/dashboard/template.html
```diff
@@ -59,7 +59,7 @@
                     </oui-tile-definition>
                     <oui-tile-definition
                         term="{{:: 'netapp_dashboard_information_quota' | translate }}"
-                        description="{{ 'netapp_dashboard_information_quota_formatted' | translate:{quota: $ctrl.storage.quota } }}"
+                        description="{{ $ctrl.storage.quota | cucBytes:0:false:'GB' }}"
```
antleblanc, line 62: **issue(blocking):** `cucBytes` filter come from `@ovh-ux/ng-ovh-cloud-universe-components`.
IMHO it doesn't make sense to have it as a peerDependency since NetApp is associated with Bare Metal Cloud section.

antleblanc (changes requested):

packages/manager/modules/netapp/src/dashboard/template.html
mohammed-zahaf: I am not sure to ge your point of view.
Currently the `"@ovh-ux/ng-ovh-cloud-universe-components"` is only declared as dependency into `packages/manager/apps/dedicated/package.json` dependecies node.
So regarding others modules (manager/modules/*), some modules which required the `cucBytes` add it as peerDependency.

What is the issue and do you have a suggestions ?

antleblanc: Did you start the standalone application `@ovh-ux/manager-netapp-app` (which is highly recommend)?
You'll noticed that `cucBytes` filter isn't working.

JeremyDec: @mohammed-zahaf it might be interesting to look at the filter located here: https://github.com/ovh/manager/tree/master/packages/manager/modules/filters/src/bandwidth. This way you could avoid relying on `@ovh-ux/ng-ovh-cloud-universe-component`

mohammed-zahaf: Thanks i add the filter module as deps.

packages/manager/modules/netapp/package.json
```diff
@@ -25,6 +25,7 @@
     "@ovh-ux/manager-core": "^11.0.0",
     "@ovh-ux/manager-models": "^1.11.1",
     "@ovh-ux/manager-ng-layout-helpers": "^2.0.0",
+    "@ovh-ux/manager-filters": "^0.1.0 || ^1.0.0",
```
antleblanc, line 28: **issue(blocking):** If this package is required, then we have to import/use it in `package/manager/modules/netapp/*`. Correct me if I'm wrong, but it's not the case here.

If package named `@ovh-ux/manager-filters` is required as a `peerDependency` for the given module `@ovh-ux/manager-netapp`, we have to add it also as a `dependency` for the standalone application `@ovh-ux/manager-netapp-app`.

:bulb: **Tips:** It's highly recommended to use the standalone application to develop features/bugfixes for product section.

```sh
$ yarn workspace @ovh-ux/manager-netapp-app run start:dev
```

packages/manager/modules/netapp/src/dashboard/template.html
antleblanc: **question:** Couldn't be replaced by filter suggested by @JeremyDec?

antleblanc (changes requested):

packages/manager/modules/netapp/package.json
mohammed-zahaf: Netapp is not part of dedicated univers ?
Because i see in dedicated and we have this deps.

packages/manager/modules/netapp/src/dashboard/template.html
marie-j, line 62: manager-filters has been added but we still use `cucBytes` ?

mohammed-zahaf: Ah i was think this module have a `cucBytes` filter.
So i am note sure this module give the right unit value.

This filter is for bandwidth (B/s) and i have to convert a quota (B).
There is a way to achieve my task with the bandwidth filter ?

mohammed-zahaf: Hello @marie-j, please read my comment here:
https://github.com/ovh/manager/pull/6251#discussion_r782844305

mohammed-zahaf: As we discuss together @JeremyDec, i changed the the `cucBytes` by `bytes`.

antleblanc (approved):

packages/manager/modules/netapp/package.json
antleblanc: NetApp and Dedicated are two separated Apps but both are using same module `@ovh-ux/manager-netapp`.

PR #6286, opened 2022-01-10
feat(netapp): add a tooltip to access column

mohammed-zahaf:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-7860

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
Add tooltip behavior to accessTo column to inform customers which value format is required.

### DIFF
--- a/packages/manager/modules/netapp/src/dashboard/volumes/dashboard/acl/template.html
+++ b/packages/manager/modules/netapp/src/dashboard/volumes/dashboard/acl/template.html
@@ -7,7 +7,7 @@
         ></abbr
         >)
     </h2>
-    <p data-translate="netapp_volumes_acl_description"></p>
+    <p id="acl_access_description" data-translate="netapp_volumes_acl_description"></p>
     <oui-button
         variant="secondary"
         disabled="$ctrl.isAdding"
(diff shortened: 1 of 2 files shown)


packages/manager/modules/netapp/src/dashboard/volumes/dashboard/acl/template.html
```diff
@@ -17,113 +17,142 @@ <h2>
...
+                ></span>
+            </th>
+            <th class="oui-table__header" scope="col">
+                <span
+                    data-translate="netapp_volumes_acl_access_to"
+                ></span>
+                <button
+                    type="button"
+                    class="oui-popover-button"
+                    oui-popover="{{:: 'netapp_volumes_acl_access_to_ip_format' | translate }}"
```
antleblanc, line 36: **question:** Why not prefer using built-in feature from Datagrid component?
see: https://ovh.github.io/ovh-ui-kit/?path=/story/design-system-components-datagrid-webcomponent--helper

packages/manager/modules/netapp/src/dashboard/volumes/dashboard/acl/template.html
```diff
@@ -17,113 +17,142 @@ <h2>
     >
         <span data-translate="netapp_volumes_acl_add"></span>
     </oui-button>
-    <oui-datagrid rows="$ctrl.acls" ng-if="$ctrl.acls.length">
-        <oui-datagrid-column
-            title=":: 'netapp_volumes_acl_access_type' | translate"
+
+    <table class="oui-table" data-ng-if="$ctrl.acls.length">
```
JeremyDec, line 21: Could the Sonarcloud report be taken in account here please?

packages/manager/modules/netapp/src/dashboard/volumes/dashboard/acl/template.html
mohammed-zahaf: It was datagrid before, so the requirement it was to add a tooltip into the header column.
And the datagrid don't allow this.

packages/manager/modules/netapp/src/dashboard/volumes/dashboard/acl/template.html
mohammed-zahaf: Yep sure i will take a look

packages/manager/modules/netapp/src/dashboard/volumes/dashboard/acl/template.html
JeremyDec: @mohammed-zahaf more precisions: At the time we develop NetApp feature, the datagrid didn't allow to set tooltip. Since then, development in the ui-kit was made to add the tooltip. So you can use, as @antleblanc mentioned, the new `helper` attribute on `oui-datagrid-column`

mohammed-zahaf: As we discuss together @JeremyDec, i have try to use the new `helper` attribute but it's not working.
I have try the two way :
`helper="{{:: 'netapp_volumes_acl_access_to_ip_format' | translate}}"` 
`helper="'netapp_volumes_acl_access_to_ip_format' | translate`

I tried to translate this key in a dedicated span, it works fine:
```
<span>{{:: 'netapp_volumes_acl_access_to_ip_format' | translate}}</span>
```

To be sure there is no others dependncy that i missed I also update do:
1. checkout to develop
2. git pull
3. yarn install
4. yarn run start:dev

Even i do these it change any things.

mohammed-zahaf: Did you already try it on your side ?

antleblanc: Noticed that there's a bug with translation and also if we are applying it in a `sortable` column.
Thanks for pointing out this @mohammed-zahaf.

Move from Datagrid to regular table here looks good then (till we fix the component).

antleblanc (approved):

varun257 (approved):

packages/manager/modules/netapp/src/dashboard/volumes/dashboard/acl/template.html
marie-j: Maybe this bug can be tracked by an issue or something ?

PR #6298, opened 2022-01-12
feat(public-cloud.database): tracking for karapace project

anooparveti:
| Question         | Answer
| ---------------- | ---
| Branch?          |  `feat/aiven-ga1`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-8131

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Addition of tracking for karapace project link

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/storages/databases/database/general-information/general-information.controller.js
+++ b/packages/manager/modules/pci/src/projects/project/storages/databases/database/general-information/general-information.controller.js
@@ -204,6 +204,15 @@ export default class {
     }
   }
 
+  trackKarapace(event) {
+    if (event.target.tagName?.toLowerCase() === 'a') {
+      this.trackDashboard(
+        'general-information::karapace-documentation-website',
+        'click',
+      );
+    }
+  }
+
   $onDestroy() {
     this.stopPollingDatabaseStatus();
     this.stopPollingNodesStatus();
--- a/packages/manager/modules/pci/src/projects/project/storages/databases/database/general-information/general-information.html
+++ b/packages/manager/modules/pci/src/projects/project/storages/databases/database/general-information/general-information.html
@@ -350,6 +350,7 @@
                 >
                     <oui-tile-description>
                         <p
+                            data-ng-click="$ctrl.trackKarapace($event)"
                             data-translate="pci_databases_general_information_kafka_rest_api_content"
                             data-translate-values="{
                                 projectLink: '<a href=&quot' + $ctrl.KARAPACE_URL + '&quot>Karapace</a>'


packages/manager/modules/pci/src/projects/project/storages/databases/database/general-information/general-information.controller.js
```diff
@@ -204,6 +204,15 @@ export default class {
     }
   }
 
+  trackKarapace(event) {
+    if (event.target === 'a') {
```
radireddy, line 208: event.target refers to html tag abject right?

anooparveti: Yes. Corrected it.

packages/manager/modules/pci/src/projects/project/storages/databases/database/general-information/general-information.html
```diff
@@ -350,6 +350,7 @@
                 >
                     <oui-tile-description>
                         <p
+                            data-ng-click="$ctrl.trackKarapace($event)"
```
JeremyDec, line 353: This will trigger the tracking on click on the whole sentence, which is not what should be done. Also having a click on a non-clickable element, should be avoided.

JeremyDec (changes requested):

packages/manager/modules/pci/src/projects/project/storages/databases/database/general-information/general-information.html
anooparveti: @JeremyDec 

In the function body added an extra check to consider only the clicks on anchor tag.  So tracking would be appropriate.
As per your second point regarding adding a click handler on non-clickable element, I dont see any other better options. The link is part of message which needs translation. Any suggestions on how this can be improved?

JeremyDec: Yes I've updated my comment in fact.
For the on-click on `<p>` element, the fact here is that the sentence contains the link. Maybe the whole phrase should be the anchor content, in order to wrap it all inside a `<a>`? Or at least be splitted to extract the link from the sentence. It requires functional approval I guess

frenautvh: did you try to manually compile your link with an ng-click directive in your controller
```js
$compile('<a href="..." ng-click="$ctrl.trackKarapace">Karapace</a>', $scope);
```
then use the result in translate values ? i'm not sure it'll work as it depends on angular translate sanitization settings but you should give it a try

marie-j: There's also `translate-compile` directive that might help achieve what you want https://angular-translate.github.io/docs/#/api/pascalprecht.translate.directive:translate

anooparveti: @JeremyDec  @marie-j  @frenautvh 
Unfortunately, both the above suggestions did not work.

radireddy (approved):

mohammed-zahaf (approved):

PR #6300, opened 2022-01-12
Feat/manager shell modal

zakariasahmane:
| Question         | Answer
| ---------------- | ---
| Branch?          | feat/tracking-plugin
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no

## Description

Make the container use the cookie policy modal instead of angular module


packages/components/ovh-shell/src/plugin/ux/components/modal.ts
```diff
@@ -0,0 +1,124 @@
...
+ * Vanilla JS Modal compatible with Bootstrap
+ * Inspired by https://github.com/KaneCohen/modal-vanilla
+ */
+const templates = {
+  container: '<div class="modal"></div>',
+  dialog: '<div class="modal-dialog"></div>',
+  content: '<div class="modal-content oui-modal oui-modal_shadow"></div>',
+  header: '<div class="modal-header"></div>',
+  headerClose:
+    '<button type="button" class="close" data-dismiss="modal" aria-label="Close"><span aria-hidden="true">×</span></button>',
```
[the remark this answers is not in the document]
zakariasahmane: Uhm, i think it should, but i don't think it's very relevant here for now, because this component has been created because we do no yet have a library of components that we can use, so it will be replaced eventually but whatever our new design system offers. 

But yes, ideally, for accessibility issues, it should be translated.

packages/components/ng-shell-tracking/src/provider.js
```diff
@@ -4,7 +4,7 @@ export default class NgAtInternet {
   }
 
   $get() {
-    this.trackingPlugin.init();
+    console.log(this.trackingPlugin);
```
frenautvh, line 7: to remove

packages/components/ovh-shell/src/plugin/tracking/index.ts
```diff
@@ -9,7 +10,12 @@ import {
 import ShellClient from '../../client/shell-client';
 
 export interface TrackingAPI {
+  init(): PromiseLike<void>;
   isTagAvailable(): PromiseLike<boolean>;
+  clearTrackQueue(): PromiseLike<void>;
+  processTrackQueue(): PromiseLike<void>;
+  initTag(): PromiseLike<void>;
```
frenautvh, line 17: i'm not sure the tracking api should know the concept of "tags" which are atinternet related, the tracking api should be as abstract as possible with only methods for tracking clicks, actions etc

packages/components/ng-at-internet/src/provider.js
```diff
@@ -18,8 +18,6 @@ import OvhAtInternet from '@ovh-ux/ovh-at-internet';
  */
 export default class NgAtInternet extends OvhAtInternet {
   $get() {
-    this.init();
```
frenautvh, line 21: are we sure this doesn't break the component ?

packages/components/ng-shell-tracking/src/provider.js
zakariasahmane: Yes, it's here just because i needed to debug on PF's

packages/components/ovh-shell/src/plugin/tracking/index.ts
zakariasahmane: I agree with u, and at first, i only included track methods in the API. But unfortunately, by replacing the atInternet module with the tracking plugin, it turns out some of these methods were already used in other modules, which means we need to expose them so we don't break legacy code

packages/components/ng-at-internet/src/provider.js
zakariasahmane: Actually, it does, and it's removed, it should never have been here. I did it in a previous PR and it was a mistake, so i'm removing it back to normal.

PR #6324, opened 2022-01-18
Feat/manager 8199: add order netapp cta

mohammed-zahaf:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-8199

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
This PR include two things:
1. Add support of simple and actions menu CTA into `manager-ng-layout-helpers`
2. Add NetApp order CTA

### DIFF
--- a/packages/manager/modules/netapp/src/routing.js
+++ b/packages/manager/modules/netapp/src/routing.js
@@ -50,6 +50,29 @@ export default /* @ngInject */ ($stateProvider) => {
         }),
       schema: /* @ngInject */ ($http) =>
         $http.get('/storage.json').then(({ data }) => data),
+
+      topbarOptions: /* @ngInject */ ($translate, goToOrder) => ({
+        cta: {
+          type: 'button',
+          displayed: true,
+          disabled: false,
+          label: $translate.instant('netapp_order_cta_label'),
+          value: $translate.instant('netapp_order_cta_value'),
+          onClick: () => {
+            goToOrder();
+          },
+        },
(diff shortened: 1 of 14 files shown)


packages/manager/modules/netapp/src/routing.js
```diff
@@ -50,6 +50,29 @@ export default /* @ngInject */ ($stateProvider) => {
...
+          type: 'button',
+          displayed: true,
+          disabled: false,
+          label: $translate.instant('netapp_order_cta_label'),
+          value: $translate.instant('netapp_order_cta_value'),
+          onClick: () => {
+            goToOrder();
+          },
+        },
+      }),
```
JeremyDec, line 65: **question**:

Maybe this logic could be set directly in `ListLayoutHelper.stateResolves` no? It looks like it can be common to other product? And by this let the user define only `goToOrder` in the resolve

mohammed-zahaf: Yes sure it could be great.

If i understand well the ```ListLayoutHelper.stateResolves``` is here to define some common configuration.
So, if i define the topbarOptions into `stateResolves` i will have to redefine it in product routing because i have to define the label and value content and the onClick behavior which could be different for each product. I think we have to let implementation into product.

Let me know if i can improve it.

radireddy (approved):

PR #6340, opened 2022-01-21
feat(module.netapp): update list page name

mohammed-zahaf:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-8588

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
Update "NetApp" title by "Enterprise File Storage"

### DIFF
--- a/packages/manager/modules/netapp/src/routing.js
+++ b/packages/manager/modules/netapp/src/routing.js
@@ -42,7 +42,7 @@ export default /* @ngInject */ ($stateProvider) => {
       apiPath: () => '/storage/netapp',
       dataModel: () => 'storage.NetAppService',
       defaultFilterColumn: () => 'id',
-      header: () => 'NetApp',
+      header: () => 'Enterprise File Storage',
       customizableColumns: () => true,
       getServiceNameLink: /* @ngInject */ ($state) => ({ id }) =>
         $state.href('netapp.dashboard', {


packages/manager/modules/netapp/src/routing.js
```diff
@@ -42,7 +42,7 @@ export default /* @ngInject */ ($stateProvider) => {
       apiPath: () => '/storage/netapp',
       dataModel: () => 'storage.NetAppService',
       defaultFilterColumn: () => 'id',
-      header: () => 'NetApp',
+      header: () => 'Enterprise File Storage',
```
zakariasahmane, line 45: The name sounds English compared to "NetApp" which is just a product name. Are you sure this is not a key that must be translated ?

mohammed-zahaf: No, it's already translated like this in others files.

zakariasahmane (approved):

packages/manager/modules/netapp/src/routing.js
frenautvh: you should use angular translate here, see : https://github.com/ovh/manager/blob/master/packages/manager/modules/emailpro/src/email-pro.routing.js#L13

antleblanc: IMHO we shouldn't use translations here because the Product name is not meant to be translated into other languages.

antleblanc (approved):

JeremyDec (approved):

PR #6398, opened 2022-02-01
feat(overthebox.autoconfig): add multipath configuration

Steffy29:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #UXCT-436

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Into OverTheBox autoconfigure, add the Network multipath configuration


packages/manager/modules/overthebox/src/overthebox/autoconfigure/multipath/autoconfigure-multipath.controller.js
```diff
@@ -0,0 +1,159 @@
...
+    }
+    if (this.interface.multipath) {
+      params.configuration.multipath = this.interface.multipath;
+    }
+    if (this.interface.routingTable) {
+      params.configuration.routingTable = this.interface.routingTable;
+    }
+    if (this.interface.type) {
+      params.configuration.type = this.interface.type;
+    }
```
frenautvh, line 103: those tests doesn't seem to be usefull since if the value is not defined then you'll set an undefined value anyway

Steffy29: By default, the values are set to blank. Those values are set from fields into a form by user if needed.

packages/manager/modules/overthebox/src/overthebox/autoconfigure/multipath/autoconfigure-multipath.controller.js
```diff
@@ -0,0 +1,159 @@
...
+
+  createInterface() {
+    this.errorMessage = '';
+    this.successMessage = '';
+
+    const params = {
+      configuration: {
+        ifname: this.interface.ifname,
+        interfaceName: this.interface.interfaceName,
+      },
```
marie-j, line 73:
```suggestion
    const params = {
      configuration: {
        ifname: this.interface.ifname,
        interfaceName: this.interface.interfaceName,
        gateway: this.interface.gateway
      },
```
to avoid that long list of `if` , it will result to the same

Steffy29: I'm not sure that I have the same result with your suggestion, some are mandatory and others not, I don't want to send a blank value to the API if the value is not set.

packages/manager/modules/overthebox/src/overthebox/autoconfigure/multipath/autoconfigure-multipath.controller.js
JeremyDec: @Steffy29 why not just declaring `this.interface = {}` on `addInterface` method call? This way you could remove those tests as all tested properties would be undefined at initialization?

packages/manager/modules/overthebox/src/overthebox/autoconfigure/translations/Messages_fr_FR.json
```diff
@@ -30,5 +30,32 @@
   "overTheBox_autoconfigure_create_failed": "Une erreur s'est produite lors de la création : {{errorMessage}}.",
   "overTheBox_autoconfigure_remove_failed": "Une erreur s'est produite lors de la suppression : {{errorMessage}}.",
   "overTheBox_autoconfigure_mandatory": "Ce champ est obligatoire",
-  "overTheBox_autoconfigure_actions": "Actions"
+  "overTheBox_autoconfigure_actions": "Actions",
+  "overTheBox_autoconfigure_interface_title": "Multipath",
+  "overTheBox_autoconfigure_mtu": "MTU",
+  "overTheBox_autoconfigure_mtu_optional": "MTU (optionel)",
```
marie-j, line 36:
```suggestion
  "overTheBox_autoconfigure_mtu_optional": "MTU (optionnel)",
```

marie-j (changes requested):

marie-j (dismissed):

JeremyDec (dismissed):

marie-j (approved):

antleblanc (approved):

PR #6607, opened 2022-03-01
Fix/manager 8716: fix csv issue

mohammed-zahaf:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-8716

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
1. fix CSV issue

### DIFF
--- a/packages/manager/modules/exchange/src/account/add/account-add.controller.js
+++ b/packages/manager/modules/exchange/src/account/add/account-add.controller.js
@@ -185,7 +185,7 @@ export default class ExchangeAccountAddController {
       lastName: this.newAccount.lastName,
       displayName: this.newAccount.displayName,
       initials: this.newAccount.initials,
-      description: this.newAccount.description,
+      description: `"${this.newAccount.description}"`,
       forwardingEmail: this.newAccount.forwardingEmail,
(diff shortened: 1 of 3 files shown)


ghyenne (approved):

packages/manager/modules/exchange/src/account/add/account-add.controller.js
```diff
@@ -185,7 +185,7 @@ export default class ExchangeAccountAddController {
       lastName: this.newAccount.lastName,
       displayName: this.newAccount.displayName,
       initials: this.newAccount.initials,
-      description: this.newAccount.description,
+      description: `"${this.newAccount.description}"`,
```
radireddy, line 188: `${this.newAccount.description}` returns new string. Why do you want to add ""?

mohammed-zahaf: Because there is an option which allow to generate a CSV file.
And for decription we can have those type of separtors (`, ; :`) in description.
Those chars had to be ignored to avoid generating CSV file.

anooparveti (approved):

PR #6668, opened 2022-03-10
fix(server.display-name): display loader during state navigation

anooparveti:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-8749

- [x] Commits are signed-off
- [ ] ~~Only FR translations have been updated~~ [n/a]
- [x] Lint has passed locally
- [ ] ~~Breaking change is mentioned in relevant commits~~ [n/a]

## Description

 Display loader during state navigation after server name update

### DIFF
--- a/packages/manager/apps/dedicated/client/app/dedicated/server/display-name/display-name.controller.js
+++ b/packages/manager/apps/dedicated/client/app/dedicated/server/display-name/display-name.controller.js
@@ -27,6 +27,7 @@ angular.module('App').controller(
       }
 
       this.loading.update = true;
+      let reload = false;
 
       return this.Server.updateDisplayName({
         serviceId: this.dedicatedServer.serviceId,
@@ -40,13 +41,7 @@ angular.module('App').controller(
             ),
             'server_dashboard_alert',
           );
-          this.$state.go(
-            '^',
-            {},
-            {
-              reload: true,
-            },
-          );
+          reload = true;
         })
         .catch(({ data }) => {
           this.Alerter.error(
@@ -61,8 +56,16 @@ angular.module('App').controller(
             ].join('. '),
             'server_dashboard_alert',
           );
-          this.$state.go('^');
         })
+        .finally(() =>
+          this.$state.go(
+            '^',
+            {},
+            {
+              reload,
+            },
+          ),
+        )
         .finally(() => {
           this.loading.update = false;
         });


packages/manager/apps/dedicated/client/app/dedicated/server/display-name/display-name.controller.js
```diff
@@ -61,8 +56,16 @@ angular.module('App').controller(
...
+            '^',
+            {},
+            {
+              reload,
+            },
+          ),
+        )
         .finally(() => {
           this.loading.update = false;
         });
```
ghyenne, line 71: Maybe we can merge the 2 .finally()?

ghyenne (approved):

packages/manager/apps/dedicated/client/app/dedicated/server/display-name/display-name.controller.js
```diff
@@ -61,8 +56,16 @@ angular.module('App').controller(
...
         })
+        .finally(() =>
+          this.$state.go(
+            '^',
+            {},
+            {
+              reload,
+            },
+          ),
+        )
```
mohammed-zahaf, line 68: @anooparveti, for migration raisons, we have to not implement $state in controllers.
If duable you have to write it into a routing file.

Also the reload have to be implement into then promise.
```suggestion
        .then(() =>
          this.refreshPage(),
        ).catch(() => {  })
```

packages/manager/apps/dedicated/client/app/dedicated/server/display-name/display-name.controller.js
anooparveti: The second `finally` depends on the promise returned by `$state.go`. So they both can not be combined.

packages/manager/apps/dedicated/client/app/dedicated/server/display-name/display-name.controller.js
anooparveti: Hello @mohammed-zahaf, I agree with your suggestion of moving `$state` to routing file but unfortunately this particular state is legacy and is implemented without `component` architecture. So it cannot be implemented as of now.

radireddy (approved):

mohammed-zahaf (approved):

JeremyDec (approved):

antleblanc (approved):

PR #6712, opened 2022-03-18
fix(sva): add fr content for companykind

ghyenne:
| Question         | Answer
| ---------------- | ---
| Branch?          | `fix/MANAGER-8521` 
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | DTRSD-75506 

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
I added fr content for companykind.


radireddy (approved):

cbourgois (approved):
It works, but I'm a bit confused because we have same translations with same keys in 2 differents files. So in this case we should load the same translations file on both side where it's used.

Not part of this PR, so I approve.

PR #6733, opened 2022-03-22
fix(exchange.account): uncheck email copy checkbox

y4nnL:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-8935

- [X] Commits are signed-off
- [X] Lint has passed locally

## Description

Uncheck the email copy checkbox if the forwarding email is empty

### DIFF
--- a/packages/manager/modules/exchange/src/account/add/account-add.controller.js
+++ b/packages/manager/modules/exchange/src/account/add/account-add.controller.js
@@ -57,6 +57,16 @@ export default class ExchangeAccountAddController {
     return this.fetchingAccountCreationOptions();
   }
 
+  $doCheck() {
+    if (!this.newAccountForm) return;
+    if (
+      !this.newAccountForm.accountInfoEmail.$invalid &&
+      !this.newAccount.forwardingEmail
+    ) {
+      this.newAccount.storeCopyOfEmail = false;
+    }
+  }
+
   fetchingAccountCreationOptions() {
     function transformAccountTypes(accountTypes) {
       return map(accountTypes, (accountType) => ({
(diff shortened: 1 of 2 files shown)


packages/manager/modules/exchange/src/account/add/account-add.controller.js
```diff
@@ -57,6 +57,16 @@ export default class ExchangeAccountAddController {
     return this.fetchingAccountCreationOptions();
   }
 
+  $doCheck() {
```
mohammed-zahaf, line 60: i think we have to avoid to use the `$` char to distinguish Angular snippet code from own.
may be it's an angular function ?

packages/manager/modules/exchange/src/account/update/account-update.controller.js
```diff
@@ -74,6 +74,16 @@ export default class ExchangeUpdateAccountCtrl {
       });
   }
 
+  $doCheck() {
```
mohammed-zahaf, line 77: Same as here :
https://github.com/ovh/manager/pull/6733/files#r832037281

packages/manager/modules/exchange/src/account/add/account-add.controller.js
mohammed-zahaf: Ah i just googled bit and yep it's an angular function.

y4nnL: Yep this function is part of the component life cycle hooks

antleblanc (approved):

JeremyDec (approved):

PR #6754, opened 2022-03-24
fix(exchange.account): broken CSV file

y4nnL:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-8716

- [X] Commits are signed-off
- [X] Lint has passed locally

## Description

When exporting exchange accounts to CSV, the file is completely broken due to malformed fields

### DIFF
--- a/packages/manager/modules/exchange/src/account/account.constants.js
+++ b/packages/manager/modules/exchange/src/account/account.constants.js
@@ -6,19 +6,51 @@ export const ACCOUNT_COUNTRY_RULE_NAME = 'country';
 
 export const ACCOUNT_WORLD_PHONE_REGEX = /^00\d{2,3}[\s\d]+$/;
 
-export const escapeDescription = (description) =>
-  typeof description === 'string' && description.length > 0
-    ? `"${description}"`
-    : null;
-
+/**
+ * Legacy
+ * @see MANAGER-8716
+ */
 export const unescapeDescription = (description) =>
   typeof description === 'string' ? description.replace(/^"+|"+$/g, '') : null;
+
+const accountPropertiesWithUnit = ['totalQuota', 'usedQuota'];
+const accountPropertiesWithStatus = ['mfa'];
+
+const accountValueToCSVString = (value) => {
+  const toStringMap = {
+    String: (v) => v,
+    Number: (v) => `${v}`,
+    Boolean: (v) => `${v}`,
+    Undefined: () => '',
+    Null: () => '',
+    Object: (v) => toStringMap.Array(Object.values(v)),
+    Array: (v) => v.map(accountValueToCSVString).join(','),
+    Date: (v) => v.toISOString(),
+  };
+  const type = Object.prototype.toString.call(value).match(/ (.+?)\]$/)[1];
+  const method = toStringMap[type] || toStringMap.String;
+  return method(value);
+};
+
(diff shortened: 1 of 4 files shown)


packages/manager/modules/exchange/src/account/export-as-csv/account-export-as-csv.controller.js
```diff
@@ -90,6 +90,8 @@ export default class ExchangeExportToCsvAccountsCtrl {
         'samaccountName',
         'taskPendingId',
         'id',
+        // https://answers.microsoft.com/fr-fr/msoffice/forum/all/excel-sortie-en-csv-avec-conversion-utf-8/0a05dc63-645d-4303-b0f5-c27513bf776e
```
marie-j, line 93: Link to an issue/solution written in english would be better 
Also why is this added above a property that doesn't seems to be transformed ?

packages/manager/modules/exchange/src/account/account.constants.js
```diff
@@ -6,19 +6,51 @@ export const ACCOUNT_COUNTRY_RULE_NAME = 'country';
...
+    String: (v) => v,
+    Number: (v) => `${v}`,
+    Boolean: (v) => `${v}`,
+    Undefined: () => '',
+    Null: () => '',
+    Object: (v) => toStringMap.Array(Object.values(v)),
+    Array: (v) => v.map(accountValuetoCSVString).join(','),
+    Date: (v) => v.toISOString(),
+  };
+  const type = Object.prototype.toString.call(value).match(/ (.+?)\]$/)[1];
```
marie-j, line 30: `typeof` is not suiting the needs here ?

marie-j (changes requested):

packages/manager/modules/exchange/src/account/account.constants.js
```diff
@@ -6,19 +6,51 @@ export const ACCOUNT_COUNTRY_RULE_NAME = 'country';
...
+ * Legacy
+ * @see MANAGER-8716
+ */
 export const unescapeDescription = (description) =>
   typeof description === 'string' ? description.replace(/^"+|"+$/g, '') : null;
 
+const accountPropertiesWithUnit = ['totalQuota', 'usedQuota'];
+const accountPropertiesWithStatus = ['mfa'];
+
+const accountValuetoCSVString = (value) => {
```
radireddy, line 19:
```suggestion
const accountValueToCSVString = (value) => {
```

packages/manager/modules/exchange/src/account/export-as-csv/account-export-as-csv.controller.js
y4nnL: None of these properties are transformed, there are simply removed from the output for different reasons
The description field breaks the output file that's why it is removed. I'll find an english version of this issue :+1:

packages/manager/modules/exchange/src/account/account.constants.js
y4nnL: Nope, typeof is not precise enough for some of the types (Array, Date and null will ouput "object")
Object.prototype.toString is more reliable as it gives the native object / primitive in the form of [object Xxxx]

radireddy (approved):

marie-j (approved):

PR #6766, opened 2022-03-25
Docs/shell api docs

zakariasahmane:
| Question         | Answer
| ---------------- | ---
| Branch?          | feat/ovh-shell-batch2
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no

## Description

Add docs for shell


packages/components/ovh-shell/README.md
```diff
@@ -1 +1,178 @@
...
+import { shell } from '@ovh-ux/shell';
+
+const shellApi = shell.initShell(yourIframe);
+```
+
+This method inits a shell and returns an API that allows you to communicate with it.
+
+The returned object contains the following methods :
+
+- `emitEvent`(`(eventId: string, data: unknown) => void`) : This method allows to emit a custom event from the Shell.
```
chloren, line 49: It is unclear to me how we can use this.

What kind of event can we pass on, and where are the event listeners defined at? Do we have to create a listener somewhere or are there a number of listener in the shell?

Also, would it be better practice to prefix our eventId with our app name to avoid mix-ups (e.g. `my-app:my-event`)? If so it should be told (or we should have an exhaustive list of events naturally listened to by the Shell).

An example could help.

packages/components/ovh-shell/README.md
```diff
@@ -1 +1,178 @@
...
+
+- `registerPlugin`(`(pluginId: string, pluginApi: Record<string, Callback>) => void`): Register a plugin by giving its id and API.
+  - `pluginId`: Your plugin `name`/`id`.
+  - `pluginAPI`: This is an object that contains your plugin API. An API is basically just multiple callbacks indentified by a string.
+
+*Example of plugin API* :
+
+```ts
+const api = {
+  enablePlugin: () => enable(),
```
chloren, line 61: The example is a bit confusing. Does it mean we necessary have to define the function `enablePlugin` and `rename`? Does it have a link with the `setPluginAvailability` function below?

We could add an example in `setPluginAvailability` to show how this plugin won't have any effect unless we've set his availability as true (e.g. have an example here where we use the plugin and it works, and another below where it doesn't work as long as we haven't set it as available).

packages/components/ovh-shell/README.md
```diff
@@ -1 +1,178 @@
...
+
+The returned object contains the following methods :
+
+- `emitEvent`(`(eventId: string, data: unknown) => void`) : This method allows to emit a custom event from the Shell.
+  - `eventId`: The `name`/`id` of your event.
+  - `data`: Contains the data you want to send with your event. It can be everything that can be parsed by the navigator, objects such as callbacks or Promises won't work.
+
+- `registerPlugin`(`(pluginId: string, pluginApi: Record<string, Callback>) => void`): Register a plugin by giving its id and API.
+  - `pluginId`: Your plugin `name`/`id`.
+  - `pluginAPI`: This is an object that contains your plugin API. An API is basically just multiple callbacks indentified by a string.
```
chloren, line 55: What do you mean? How do we use it and what does the string have to do with it?

packages/components/ovh-shell/README.md
```diff
@@ -1 +1,178 @@
...
+This method inits a shell and returns an API that allows you to communicate with it.
+
+The returned object contains the following methods :
+
+- `emitEvent`(`(eventId: string, data: unknown) => void`) : This method allows to emit a custom event from the Shell.
+  - `eventId`: The `name`/`id` of your event.
+  - `data`: Contains the data you want to send with your event. It can be everything that can be parsed by the navigator, objects such as callbacks or Promises won't work.
+
+- `registerPlugin`(`(pluginId: string, pluginApi: Record<string, Callback>) => void`): Register a plugin by giving its id and API.
+  - `pluginId`: Your plugin `name`/`id`.
```
chloren, line 54: Should all plugin name/id be unique? We should have a list of plugins reserved by the shell.

Maybe it would be good practice to register plugin id with the app as prefix (e.g. `my-app:my-plugin`) to avoid errors (both for projects using the Shell that could try to redefine an existing plugin and for the shell that could have new plugins and redefine a plugin with a name that another project is already using).

packages/components/ovh-shell/README.md
```diff
@@ -1 +1,178 @@
...
+
+const { ux } = plugin;
+
+ux.showAccountSidebar();
+```
+
+#### `tracking`
+
+The tracking API exposes everything that the library `ovh-at-internet` exposes with the same signature.
+So basically, you can use all the functions exposed by `ovh-at-internet` the same way via this plugin.
```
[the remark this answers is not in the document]
zakariasahmane: That's a good idea, but we don't have the documentation for that for now... Maybe it could be the time to write it too, but the lib is just a wrapper of an external lib that we use in private, so i don't know if we can expose their docs ?

packages/components/ovh-shell/README.md
zakariasahmane: Example : 

```
const api = {
myApiIdString: myCallback,
myApiIdString2: myOtherCallback
}
```

Just a Record<string, Callback>, i'll add the example in the docs so it's more clear.

antleblanc (changes requested):
@zakariasahmane Could you please resolve existing conflicts? Thank you!

PR #6781, opened 2022-03-29
fix MLS-2743 (pci.notebooks): filter notebook volumes before fetching command

jperchoc:
MLS-2743

| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MLS-2743

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

The command showed on a notebook's dashboard does not work anymore, due to an incorrect parameter send to the api.
Before POSTing the spec object on the command route:
- Drop volumes inside spec whose `internal` parameter is `true`
- Don't POST that `internal` parameter inside the spec payload

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/notebooks/components/configuration-command/configuration-command.controller.js
+++ b/packages/manager/modules/pci/src/projects/project/notebooks/components/configuration-command/configuration-command.controller.js
@@ -1,3 +1,4 @@
+import omit from 'lodash/omit';
 import { DOCUMENTATION_LINK } from './configuration-command.constants';
 
 export default class {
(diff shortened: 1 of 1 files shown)


frenautvh (approved):

packages/manager/modules/pci/src/projects/project/notebooks/components/configuration-command/configuration-command.controller.js
```diff
@@ -17,8 +18,32 @@ export default class {
     this.getConfigurationCommand();
   }
 
+  getFilteredVolumes() {
+    return this.notebookSpecs.volumes.flatMap((volume) => {
+      let toRemove = false;
+      let internalKey;
+      Object.keys(volume).forEach((key) => {
```
marie-j, line 25: I wonder if `_.filter`. or `_.find` can help simplify logic here

jperchoc: Hi @marie-j , as seen together, using `_.filter` or `_.find` does not seem to simplify, as we'd need to map the result of filter and loop again through all properties to achieve the same result.

antleblanc (approved):

PR #6804, opened 2022-04-01
fix(exchange.account): add exchange guid in update form

ghyenne:
| Question         | Answer
| ---------------- | ---
| Branch?          | `MANAGER-8940` 
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-8940 

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
I added exchange guid in update form.

### DIFF
--- a/packages/manager/modules/exchange/src/account/update/account-update.html
+++ b/packages/manager/modules/exchange/src/account/update/account-update.html
@@ -143,12 +143,22 @@
                             ></span>
                         </div>
                         <div class="oui-field__helper">
-                            <span data-translate="exchange_account_update_emailAddress_hint"></span>
+                            <span
+                                data-translate="exchange_account_update_emailAddress_hint"
+                            ></span>
                             <ul>
-                                <li data-translate="exchange_account_update_emailAddress_hint_part_1"></li>
-                                <li data-translate="exchange_account_update_emailAddress_hint_part_2"></li>
-                                <li data-translate="exchange_account_update_emailAddress_hint_part_3"></li>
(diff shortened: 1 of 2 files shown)


packages/manager/modules/exchange/src/account/update/account-update.html
```diff
@@ -487,6 +497,24 @@
                             </div>
                         </div>
                     </div>
+
+                    <!--ExchangeGuid-->
+                    <div class="row">
+                        <div class="col-md-6 oui-field mb-4">
+                            <label
+                                class="oui-field__label oui-label"
+                                for="accountExchangeGuid"
```
radireddy, line 506: maybe not required? is there a field with this ID?

radireddy (approved):

packages/manager/modules/exchange/src/account/update/account-update.html
anooparveti, line 506: **question**: Is `for` attribute required here? because I dont see any for element with `id = accountExchangeGuid`

ghyenne: No it is not necessary because it is not a field but a data return.
I removed it.

ghyenne: No it is not necessary because it is not a field but a data return.
I removed it.

anooparveti (approved):

antleblanc (approved):

PR #6806, opened 2022-04-01
fix(exchange.account): add required field in the label for displayname

ghyenne:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/MANAGER-8230` 
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-8910 

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
I added required field in the label for displayname

### DIFF
--- a/packages/manager/modules/exchange/src/account/update/account-update.html
+++ b/packages/manager/modules/exchange/src/account/update/account-update.html
@@ -34,9 +34,9 @@
                         data-ng-if="$ctrl.updateAccountOptions.availableTypes.length > 1"
                     >
                         <label
-                            class="oui-field__label oui-label required"
+                            class="oui-field__label oui-label"
                             for="selectedAccountType"
-                            data-translate="exchange_ACTION_update_account_step1_type_label"
+                            data-translate="exchange_account_update_accountType_label"
                         ></label>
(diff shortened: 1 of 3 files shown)


packages/manager/modules/exchange/src/account/update/account-update.html
```diff
@@ -220,7 +230,7 @@
                         <!--Display name-->
                         <oui-field
                             class="col-md-6"
-                            label="{{:: 'exchange_account_update_displayName_label' | translate }}"
+                            label="{{:: 'exchange_account_update_displayName_label' | translate }} *"
```
marie-j, line 233: Even though this is a usual hint, I'm not sure it is sufficient to indicate field is mandatory. Something to mention that the `*` means mandatory fields could be great to add

radireddy (approved):

packages/manager/modules/exchange/src/account/update/account-update.html
ghyenne: I have modified. I have standardized with the creation form which indeed puts "(obligatoire)" and not * as the modification form used to.

mohammed-zahaf (approved):

PR #6918, opened 2022-04-19
fix(telecom): load targets for transfer of security deposit

anooparveti:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-80172

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Existing behaviour: While transferring Security deposit, when user selects the source, API calls are made on the list of targets to determine if Security Deposit can be transferred to the specified target. If the targets list is huge, the API calls are rejected by the server with HTTP status code 429.
To avoid making more calls at the same time, the implementation has been modified to fetch data in chunks of 10.

### DIFF
--- a/packages/manager/apps/telecom/src/app/telecom/telephony/billingAccount/billing/depositMovement/deposit-movement.controller.js
+++ b/packages/manager/apps/telecom/src/app/telecom/telephony/billingAccount/billing/depositMovement/deposit-movement.controller.js
@@ -5,6 +5,7 @@ import map from 'lodash/map';
 import reduce from 'lodash/reduce';
 import set from 'lodash/set';
 import sortBy from 'lodash/sortBy';
+import cloneDeep from 'lodash/cloneDeep';
 
 export default /* @ngInject */ function TelecomTelephonyBillingAccountBillingDepositMovementCtrl(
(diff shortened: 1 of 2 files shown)


packages/manager/apps/telecom/src/app/telecom/telephony/billingAccount/billing/depositMovement/deposit-movement.controller.js
```diff
@@ -74,6 +75,50 @@ export default /* @ngInject */ function TelecomTelephonyBillingAccountBillingDep
...
+    let result = {};
+    while (targetsArrCopy.length) {
+      // eslint-disable-next-line no-await-in-loop
+      const resultOfChunk = await getCanTransferSecurityDeposit(
+        targetsArrCopy.splice(0, 10),
+      );
+      result = Object.assign(result, resultOfChunk);
+    }
+    return result;
+  }
```
mohammed-zahaf, line 120: If duable, it's better to complete this task in routing file.
It will avoid us the `async await` by using resolve behavior.

PR #7016, opened 2022-04-29
fix(container): add feature flipping for chatbot

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | feat/product-nav-reshuffle
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | 

## Description

add feature flipping for chatbot in container

### DIFF
--- a/packages/manager/apps/container/src/core/container/ContainerProvider.jsx
+++ b/packages/manager/apps/container/src/core/container/ContainerProvider.jsx
@@ -1,5 +1,4 @@
 import React, { useContext, useEffect, useState } from 'react';
-import PropTypes from 'prop-types';
 import { useReket } from '@ovh-ux/ovh-reket';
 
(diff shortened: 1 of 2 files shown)


packages/manager/apps/container/src/core/container/ContainerProvider.jsx
```diff
@@ -21,27 +20,33 @@ export const ContainerProvider = ({ children }) => {
   // user choice about using or not the beta
   const [useBeta, setUseBeta] = useState(false);
 
+  // is the chatbot enabled ?
+  const [hasChatbot, setHasChatbot] = useState(false);
```
chloren, line 24: Maybe we should rename this value instead of having a comment explaining what this is

```
const [isChatbotEnabled, setIsChatbotEnabled] = useState(false);
```

PR #7079, opened 2022-05-09
fix(container): pnr mobile header

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | feat/product-nav-reshuffle
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-9247

## Description

make product nav reshuffle header responsive

### DIFF
--- a/packages/manager/apps/container/src/container/common/language/index.tsx
+++ b/packages/manager/apps/container/src/container/common/language/index.tsx
@@ -1,4 +1,5 @@
 import React, { useEffect, useRef, useState } from 'react';
+import { useMediaQuery } from 'react-responsive';
 
 import useClickAway from 'react-use/lib/useClickAway';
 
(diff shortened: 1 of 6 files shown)


packages/manager/apps/container/src/container/common/language/index.tsx
```diff
@@ -22,6 +24,9 @@ function LanguageMenu({
   const shell = useShell();
   const [show, setShow] = useState(false);
   const handleRootClose = () => setShow(false);
+  const isSmallDevice = useMediaQuery({
```
cbourgois, line 27: Maybe we can also done the same things using only CSS and 2 span for language label.

cbourgois, line 27: Maybe we can also done the same things using only CSS and 2 span for language label.

cbourgois (approved):

PR #6385, opened 2022-01-31
feat(hosting.offer): implement detach start10M

lucas-chaigne:
ref: WEB-11147

Signed-off-by: Lucas Chaigne <lucas.chaigne@ovhcloud.com>

| Question         | Answer
| ---------------- | ---
| Branch?          | `develop`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Feat #WEB-11147
| License          | BSD 3-Clause

- [x] Try to keep pull requests small so they can be easily reviewed.
- [x] Commits are signed-off
- [x] Only FR translations have been updated
- [x] Branch is up-to-date with target branch
- [x] Lint has passed locally
- [x] Standalone app was ran and tested locally
- [x] Ticket reference is mentioned in linked commits (internal only)
- [ ] Breaking change is mentioned in relevant commits

## Description

With the purchase of a domain name, customer can activate a free hosting offer, called start10m. 
As of now, customer cannot upgrade this offer to a standard one (ex: perso2014). This PR aim is to give the possibility to perform such action.
Technically, it is not the same process as a classical upgrade (is is a detach operation, using different api calls) but it must appear similar to the customer.
After testing, we encountered an issue: there is a distinction between legacy and "new" start10m offers : for now, legacy cannot be detached, so we limit the detach of start10m to "new" offers (start10m-addon-v1).

PR #7165, opened 2022-05-19
fix(hosting.offer): implement detach addon of start10M

lucas-chaigne:
Signed-off-by: Lucas Chaigne <lucas.chaigne@ovhcloud.com>

| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #WEB-12035

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Start10m is a free hosting offered to customer when ordering a domain. We want this customer to be able to upgrade it to a paid offer (perso2014). It was done through https://github.com/ovh/manager/pull/6385
However we add an issue: if the start10m has an addon (free email), we couldn't detach it. This PR aim is to solve this issue.

### DIFF
--- a/packages/manager/apps/web/client/app/hosting/offer/upgrade/hosting-offer-upgrade.controller.js
+++ b/packages/manager/apps/web/client/app/hosting/offer/upgrade/hosting-offer-upgrade.controller.js
@@ -110,40 +110,50 @@ angular.module('App').controller(
       );
     }
 
-    getDetachActionsOptions() {
-      return {
-        addons: [
-          {
+    getDetachRequestOptions(serviceId, planCode) {
+      return this.ovhManagerProductOffersActionService
+        .getDetachPlancodeInformationOptions(serviceId, planCode)
+        .then((data) => {
+          const addons = data.map((e) => ({
             duration: DETACH_DEFAULT_OPTIONS.durationCode,
-            planCode: this.model.offer.value,
+            planCode: e.plans[0].planCode,
             pricingMode: DETACH_DEFAULT_OPTIONS.pricingMode,
             quantity: DETACH_DEFAULT_OPTIONS.quantity,
-            serviceId: this.serviceId,
-          },
-        ],
-        duration: DETACH_DEFAULT_OPTIONS.durationCode,
-        pricingMode: DETACH_DEFAULT_OPTIONS.pricingMode,
-        quantity: DETACH_DEFAULT_OPTIONS.quantity,
-      };
+            serviceId: e.serviceId,
+          }));
+
+          return {
+            addons,
+            duration: DETACH_DEFAULT_OPTIONS.durationCode,
+            pricingMode: DETACH_DEFAULT_OPTIONS.pricingMode,
+            quantity: DETACH_DEFAULT_OPTIONS.quantity,
+          };
+        });
+    }
+
+    getDetachPrices(serviceId, planCode) {
+      return this.getDetachRequestOptions(serviceId, planCode)
+        .then((detachOptions) => {
+          return this.ovhManagerProductOffersActionService.simulate(
+            planCode,
+            serviceId,
+            DETACH_DEFAULT_OPTIONS.type,
+            detachOptions,
+          );
+        })
+        .then((data) => {
+          return [
+            {
+              ...data.order,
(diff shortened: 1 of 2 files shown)


packages/manager/apps/web/client/app/hosting/offer/upgrade/hosting-offer-upgrade.controller.js
```diff
@@ -110,40 +110,55 @@ angular.module('App').controller(
...
-    getDetachActionsOptions() {
-      return {
-        addons: [
-          {
+    getDetachRequestOptions(serviceId, planCode) {
+      return this.ovhManagerProductOffersActionService
+        .getDetachPlancodeInformationOptions(serviceId, planCode)
+        .then((data) => {
+          const addons = [];
+          if (data.length === 1) {
```
marie-j, line 118: addons is plural but we can't have more than 2 or at least we don't handle the case ? Maybe we should `map` data no ?

packages/manager/apps/web/client/app/hosting/offer/upgrade/hosting-offer-upgrade.controller.js
```diff
@@ -110,40 +110,55 @@ angular.module('App').controller(
...
-            details.duration = DETACH_DEFAULT_OPTIONS.durationText;
-            durationsTab.push(details);
+            detachOptions,
+          );
+        })
+        .then((data) => {
+          const detachPrices = [];
+          const details = angular.copy(data.order);
+          details.duration = DETACH_DEFAULT_OPTIONS.durationText;
+          detachPrices.push(details);
```
marie-j, line 153: I'm having trouble understanding the use of an array for one element

packages/manager/apps/web/client/app/hosting/offer/upgrade/hosting-offer-upgrade.controller.js
```diff
@@ -110,40 +110,55 @@ angular.module('App').controller(
...
-          .then((data) => {
-            const durationsTab = [];
-            const details = angular.copy(data.order);
-            details.duration = DETACH_DEFAULT_OPTIONS.durationText;
-            durationsTab.push(details);
+            detachOptions,
+          );
+        })
+        .then((data) => {
+          const detachPrices = [];
```
marie-j, line 150:
```suggestion
        return [{
        ...data.order,
        duration: DETACH_DEFAULT_OPTIONS.durationText
        }]
```
no need to create an array variable then push, directly return the value

packages/manager/apps/web/client/app/hosting/offer/upgrade/hosting-offer-upgrade.controller.js
lucas-chaigne: I agree this is weird, but it comes from the api design. 
We can only have 0 or one addon to be handled in detach.
But the detach apis do not reflect this design (paramater "addons" is an array, not an object)

packages/manager/apps/web/client/app/hosting/offer/upgrade/hosting-offer-upgrade.controller.js
lucas-chaigne: the controller in general and the function getPrices is used for both classical upgrade and start10m detach.
the function getPrices returns an array (there can be several options for a classical upgrade). getDetachPrices is used to return a value in getPrices in case of detach, that's why it returns an array of 1 element.

packages/manager/apps/web/client/app/hosting/offer/upgrade/hosting-offer-upgrade.controller.js
lucas-chaigne: thanks for the review, it is corrected

packages/manager/apps/web/client/app/hosting/offer/upgrade/hosting-offer-upgrade.controller.js
lucas-chaigne: yes, you are right. even if for the moment we only either 0 or one addon, we can easily be more flexible

marie-j (approved):

antleblanc (approved):

PR #7179, opened 2022-05-19
fix(sidebar): hide services when no count

zakariasahmane:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/product-nav-reshuffle`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | MANAGER-9399

## Description

hide services when none

### DIFF
--- a/packages/manager/apps/container/src/container/nav-reshuffle/sidebar/index.tsx
+++ b/packages/manager/apps/container/src/container/nav-reshuffle/sidebar/index.tsx
@@ -44,6 +44,23 @@ function Sidebar(): JSX.Element {
 
   const logoLink = navigationPlugin.getURL('hub', '#/');
 
+  const shouldHideElement = (node: Node) => {
+    if (betaVersion === 2) {
+      if (node.id === 'services') return false;
+      if (node.count === false) return false;
+      if (node.hideIfEmpty === false) return false;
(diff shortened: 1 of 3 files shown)


antleblanc (approved):

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/index.tsx
```diff
@@ -44,6 +44,23 @@ function Sidebar(): JSX.Element {
 
   const logoLink = navigationPlugin.getURL('hub', '#/');
 
+  const shouldHideElement = (node: Node) => {
+    if (betaVersion === 2) {
+      if (node.id === 'services') return false;
+      if (node.count === false) return false;
```
marie-j, line 50: count is a boolean not a number ? That's confusing

zakariasahmane: Hey @marie-j 

It's actually both. If you take a look at the `node.ts` file, you will see that it's an optionnal attribute, that can either be a boolean or a number.
 
I don't recall why it was both, but i think that it's related to the json config of services we have, we write "false" to avoid showing the element before starting any count or something like that.

PR #7261, opened 2022-06-02
feat(billing.history): hide HTML invoice for India customers

anooparveti:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-8861

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Hide 'HTML Version' invoice from Indian customers


ghyenne (approved):

packages/manager/modules/billing/src/main/history/billing-main-history.controller.js
```diff
@@ -10,13 +11,18 @@ export default class BillingMainHistoryCtrl extends ListLayoutHelper.ListLayoutC
...
     ouiDatagridService,
   ) {
     super($q, ouiDatagridService);
     this.$translate = $translate;
     this.$uibModal = $uibModal;
     this.Alerter = Alerter;
     this.atInternet = atInternet;
+
+    this.displayHtmlInvoice = !HIDE_HTML_INVOICE_FOR_COUNTRIES.includes(
+      coreConfig.getUser()?.country,
```
mohammed-zahaf, line 24: I think we should use `ovhSubsidiary` instead of `country`.
Are you sure we must display invoices accourding `country` ?

Because we can edit `country` field into account edit page.

anooparveti: Yes, we need to display based on the `country`. Because for Indian customers, we cannot rely on `ovhSubsidiary`.

mohammed-zahaf: Ok, thanks

mohammed-zahaf (approved):

packages/manager/modules/billing/src/main/history/billing-main-history.constants.js
```diff
@@ -0,0 +1,5 @@
+export const HIDE_HTML_INVOICE_FOR_COUNTRIES = ['IN'];
```
marie-j, line 1: Please use feature flipping

marie-j, line 1: Please use feature flipping instead

marie-j (changes requested):

packages/manager/modules/billing/src/main/history/billing-main-history.constants.js
anooparveti: @marie-j, the logic should be implemented based on the user country and in this scenario, we cannot depend on subsidiary. Since feature-flipping does not support checking for the country, the implementation has been done in this manner.

ghyenne (dismissed):

sachinrameshn (approved):

radireddy (dismissed):

antleblanc:
@anooparveti Could you please take into account the request changes asked by @marie-j?

anooparveti:
> @anooparveti Could you please take into account the request changes asked by @marie-j?

Hello @antleblanc, 

I have given the reply https://github.com/ovh/manager/pull/7261#discussion_r887807157
However, let me know if we can extend the functionality of feature-flipping to support checking the feature availability based on the country, I would be more than happy to do that.

Let me know your opinion.

cc- @marie-j

antleblanc:
@anooparveti Seen together. We should be able to do it via feature flipping mechanism by adding a new `country` rule.
This will allow prevent us to manager access by modifying the front and redeploy it.

ghyenne (approved):

lizardK (approved):

antleblanc (approved):

antleblanc (dismissed):

antleblanc (approved):

PR #7423, opened 2022-06-23
fix (pci.ai.training): filter internal key on jobs volumes when resubmitting

jperchoc:
DATATR-56

Signed-off-by: Jonathan Perchoc <jonathan.perchoc@ovhcloud.com>

| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DATATR-56

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Currently, when we relaunch a job on AI Training, the manager tries to fill the `internal` parameter whereas it should not.

Good behavior should be:

* drop the volume whose `internal` parameter = `true`
* in any other case, we send the volume but we remove the `internal` parameter

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/training/job.service.js
+++ b/packages/manager/modules/pci/src/projects/project/training/job.service.js
@@ -30,20 +30,25 @@ export default class PciProjectTrainingJobService {
       // API does not accept volumes' internal property. We remove it before sending the request.
       toSubmit.volumes = jobSpec.volumes.flatMap((volume) => {
         let toRemove = false;
-        let internalKey;
+        const internalKeys = [];
+        let filteredVolume = volume;
         Object.keys(volume).forEach((key) => {
           if (volume[key] && volume[key].internal !== undefined) {
             if (volume[key].internal === true) {
               toRemove = true;
             } else {
-              internalKey = key;
+              internalKeys.push(key);
             }
           }
         });
         if (toRemove) {
           return [];
         }
-        return internalKey ? omit(volume, `${internalKey}.internal`) : volume;
+        // remove all 'internal' keys from object
+        internalKeys.forEach((key) => {
+          filteredVolume = omit(filteredVolume, `${key}.internal`);
+        });
+        return filteredVolume;
       });
     }
     return this.submit(projectId, toSubmit);
--- a/packages/manager/modules/pci/src/projects/project/training/job.service.js
+++ b/packages/manager/modules/pci/src/projects/project/training/job.service.js
@@ -1,3 +1,4 @@
+import omit from 'lodash/omit';
 import Job from './job.class';
 
 export default class PciProjectTrainingJobService {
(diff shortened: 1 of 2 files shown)


packages/manager/modules/pci/src/projects/project/training/job.service.js
```diff
@@ -25,6 +26,26 @@ export default class PciProjectTrainingJobService {
...
+        let toRemove = false;
+        let internalKey;
+        Object.keys(volume).forEach((key) => {
+          if (volume[key] && volume[key].internal !== undefined) {
+            if (volume[key].internal === true) {
+              toRemove = true;
+            } else {
+              internalKey = key;
+            }
+          }
```
anooparveti, line 41: **question:** Can only one `key` has `internal` property in the `volume` obj? If multiple keys can have internal property, this logic would only remove from the last key where internal is not set to true.

jperchoc: Good point. As far as I know, we should not have multiple `internal` keys on volumes. However, I have changed the function to remove all of them, just in case: https://github.com/ovh/manager/pull/7423/commits/6f43fa3401e12a0af4f367ce703845f5136344c2

antleblanc (approved):

packages/manager/modules/pci/src/projects/project/training/job.service.js
```diff
@@ -25,6 +26,31 @@ export default class PciProjectTrainingJobService {
...
+    if (jobSpec.volumes) {
+      // API does not accept volumes' internal property. We remove it before sending the request.
+      toSubmit.volumes = jobSpec.volumes.flatMap((volume) => {
+        let toRemove = false;
+        const internalKeys = [];
+        let filteredVolume = volume;
+        Object.keys(volume).forEach((key) => {
+          if (volume[key] && volume[key].internal !== undefined) {
+            if (volume[key].internal === true) {
+              toRemove = true;
```
JeremyDec, line 38:
```suggestion
              return [];
```

packages/manager/modules/pci/src/projects/project/training/job.service.js
```diff
@@ -25,6 +26,31 @@ export default class PciProjectTrainingJobService {
...
+            } else {
+              internalKeys.push(key);
+            }
+          }
+        });
+        if (toRemove) {
+          return [];
+        }
```
JeremyDec, line 46: With above suggestion applied
```suggestion
```

packages/manager/modules/pci/src/projects/project/training/job.service.js
jperchoc: Returning `[]` in the forEach loop does not work here, as the forEach returns `void` anyway. So without the `toRemove` flag, we will still have the volume being added to the volumes to submit instead of filtering it.

JeremyDec (approved):

PR #7435, opened 2022-06-27
Fix Apple related issues

zakariasahmane:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | DTRSD-90803, MANAGER-9433

## Description

1 - Under Safari on BigSur and before, the flex is not working as intended, it need the height to be setup explicitly. 
2 - Under any apple device, on hub, the navbar is hidden because of a scroll to top behavior. It has been removed since it doesn't serve any purpose.


ghyenne (approved):

packages/manager/apps/container/src/container/legacy/template.module.scss
```diff
@@ -9,6 +9,7 @@
 
   &_content {
     flex: 1 1 auto;
+    height: 100%; /* For Safari on BigSur or before */
```
antleblanc, line 12: Release name is `Big Sur` (not `BigSur`) :eyes:.
It might be more interesting to specify which Safari version is causing the issue rather than the OS version.

antleblanc (changes requested):

packages/manager/apps/container/src/container/legacy/template.module.scss
zakariasahmane: Apparently the problem comes from the OS not the Safari version, as i am the only one who can reproduce it because i am on Catalina.

antleblanc: Issue should be related to a given version of Safari, not OS related (because we can upgrade Safari in Big Sur (at least to a given version).

antleblanc:
@zakariasahmane Could you please target following branch `release/run-2022-w26`?
Thank you!

cbourgois (approved):

antleblanc (approved):

PR #7439, opened 2022-06-27
fix(container): sidebar ssh key for US region

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | feat/product-nav-reshuffle-update
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-9574

## Description

add ssh menu entry to sidebar for US region


antleblanc (approved):

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/utils.ts
```diff
@@ -37,13 +38,16 @@ export function initTree(
     ) {
       return all;
     }
+    if (node?.region && ![].concat(node.region).includes(region)) {
```
zakariasahmane, line 41: Since we maintain the `node.region` object, wouldn't it be better to just be a `string[]` instead of `string | string[]`, it would make the condition easier to read and to maintain

PR #7495, opened 2022-07-06
refactor(nasha): delete unnecessary files & code

y4nnL:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/nasha-refresh-task-tracker-integration` for hotfix 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #N/A

- [X] Commits are signed-off
- [X] Lint has passed locally
- [ ] ~~Ticket reference is mentioned in linked commits (internal only)~~


packages/manager/modules/nasha/src/dashboard/partition/accesses/accesses.styles.scss
```diff
@@ -11,4 +11,8 @@
   .oui-datagrid-responsive {
     overflow: inherit;
   }
+
+  oui-datagrid.hide-pagination oui-pagination {
+    display: none !important;
+  }
```
anooparveti, line 17: **question:** Can we try and use `ng-if` instead of adding a new class?

y4nnL: I agree this would be better but is this possible when using oui-datagrid as a component ?

anooparveti: I did not see any option to hide the pagination in `data-grid`.

ghyenne (approved):

PR #7634, opened 2022-07-21
fix(cookie-policy): use sources instead of build

cbourgois:
| Question         | Answer
| ---------------- | ---
| Branch?          | `release/stack-clean-apps`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix MANAGER-9745,

## Description

Stop building `@ovh-ux/manager-cookie-policy` as this feature is now built-in in container and this package is only used in `@ovh-ux/sign-up-app`.

### DIFF
--- a/packages/manager/modules/cookie-policy/package.json
+++ b/packages/manager/modules/cookie-policy/package.json
@@ -19,22 +19,7 @@
     "directory": "packages/manager/modules/cookie-policy"
   },
   "license": "BSD-3-Clause",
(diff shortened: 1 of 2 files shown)


frenautvh (approved):
reading the package.json i'm wondering why we are using keywords since the package is private anyway, not really the scope of this PR but it should be removed as it's useless

zakariasahmane (approved):

PR #7653, opened 2022-07-25
Fix/dtrsd 93524

JeremyDec:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-93524

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager/modules/models/user/User.class.js
+++ b/packages/manager/modules/models/user/User.class.js
@@ -31,4 +31,12 @@ export default class User {
   canHaveInvoicesByPostalMail() {
     return this.billingCountry === 'FR' && this.legalform === 'individual';
   }
+
+  isCorporation() {
+    return this.legalform === 'corporation';
+  }
+
+  isPersonalCorporation() {
+    return this.legalform === 'personalcorporation';
+  }
 }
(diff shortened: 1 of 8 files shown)


anooparveti (approved):

packages/manager/modules/models/user/User.class.js
```diff
@@ -31,4 +31,8 @@ export default class User {
   canHaveInvoicesByPostalMail() {
     return this.billingCountry === 'FR' && this.legalform === 'individual';
   }
+
+  isRetractationAllowed() {
```
marie-j, line 35: This looks really generic, are we sure it can be applied to something else than telephony product ? 
Probably should use a `isCorporation()` method

marie-j (changes requested):

packages/manager/modules/models/user/User.class.js
JeremyDec: Good point, name should just reflect what the function does

anooparveti (approved):

cbourgois (approved):

packages/manager/modules/models/user/User.class.js
```diff
@@ -31,4 +31,8 @@ export default class User {
   canHaveInvoicesByPostalMail() {
     return this.billingCountry === 'FR' && this.legalform === 'individual';
   }
+
+  isNotCorporation() {
```
marie-j, line 35: Why not use the `isCorporation` and just reverse the check `!user.isCorporation()` ? Imo it seems more logical to check something `is` that `is not`

JeremyDec: Done

cbourgois (approved):

marie-j (approved):

anooparveti (approved):

PR #7656, opened 2022-07-25
fix(nasha.snapshots): snapshot name validation

y4nnL:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/nasha-refresh`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-9688

- [X] Commits are signed-off
- [X] Lint has passed locally

### DIFF
--- a/packages/manager/modules/nasha/src/dashboard/partition/snapshots/snapshots.template.html
+++ b/packages/manager/modules/nasha/src/dashboard/partition/snapshots/snapshots.template.html
@@ -127,7 +127,7 @@
             </tr>
             <!-- Custom snapshot form row -->
             <tr
-                class="oui-datagrid__row"
+                class="oui-datagrid__row form"
                 data-ng-if="$ctrl.isCustomSnapshotFormShown"
             >
                 <td class="oui-datagrid__cell">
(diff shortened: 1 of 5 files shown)


packages/manager/modules/nasha/src/dashboard/partition/snapshots/snapshots.template.html
```diff
@@ -127,7 +127,7 @@ <h4 data-translate="nasha_dashboard_partition_snapshots_subtitle"></h4>
             </tr>
             <!-- Custom snapshot form row -->
             <tr
-                class="oui-datagrid__row"
+                class="oui-datagrid__row form"
```
marie-j, line 130: Any reason to use the styles instead of the component directly ?

y4nnL: You mean `oui-datagrid` ?
Yes there are two reasons actually :
- It does not provide a property to hide the pagination (doing this using css is pretty uggly IMO)
- The form row which is hidden by default. Instead of pushing / shifting the array with fake data and split the form across multiple `oui-datagrid-column`s, using styles allows me to simply ng-if the corresponding row

anooparveti (approved):

sachinrameshn (approved):

PR #7668, opened 2022-07-26
fix(storages.containers): display public-url or virtual-host

mohammed-zahaf:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/MANAGER-9093`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-9724

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
1. display `publicUrl` for SWIFT container
2. display `virtualHost` for S3 buckets

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/storages/containers/container/container.html
+++ b/packages/manager/modules/pci/src/projects/project/storages/containers/container/container.html
@@ -93,7 +93,7 @@
                 >
                     <oui-clipboard
                         data-name="containerStaticUrl"
-                        data-model="$ctrl.container.staticUrl"
+                        data-model="$ctrl.container.staticUrl || $ctrl.container.virtualHost"
                     ></oui-clipboard>
                 </oui-field>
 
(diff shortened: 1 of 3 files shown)


sachinrameshn (approved):

packages/manager/modules/pci/src/projects/project/storages/containers/container/container.html
```diff
@@ -93,7 +93,7 @@
                 >
                     <oui-clipboard
                         data-name="containerStaticUrl"
-                        data-model="$ctrl.container.staticUrl"
+                        data-model="$ctrl.container.staticUrl || $ctrl.container.virtualHost"
```
y4nnL, line 96: **question**: no issue to bind a model to a key that may not exist in the container object ?

mohammed-zahaf: Well see, i don't notice any issue from my side.
It works for both case.

Also, as you know it's not a `data-ng-model` directive, and the clipboard is not editable.
So, i think no issues.

anooparveti (approved):

ghyenne (approved):

PR #7757, opened 2022-08-09
fix(sso-auth): add target top for signup new redirection

ghyenne:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | DTRSD-95297 

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
I added target top for signup new redirection if nic have partiel registration.
I also fixed a console message related to the variable newZipModel that not tested.

### DIFF
--- a/packages/components/ng-ovh-sso-auth/src/provider.js
+++ b/packages/components/ng-ovh-sso-auth/src/provider.js
@@ -547,7 +547,7 @@ export default function() {
             }
           }
 
-          $window.location.assign(
+          $window.top.location.assign(
             destUrl +
               (urlPart.indexOf('?') > -1 ? '&' : '?') +
               params.join('&'),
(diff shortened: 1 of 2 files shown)


JeremyDec (approved):

zakariasahmane:
What is the purpose of the target top here? If i'm not mistaken, signup is not in the container or in an iframe, why do we need to target top ?

ghyenne:
> What is the purpose of the target top here? If i'm not mistaken, signup is not in the container or in an iframe, why do we need to target top ?

If we don't put a "top", we stay on the manager's loader and we already have the navbar while the account is not yet finalized.

zakariasahmane (approved):

anooparveti (approved):

PR #7777, opened 2022-08-11
Fix/tracking

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | release/RUN-2022-w32
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-8855

## Description

- fix site name attribute
- avoid sending tracking data from the container before configuration has been set by the current application


antleblanc (approved):

anooparveti (approved):

packages/manager/apps/container/src/container/index.tsx
```diff
@@ -12,28 +12,34 @@ import SSOAuthModal from '@/sso-auth-modal/SSOAuthModal';
 export default function Container(): JSX.Element {
   const { isLoading, betaVersion, useBeta } = useContainer();
   const shell = useShell();
-  const tracking = shell.getPlugin('tracking');
   const isNavReshuffle = betaVersion && useBeta;
+
   if (isLoading) {
     return <></>;
   }
```
anooparveti, line 19: **nitpick:** Can be moved to return statement. This could fix the reported SonarCloud issue

frenautvh: you're right i fixed it, the code is cleaner like that, thanks

PR #7808, opened 2022-08-17
feat(react-app): add dashboard

Jisay:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/core-dev`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | yes
| Tickets          | MANAGER-9859


packages/manager/core/app/src/components/Dashboard/DashboardTile.tsx
```diff
@@ -0,0 +1,42 @@
...
+
+export type DashboardTileProps = {
+  tile: DashboardTileType;
+};
+
+export default function DashboardTile({
+  tile,
+}: DashboardTileProps): JSX.Element {
+  const [isLoading, setIsLoading] = useState(true);
+  const [data, setData] = useState<unknown>();
```
frenautvh, line 16: if all of your tile components are using some data but you don't know in advance what the data will be, it can be good to use templates instead of unknown :
```ts
DashboardTile<T>
const [data, setData] = useState<T>();
```
same for other components :
```ts
DashboardTileList<T> 
```
and for example if you know that the data will always have a description you can then provides some typing by using :
```ts
interface DashboardTileData {
  description: string;
};
DashboardTileList<T extends DashboardTileData> 
```
that way your component will work for abstract data but you can add your own constraints (for example i want my component to work with any object that as a "description" attribute)

packages/manager/core/app/src/components/Dashboard/DashboardTileLoading.tsx
```diff
@@ -0,0 +1,21 @@
+import React from 'react';
+import { TileSection } from '@ovh-ux/manager-themes';
+import { SkeletonText } from '@chakra-ui/react';
+
+export default function DashboardTileLoading({ count = 4 }: any): JSX.Element {
```
frenautvh, line 5: -any :)

packages/manager/core/app/src/pages/Dashboard.tsx
```diff
@@ -1,7 +1,145 @@
...
-export default function Dashboard(): JSX.Element {
+import Nutanix, {
+  getNutanix,
+  getServiceInfos,
+  getServer,
+  getTechnicalDetails,
+  Server,
+} from '@/api/nutanix';
+import Service, { getServiceDetails, TechnicalDetails } from '@/api/service';
+import Dashboard, { TileTypesEnum } from '../components/Dashboard';
```
frenautvh, line 15: you can use
```ts
import Dashboard, { TileTypesEnum } from '@/components/Dashboard';
```
to avoid all the `..`

packages/manager/core/app/src/pages/Details.tsx
```diff
@@ -0,0 +1,45 @@
...
+      name: 'nodes',
+      title: t('tab_nodes'),
+      to: 'nodes',
+    },
+  ];
+
+  const [activeIndex, setActiveIndex] = useState(0);
+
+  return (
+    <Tabs
```
frenautvh, line 25: you want to use the isLazy option to avoid making network calls for tabs that are not displayed, see : https://chakra-ui.com/docs/components/tabs#lazily-mounting-tab-panels

packages/manager/core/app/src/pages/Details.tsx
```diff
@@ -0,0 +1,45 @@
...
+      variant="light"
+      index={activeIndex}
+      onChange={(e) => setActiveIndex(e)}
+    >
+      <TabList>
+        {tabs.map((tab, index) => (
+          <Tab as={NavLink} key={tab.name} to={tab.to}>
+            {({ isActive }: { isActive: boolean }) => {
```
frenautvh, line 33: i don't understand the usage of this method here ?
in the documentation it doesn't looks like they have to do that to handle tab change

packages/manager/core/app/src/public/translations/.editorconfig
```diff
@@ -0,0 +1,12 @@
+# EditorConfig is awesome: https://EditorConfig.org
```
frenautvh, line 1: to remove ?

packages/manager/core/app/src/public/translations/dashboard/Messages_fr_FR.json
```diff
@@ -0,0 +1,22 @@
...
+    "tile_general_item_admin_interface_link": "Accéder à Prism Central",
+    "tile_general_item_license": "Licence",
+    "tile_general_item_deployment_mode": "Mode de déploiement",
+    "tile_general_item_replication_factor": "Facteur de réplication",
+    "tile_general_item_datacenter": "Datacentre",
+    "tile_general_item_rack": "Baie",
+    "tile_general_title": "Informations générales",
+    "tile_licenses_title": "Licences Nutanix",
+    "tile_billing_title": "Abonnement",
+    "tile_supportl_title": "Support",
```
frenautvh, line 18: supportl typo

packages/manager/core/app/src/components/Dashboard/DashboardTileLoading.tsx
Jisay: This component will be useless. With @zakariasahmane, we discussed on a props that will be added to the `Tile` component that will handle the loading.

packages/manager/core/app/src/pages/Details.tsx
Jisay: I will change the way the activation of the current tab is done

packages/manager/core/app/src/api/nutanix.ts
```diff
@@ -32,6 +40,12 @@ export type Nutanix = {
   };
 };
 
+export type Server = {
+  datacenter: string;
```
marie-j, line 44: Could it make sense to have an enum ?

packages/manager/core/app/src/api/nutanix.ts
```diff
@@ -76,4 +90,79 @@ export async function fetchNutanixMetaInfos(
   return { region: data.datacenter };
 }
 
+export async function getNutanix(serviceName: string): Promise<Nutanix> {
+  const response = await fetch(`/engine/api/nutanix/${serviceName}`);
+  return response.json();
+}
+
+export async function getServiceInfos(
```
marie-j, line 98: Can it be replaced with `/services` ?

packages/manager/core/app/src/api/nutanix.ts
```diff
@@ -76,4 +90,79 @@ export async function fetchNutanixMetaInfos(
...
+  serviceName: string,
+): Promise<ServiceInfos> {
+  const response = await getInfos('nutanix', serviceName);
+  return response;
+}
+
+export async function getServer(nodeId: string): Promise<Server> {
+  const response = await fetch(`/engine/2api/sws/dedicated/server/${nodeId}`);
```
marie-j, line 106: Maybe as a TODO we could think of removing this 2API 😬

packages/manager/core/app/src/api/service.ts
```diff
@@ -0,0 +1,68 @@
...
+};
+
+export type TechnicalDetails = {
+  baremetalServers?: BareMetalServersDetails;
+  nutanixCluster?: NutanixClusterDetails;
+  serviceId?: number;
+};
+
+export async function getServiceDetails(serviceId: number): Promise<Service> {
+  const response = await fetch(`/engine/api/services/${serviceId}`);
```
marie-j, line 47:
```suggestion
  const response = await fetch(`/engine/apiv6/services/${serviceId}`);
```
?

packages/manager/core/app/src/api/nutanix.ts
Jisay: The use /serviceInfos is to get the `serviceId` with given `serviceName`. I need it to get this information and then call /services.

PR #7852, opened 2022-08-23
fix(nasha): add and change some tracking

ghyenne:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/nasha-refresh `
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          |  MANAGER-9031 

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
I added and changed some tracking


sachinrameshn (approved):

packages/manager/modules/nasha/src/components/task-tracker/task-tracker.utils.js
```diff
@@ -25,6 +26,8 @@ export const createTaskTrackerStateOptions = (queryParams = []) => ({
         .filter((key) => queryParams.includes(key))
         .reduce((set, key) => ({ ...set, [key]: params[key] }), {});
     },
+    trackingData: /* @ngInject */ ($transition$) =>
+      $transition$.params().trackingData,
```
anooparveti, line 30: **question:** If state param is not put in the URL, it can be lost after refresh. So is this case taken care?

ghyenne: If there is a refresh of the page, there is no tracking in this case (and no error).

anooparveti (approved):

PR #7908, opened 2022-08-31
feat(react-app): add error management on dashboard tile

Jisay:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/core-dev`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | none

## Description

Add error management on dashboard tiles.

## Todos

- [ ] rebase and use queries since #7941 is merged (@chloren)


antleblanc (changes requested):
Some checks were not successful

>  CDS/UXMAN-manager-build-v2 — build-v2: Fail 

Could you please have a look?
Thank you!

packages/manager/core/app/src/api/service.ts
```diff
@@ -153,32 +154,32 @@ export type GenericProductDefinition = {
 };
 
 export async function getServiceDetails(serviceId: number): Promise<Service> {
-  const response = await fetch(`/engine/apiv6/services/${serviceId}`);
-  return response.json();
+  const { data } = await apiClient.v6.get(`/services/${serviceId}`);
```
marie-j, line 157: I'm wondering as we are always repeating the same pattern, does it make sense for apiClient to directly return data instead of an object ?

Jisay: Good point. Need to be discussed :-)
Cc @frenautvh, @cbourgois

PR #8014, opened 2022-09-16
fix(pci.project): cannot create gateway with existing network

vks-ovh:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/l3-services` 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-10012

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Gateway creation doesnot happen with existing private network.

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/public-gateways/add/add.controller.js
+++ b/packages/manager/modules/pci/src/projects/project/public-gateways/add/add.controller.js
@@ -45,6 +45,7 @@ export default class PciPublicGatewaysAddController {
     this.gatewayName = null;
     this.loadingDefaultValues = false;
     this.isCustomNetwork = false;
+    this.isCreatingGateway = false;
 
     this.selectedPrivateNetwork = this.getDefaultSelectValue(
       'pci_projects_project_public_gateways_add_select_private_network',
@@ -262,7 +263,7 @@ export default class PciPublicGatewaysAddController {
       .then(() => this.onSuccess())
       .catch((err) => this.onError(err))
       .finally(() => {
-        this.loading = false;
+        this.isCreatingGateway = false;
       });
   }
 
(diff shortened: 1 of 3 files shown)


packages/manager/modules/pci/src/projects/project/public-gateways/service/public-gateways.service.js
```diff
@@ -144,4 +169,18 @@ export default class publicGatewaysServiceClass {
       };
     });
   }
+
+  checkGatewayStatus(serviceName, operationId, namespace) {
```
anooparveti, line 173: **nitpick** Since the function checks the status of gateway and private network creation, it is a generic function. So generic name would be better.
```suggestion
  checkOperationStatus(serviceName, operationId, namespace) {
```

anooparveti (changes requested):

packages/manager/modules/pci/src/projects/project/public-gateways/service/public-gateways.service.js
vks-ovh: updated as suggested.

packages/manager/modules/pci/src/projects/project/public-gateways/add/add.controller.js
```diff
@@ -281,15 +282,15 @@ export default class PciPublicGatewaysAddController {
       .then(() => this.onSuccess())
       .catch((err) => this.onError(err))
       .finally(() => {
-        this.loading = false;
+        this.createGateway = false;
```
sachinrameshn, line 285: Question: It should be this.isCreatingGateway = false ?

sachinrameshn (approved):

PR #8114, opened 2022-09-30
feat(webpack): add html minifier in webpack config

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | develop
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-10134

## Description

Minify html in production, bundle sizes should be sightly reduced

### DIFF
--- a/packages/manager/tools/webpack-config/src/loaders/html-minifier.js
+++ b/packages/manager/tools/webpack-config/src/loaders/html-minifier.js
@@ -1,0 +1,18 @@
+const { minify } = require('html-minifier');
+// see : https://github.com/kangax/html-minifier
+
+module.exports = function translationsJson(source) {
+  if (this.mode !== 'production') {
+    return source;
+  }
+  return minify(source, {
+    continueOnParseError: true,
+    collapseWhitespace: true,
+    removeAttributeQuotes: true,
+    removeComments: true,
+    removeRedundantAttributes: true,
+    removeScriptTypeAttributes: true,
+    removeStyleLinkTypeAttributes: true,
+    useShortDoctype: true,
+  });
+};
--- a/yarn.lock
+++ b/yarn.lock
@@ -7727,7 +7727,7 @@ callsites@^3.0.0:
   resolved "https://registry.yarnpkg.com/callsites/-/callsites-3.1.0.tgz#b3630abd8943432f54b3f0519238e33cd7df2f73"
   integrity sha512-P8BjAsXvZS+VIDUI11hHCQEv74YT67YUi5JJFNWIqL235sBmjX4+qx9Muvls5ivyNENctx46xQLQ3aTuE7ssaQ==
 
-camel-case@3.0.x:
+camel-case@3.0.x, camel-case@^3.0.0:
   version "3.0.0"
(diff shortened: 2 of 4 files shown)


antleblanc (changes requested):
Changes looks good to me.

`@ovh-ux/manager-webpack-config` is a public package.
According to the current diff, looks like a new feature is introduced.

Shouldn't the commit type be `feat` rather than `build`?

packages/manager/tools/webpack-config/src/loaders/html-minifier.js
```diff
@@ -0,0 +1,18 @@
+const { minify } = require('html-minifier');
+// see : https://github.com/kangax/html-minifier
+
+module.exports = function translationsJson(source) {
+  if (this.mode !== 'production') {
+    return source;
+  }
+  return minify(source, {
+    continueOnParseError: true,
```
marie-j, line 9: Out of curiosity why do we need this ? On what cases is it failing and can't we fix HTML to be valid instead ?

yarn.lock
```diff
@@ -11781,11 +11788,16 @@ flat-cache@^2.0.1:
     rimraf "2.6.3"
     write "1.0.3"
 
-flatpickr@4.6.9, flatpickr@^4.6.3, flatpickr@~4.5.2:
+flatpickr@^4.6.3:
   version "4.6.9"
   resolved "https://registry.yarnpkg.com/flatpickr/-/flatpickr-4.6.9.tgz#9a13383e8a6814bda5d232eae3fcdccb97dc1499"
   integrity sha512-F0azNNi8foVWKSF+8X+ZJzz8r9sE1G4hl06RyceIaLvyltKvDl6vqk9Lm/6AUUCi5HWaIjiUbk7UpeE/fOXOpw==
 
+flatpickr@~4.5.2:
```
marie-j, line 11796: Maybe we need to rerun `yarn install` to remove this

packages/manager/tools/webpack-config/src/loaders/html-minifier.js
frenautvh: My reasoning was that minification plugin should not have the role of html validator. Since the scope of the tool is to optimize the bundle sizes. That's why i choosed the most permissive option, which uses the handle parse error behavior here : https://html.spec.whatwg.org/multipage/parsing.html#parse-errors. For syntax validation i prefer to rely on linting tools (which provide a realtime feedback during development). What do you think about it ?

Jisay (approved):

frenautvh:
changed commit type to "feat", ran "yarn install" to avoid yarn.lock diffs

antleblanc (approved):

packages/manager/tools/webpack-config/src/loaders/html-minifier.js
marie-j: It makes sense

marie-j (approved):

cbourgois:
Rebased to target `release/core-stack-w41`

cbourgois (approved):

mohammed-zahaf (approved):

JeremyDec (approved):

PR #8389, opened 2022-11-16
fix(dedicated.nutanix): add type of pack as link on list page

vks-ovh:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/dev-nutanix`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-10373 MANAGER-10382

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager/modules/nutanix/src/routing.js
+++ b/packages/manager/modules/nutanix/src/routing.js
@@ -3,6 +3,7 @@ import statusTemplate from './templates/status.html';
 import prismUrl from './templates/prismUrl.html';
 import serviceLink from './templates/serviceLink.html';
 import localizationTemplate from './templates/localization.html';
+import { CLUSTER_LIST_TITLE } from './constants';
 
 export default /* @ngInject */ ($stateProvider) => {
   $stateProvider.state('nutanix.index', {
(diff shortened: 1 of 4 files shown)


packages/manager/modules/nutanix/src/routing.js
```diff
@@ -61,10 +62,10 @@ export default /* @ngInject */ ($stateProvider) => {
             property: 'targetSpec.nodes',
           },
           {
-            title: $translate.instant('nutanix_cluster_list_licence'),
-            property: 'targetSpec.license',
+            title: CLUSTER_LIST_TITLE.TYPE_OF_PACK,
```
ghyenne, line 65: I think that for the user it is better to keep something translated. Because `Type of pack` can easily be translated.

vks-ovh: Hi @ghyenne, we don't have to translate `Type of pack` that's why we kept it under constant. Its going to be same in all the languages.

anooparveti (approved):

sachinrameshn (approved):

PR #8620, opened 2022-12-22
feat(core): add logger plugin

Jisay:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/core-react-app-generator`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | -

## Description

Add a logger plugin that enable log into the console through the shell when debug mode is activated (a key in localStorage).


packages/manager/core/generator/app/templates/src/hooks/useLogger.tsx
```diff
@@ -0,0 +1,8 @@
+import { useShell } from '@ovh-ux/manager-react-core-application';
+
+export function useLogger() {
```
cbourgois, line 3: Maybe we should have a specific package that contains all shell hooks.

cbourgois (approved):

PR #8784, opened 2023-01-20
fix(telecom.sms): disable the add and remove ip if update ip status

ghyenne:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/telecom_sms_smpp_ga` 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-10805 

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
I disabled the add and remove ip if update ip status

### DIFF
--- a/packages/manager/modules/sms/src/sms/options/smppParameter/smppParameter.controller.js
+++ b/packages/manager/modules/sms/src/sms/options/smppParameter/smppParameter.controller.js
@@ -1,19 +1,45 @@
 export default class SmsSmppParameterCtrl {
   /* @ngInject */
-  constructor(SmsService) {
+  constructor(SmsService, $timeout, $state) {
     this.smsService = SmsService;
+    this.$timeout = $timeout;
+    this.$state = $state;
   }
 
   $onInit() {
+    this.timer = null;
     this.OPTION_SMPP_TRACKING_PREFIX = 'sms::service::configure-smpp::';
     this.loading = true;
     return this.smsService
       .getSmppSettings(this.serviceName)
       .then((result) => {
         this.smppSettings = result;
+        if (this.smppSettings.status === 'UPDATING_IP') {
+          this.checkStatus();
+        }
       })
       .finally(() => {
         this.loading = false;
+      });
+  }
+
+  $onDestroy() {
+    this.$timeout.cancel(this.timer);
+  }
+
+  checkStatus() {
+    return this.smsService
+      .getSmppSettings(this.serviceName)
+      .then(({ status }) => {
+        this.smppSettings.status = status;
+        if (
+          status === 'UPDATING_IP' &&
+          this.$state.current.name === 'sms.service.options.smppParameter'
+        ) {
+          this.timer = this.$timeout(() => this.checkStatus(), 5000);
+          return null;
+        }
+        return status;
       });
   }
 }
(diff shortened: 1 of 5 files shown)


packages/manager/modules/sms/src/sms/options/smppParameter/smppParameter.controller.js
```diff
@@ -1,19 +1,44 @@
...
+          this.checkStatus();
+        }
       })
       .finally(() => {
         this.loading = false;
       });
   }
+
+  $onDestroy() {
+    this.checkStatus = () => false;
```
anooparveti, line 27: **question:** Why redefining the function in `onDestroy`?  Doesn't smell good.

ghyenne: Without this, if the user goes to the page with a service in UPDATE_IP, it starts the function (which is desired) but if he goes out and comes back several times in a few seconds, the function is started several times, whereas in this case the polling function is killed if he goes out of the page.

packages/manager/modules/sms/src/sms/options/smppParameter/smppParameter.controller.js
```diff
@@ -1,19 +1,44 @@
...
+  checkStatus() {
+    return this.smsService
+      .getSmppSettings(this.serviceName)
+      .then(({ status }) => {
+        this.smppSettings.status = status;
+        if (
+          status === 'UPDATING_IP' &&
+          this.$state.current.name === 'sms.service.options.smppParameter'
+        ) {
+          return this.$timeout(() => this.checkStatus(), 5000);
```
anooparveti, line 39: **nitpick:** `$interval` is appropriate for this use-case. instead of `$timeout`. Could you please change the implementation?
Let me know if you need any information.

anooparveti (changes requested):

packages/manager/modules/sms/src/sms/options/smppParameter/smppParameter.controller.js
ghyenne: I don't have the same expected result.

packages/manager/modules/sms/src/sms/options/smppParameter/smppParameter.controller.js
ghyenne: I did modify $onDestroy

anooparveti (approved):

packages/manager/modules/sms/src/sms/options/smppParameter/smppParameter.controller.js
```diff
@@ -1,19 +1,46 @@
...
-  constructor(SmsService) {
+  constructor(SmsService, $timeout, $state) {
     this.smsService = SmsService;
+    this.$timeout = $timeout;
+    this.$state = $state;
   }
 
   $onInit() {
+    this.timer = null;
+    this.checkStatus = this.checkStatus.bind(this);
```
y4nnL, line 11: This line could be removed safely
https://github.com/ovh/manager/pull/8784/files#diff-38f929e0b6174bcb765649cbb60135040a628b126368975cf87472a314186a22R27
When you call `$ctrl.checkStatus()`, `checkStatus`'s context is always `$ctrl`

y4nnL (approved):

PR #8855, opened 2023-02-02
feat/iam

y4nnL:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-10393

- [X] Commits are signed-off
- [X] Lint has passed locally


ghyenne (dismissed):

anooparveti (dismissed):

packages/manager/modules/iam/package.json
```diff
@@ -0,0 +1,33 @@
...
+    "@ovh-ux/ng-ui-router-breadcrumb": "^1.1.8",
+    "@ovh-ux/ui-kit": "^6.4.0",
+    "@uirouter/angularjs": "^1.0.23",
+    "angular": "^1.7.5",
+    "angular-translate": "^2.18.1",
+    "lodash-es": "^4.17.15",
+    "oclazyload": "^1.1.0",
+    "ovh-ui-kit-bs": "^4.2.0"
+  },
+  "resolveAlias": {
```
frenautvh, line 26: i think you should extend your webpack config file and setup your aliases here, just like in the doc : https://webpack.js.org/configuration/resolve/

packages/manager/tools/webpack-config/src/webpack.common.js
```diff
@@ -23,9 +24,21 @@ const cacheLoader = {
   },
 };
 
+const readClosestPackage = (entry) => {
```
frenautvh, line 27: i don't think you need to do that you just need to define your aliases in your own webpack config by extending this default configuration

frenautvh (changes requested):

packages/manager/modules/iam/package.json
y4nnL: I don't think that's a good idea because we separate modules from apps, that leads to a lot of code repetition. The module needs the aliases, so it is his responsibility to declare them, I can't find a cleaner way of doing it ([others had the same idea](https://github.com/Rush/link-module-alias)). The apps that import the module have to resolve them for the module to work. They should not know about the details of the resolution

packages/manager/tools/webpack-config/src/webpack.common.js
y4nnL: This is exactly what they - the apps that import the IAM module - do [here](https://github.com/ovh/manager/pull/8855/files#diff-7efd7f75a5e0dac4f9df27bb7504c7c1d7ac723943f160a3d211f30f3ccdef55R53) and [here](https://github.com/ovh/manager/pull/8855/files#diff-bca528868988736434df1ac5cf88ff4553ae365eac1db02cdc4bad69d41aab6fR13). Because we use a common webpack file to do common stuff, and because module could be imported by many apps (IAM and Dedicated in our case), I created a common option instead of duplicating the code

That way :
- The module defines its own aliases, which is the best thing to do since it uses the aliases
- Applications that import the module set the common `resolveAlias` option because the module requires them to (think of it as a peer dependency) and don't know about the details
- The common webpack file only handles the option, it is now pure algorithm (I was asked by @antleblanc to remove all business logic in a previous version, which I did)

qpavy (approved):

ghyenne (approved):

sachinrameshn (approved):

PR #9011, opened 2023-03-06
feat(iam): make services a module

y4nnL:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/iam`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-10394

- [X] Commits are signed-off
- [X] Lint has passed locally

### DIFF
--- a/packages/manager/modules/iam/src/services/index.js
+++ b/packages/manager/modules/iam/src/services/index.js
@@ -1,9 +1,16 @@
-import APIV2Service from './apiv2.service';
+import angular from 'angular';
+
+import Apiv2Service from './apiv2.service';
 import GuideService from './guide.service';
 import PolicyService from './policy.service';
 
-export default {
-  APIV2Service,
-  GuideService,
-  PolicyService,
-};
+export const moduleName = 'ovhManagerIAMServices';
+
+angular
+  .module(moduleName, [])
+  .service('Apiv2Service', Apiv2Service)
+  .service('GuideService', GuideService)
+  .service('PolicyService', PolicyService)
+  .run(/* @ngTranslationsInject:json ./translations */);
+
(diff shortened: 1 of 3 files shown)


packages/manager/modules/iam/src/services/index.js
```diff
@@ -1,9 +1,16 @@
...
-  PolicyService,
-};
+export const moduleName = 'ovhManagerIAMServices';
+
+angular
+  .module(moduleName, [])
+  .service('Apiv2Service', Apiv2Service)
+  .service('GuideService', GuideService)
+  .service('PolicyService', PolicyService)
+  .run(/* @ngTranslationsInject:json ./translations */);
```
anooparveti, line 14: **nitpick:** Services are meant to make HTTP calls, transform result and return. Why do we need translations in `services` folder. Could you please maintain translations along with the state. This makes services reusable,

y4nnL: > Why do we need translations in `services` folder

For two main reasons:
- You said it yourself, theses services transform data and return it. These transformations include a new mechanism from the v2 API that return specific error classes from POST calls. These classes are parsed and translated directly into the `Apiv2Service`. It would make no sense if every component that uses this service has to translate the same errors again and again. Please have a look at https://github.com/ovh/manager/commit/9694739e916dcfeccd980c2279ee9dde938874ad#diff-6e255867de5e16e5bf8df9e18fe0086e04f98dd91066921f9b51f7d41acc472cR84-R127
- `services` is an angular module, as such it brings its own translations, so it is isolated and could be reused like any other modules

anooparveti (approved):

PR #9030, opened 2023-03-09
fix(ldp): update en_gb translation related to encryption keys

pfrayer:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

A recent english translation is erroneous (`Surname` instead of `Name`). It refers to an object (an encryption key), and not a real person, so `Name` should be the good translation. Other languages translations seems OK.

## Related

PR #8249

### DIFF
--- a/packages/manager/modules/dbaas-logs/src/logs/translations/Messages_en_GB.json
+++ b/packages/manager/modules/dbaas-logs/src/logs/translations/Messages_en_GB.json
@@ -730,7 +730,7 @@
   "logs_encryption_keys_goto_1": "Manage my",
   "logs_encryption_keys_goto_2": "PGP encryption keys",
   "logs_encryption_keys_add": "Add a PGP encryption key",
-  "logs_encryption_keys_col_title": "Surname",
+  "logs_encryption_keys_col_title": "Name",
   "logs_encryption_keys_col_fingerprint": "Fingerprint",
   "logs_encryption_keys_col_algorithm": "Algorithm",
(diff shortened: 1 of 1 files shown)


anooparveti (changes requested):
Hello @pfrayer,
We are not supposed to modify the translations files apart from `Message_fr_FR.json`. Could you please handle this change by raising a translation request?

Procedure:

1. Create a branch from `master`
2. Push it to origin
3. Prepare a `yaml` config 
4. Request a translation

Let me know if you have any queries.

pfrayer:
Hello @anooparveti ,

Should we really go to the entire translation process to just update one word ? If so, the translation process will result in the exact content of this PR.

anooparveti:
> 

Hello @pfrayer,
Yes it is required to go through the trans process no matter the size of the change. This is required because all the French content is cached and the next time if we request trans for the existing content, trans process will serve it from the cache. So when the change goes through trans process, we can make sure cache is updated.

pfrayer:
Fixed in [f6a075eecd6492a61bfc8f786a022e9a24203780](https://github.com/ovh/manager/commit/f6a075eecd6492a61bfc8f786a022e9a24203780#diff-13f5687a3bbd8fac6a111c8346b2212338918dbf40b8a7a320950d360a866354)

PR #9502, opened 2023-06-21
fix(iam): set resource group form model

y4nnL:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/iam`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-10401,  MANAGER-10518

- [X] Commits are signed-off
- [X] Lint has passed locally

### DIFF
--- a/packages/manager/modules/iam/src/components/createResourceGroup/createResourceGroup.controller.js
+++ b/packages/manager/modules/iam/src/components/createResourceGroup/createResourceGroup.controller.js
@@ -49,8 +49,7 @@ export default class CreateResourceGroupController {
      */
     this.model = {
       name: '',
-      resources: [],
-      resourceTypes: [],
+      resources: { selection: [], types: [] },
     };
 
     /**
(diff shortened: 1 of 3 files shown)


packages/manager/modules/iam/src/components/createResourceGroup/createResourceGroup.controller.js
```diff
@@ -186,13 +185,30 @@ export default class CreateResourceGroupController {
...
+                this.model.resources.selection?.filter(
+                  ({ id }) =>
+                    !deletedResources.find((resource) => id === resource.id),
+                ) || [],
+            };
+          }
+        }),
+      )
+      .then(() => {
+        this.deletion = null;
```
darsene, line 208: What about if `this.deletion` fails, `this.deletion` should also be set to null or not ?

y4nnL: The promise can not fail with the following line
```javascript
this.deletion.resolve(success);
```
But I agree this should be set in a `finally` statement

PR #9521, opened 2023-06-26
feat(tracking): add tracking on ods components

aboungnaseng-ovhcloud:
ref: MANAGER-11442

| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes/no
| New feature?     | yes/no
| Breaking change? | yes/no
| Tickets          | Fix #... 

- [ ] Commits are signed-off
- [ ] Lint has passed locally


packages/manager/core/application/src/ovh-tracking.tsx
```diff
@@ -0,0 +1,131 @@
...
+        page: { name: page },
+        complete_page_name: page,
+        level2: '81',
+        type: 'action',
+      });
+    });
+  };
+
+  // Send tracking click for the ODS_COMPONENTS_GROUP
+  const ovhTrackingParenNode = (event) => {
```
anooparveti, line 66:
```suggestion
  const ovhTrackingParentNode = (event) => {
```

packages/manager/core/application/src/ovh-tracking.tsx
```diff
@@ -0,0 +1,131 @@
...
+      if (tabsEvents.length > 0)
+        tabsEvents.forEach((element) => {
+          element.removeEventListener(
+            'odsValueChange',
+            ovhTrackingParenNode,
+            false,
+          );
+        });
+    };
+  }, [location]);
```
anooparveti, line 108: **question:**  Default tracking is added to ODS components when the location changes. But what if some ODS components are added to the DOM after some user actions on the current screen.

For ex,
1. User lands on the config page
2. Enters the value in `Field A`
3. Based on the response to `Field A`, we decide to display either `Field B` or `Field C`.

So with this implementation, are the fields displayed conditionally are not tracked right as there is no change in the location?

packages/components/ovh-at-internet/src/ovh-at-internet.ts
```diff
@@ -297,13 +307,25 @@ export default class OvhAtInternet extends OvhAtInternetConfig {
...
+        const pageTrackingData2 = getPageTrackingData(data?.page);
+        tracking = {
+          ...tracking,
+          page: pageTrackingData2.page,
+          page_chapter1: pageTrackingData2.page_chapter1,
+          page_chapter2: pageTrackingData2.page_chapter2,
+          page_chapter3: pageTrackingData2.page_chapter3,
+          page_complete: data?.page?.name,
+        };
+      }
```
anooparveti, line 327: **question:** Can you please explain me this? Why do we need to execute `getPageTrackingData` on `data` as well as `data.page`?

aboungnaseng-ovhcloud: sure i will update it

packages/components/ovh-at-internet/src/ovh-at-internet.ts
```diff
@@ -115,6 +118,10 @@ export default class OvhAtInternet extends OvhAtInternetConfig {
   }
 
   shouldUsePianoAnalytics() {
+    if (window.location?.hostname === 'localhost') {
+      return true;
+    }
```
anooparveti, line 123: Use environment variable also to enable tracking in localhost.
Environment variable can be defined in `packages/manager/tools/dev-server/bin/manager-dev-server.js`

chipp972 (approved):

packages/manager/core/application/src/ovh-routing.tsx
```diff
@@ -36,6 +45,16 @@ function HidePreloader(): JSX.Element {
 
 function OvhContainerRoutingSync(): JSX.Element {
   const location = useLocation();
+  const outlet = useOutlet();
+  const navigate = useNavigate();
+  useEffect(() => {
+    setTimeout(() => {
+      const child = outlet.props?.children?.props?.children?.props?.children;
```
anooparveti, line 52: **question:** Why is this nesting required?
Also, this seems like a bad approach as `props.childern` can have multiple levels of children

packages/manager/core/application/src/ovh-tracking.tsx
```diff
@@ -0,0 +1,131 @@
...
+    const path = myStateRef.current.pathname.split('/')[1];
+    env.then((response) => {
+      const { applicationName, universe } = response;
+      const name = `${applicationName}::${path || 'homepage'}::${value}`;
+      const page = `${universe}::${applicationName}::${path || 'homepage'}`;
+      tracking.trackClick({
+        name,
+        page: { name: page },
+        complete_page_name: page,
+        level2: '81',
```
anooparveti, line 59: **nitpick:** `level2` is set based on the universe. So it should not be hard-coded

anooparveti (changes requested):

packages/manager/core/application/src/ovh-tracking.tsx
aboungnaseng-ovhcloud: For the moment, we didn't have the key in the useShell(), that why I hard coded this

In the this file 'packages/components/ovh-at-internet/src/constants.ts' you have a constant AT_INTERNET_LEVEL2, but we can't get the key nowhere

packages/manager/core/application/src/ovh-routing.tsx
```diff
@@ -107,16 +115,29 @@ export function createAppRouter() {
...
-  const routes = [...regularRoutes, { path: '*', element: <NotFound /> }];
+  const newRoutes = regularRoutes.map((route, index) => {
+    if (route.path === 'dashboard') {
+      return {
+        ...route,
+        path: '*',
+        errorElement: <NotFound />,
+        error: <NotFound />,
+      };
+    }
```
anooparveti, line 126: **question:** Why is it different for `dashboard`?

packages/manager/core/application/src/ovh-tracking.tsx
```diff
@@ -0,0 +1,131 @@
...
+  const OvhTrackPage = (loc: any) => {
+    const path = loc?.pathname.split('/')[1];
+    const partsReplace = loc.pathname.substring(1).replaceAll('/', '::');
+    env.then((response) => {
+      const { applicationName, universe } = response;
+      const name = `${universe}::${applicationName}::${partsReplace ||
+        'homepage'}`;
+      tracking.trackPage({
+        name,
+        level2: '81',
```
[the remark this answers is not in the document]
anooparveti: If the key is not present in `useShell()`, you can add it.

packages/manager/core/application/src/ovh-tracking.tsx
anooparveti: If the key is not present in `useShell()`, you can add it.

packages/manager/core/application/src/ovh-routing.tsx
aboungnaseng-ovhcloud: Its different for dashboard to stay iso as angular, to have the services name in first path 

https://www.ovh.com/manager/#/dedicated/nasha/{service}

and not  :

https://www.ovh.com/manager/#/dedicated/nasha/dashboard/{service}

the package generouted didn't allow us to us parameter in first path so we need to made a fix

PR #9694, opened 2023-07-21
feat(web.database): remove deprecated calls

mlcdf:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | WEB-15524

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Hello,

This PR remove calls from deprecated API and replace them via the new API. That's pretty much it.

This is my first PR on the manager, so let me know if I did anything wrong.

### DIFF
--- a/packages/manager/apps/web/client/app/private-database/database/ram/update/private-database-database-ram-update.controller.js
+++ b/packages/manager/apps/web/client/app/private-database/database/ram/update/private-database-database-ram-update.controller.js
@@ -1,5 +1,4 @@
 import get from 'lodash/get';
-import remove from 'lodash/remove';
 
 angular.module('App').controller(
   'PrivateDatabaseChangeRamCtrl',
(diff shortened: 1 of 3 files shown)


darsene (changes requested):

darsene:
@mlcdf : when your pull request doesn't have any change about translation, do not check "Only FR translations have been updated" and strike it out

packages/manager/apps/web/client/app/private-database/private-database.service.js
```diff
@@ -485,62 +484,24 @@ export default class PrivateDatabase {
...
     return this.$http
-      .post(`${this.swsProxypassOrderPath}/${serviceName}/ram/${duration}`, {
-        ram,
+      .post(`${this.swsProxypassUpgradePath}/${serviceName}/${planCode}`, {
+        params: {
+          quantity: 1,
+        },
       })
-      .then((response) => response.data);
+      .then((response) => response.data.order);
```
anooparveti, line 504: **suggestion:** It is better to return only data from here as it will improve the reusability of the function if we need to consume any other info apart from `order` in other places.
```suggestion
      .then((response) => response.data);
```

darsene (approved):

packages/manager/apps/web/client/app/private-database/private-database.service.js
mlcdf: The returned payload is :
```
{
  "order": {
    ...
  },
  "operation": null
}
```

`operation` being null, it does see the point of returning it.

packages/manager/apps/web/client/app/private-database/database/ram/update/private-database-database-ram-update.controller.js
```diff
@@ -40,38 +39,42 @@ angular.module('App').controller(
...
+              this.database.infrastructure !== 'legacy' ||
+              capacity !== 2048 ||
+              +capacity !== +this.database.ram.value
+            ) {
+              
+                this.data.availableRam.push({
+                  capacity,
+                  planCode: e.planCode,
+            });
+            }
```
anooparveti, line 76: **nitpick:** Please use meaning variable name
```suggestion
          results.forEach((plan) => {
            const capacity = parseInt(plan.planCode.match(/[0-9]+/)?.[0], 10);

            if (
              this.database.infrastructure !== 'legacy' ||
              capacity !== 2048 ||
              +capacity !== +this.database.ram.value
            ) {
              
                this.data.availableRam.push({
                  capacity,
                  planCode: plan.planCode,
            });
            }
```

anooparveti (changes requested):

packages/manager/apps/web/client/app/private-database/database/ram/update/private-database-database-ram-update.controller.js
mlcdf: Fixed

darsene (approved):

packages/manager/apps/web/client/app/private-database/private-database.service.js
anooparveti: Yes, if we need to use `operation` from the response in other places, we cannot re-use this function in the service as it only returns the `order`. So I suggest to return the full response from here and filter for orders in the controller

anooparveti (approved):

PR #9955, opened 2023-09-06
feat(iam): improve policies display for selected actions

darsene:
ref: MANAGER-11345

| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-11345

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Auto-select and disable all actions covered by a custom action with a regex pattern (containing wildcard character)

### DIFF
--- a/packages/manager/modules/iam/src/components/actionSelect/ActionTrees.class.js
+++ b/packages/manager/modules/iam/src/components/actionSelect/ActionTrees.class.js
@@ -1,7 +1,8 @@
 /**
  * @typedef {{
  *   resourceType: string
- *   selected: boolean
+ *   selected: boolean,
+ *   embedded: boolean,
  *   value: string
  * }} Action
  *
@@ -44,7 +45,11 @@
  * }} RawAction
  */
 
-import { CUSTOM_RESOURCE_TYPE } from '../../iam.constants';
+import {
+  CUSTOM_ACTION_WILDCARD_PATTERN,
+  CUSTOM_RESOURCE_TYPE,
+  WILDCARD,
+} from '../../iam.constants';
 
 export default class ActionTrees extends Array {
   /**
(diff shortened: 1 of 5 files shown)


packages/manager/modules/iam/src/components/actionSelect/ActionTrees.class.js
```diff
@@ -95,14 +99,63 @@ export default class ActionTrees extends Array {
...
     return this.actions.find((action) => action.value === value);
   }
 
+  /**
+   * Find all action provided by the wildcard action value
+   * @param {string} value
+   * @returns {Action[]}
+   */
+  findAllActions(value) {
+    const wildcard = '*';
```
anooparveti, line 146: **nitpick:** `*` is used a lot of times. Define it as a constant

packages/manager/modules/iam/src/components/createPolicy/createPolicy.controller.js
```diff
@@ -389,6 +389,9 @@ export default class CreatePolicyController {
         allow: this.model.actions.isWildcardActive
           ? [{ action: '*' }]
           : this.model.actions.selection
+              // actions tagged as embedded are covered by custom actions
+              // there is no need to persist them
+              .filter(({ embedded }) => !embedded)
```
anooparveti, line 394: **nitpick(perf):** There is a `filter` running on line 396. Re-use the existing filter to avoid multiple looping .

darsene: The second `filter` is only used to remove duplicated actions but at this step, the property `embedded` doesn't exist anymore. Indeed, the array is reduced to object as `[{action: 'actionValue'}]` therefore I can't use that `filter`

packages/manager/modules/iam/src/components/actionSelect/ActionTrees.class.js
```diff
@@ -95,14 +100,62 @@ export default class ActionTrees extends Array {
     return { action: newAction, created: true };
   }
 
+  /**
+   * Check all actions matching to a regex action
+   * If some actions exist, they're tagged as embedded
+   * @param {string} action
+   * @param {Boolean} selected
```
BrunoMarquesOVH, line 107: Comments says that the function takes a selected param, but it doesn't seem so. Maybe remove it?

darsene: good catch! Fixed

BrunoMarquesOVH (approved):

anooparveti (approved):

anooparveti (approved):

BrunoMarquesOVH (approved):

PR #10040, opened 2023-09-27
fix(web): fix domain hosting order without payment method

darsene:
ref: DTRSD-121236

| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-121236

- [x] Commits are signed-off
- ~Only FR translations have been updated~
- [x] Lint has passed locally
- ~Breaking change is mentioned in relevant commits~

## Description

Fix domain hosting order when the user doesn't have a payment method set by default

### DIFF
--- a/packages/manager/apps/web/client/app/domain/webhosting/order/steps/payment/domain-webhosting-order-steps-payment.html
+++ b/packages/manager/apps/web/client/app/domain/webhosting/order/steps/payment/domain-webhosting-order-steps-payment.html
@@ -2,7 +2,7 @@
     data-header="{{:: 'domain_webhosting_order_payment_step_title' | translate }}"
     data-on-focus="$ctrl.getCheckout()"
     data-loading="$ctrl.stepper.loadingCheckout"
-    data-valid="$ctrl.defaultPaymentMethod != null && $ctrl.agreeContracts"
+    data-valid="$ctrl.agreeContracts"
     data-prevent-next="true"
     data-on-cancel="$ctrl.goBackToDashboard()"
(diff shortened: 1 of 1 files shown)


JacquesLarique (approved):

packages/manager/apps/web/client/app/domain/webhosting/order/steps/payment/domain-webhosting-order-steps-payment.html
```diff
@@ -2,7 +2,7 @@
     data-header="{{:: 'domain_webhosting_order_payment_step_title' | translate }}"
     data-on-focus="$ctrl.getCheckout()"
     data-loading="$ctrl.stepper.loadingCheckout"
-    data-valid="$ctrl.defaultPaymentMethod != null && $ctrl.agreeContracts"
+    data-valid="$ctrl.agreeContracts"
```
anooparveti, line 5: **question:** Do we have validation from PO to remove the check if default payment method does not exist?

ghyenne (approved):

packages/manager/apps/web/client/app/domain/webhosting/order/steps/payment/domain-webhosting-order-steps-payment.html
darsene: I didn't ask but the current behavior blocks users if they want to pay with another payment method or if they have not.
AS UI allows both choices, the button has to do the same.

PR #10102, opened 2023-10-10
feat(generator): useFeatureAvailability hook

chipp972:
ref: MANAGER-12296

| Question         | Answer
| ---------------- | ---
| Branch?          | `dev-transformation` 
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-12296

- [x] Commits are signed-off
- [x] Lint has passed locally
- [ ] ~~Standalone app was ran and tested locally~~

## Related

Create a hook `useFeatureAvailability` in core package
Will be used to determine if a feature is available for the current user

### DIFF
--- a/packages/manager/core/application/src/hooks/useFeatureAvailability.ts
+++ b/packages/manager/core/application/src/hooks/useFeatureAvailability.ts
@@ -1,0 +1,47 @@
+import { useQuery } from '@tanstack/react-query';
+import { apiClient } from '@ovh-ux/manager-core-api';
+
+export const featureAvailabilityBaseQueryKey = 'feature-availability-';
+
+export type UseFeatureAvailabilityParams = {
+  parent: string;
+  featureList?: string[];
+};
+
+const getFeaturesListToCheck = ({
+  parent,
+  featureList = [],
+}: UseFeatureAvailabilityParams) =>
+  [parent].concat(featureList.map((name) => `${parent}:${name}`));
+
+/**
+ * @example
+ * const {data, error, loading} = useFeatureAvailability(
+ *   {parent: 'dedicated-nasha', featureList: ['feature1', 'feature2']},
+ *   {parent: 'app2'},
+ *   {parent: 'app3', featureList: ['feat1']}
+ * );
+ * const isAppAvailable = data.['dedicated-nasha'];
+ * const isFeature1Available = data.feature1;
+ * const isFeature2Available = data.feature2;
+ * const isApp2Available = data.app2;
+ * const isApp3Available = data.app3;
+ * const isApp3Feature1Available = data.feat1;
+ */
+export const useFeatureAvailability = (
+  ...featureList: UseFeatureAvailabilityParams[]
+) => {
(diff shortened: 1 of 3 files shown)


Kranysys (approved):

packages/manager/core/application/src/hooks/useFeatureAvailability.ts
```diff
@@ -0,0 +1,35 @@
...
+    const result = await apiClient.aapi.get(
+      `/feature/${featuresListToCheck.join(',')}/availability`,
+    );
+    return result.data || {};
+  };
+
+  return useQuery<Record<string, boolean>>(
+    [`${featureAvailabilityBaseQueryKey}-${featuresListToCheck.join('-')}`],
+    fetchFeatureAvailabilityData,
+  );
```
anooparveti, line 34: **nitpick:** We need not limit the feature-flipping to query related to one particular app. A query can contain features from multiple apps.

anooparveti (changes requested):

packages/manager/core/application/src/hooks/useFeatureAvailability.ts
chipp972: Yes it is currently a possibility
But, except in the container, is it a good idea to be able to add coupling between different apps ?
I think it doesn't make sense that the availability of a feature in a particular app impacts another app.

Additionally, the calls are cached in react-query so, even if the container has to make multiple calls, the apps using the hook will be able to use the result directly, instead of doing a feature availability call again when opening the app

anooparveti: @chipp972, I think this is the feature we can keep because we can have multiple products depending on each other (especially on the PCI, for example, Instance/Floating IP/Gateway/Private Network/Load Balancer are related to each other and we may be in a situation to use this.) More over API doesn't restrict us to have only `json` for each product. So just for Baremetal server, you can have multiple json grouped based on any particular features.

chipp972: Ok then I put the possibility to passe multiple objects to create one request

packages/manager/core/application/src/hooks/useFeatureAvailability.ts
```diff
@@ -0,0 +1,49 @@
...
+  ...appFeatureList: UseFeatureAvailabilityParams[]
+) => {
+  const featuresListToCheck = appFeatureList.flatMap(getFeaturesListToCheck);
+
+  const fetchFeatureAvailabilityData = async () => {
+    const result = await apiClient.aapi.get(
+      `/feature/${featuresListToCheck.join(',')}/availability`,
+    );
+    return result.data || {};
+  };
```
anooparveti, line 43: Using the property `applicationName` can be misleading because we have other features like `advices` `liveChat` which are not full-fledged applications but rather a parent feature.

chipp972: renamed to `parent`

PR #10104, opened 2023-10-10
feat(dedicated): reload properly manager after language change

BrunoMarquesOVH:
ref: MANAGER-11725

| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-11725

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

reload properly manager after language change

### DIFF
--- a/packages/manager/apps/dedicated/client/app/account/user/components/newAccountForm/new-account-form-controller.js
+++ b/packages/manager/apps/dedicated/client/app/account/user/components/newAccountForm/new-account-form-controller.js
@@ -407,7 +407,7 @@ export default class NewAccountFormController {
           this.originalManagerLanguage !== this.model.managerLanguage
         ) {
           this.coreConfig.setUserLocale(this.model.managerLanguage);
-          window.location.reload();
+          window.parent.location.reload();
         } else if (this.onSubmit) {
           this.$location.search('isUpdated', true);
           this.onSubmit();


packages/manager/apps/dedicated/client/app/account/user/components/newAccountForm/new-account-form-controller.js
```diff
@@ -407,7 +407,7 @@ export default class NewAccountFormController {
           this.originalManagerLanguage !== this.model.managerLanguage
         ) {
           this.coreConfig.setUserLocale(this.model.managerLanguage);
-          window.location.reload();
+          window.parent.location.reload();
```
anooparveti, line 410: We need to add an utility method in the container/shell to make the reload. 
Current implementation will not be appropriate if we run the dedicated app without container.

anooparveti, line 410: We need to add an utility method in the container/shell to make the reload. 
Current implementation will not be appropriate if we run the dedicated app without container.

anooparveti (changes requested):

packages/manager/apps/dedicated/client/app/account/user/components/newAccountForm/new-account-form-controller.js
BrunoMarquesOVH: Hello Anoop,
I did this because "window.parent" returns the same window object if you are in an iFrame or not. So it will work the same, with or without the container.

ghyenne (approved):

PR #10108, opened 2023-10-11
feat(web): add DKIM configuration in exchange domain configuration

darsene:
ref: MANAGER-12205

| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/dkim-b2`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-12205

- [x] Commits are signed-off
- [x] Lint has passed locally
~Breaking change is mentioned in relevant commits~

## Description

Add configuration and activation of DKIM during the configuration of an OVH exchange domain


packages/manager/modules/exchange/src/domain/add/domain-add.controller.js
```diff
@@ -156,6 +157,9 @@ export default class ExchangeAddDomainController {
 
       delete this.model.attachOrganization2010;
     }
+    const isOvhDomain = this.model.domainType === this.OVH_DOMAIN;
+    this.model.autoEnableDKIM = this.model.autoEnableDKIM && isOvhDomain;
```
anooparveti, line 161: **question:** What is the expected behavior here?
If `this.model.autoEnableDKIM` is false, the expression always evaluates to false. and do we need to consider `isOvhDomain` only is `this.model.autoEnableDKIM` is true?

anooparveti (changes requested):

packages/manager/modules/exchange/src/domain/add/domain-add.html
```diff
@@ -475,6 +483,15 @@ <h3 data-translate="exchange_tab_TASKS_ADD_DOMAIN"></h3>
                     data-ng-bind="('exchange_tab_domain_add_step3_configuration_' + ctrl.model.srvParam) | translate"
                 ></dd>
 
+                <dt
+                    data-ng-if="ctrl.model.domainType !== ctrl.nonOvhDomain"
```
ghyenne, line 487: You have OVH_DOMAIN, nonOvhDomain and isOvhDomain. Maybe we can reuse or standardized to avoid getting lost between affirmation and double negation.

packages/manager/modules/exchange/src/domain/add/domain-add.controller.js
darsene: `this.model.autoEnableDKIM `can be `true` or `false` if  `isOvhDomain=true` and `this.model.autoEnableDKIM` has to be `false` if `sOvhDomain=false`.
Because of the wizzard navigation, model is cleaned just before validation

packages/manager/modules/exchange/src/domain/add/domain-add.html
darsene: `isOvhDomain` doesn't exist actually that's why the code looks not unfiorm. But Ok I'll do it

darsene: fixed !

ghyenne (approved):

anooparveti (approved):

PR #10139, opened 2023-10-17
feat(iam): enhancing actions filtering

BrunoMarquesOVH:
ref: MANAGER-11332

| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-11332

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Enhancing actions filtering while creating IAM policy

## Related

- [x] CMT-5114


packages/manager/modules/iam/src/components/actionSelect/actionSelect.controller.js
```diff
@@ -367,19 +367,56 @@ export default class ActionSelectController {
...
+    if (selectionCount && searchQuery) {
+      if (searchCount <= 1) {
+        return selectionCount === 1
+          ? `${prefix}_selection_one_result_one`
+          : `${prefix}_selection_many_result_one`;
+      }
+      return selectionCount === 1
+        ? `${prefix}_selection_one_result_many`
+        : `${prefix}_selection_many_result_many`;
+    }
```
darsene, line 386: Can be factorized as you return the same keys

packages/manager/modules/iam/src/components/actionSelect/actionSelect.controller.js
```diff
@@ -367,19 +367,56 @@ export default class ActionSelectController {
...
     if (selectionCount) {
       return selectionCount === 1
         ? `${prefix}_selection_one`
         : `${prefix}_selection_many`;
     }
+    if (searchQuery) {
+      return searchCount <= 1
+        ? `${prefix}_result_one`
+        : `${prefix}_result_many`;
+    }
```
darsene, line 396: Same here, can be factorized

packages/manager/modules/iam/src/components/actionSelect/actionSelect.controller.js
```diff
@@ -367,19 +367,56 @@ export default class ActionSelectController {
...
     return actions.length === 1
       ? `${prefix}_no_selection_one`
       : `${prefix}_no_selection_many`;
   }
 
+  filterActions(actionTree) {
+    angular.forEach(actionTree.categories, (category) => {
```
darsene, line 403:
```suggestion
    actionTree.categories.forEach( (category) => {
```
No need to use angular functions

packages/manager/modules/iam/src/components/actionSelect/actionSelect.controller.js
```diff
@@ -367,19 +367,56 @@ export default class ActionSelectController {
...
+      category.filteredActions = category?.actions?.filter(
+        (action) =>
+          action?.value
+            .toLowerCase()
+            .indexOf(actionTree.searchQuery.toLowerCase()) > -1,
+      );
+    });
+    this.actionTrees[
+      findIndex(
+        this.actionTrees,
```
darsene, line 414:
```suggestion
           this.actionTrees.findIndex
```
No need of `lodash here`

packages/manager/modules/iam/src/components/actionSelect/actionSelect.controller.js
```diff
@@ -388,4 +425,20 @@ export default class ActionSelectController {
...
+   * @param {string} searchQuery
+   * @returns {Number}
+   */
+  static countActionsMatchingSearch(actions, searchQuery) {
+    if (!actions || !searchQuery) {
+      return 0;
+    }
+    return actions.filter(
+      (action) =>
+        action?.value.toLowerCase().indexOf(searchQuery.toLowerCase()) > -1,
```
darsene, line 441:
```suggestion
        action.value.toLowerCase().indexOf(searchQueryLowerCase) > -1,
```
Create a constant `const searchQueryLowerCase = searchQuery.toLowerCase()` to avoid repeating the operation in the loop

packages/manager/modules/iam/src/components/actionSelect/actionSelect.template.html
```diff
@@ -102,7 +103,7 @@
         <!-- <NoCategoryActions> -->
         <oui-checkbox
             class="ml-4"
-            data-ng-repeat="action in actionTree.actions track by action.value"
+            data-ng-repeat="action in actionTree.filteredActions track by action.value"
```
darsene, line 106: Did you check that all is still working for embedded actions ? -> `action.embedded = true`

packages/manager/modules/iam/src/components/actionSelect/translations/Messages_fr_FR.json
```diff
@@ -3,6 +3,12 @@
   "iam_action_select_action_search_label": "Rechercher une action",
   "iam_action_select_all_actions_description": "Toutes les actions existantes et à venir sur tous les types de produits seront autorisées",
   "iam_action_select_all_actions_label": "Autoriser toutes les actions",
+  "iam_action_select_category_count_selection_many_result_one": "({{ selected }} actions sélectionnées | {{ result }} résultat / {{ total }})",
+  "iam_action_select_category_count_selection_many_result_many": "({{ selected }} actions sélectionnées | {{ result }} résultats / {{ total }})",
+  "iam_action_select_category_count_selection_one_result_one": "({{ selected }} action sélectionnée | {{ result }} résultat / {{ total }})",
+  "iam_action_select_category_count_selection_one_result_many": "({{ selected }} action sélectionnée | {{ result }} résultats / {{ total }})",
+  "iam_action_select_category_count_result_many": "({{ result }} résultats / {{ total }})",
+  "iam_action_select_category_count_result_one": "({{ result }} résultat / {{ total }})",
```
darsene, line 11: You should consider to use` data-ng-pluralize` to simplify plural management

darsene (changes requested):

packages/manager/modules/iam/src/components/actionSelect/actionSelect.controller.js
```diff
@@ -367,19 +367,56 @@ export default class ActionSelectController {
...
+        : `${prefix}_result_many`;
+    }
     return actions.length === 1
       ? `${prefix}_no_selection_one`
       : `${prefix}_no_selection_many`;
   }
 
+  filterActions(actionTree) {
+    angular.forEach(actionTree.categories, (category) => {
+      // eslint-disable-next-line no-param-reassign
```
anooparveti, line 404: Use `Array.prototype.map` but dont disable the ESLint rules

packages/manager/modules/iam/src/components/actionSelect/ActionTrees.class.js
```diff
@@ -363,13 +364,15 @@ export default class ActionTrees extends Array {
...
+      }));
       Object.assign(category, {
-        actions: actions.map((action) => ({
-          ...action,
-          id: `${actionTree.value}_${category.value}_${action.value}`
-            .replace(/[^\w]/g, '-')
-            .toLocaleLowerCase(),
-        })),
+        actions: currentActions,
+        filteredActions: currentActions,
```
anooparveti, line 375: **nitpick:** Why maintain the same array twice in the same object?

anooparveti (changes requested):

packages/manager/modules/iam/src/components/actionSelect/actionSelect.controller.js
BrunoMarquesOVH: Updated

packages/manager/modules/iam/src/components/actionSelect/actionSelect.controller.js
BrunoMarquesOVH: Updated

packages/manager/modules/iam/src/components/actionSelect/actionSelect.controller.js
BrunoMarquesOVH: Done

packages/manager/modules/iam/src/components/actionSelect/actionSelect.controller.js
BrunoMarquesOVH: Done

packages/manager/modules/iam/src/components/actionSelect/actionSelect.controller.js
BrunoMarquesOVH: Done

packages/manager/modules/iam/src/components/actionSelect/actionSelect.controller.js
BrunoMarquesOVH: Hello, Array.map does not solve the issue. 
The problem is that this rule is used to prevent unintentionnaly mutate a parameter. But this is what It is meant for. I want for each category to update the array of objects filtered according to filter, which seems normal to me. As stated here https://stackoverflow.com/questions/35637770/how-to-avoid-no-param-reassign-when-setting-a-property-on-a-dom-object, a solution could be to use an intermediary variable but I is not what I want, since I want to update the original value. The only other viable solution is to disable the rule, which is use pretty often in the codebase. Unless you see another solution which I don't.

packages/manager/modules/iam/src/components/actionSelect/ActionTrees.class.js
BrunoMarquesOVH: Hello, this is just initialization of the value. Whenever a user uses a filter, it is updated. I need to have an intermediate value because I need a reference with all actions so I can reset to original state actions if user deletes the filter, or updates it. Seems ok to me.

packages/manager/modules/iam/src/components/actionSelect/translations/Messages_fr_FR.json
BrunoMarquesOVH: Hello, from what I see in documentation of ng-pluralize (https://docs.angularjs.org/api/ng/directive/ngPluralize), it only supports one "count" data. But in this use case, we need two mixed count: we can have both selected items and filtered items. I don't see how to simply this code with ng-pluralize so. And I only extended the previous code that was already using this syntax, which by the way, in that case, would have make sense. But with two count variables, I don't see how to overcome this with ng-pluralize.

packages/manager/modules/iam/src/components/actionSelect/actionSelect.template.html
BrunoMarquesOVH: Yes should be ok

darsene (approved):

packages/manager/modules/iam/src/components/actionSelect/actionSelect.controller.js
anooparveti: Please use a temporary variable. Otherwise, you can accept an array as a param and return a new modified array and assign it to whatever variable you want.

BrunoMarquesOVH: Done

darsene (dismissed):

anooparveti (dismissed):

PR #10172, opened 2023-10-24
Feat/manager 12441 Query URL

Kranysys:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/EPIC-MANAGER-11852`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-12441

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

When user add filters and input text update url (params: q, univers, categories)


packages/manager/apps/catalog/src/pages/index.tsx
```diff
@@ -17,6 +18,8 @@ import Errors from '@/components/Error/Errors';
 
 export default function CatalogRevamp() {
```
chipp972, line 19:
```suggestion
export default function Catalog() {
```

packages/manager/apps/catalog/src/components/Filters/Filters.tsx
```diff
@@ -31,12 +32,28 @@ const Filters: React.FC<FiltersProps> = ({
...
+    const categoriesURL = params.get('categories');
+    if (categoriesURL) {
+      setSelectedCategories(categoriesURL.split(','));
+      setParentSelectedCategories(categoriesURL.split(','));
+    }
+    if (universesURL) {
+      setSelectedUniverses(universesURL.split(','));
+      setParentSelectedUniverses(universesURL.split(','));
+    }
+  }, [location.search, products]);
```
chipp972, line 55: why put products in the dependency array ?

Kranysys: it's an asynchronous effect, we get `location.search` before the 2API request that load the products, adding `products` seems to fix this issue. That's why.

chipp972: You didn't describe the issue actually
What is the issue ?

chipp972: If the variable are irrelevant while the products are not fetched, there should be a condition if(products.length > 0) for example

packages/manager/apps/catalog/src/pages/index.tsx
```diff
@@ -27,6 +30,15 @@ export default function CatalogRevamp() {
     searchText,
   });
 
+  useEffect(() => {
+    const params = new URLSearchParams();
+    if (searchText) params.append('q', searchText);
+    if (categories.length > 0)
+      params.append('categories', categories.join(','));
+    if (universes.length > 0) params.append('universes', universes.join(','));
+    navigate({ search: params.toString() });
```
chipp972, line 39: I feel that we should put this in a function and the content of the useEffect of the Filters.tsx file because, if for any reason, we change the parameter or we add new parameters, it will be easier to refactor if the functions are next to each other.
They will be easy to unit test, also.

packages/manager/apps/catalog/src/components/Filters/Filters.tsx
```diff
@@ -31,12 +32,28 @@ const Filters: React.FC<FiltersProps> = ({
...
+    const universesURL = params.get('universes');
+    const categoriesURL = params.get('categories');
+    if (categoriesURL) {
+      setSelectedCategories(categoriesURL.split(','));
+      setParentSelectedCategories(categoriesURL.split(','));
+    }
+    if (universesURL) {
+      setSelectedUniverses(universesURL.split(','));
+      setParentSelectedUniverses(universesURL.split(','));
+    }
```
chipp972, line 54: I feel that we should put this in a function and the content of the useEffect of the index.tsx file because, if for any reason, we change the parameter or we add new parameters, it will be easier to refactor if the functions are next to each other.
They will be easy to unit test, also.

You can call it `getFilterParamsFromUrl` and return something like `{universes: string[]' categories: string[]}`

packages/manager/apps/catalog/src/pages/index.tsx
Kranysys: Done

packages/manager/apps/catalog/src/components/Filters/Filters.tsx
Kranysys: I did many tests, products variable is mandatory in this useEffect as it is important that all the products are fetched to apply the filters from params from URL. I added your recommendation of `if(products.length > 0)` in filters.tsx and also in index.tsx is it important that all the products are fetched before we apply the params from the checkboxes and input.

packages/manager/apps/catalog/src/pages/index.tsx
Kranysys: Done, getSearchUrlFromFilterParams created ✅

packages/manager/apps/catalog/src/components/Filters/Filters.tsx
Kranysys: Done, getFilterParamsFromUrl created ✅

chipp972: Ok nice
Can you move it to the same file as the other function that create url from params ?

packages/manager/apps/catalog/src/pages/index.tsx
chipp972: Ok nice
Can you move it to the same file as the other function that extracts params from url ?
Like an helper or utils file

PR #10231, opened 2023-11-03
fix(sign-up): not siret yet path : tracking implementation

dectotam:
ref: MANAGER-12543

| Question         | Answer
| ---------------- | ---
| Branch?          | feat/signup-siret
| Bug fix?         | no
| New feature?     | no
| Breaking change? |no
| Tickets          | Feat MANAGER-12543

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Tracking implementation for the "Not a SIRET yet" path in signup app.

### DIFF
--- a/packages/manager/modules/sign-up/src/form/activity/activity.html
+++ b/packages/manager/modules/sign-up/src/form/activity/activity.html
@@ -143,6 +143,7 @@
             data-required
             data-label="{{:: 'sign_up_activity_corporation_created_title' | translate }}"
             data-values="[true]"
+            data-ng-click="$ctrl.trackClick($ctrl.COMPANY_CREATED_PREFIX)"
         >
             <oui-select-picker-description>
                 <span
@@ -159,6 +160,7 @@
             data-on-change="$ctrl.resetCorporationData()"
             data-label="{{:: 'sign_up_activity_corporation_not_created_title' | translate }}"
             data-values="[false]"
+            data-ng-click="$ctrl.trackClick($ctrl.COMPANY_NOT_CREATED_PREFIX)"
         >
             <oui-select-picker-description>
                 <span
--- a/packages/manager/modules/sign-up/src/form/activity/constants.js
+++ b/packages/manager/modules/sign-up/src/form/activity/constants.js
@@ -67,6 +67,12 @@ export const COUNTRIES_VAT_LABEL = {
   MA: 'ICE',
 };
 
+const SIRET_TRACKING_PREFIX = 'accountcreation::company-search';
+export const COMPANY_CREATED_PREFIX = `${SIRET_TRACKING_PREFIX}::company-created`;
+export const COMPANY_NOT_CREATED_PREFIX = `${SIRET_TRACKING_PREFIX}::company-not-created`;
+
 export default {
+  COMPANY_CREATED_PREFIX,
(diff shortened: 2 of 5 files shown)


packages/manager/modules/sign-up/src/form/activity/constants.js
```diff
@@ -67,6 +67,12 @@ export const COUNTRIES_VAT_LABEL = {
   MA: 'ICE',
 };
 
+export const SIRET_TRACKING_PREFIX = 'accountcreation::company-search';
+export const COMPANY_CREATED_PREFIX = `${SIRET_TRACKING_PREFIX}::company-created`;
+export const COMPANY_NOT_CREATED_PREFIX = `${SIRET_TRACKING_PREFIX}::company-not-created`;
+
 export default {
+  COMPANY_CREATED_PREFIX,
+  COMPANY_NOT_CREATED_PREFIX,
```
anooparveti, line 76:
```suggestion
  SIRET_TRACKING_PREFIX,
  COMPANY_CREATED_PREFIX,
  COMPANY_NOT_CREATED_PREFIX,
```

packages/manager/modules/sign-up/src/form/activity/activity.html
```diff
@@ -159,6 +160,7 @@
             data-on-change="$ctrl.resetCorporationData()"
             data-label="{{:: 'sign_up_activity_corporation_not_created_title' | translate }}"
             data-values="[false]"
+            data-ng-click="$ctrl.trackClick($ctrl.COMPANY_NOT_CREATED_PREFIX)"
```
darsene, line 163: you can use directives `data-track-on`, `data-track-type` and `data-track-name` to avoid creating a specific function in the controller

packages/manager/modules/sign-up/src/form/activity/controller.js
```diff
@@ -61,4 +68,11 @@ export default class OvhSignUpActivityCtrl {
       this.signUpFormCtrl.model.country === 'FR'
     );
   }
+
+  trackClick(name) {
+    this.atInternet.trackClick({
+      name,
+      type: 'action',
+    });
```
darsene, line 76: Should be remove if you use tracking directives

packages/manager/modules/sign-up/src/form/activity/constants.js
dectotam: Why should SIRET_TRACKING_PREFIX be exported ? 
It is only used in the constants file to build COMPANY_CREATED_PREFIX and COMPANY_NOT_CREATED_PREFIX

packages/manager/modules/sign-up/src/form/activity/activity.html
dectotam: Thanks, I tried to do so but I got an error because it implies that several directives (ouiSelectPicker and trackOn) on the custom component ask for an isolated scope. In this case we have to use the function.

packages/manager/modules/sign-up/src/form/activity/constants.js
anooparveti: You're exporting it during the definition as well.

`export const SIRET_TRACKING_PREFIX ...`

anooparveti (approved):

PR #10309, opened 2023-11-15
feat(ui-manager): Create common react ui folder for manager

lionel95200x:
| Question         | Answer
| ---------------- | ---
| Branch?          | `dev-transformation` 
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #... 

- [x] Commits are signed-off
- [x] Lint has passed locally


packages/react-super-components/playwright.config.ts
```diff
@@ -0,0 +1,20 @@
+import { defineConfig } from '@playwright/test'
```
chipp972, line 1: We can remove this file

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
```diff
@@ -0,0 +1,127 @@
...
+  category: string;
+  tileTitle: string;
+  tileDescription: string;
+  dataTracking?: string;
+  locale?: any;
+  badges?: Badge[];
+  footer?: React.ReactElement;
+}
+
+const ScTile = ({
```
chipp972, line 32:
```suggestion
export const ScTile = ({
```

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
```diff
@@ -0,0 +1,127 @@
...
+import { OsdsText } from '@ovhcloud/ods-components/text/react';
+import { OsdsChip } from '@ovhcloud/ods-components/chip/react';
+import { OsdsTile } from '@ovhcloud/ods-components/tile/react';
+import { OsdsLink } from '@ovhcloud/ods-components/link/react';
+import { OsdsIcon } from '@ovhcloud/ods-components/icon/react';
+
+import { ODS_ICON_NAME, ODS_ICON_SIZE } from '@ovhcloud/ods-components/icon';
+import { ODS_TEXT_LEVEL, ODS_TEXT_SIZE } from '@ovhcloud/ods-components/text';
+
+interface Badge {
```
chipp972, line 13:
```suggestion
export interface Badge {
```

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
```diff
@@ -0,0 +1,127 @@
...
+            color={ODS_THEME_COLOR_INTENT.primary}
+          >
+            {category}
+            <span className="inline-flex ml-3">
+              {badges?.map((b) => (
+                <OsdsChip
+                  className="mr-1"
+                  key={b.text}
+                  color={b.color}
+                  size="sm"
```
chipp972, line 76: Use ODS variables

package.json
```diff
@@ -11,6 +11,8 @@
     "packages/manager/core/*",
     "packages/manager/modules/*",
     "packages/manager/tools/*",
+    "packages/react-super-components",
+    "packages/react-super-components/*",
```
chipp972, line 15: there is no workspaces inside the folder

lerna.json
```diff
@@ -7,6 +7,8 @@
     "packages/manager/core/*",
     "packages/manager/modules/*",
     "packages/manager/tools/*",
+    "packages/react-super-components",
+    "packages/react-super-components/*",
```
chipp972, line 11: there is no workspace inside the folder

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
```diff
@@ -0,0 +1,127 @@
...
+  isExternalHref?: boolean;
+  imgSrc?: string;
+  imgAlt?: string;
+  category: string;
+  tileTitle: string;
+  tileDescription: string;
+  dataTracking?: string;
+  locale?: any;
+  badges?: Badge[];
+  footer?: React.ReactElement;
```
chipp972, line 29: use children instead

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
lionel95200x: Why export it's only use in this file ?

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
```diff
@@ -0,0 +1,127 @@
...
+  text: string;
+  color: ODS_THEME_COLOR_INTENT;
+}
+
+export interface ScTileProps {
+  href?: string;
+  isExternalHref?: boolean;
+  imgSrc?: string;
+  imgAlt?: string;
+  category: string;
```
chipp972, line 22: category should be inside text also

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
chipp972: because when you use the component, you want to know the types of its props :

```
const badges: Badge[] = [{text: 'blabla', color: ODS_THEME_COLOR_INTENT.primary}]
<ScTile badges={badges} {...} />
```

And you should use `type` instead of `interface` since we don't use OOP in React. `interface`s won't be `implemented` by `class`es.

lionel95200x: For me we should not export interface that we will not use in other file to avoid exporting useless class
Even if it's not export the following code will trigger an error `text is missing in badges`

```
const badges = [{textError: 'blabla', color: ODS_THEME_COLOR_INTENT.primary}]

<ScTile badges={badges} {...} />

```
And for the interface VS type, following the exemple of React doc `https://react.dev/learn/typescript` they use interface instead of class

chipp972: "For me we should not export interface that we will not use in other file to avoid exporting useless class"
It's a good way to think when coding an app

But when it's a library we should export everything that is exposed to the user of the lib. If we build dynamically the list of badges and we create a function for this, it's better to be able to import the type directly instead of duplicating it in the code that uses the library :

```
const createBadgeList = (someParams, someParams2): Badge[] => {...};
```

chipp972: And concerning the debate about type vs interface : 
https://blog.logrocket.com/types-vs-interfaces-typescript/

I think we should use types because declaration merging can be dangerous in our case :
If you declare and export multiple times the same interface, they will merge and you will have a dynamically created interface (but you may not know where the other interface comes from and it can create conflicts). If you do the same with types, Typescript gives an error.

Additionally, types have some interesting features that interfaces don't have, as explained in the blog post.

packages/react-super-components/src/components/msc-tile/msc-tile.stories.tsx
```diff
@@ -0,0 +1,109 @@
...
+Primary.args = {
+  ...defaultProps,
+}
+
+export const WithFooter = Template.bind({})
+WithFooter.args = {
+  ...defaultProps,
+  footer: (
+    <button color="primary" className="mb-1">
+      Commander
```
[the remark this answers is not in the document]
lionel95200x: It's on storybook `stories` file

yarn-error.log
frenautvh, line 1: to remove

packages/react-super-components/yarn.lock
frenautvh, line 1: you shouldn't have a yarn.lock here

packages/react-super-components/src/translation/translations.ts
frenautvh, line 1: it looks like you are implementing a custom translation system, why not relying on i18next for this ?

packages/react-super-components/src/translation/translations.ts
```diff
@@ -0,0 +1,37 @@
...
+export type Translations = {
+  see_more_label: string;
+};
+
+/**
+ * Import dynamically the right translation file
+ */
+export async function getTranslations(locale?: Locale): Promise<Translations> {
+  try {
+    switch (locale) {
```
aboungnaseng-ovhcloud, line 12: `return await import(`./translations/Messages_${locale?.replace('-', '_')}.json`);`

aboungnaseng-ovhcloud (approved):

packages/react-super-components/src/translation/translations.ts
aboungnaseng-ovhcloud, line 12:
```suggestion
    return await import(./translations/Messages_${locale?.replace('-', '_')}.json);
```

aboungnaseng-ovhcloud (approved):

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
```diff
@@ -0,0 +1,130 @@
...
+            className="block mb-4"
+            level={ODS_TEXT_LEVEL.body}
+            size={ODS_TEXT_SIZE._400}
+            color={ODS_THEME_COLOR_INTENT.default}
+          >
+            {description}
+          </OsdsText>
+          <OsdsLink
+            tabIndex={hasFooterContent ? 0 : -1}
+            data-tracking={dataTracking}
```
cbourgois, line 105:
```suggestion
```

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
```diff
@@ -0,0 +1,130 @@
...
+export interface ScTileProps {
+  href?: string;
+  isExternalHref?: boolean;
+  imgSrc?: string;
+  imgAlt?: string;
+  texts: {
+    title: string;
+    description: string;
+    category: string;
+  };
```
cbourgois, line 29: Difficult to understand what is the proper way to structure props: 

It could be `imgSrc` & `imgAlt`, why not `img: { src, alt }` ? 
it could be `texts: { title, description, category }` but why not `title`, `description` & `category` ?

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
```diff
@@ -0,0 +1,130 @@
...
+  dataTracking?: string;
+  badges?: Badge[];
+  children?: React.ReactElement;
```
chipp972, line 32: There is a `React.PropsWithChildren` for this

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
```diff
@@ -0,0 +1,130 @@
...
+  badges,
+  children,
+  texts,
+}: ScTileProps) => {
+  const { title, description, category } = texts;
+  const hasFooterContent = !!children;
+  const { t } = useDynamicTranslation('tile');
+
+  return (
+    <div className="tile-container">
```
chipp972, line 50: Actually, it should be a link
There is a condition in the stencil code to determine if it is a link or not based on the footerContent. It's because, from an accessibility standpoint, there should not be actionnable elements inside actionnable elements (link and buttons inside a link).
But the default behavior is a clickable tile that action the link.

packages/react-super-components/src/translation/translations.ts
frenautvh: seen together since you are using i18next addResourceBundle it's fine

packages/react-super-components/src/translation/translations.ts
```diff
@@ -0,0 +1,17 @@
+import { Locale } from '@ovhcloud/msc-utils';
+
+export type Translations = {
+  see_more_label: string;
+};
+
+export async function getTranslations(locale?: Locale): Promise<Translations> {
+  try {
+    return await import(
+      `./translations/Messages_${locale?.replace('-', '_')}.json`
```
chipp972, line 10: You can't make a dynamic import with a variable inside it and you don't need a getTranslations actually

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
lionel95200x: Yes good catch i decide to split also `img: { src, alt }` for me it's better to group props by logic to avoid a component with 30 props, but i am open to discussion if you prefer another way

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
lionel95200x: No `PropsWithChildren` return an object with all props { ... props, children } and children also 
[Link here
](https://blog.logrocket.com/react-children-prop-typescript/)

packages/react-super-components/src/components/msc-tile/msc-tile.tsx
chipp972: We don't track links usually ?

packages/ui-manager/src/translation/useDynamicTranslation.ts
```diff
@@ -0,0 +1,23 @@
...
+import getTranslations from './translations';
+import { useTranslation } from 'react-i18next';
+import i18next from 'i18next';
+import { Locale } from '@ovhcloud/msc-utils';
+
+export const useDynamicTranslation = (namespace: string) => {
+  const { t, i18n } = useTranslation(namespace);
+
+  useEffect(() => {
+    const loadTrad = async () => {
```
frenautvh, line 11: if the translation file doesn't exist you should load the default fallback language translation file (so i18next can fallback to the default language)

chipp972 (approved):

packages/ui-manager/src/translation/useDynamicTranslation.ts
lionel95200x: @frenautvh we already have in the app a default fallback language as 
```
          fallbackLng: 'fr_FR',
````

In `manager/packages/manager/apps/container/src/index.tsx`

Comment (frenautvh) [reply packages/ui-manager/src/translation/useDynamicTranslation.ts:11]: seen together, there shouldn't have missing translations for components so we can skip the handling of fallback languages for now

File: packages/ui-manager/README.md [hunk 17]
```diff
@@ -0,0 +1,32 @@
...
+1. Create a new components foler in `src/components`
+2. Add your `stories.tsx` file
+3. Add your e2e test
+
+## Import component to another project
+
+The react super component is not builded so you can import directly selected components
+Just update your `tsconfig.json` :
+`         "@react-sc/*": [
+        "../../../react-super-components/src/components/"
```
aboungnaseng-ovhcloud, line 19: Update the path

chipp972 (approved):

chipp972 (approved):

chipp972 (approved):

aboungnaseng-ovhcloud (dismissed):

anooparveti (changes requested):
`ui-manager` does not convey that it is the folder for having `super-components`

packages/manager-components/.eslintrc.json
```diff
@@ -0,0 +1,6 @@
+{
+  "extends": "../../.eslintrc.js",
+  "rules": {
+    "no-underscore-dangle": "off"
```
anooparveti, line 4: **question:** Any particular reason to turn off this rule?

packages/manager-components/README.md
```diff
@@ -0,0 +1,25 @@
+# ⚛️⚡ Manager Components Library
+
+## Features
+
+Handle all React component use in manager
```
anooparveti, line 5:
```suggestion
Hosts the components used in manager
```

packages/manager-components/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+## Add new component
+
+1. Create a new components foler in `src/components`
+2. Add your `stories.tsx` file
+3. Add your e2e test
+
+## Import component to another project
+
+The react super component is not builded so you can import directly from the workspace like this :
```
anooparveti, line 16: No need for specifying "react super component", just "component" is sufficient
```suggestion
The component is not builded so you can import directly from the workspace like this :
```

packages/manager-components/src/components/tile/tile.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+  argTypes: {
+    texts: {
+      title: { control: 'text', description: 'Title of the tile' },
+      category: {
+        description: 'Category of the tile',
+        control: 'text',
+      },
+    },
+    href: { control: 'text', description: 'URL of the tile and link' },
+    isExternalHref: {
```
anooparveti, line 40: **nitpick:** `is` is not required
```suggestion
    externalHref: {
```

packages/manager-components/src/components/tile/tile.stories.tsx
```diff
@@ -0,0 +1,92 @@
...
+        description: 'Alternative label of the image',
+      },
+    },
+    badges: {
+      description:
+        'Display examples of badges in the story (in the actual code there is a badge slot)',
+    },
+    dataTracking: {
+      description: 'Tracking label sent when the tile or the link is clicked',
+    },
```
anooparveti, line 64: We can have multiple links/buttons on the tile. So I dont see the need to include `dataTracking`. And, fyi, we dont track a click on tile, we only track clicks on links & buttons.

packages/manager-components/src/components/tile/tile.stories.tsx
```diff
@@ -0,0 +1,92 @@
+import React from 'react';
+import { ScTile, ScTileProps } from './tile';
```
anooparveti, line 2: **question:** What does `Sc` indicate? Is it a super-component?

packages/manager-components/src/components/tile/tile.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const { title, description, category } = texts;
+  const { t } = useDynamicTranslation('tile');
+
+  return (
+    <a
+      className="msc-tile-wrapper"
+      target={OdsHTMLAnchorElementTarget._blank}
+      rel={OdsHTMLAnchorElementRel.noopener}
+      href={href}
+    >
```
anooparveti, line 57: **question:** Why wrapping a tile around anchor tag? I dont think it will be correct with respect to accessibilty.

packages/manager-components/src/components/tile/tile.stories.tsx
lionel95200x: Just a good practice to prefix boolean by `is`

packages/manager-components/src/components/tile/tile.tsx
lionel95200x: Cause all the card is clickable

anooparveti: Cards are not clickable. Can you show me an example where it is clickable?

packages/manager-components/src/components/tile/tile.stories.tsx
anooparveti: Did we not agree that we will not have Prefix to the components name?

anooparveti: And even if we do so, `Sc` doesn't make sense to me. It can be `Mc` to indicate "Manager Components"

lionel95200x: Fixed

anooparveti (approved):

aboungnaseng-ovhcloud (approved):

PR #10369, opened 2023-11-22
feat(dedicated.network-security): load traffic from event select

Steffy29:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #UXCT-502

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Load traffic graph from event selection on events datagrid.

### DIFF
--- a/packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.controller.js
+++ b/packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.controller.js
@@ -24,11 +24,6 @@ export default class TrafficController {
     this.isPPs = false;
     this.isServiceSelected = false;
     this.subnet = '';
-    if (this.getSubnet()) {
-      this.subnet = this.getSubnet();
-      this.selectedIp = this.subnet;
-      this.model = this.selectedIp;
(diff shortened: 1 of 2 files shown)


sachinrameshn (approved):

packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.controller.js
```diff
@@ -37,6 +32,12 @@ export default class TrafficController {
       this.services = data;
       return data;
     });
+    if (this.getSubnet()) {
+      this.subnet = this.getSubnet();
+      this.selectedIp = this.subnet;
+      this.model = this.selectedIp;
```
ghyenne, line 38: Why you need 3 variables with same value ?

Steffy29: One for the select list display into the html, one for the call API and the last for the model else it doesn't work.

ghyenne (approved):

PR #10452, opened 2023-12-04
feat(sidebar): hide terminated pci projects

y4nnL:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/pci-discovery`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #TAPC-395

- [X] Commits are signed-off
- [X] Lint has passed locally


packages/manager/apps/container/src/container/common/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+export type PciProject = {
+  access: string;
+  creationDate: Date;
+  description: string;
+  expiration: Date;
+  manualQuota: boolean;
+  orderId: unknown;
+  planCode: string;
+  projectName: string;
+  project_id: string;
```
dectotam, line 12: maybe use camelCase also on projectId ?

dectotam (approved):

packages/manager/apps/container/src/container/common/utils.ts
darsene: @y4nnL , it's due to an API constraint ?

y4nnL: This type is basically here to describe the API response so the compiler does not complain. If camel case is mandatory we could cast the response even though it's not ideal. You tell me

dectotam: Ok thanks for the explanation :)

dectotam (approved):

sachinrameshn (approved):

PR #10456, opened 2023-12-04
feat(web.domain): datagrid - add the server details

dectotam:
| Question         | Answer
| ---------------- | ---
| Branch?          | feat/domain-optim
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Feat MANAGER-11783

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

datagrid - add the server details
adding text-accordion component to ng-ovh-utils


packages/manager/apps/web/client/app/domain/list/list-domain-layout.controller.js
```diff
@@ -151,6 +151,14 @@ export default class ListDomainLayoutCtrl extends ListLayoutHelper.ListLayoutCtr
     );
   }
 
+  static getDNSServers(nameServers) {
+    return nameServers
+      .map((item) => {
+        return item.name ? item.name : '';
+      })
```
darsene, line 158:
```suggestion
      .map((item) => item.name || '')
```

packages/manager/apps/web/client/app/domain/list/style.scss
```diff
@@ -0,0 +1,9 @@
+.list-domains-datagrid {
+  .domains-nameserver-column {
+    width: 15vw;
+  }
```
darsene, line 4: Cannot be managed with bootstrap css ?

packages/components/ng-ovh-utils/src/textAccordion/textAccordion.js
darsene, line 1: Why do you create a new component ? There are already some accordions in the manager

qpavy: The component already exist in octavia branch under pci-universe-componant, Amandine move this into a more generic module to allow everyone to use it if needed ! We will make the change on Octavia side as well and if everything works well we should also remove the legacy one inside pci module and use this new one

packages/manager/apps/web/client/app/domain/list/style.scss
qpavy: I think the bootstrap sizing is relative to the parent but we need to be relative to screen size

packages/manager/apps/web/client/app/domain/list/template.html
```diff
@@ -1,332 +1,344 @@
...
+            >
+            </span>
+        </oui-datagrid-column>
+
+        <oui-datagrid-column
+            data-title="::'domains_column_domains_name_server' | translate"
+            data-property="nameServers"
+        >
+            <text-accordion
+                class="domains-nameserver-column"
```
qpavy, line 271: Is it working to put the class on the component and not on his parent ?

packages/components/ng-ovh-utils/src/textAccordion/controller.js
y4nnL, line 1: You could have used the "angular-inview": "^2.2.0" package instead of rewriting the whole mechanism. I see a performance problem with this implementation : each resize triggers a new visibility check through the timeout. If you intend to keep it, you should throttle the listener
```javascript
import { throttle } from 'lodash';
this.boundOnResize = throttle(() => this.onResize(), 100)
```
The github project
https://github.com/thenikso/angular-inview

How I used it to lazy load items in a list inside a modal
`/packages/manager/apps/telecom/src/components/telecom/telephony/service/choice/choice.html:40`

packages/components/ng-ovh-utils/src/textAccordion/template.html
```diff
@@ -0,0 +1,20 @@
+<div
+    class="text-accordion"
+    data-ng-class="{'text-wrap' : textAccordionCtrl.expanded }"
+>
+    <span
+        type="button"
```
y4nnL, line 6: There is no `type` attributes on the `span` element

packages/components/ng-ovh-utils/src/textAccordion/directive.js
```diff
@@ -0,0 +1,14 @@
+import template from './template.html';
+
+export default /* @ngInject */ () => {
+  return {
+    restrict: 'E',
+    transclude: true,
+    controller: 'textAccordionController',
+    controllerAs: 'textAccordionCtrl',
```
y4nnL, line 8: We should maintain the same design with regard to how we import / name the controllers
```javascript
import controller from './controller';

export default /* @ngInject */ () => {
  return {
    controller,
    controllerAs: '$ctrl',

y4nnL (changes requested):

packages/components/ng-ovh-utils/src/textAccordion/controller.js
dectotam: Hello, I didn't rewrite anything : the text accordion component already exists on octavia branch, I just moved it in ng-ovh-utils. Here the goal is to use it only : if a refacto is needed we'll need to have time allocated for this

qpavy: And the octavia branch is a copy of the legacy text-accordion component inside pci module

qpavy (approved):

darsene (approved):

PR #10457, opened 2023-12-04
feat(vrack-services): subnet onboarding page

chipp972:
ref: MANAGER-12337

| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes/no
| New feature?     | yes/no
| Breaking change? | yes/no
| Tickets          | Fix #... 

- [ ] Commits are signed-off
- [ ] Lint has passed locally


packages/manager/apps/vrack-services/src/components/layout-helpers/PageLayout.tsx
```diff
@@ -1,11 +1,16 @@
...
 
-export type PageLayoutProps = React.PropsWithChildren<unknown>;
+export type PageLayoutProps = React.PropsWithChildren<{
+  noBreacrumb?: boolean;
+}>;
 
-export const PageLayout: React.FC<PageLayoutProps> = ({ children }) => (
+export const PageLayout: React.FC<PageLayoutProps> = ({
+  children,
+  noBreacrumb,
```
lionel95200x, line 10: showBreadcrumb ?

lionel95200x: You could use `showBreadrumb = true` as props and optional so it's the default value(and most frequant) and if user want to disable it `showBreadcrumb={false}`, just a good practice to not use negative display props

chipp972: Could you show me the reference saying it's a bad practice to use negative prop names ?

JSX syntax looks a bit like HTML and in this regard, a prop with `true` default value is non-intuitive : 
https://legacy.reactjs.org/docs/jsx-in-depth.html#props-default-to-true

Having to specify a prop to pass it `false` looks like an anti-pattern to me; Way more than having a negative prop name.

PR #10488, opened 2023-12-06
feat(octavia): add ip instances step two

JacquesLarique:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/octavia-batch-1`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-12304

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Added the second step of the view add instance IP

### DIFF
--- a/packages/manager/modules/octavia-load-balancer/src/load-balancers/load-balancer/pools/detail/members/add-ip-instance/style.less
+++ b/packages/manager/modules/octavia-load-balancer/src/load-balancers/load-balancer/pools/detail/members/add-ip-instance/style.less
@@ -2,3 +2,9 @@
   max-height: 50dvh;
   overflow-y: scroll;
 }
+
+.box-member {
+  .oui-field {
+    margin-bottom: 0;
+  }
+}
--- a/packages/manager/modules/octavia-load-balancer/src/load-balancers/load-balancer/pools/detail/members/add-ip-instance/template.html
+++ b/packages/manager/modules/octavia-load-balancer/src/load-balancers/load-balancer/pools/detail/members/add-ip-instance/template.html
@@ -45,7 +45,7 @@
                     >
                         <oui-checkbox
                             data-name="{{'instance-' + $index}}"
-                            data-model="$ctrl.model.checked[$index]"
+                            data-model="instance.checked"
                         >
                             <span data-ng-bind="instance.label"></span>
                         </oui-checkbox>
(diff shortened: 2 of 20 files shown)


packages/manager/modules/octavia-load-balancer/src/load-balancers/load-balancer/pools/detail/members/add-ip-instance/style.less
```diff
@@ -2,3 +2,9 @@
   max-height: 50dvh;
   overflow-y: scroll;
 }
+
+.box-member {
+  .oui-field {
+    margin-bottom: 0;
```
darsene, line 8: you can use css bootsrap 
```suggestion
   @extend .mb-0;
```

packages/manager/modules/octavia-load-balancer/src/load-balancers/load-balancer/pools/detail/members/add-ip-instance/template.html
```diff
@@ -63,9 +63,82 @@
...
+                        class="oui-input"
+                        type="number"
+                        name="{{:: 'member-port-' + memberIndex}}"
+                        data-ng-model="member.protocolPort"
+                        data-ng-model-options="{ allowInvalid: true }"
+                        data-min="1"
+                        data-max="65535"
+                        data-ng-disabled="$ctrl.duplicatePort && memberIndex !== 0"
+                        data-ng-change="$ctrl.onPortChange(memberIndex)"
+                        data-ui-validate="{ existing: '!$ctrl.isIpAlreadyAssociatedWithPort(member.address, $value)' }"
```
darsene, line 122: Why not using `valid` attributes from stepper component instead ? Adding a new library just for one field validation seems a bit heavy

JacquesLarique: Using valid will disabled the "confirm" button and will not display a custom error.
I agree that adding a library (not that big btw) is not great, but to achieve a custom validation in this case the solution is pretty ugly: the only way to access the input to change its validity is by adding an encapsulating form to be able to access the one behind step-form and finaly access the input control.

darsene (approved):

packages/manager/modules/octavia-load-balancer/src/load-balancers/load-balancer/pools/detail/members/add-ip-instance/style.less
JacquesLarique: Extending mb-0 doesn't not work with less files, and switching to scss require to import 4 bootstrap4 files, so this will stay as is

antonymarion (approved):

PR #10557, opened 2023-12-18
feat(dedicated): implement os-dynamic form in ovh and gabarit wizards

JayBeeDe:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | yes
| Breaking change? | no

- [x] Commits are signed-off
- [x] Lint has passed locally
~Breaking change is mentioned in relevant commits~

## Description

- feat: Introduce OS-dynamic form in the OVH and gabarit OS installation wizards
- minor fix: Prevent start installation when a field is invalid at step 4/4 of the OVH and gabarit OS installation wizards

## Related

ref: MANAGER-10557


packages/manager/apps/dedicated/client/app/css/less/dedicated/server/installation.less
```diff
@@ -105,3 +105,19 @@
...
+
+.fix-help-block .help-block {
+  // this is to compensate the move down of the next questions when an error is displayed
+  margin-bottom: -25px;
+  height: 20px;
+}
+
+.fix-help-block .oui-field__footer {
+  // this is to hide the native lib errors because they are customized
+  display: none;
```
JayBeeDe, line 122: If you have a better idea to hide the lib native error message (because I use custom error messages) I will be very happy

packages/manager/modules/bm-server-components/src/general-information/inputs/inputs.class.js
```diff
@@ -0,0 +1,35 @@
+export default class Inputs {
```
JayBeeDe, line 1: I have created a class because I wanted to have a shared function used within the server-installation-gabarit.controller as well as the server-installation-ovh.controller.

packages/manager/modules/bm-server-components/src/general-information/installation/gabarit/server-installation-gabarit.controller.js
```diff
@@ -134,6 +135,24 @@ export default class ServerInstallationGabaritCtrl {
     });
   }
 
+  static setSizeModalDialog(bigSize) {
```
JayBeeDe, line 138: I'm wondering whether we should put this function in a class or anything else so that we can reuse it in several controllers without duplicating the code

packages/manager/modules/bm-server-components/src/general-information/inputs/controller.js
```diff
@@ -0,0 +1,75 @@
...
+  $onInit() {
+    angular.forEach(this.installation.inputs, (value, key) => {
+      if (value.type === 'enum' && !value.mandatory) {
+        this.installation.inputs[key].enum = [''].concat(value.enum);
+        // you have an optional select dropdown so you want to be able to unset to none
+      }
+    });
+
+    // inputs
+    // let inputsIndex = 0;
```
dectotam, line 22:
```suggestion
```

packages/manager/modules/bm-server-components/src/general-information/inputs/controller.js
```diff
@@ -0,0 +1,75 @@
...
+        this.installation.input[inputItem.name] = +inputItem.default;
+        this.installation.input[inputItem.name] =
+          inputItem.default === null ? 0 : +inputItem.default;
+      } else if (inputItem.type === 'boolean') {
+        this.installation.input[inputItem.name] =
+          inputItem.default !== null && inputItem.default === 'true';
+      } else {
+        this.installation.input[inputItem.name] =
+          inputItem.default === null ? '' : inputItem.default?.toString();
+      }
```
dectotam, line 37:
```suggestion
  let input;
  switch (inputItem.type) {
    case 'keyValue':
      input = [];
      break;
    case 'number':
      input = inputItem.default === null ? 0 : +inputItem.default;
      break;
    case 'boolean':
      input = inputItem.default !== null && inputItem.default === 'true';
      break;
    default:
      input = inputItem.default === null ? '' : inputItem.default?.toString();
      break;
  }

  this.installation.input[inputItem.name] = input;
```

packages/manager/modules/bm-server-components/src/general-information/inputs/controller.js
```diff
@@ -0,0 +1,75 @@
...
+      indexItem !== -1 &&
+      this.installation.inputs[indexItem].default &&
+      this.installation.inputs[indexItem].default.replace(', ', ',').split(',')
+        .length === 2
+    ) {
+      return this.installation.inputs[indexItem].default
+        .replace(', ', ',')
+        .split(',')[mode === 'value' ? 1 : 0];
+    }
+    return '';
```
dectotam, line 58:
```suggestion
    let keyValueInit = '';
    if (this.installation.input[name].length == 0) {
      const indexItem = findLastIndex(this.installation.inputs, {
        name,
      });
      const inputItem = this.installation.inputs[indexItem];

      if (inputItem?.default) {
        const splitDefaults = inputItem.default.split(/, ?/);
        
        if (splitDefaults.length === 2) {
          keyValueInit = splitDefaults[mode === 'value' ? 1 : 0];
        }
      }
    }

    return keyValueInit;
```

dectotam (changes requested):

packages/manager/modules/bm-server-components/src/general-information/installation/gabarit/server-installation-gabarit.controller.js
JacquesLarique: I am not able to find a proper way of doing this as creating a new component to handle modal doesn't seem pertinent, as we already have one in UI-kit but unfortunately it does not allow changing its size and I'm not sure it can be used the way your modal is used (the modal's being declared in an html which is not the one linked to the controller who's managing it)
There's also a modal in the shell, but I think using it would result in a modal covering all the page and not just the angularjs application.

packages/manager/apps/dedicated/client/app/css/less/dedicated/server/installation.less
JacquesLarique: The rules to display errors in a oui-field are the following:

- a form is encapsulating the field
- the input is filled or it has lost focus or the form has been submitted

So the only way (other than using css) to not diplay it is to remove the encapsulating form (which may not be possible when using wizard)

Your solution seems OK to me

packages/manager/apps/dedicated/client/app/css/less/dedicated/server/installation.less
```diff
@@ -105,3 +105,19 @@
...
+
+.form-control.oui-input_l,
+.form-control.oui-input_s {
+  width: inherit;
+}
+
+.fix-help-block .help-block {
+  // this is to compensate the move down of the next questions when an error is displayed
+  margin-bottom: -25px;
+  height: 20px;
```
JacquesLarique, line 117:
```suggestion
  margin-bottom: -1.563rem;
  height: 1.25rem;
```

packages/manager/modules/bm-server-components/src/general-information/inputs/controller.js
```diff
@@ -0,0 +1,75 @@
...
+    this.$translate = $translate;
+  }
+
+  $onInit() {
+    angular.forEach(this.installation.inputs, (value, key) => {
+      if (value.type === 'enum' && !value.mandatory) {
+        this.installation.inputs[key].enum = [''].concat(value.enum);
+        // you have an optional select dropdown so you want to be able to unset to none
+      }
+    });
```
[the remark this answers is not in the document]
JayBeeDe: done

packages/manager/modules/bm-server-components/src/general-information/installation/gabarit/server-installation-gabarit.controller.js
JayBeeDe: ok let's keep that

packages/manager/modules/bm-server-components/src/general-information/inputs/controller.js
```diff
@@ -0,0 +1,79 @@
+import remove from 'lodash/remove';
+
+export default class BmServerComponentsInputsController {
+  /* @ngInject */
+  constructor($q, $scope, $stateParams, $translate) {
+    this.$q = $q;
+    this.$scope = $scope;
+    this.$stateParams = $stateParams;
+    this.$translate = $translate;
+  }
```
JacquesLarique, line 10: These injections don't seem to be used, you should be able to remove this constructor

packages/manager/modules/bm-server-components/src/general-information/inputs/template.html
```diff
@@ -0,0 +1,112 @@
+<!-- dynamic inputs form depending on the OS template -->
```
JacquesLarique, line 1: Some formatting would be nice on this file, this should have been done with the pre-commit hook from husky.

packages/manager/modules/bm-server-components/src/general-information/inputs/controller.js
JayBeeDe: What if findLastIndex returns -1 (no found) ?
@dectotam can you re-review this part of the new updated version ?

packages/manager/modules/bm-server-components/src/general-information/inputs/controller.js
JayBeeDe: done

JacquesLarique (dismissed):

packages/manager/modules/bm-server-components/src/general-information/inputs/controller.js
dectotam: If lastIndex is -1, then inputItem will be undefined, so the condition inputItem?.default will be false, and keyValueInit will be returned as default empty value ''.

JayBeeDe: I just disagree with the `let keyValueInit = '';` and single `return keyValueInit` because it makes too many nested if compared to a first
```js
if (this.installation.input[name].length === 0) {
    return '';
}
```
then we save an if indentation `if (this.installation.input[name].length !== 0) {` and the nested if conditions that we  need to put inside

JayBeeDe: @dectotam check my change, it's not exactly your suggestion because of my previous comment and @JacquesLarique's suggestion regarding `findLastIndex` function's deprecated `lodash` depedency

dectotam (dismissed):

JacquesLarique (approved):

JacquesLarique (approved):

packages/manager/modules/bm-server-components/src/general-information/inputs/template.html
```diff
@@ -0,0 +1,247 @@
...
+        aria-hidden="true"
+    >
+    </span>
+    <p
+        class="oui-message__body"
+        data-translate="server_configuration_installation_inputs_info"
+        data-translate-values="{ t0: $ctrl.installation.selectDistribution.description }"
+    ></p>
+</div>
+<p>&nbsp;</p>
```
ghyenne, line 15:
```suggestion
```
I don't think is really needed, maybe you can add a class on previus element if you want an espace between the oui-message and following of form

packages/manager/modules/bm-server-components/src/general-information/inputs/template.html
```diff
@@ -0,0 +1,247 @@
...
+            data-sshkey-switch-key="{{$ctrl.installation.input[input.name]}}"
+        ></div>
+        <textarea
+            data-ng-switch-when="sshPubKey"
+            class="oui-input form-control"
+            data-ng-model="$ctrl.installation.input[input.name]"
+            data-ng-required="{{input.mandatory}}"
+            rows="15"
+            name="{{input.name}}"
+            data-ng-pattern="/^(ssh-rsa|ecdsa-sha\d+-nistp\d+|ssh-ed\d+)\s+(AAAA[a-zA-Z0-9\/=+]+)(\s+(\S{1,128}))*?$/"
```
ghyenne, line 103: It can be move in a constant file.

packages/manager/modules/bm-server-components/src/general-information/inputs/template.html
```diff
@@ -0,0 +1,247 @@
...
+                        <oui-field label="Key">
+                            <input
+                                class="oui-input form-control"
+                                type="text"
+                                name="key"
+                                data-ng-model="key"
+                                data-ng-required="input.mandatory && $ctrl.installation.input[input.name].length === 0"
+                                data-ng-init="key = $ctrl.keyValueInit(input.name, 'key')"
+                                placeholder="{{:: 'server_configuration_installation_inputs_item_placeholder' | translate:{ t0: 'key' } }}"
+                                data-ng-pattern="/^[a-z0-9]{1,256}$/i"
```
ghyenne, line 127: For the all pattern you can be move in constant file.

ghyenne (approved):

JacquesLarique (approved):

ghyenne (dismissed):

JacquesLarique (approved):

ghyenne (dismissed):

ghyenne (approved):

JacquesLarique (dismissed):

darsene (approved):

JacquesLarique (approved):

dectotam (approved):

PR #10661, opened 2024-01-09
fix(ip): create edge firewall rule

antonymarion:
ref: DTRSD-125797

| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #DTRSD-125797

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

CreateRule function was breaking because of Enums. UI dropdown choice was not displaying well.

### DIFF
--- a/packages/manager/apps/dedicated/client/app/ip/components/list/firewall/ip-ip-firewall.controller.js
+++ b/packages/manager/apps/dedicated/client/app/ip/components/list/firewall/ip-ip-firewall.controller.js
@@ -120,7 +120,7 @@ export default /* @ngInject */ function IpFirewallCtrl(
           const newResult = result;
           const map = {
             key: name,
-            name: parseInt(name.replace('_', ''), 10),
+            name: key,
           };
           newResult[key] = map;
         }),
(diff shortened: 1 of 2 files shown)


packages/manager/apps/dedicated/client/app/ip/components/list/firewall/ip-ip-firewall.controller.js
```diff
@@ -120,7 +120,7 @@ export default /* @ngInject */ function IpFirewallCtrl(
           const newResult = result;
           const map = {
             key: name,
-            name: parseInt(name.replace('_', ''), 10),
+            name: key,
```
ghyenne, line 123: Why key is not key and name isn't name. It isn't logical.

antonymarion: I am going to use { key, name } instead it will work and be better for code

PR #10672, opened 2024-01-10
feat(chakra): replace link color in sidebar

stif59100:
ref:MANAGER-13021

| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-13021

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

I replaced the color for icon on sidebar after remove chakra ui

### DIFF
--- a/packages/manager/apps/container/src/container/legacy/server-sidebar/index.module.scss
+++ b/packages/manager/apps/container/src/container/legacy/server-sidebar/index.module.scss
@@ -241,5 +241,6 @@
   text-align: left;
 }
 
-
-
+.iconStyleColor{
+  color:#4D5592;
+}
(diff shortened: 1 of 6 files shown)


aboungnaseng-ovhcloud (approved):

vovh (commented):
Lionel's suggestion was relevant and to make it even cleaner I would tend to create a file/component in its own for a better understanding. I'm sorry to bother you with this.

But I find it a little strange to import a function from another universe that has no specific relation to this said universe.

You can create a file in a utils folder, for example, where you declare the function you want to import.

We can do something simpler without a folder eventually but I find that in terms of cognitive complexity, import a function of the "Dedicated" universe

`import { getIcon } from './DedicatedSidebar';`

makes little sense.

I would tend to make a separate file

`import { getIcon } from './SidebarUtils';`

For example, to avoid having a mention of DedicatedSidebar in all other files, what do you think?

PR #10686, opened 2024-01-11
feat(dedicated.cluster): add cluster nodes tab

darsene:
ref: MANAGER-12271

| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/multi-az-feature`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-12271

- [x] Commits are signed-off
- [x] Lint has passed locally
- ~Breaking change is mentioned in relevant commits~


qpavy (approved):

antonymarion (approved):

packages/manager/apps/dedicated/client/app/dedicated/dedicated-server/clusters/cluster/cluster.service.js
```diff
@@ -22,4 +26,23 @@ export default class Cluster {
...
+    const serverNames = cluster.nodes?.map(({ serverName }) => serverName);
+    const nodes = await this.$q.all(
+      serverNames.map((serverName) =>
+        this.Server.getSelected(serverName).then(
+          (swsResponse) => new DedicatedServer(swsResponse),
+        ),
+      ),
+    );
+
+    return nodes;
```
anooparveti, line 46: **nitpick:** Is `async` &. `await` working as expected because I have noticed an issue few times where we had to explicitly trigger the Digest cycle to update the view.
IMO, better to switch to `.then`

qpavy: In this case it should work as expected because it is used inside resolve so digest cycle is triggered anyway at component init

PR #10699, opened 2024-01-16
feat(catalog): mock login & data for playwright tests

vovh:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #13037

- [x] Commits are signed-off
 ~~[ ] Only FR translations have been updated~~
- [x] Lint has passed locally
 ~~[ ] Breaking change is mentioned in relevant commits~~


.github/workflows/playwright.yml
```diff
@@ -0,0 +1,28 @@
...
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v3
+    - uses: actions/setup-node@v3
+      with:
+        node-version: 18
+    - name: Setup yarn
+      run: npm install -g yarn
+    - name: Install dependencies
+      run: yarn install --frozen-lokfile
```
antleblanc, line 16: Invalid option.

**suggestion:** You can refer to an existing GitHub Action to be able to install project dependencies.

see: https://github.com/ovh/manager/blob/21eac231d2517524ba91a3dc88b5c5e163156aa0/.github/workflows/linter.yaml#L13-L28

antleblanc (changes requested):

.github/workflows/playwright.yml
vovh: Refactored to use suggested existing syntax.

.github/workflows/playwright.yml
```diff
@@ -0,0 +1,33 @@
...
+        yarn install --frozen-lockfile
+    - name: Build dependencies
+      run: | 
+        yarn build
+    - name: Install Playwright Browsers
+      run: | 
+        npx playwright install --with-deps
+    - name: Run Playwright catalog tests
+      run: |
+        yarn run test:playwright:ci
```
antleblanc, line 33: **question:** Would it possible to run this command only on apps that has recently changed?

vovh: We're currently discussing into our team about the process because it's indeed too resource consuming.
I'm gonna put this PR as draft for now

PR #10760, opened 2024-01-24
fix(public-cloud): list priavte network with vlanid 0

sachinrameshn:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | DTRSD-126315 

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/private-networks/private-networks.service.js
+++ b/packages/manager/modules/pci/src/projects/project/private-networks/private-networks.service.js
@@ -19,13 +19,19 @@ export default class {
     this.OvhApiCloudProjectNetworkPrivateSubnet = OvhApiCloudProjectNetworkPrivateSubnet;
   }
 
-  getPrivateNetworks(serviceName) {
+  getPrivateNetworks(serviceName, customerRegions = []) {
     return this.$http
       .get(`/cloud/project/${serviceName}/aggregated/network`)
       .then(({ data }) => {
         const privateNetworks = {};
+        const localZones =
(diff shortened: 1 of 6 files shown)


anooparveti (approved):

qpavy (approved):

qpavy (dismissed):

anooparveti (approved):

qpavy (approved):

packages/manager/modules/pci/src/projects/project/private-networks/private-networks.service.js
```diff
@@ -19,13 +19,19 @@ export default class {
...
       .then(({ data }) => {
         const privateNetworks = {};
+        const localZones = customerRegions?.filter(({ type }) =>
+          type.includes('localzone'),
+        );
         data.resources.forEach((network) => {
-          if (network.visibility === 'private' && network.vlanId) {
+          if (
+            network.visibility === 'private' &&
+            !localZones?.some((region) => region.name === network.region)
```
JacquesLarique, line 33: Won't this check fail if localZones is undefined (only case I see now would be that customerRegions is null)
From what I understand of the code we want to handle the case where network is private and it is not located in a localZone
So I guess if localZones is undefined (meaning we don't have localZones) we should pass this test

JacquesLarique (approved):

qpavy (approved):

PR #10861, opened 2024-02-07
feat(project.flavor-list): add badge for the new flavor selector

brofik:
ref: TAPC-525

| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #TAPC-525 

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager/modules/pci/src/components/project/flavors-list/flavor-group.scss
+++ b/packages/manager/modules/pci/src/components/project/flavors-list/flavor-group.scss
@@ -1,0 +1,7 @@
+.flavors-list {
+  .new-flavor-badge {
+    position: relative;
+    top: -2.5rem;
+    float: right;
+  }
+}
--- a/packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.html
+++ b/packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.html
@@ -1,6 +1,6 @@
 <oui-spinner data-size="s" data-ng-if="$ctrl.isLoading"></oui-spinner>
 
-<div data-ng-if="!$ctrl.isLoading">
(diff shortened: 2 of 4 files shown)


dectotam (approved):

packages/manager/modules/pci/src/components/project/flavors-list/flavor-group.scss
```diff
@@ -0,0 +1,7 @@
+.flavors-list {
+  .new-flavor-badge {
+    position: relative;
+    top: -2.5rem;
+    float: right;
+  }
+}
```
sachinrameshn, line 7: Question: Cant we add class directly on span ? Why are we adding it from top div ?

packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.html
```diff
@@ -41,6 +41,13 @@
                         data-required
                     >
                         <oui-select-picker-section>
+                            <span
+                                class="oui-badge oui-badge_new new-flavor-badge"
```
anooparveti, line 45: **nitpick:** Avoid custom css.
```suggestion
                                class="oui-badge oui-badge_new float-right position-relative"
```

anooparveti (changes requested):

packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.html
brofik: The bootstrap CSS class does not enable the definition of precise value for the `top` property. 
That why I had to use the custom CSS.

anooparveti (approved):

PR #10907, opened 2024-02-13
fix(security.dashboard): display timestamp to locale time

Steffy29:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #UXCT-524

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

- Display timestamp to locale time into graph
- Display UTC time to inform users

### DIFF
--- a/packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.constant.js
+++ b/packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.constant.js
@@ -53,6 +53,30 @@ export const CHART = {
     tooltips: {
       mode: 'label',
       intersect: false,
+      callbacks: {
+        title: (context) => {
+          // Retrieve user language
+          const userLanguage = navigator.language;
+
+          // Configure dateTime format to display like DD/MM/YYYY hh:mm:ss (for FR) or MM/DD/YY, hh:mm:ss (for US)
+          const dateTimeFormat = new Intl.DateTimeFormat(userLanguage, {
+            hourCycle: 'h23',
+            timeStyle: 'medium',
+            dateStyle: 'short',
+          });
+
+          // Retrieve user local time zone
+          const timeZoneOffset =
+            (new Date(context[0].label).getTimezoneOffset() / 60) * -1;
+          const sign = timeZoneOffset > 0 ? '+' : '';
+
+          // Format title like DD/MM/YYYY hh:mm:ss UTC+1 (for FR) or MM/DD/YY, hh:mm:ss UTC-2 (for US)
+          const title = `${dateTimeFormat.format(
+            new Date(context[0].label),
+          )} UTC${sign}${timeZoneOffset}`;
+          return title;
+        },
+      },
     },
     scales: {
       yAxes: [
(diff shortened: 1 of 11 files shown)


packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.constant.js
```diff
@@ -53,6 +53,15 @@ export const CHART = {
     tooltips: {
       mode: 'label',
       intersect: false,
+      callbacks: {
+        title: (context) => {
+          const timeZoneOffset =
+            (new Date(context[0].label).getTimezoneOffset() / 60) * -1;
```
qpavy, line 59: Just to be sure there won't be an issue, getTimezoneOffset is getting the timezone offset of the local zone and not the zone of the date:
For example: if the date is 2024-06-06T10:00:000+01:00 but the code is executed on the timezone UTC+2, the computed timezoneOffset will be 120 and not 60. 
Only the date is used which allow to get the good offset depending if you are in Daylight saving time or not.

JacquesLarique, line 59: Are we sure `context[0].label` will always be a valid Date ?
If not it would result in displaying NaN in the tooltip

packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.controller.js
```diff
@@ -281,7 +290,11 @@ export default class TrafficController {
       delete this.options.scales.yAxes[0].stacked;
     }
 
-    this.labels = this.results?.timestamps?.map((value) => value);
+    // Convert timestamps UTC to locale time zone
+    this.labels = this.results?.timestamps?.map((value) => {
+      const timestamp = moment(new Date(value)).format('YYYY-MM-DD HH:mm:ss');
```
qpavy, line 295: you can use intl datetime format as it also format depending of the language

packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.constant.js
Steffy29: Yes we're sure it's a valid date.

Steffy29: We need to display the date in the local zone of the user. For the moment, the date displayed is, for example, 2024-02-08T05:55:00Z which is not readable for users.

JacquesLarique (dismissed):

packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.constant.js
```diff
@@ -53,6 +53,23 @@ export const CHART = {
...
+            timeStyle: 'medium',
+            dateStyle: 'short',
+          });
+          const timeZoneOffset =
+            (new Date(context[0].label).getTimezoneOffset() / 60) * -1;
+          const sign = timeZoneOffset > 0 ? '+' : '';
+          const title = `${dateTimeFormat.format(
+            new Date(context[0].label),
+          )} UTC${sign}${timeZoneOffset}`;
+          return title;
```
anooparveti, line 70: Can you please add a comment explaining what you're doing here?

JacquesLarique (approved):

qpavy (dismissed):

anooparveti (dismissed):
@Steffy29, can you use https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toLocaleString?
Let me know if you have any scenarios which is not fulfilled by this?

Steffy29:
@anooparveti I'm not sure that your suggestion answer my need. I want to display date into local time with a label to inform timezone from UTC (like UTC+1 for France) and it seems that the Date.toLocaleString, you display the date into UTC format with locale format.

anooparveti (dismissed):

ghyenne (approved):

ghyenne (approved):

PR #10909, opened 2024-02-13
fix(dedicated): fix kvm order when datacenter name is eri

qpavy:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #INC0019752

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

fix order kvm for server in eri datacenter

### DIFF
--- a/packages/manager/modules/bm-server-components/src/ipmi/constants.js
+++ b/packages/manager/modules/bm-server-components/src/ipmi/constants.js
@@ -26,8 +26,13 @@ export const KVM_PLAN_CODE = 'usb-kvm-ip';
 export const KVM_ORDER_TRACKING_PREFIX =
   'dedicated::dedicated-server::server::ipmi::order-kvm';
 
+export const DATACENTER_AGORA_NAME = {
+  eri: 'lon',
+};
+
 export default {
   getIpmiGuideUrl,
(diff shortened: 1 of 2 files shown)


packages/manager/modules/bm-server-components/src/ipmi/order-kvm/controller.js
```diff
@@ -13,7 +13,9 @@ export default class BmServerComponentsOrderKvmController {
   $onInit() {
     this.serviceName = this.server.name;
     const [datacenter] = this.server.datacenter.split('_');
-    this.datacenter = datacenter.toLowerCase();
+    this.datacenter = this.IpmiService.constructor.getRealDCAgoraName(
+      datacenter.toLowerCase(),
+    );
```
darsene, line 18: No need to define the mapping here, as it concerns kvm order, it should be done just before calling the API, therefore directly in `IpmiService`. By the way, the mapping function is alredy defined in the service so it sounds more logical

darsene (approved):

dectotam (approved):

anooparveti (approved):

PR #10911, opened 2024-02-14
fix(pci.project): display one antifraud banner at a time

y4nnL:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/pci-discovery`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #TAPC-597

- [X] Commits are signed-off
- [X] Lint has passed locally

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/project.service.js
+++ b/packages/manager/modules/pci/src/projects/project/project.service.js
@@ -55,33 +55,27 @@ export default class {
               followUp: this.getOrderFollowUp(order.orderId),
             })
             .then(({ followUp, vouchers: [firstVoucher] }) => {
-              const validatingStepHistory = followUp.find(
-                ({ step }) => step === VALIDATING,
-              )?.history;
+              const labels =
+                followUp
+                  .find(({ step }) => step === VALIDATING)
+                  ?.history.map(({ label }) => label) || [];
 
               status.voucher = firstVoucher?.voucher || '';
-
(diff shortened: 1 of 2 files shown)


qpavy (approved):

packages/manager/modules/pci/src/projects/project/project.service.js
```diff
@@ -70,13 +70,15 @@ export default class {
...
-                }
+              }
+              if (status.isRefusedByAntiFraud) {
+                status.isManuallyReviewedByAntiFraud = false;
+              }
+              if (
+                status.isRefusedByAntiFraud ||
+                status.isManuallyReviewedByAntiFraud
+              ) {
+                status.isActivating = false;
```
anooparveti, line 81: **question:** Can you explain what you're trying to do here?

y4nnL: Sure, the status object we build in this method reflects the steps a pci order goes through. Each property is a boolean which is true is the associated step exists in the order. But some properties should be cleared in case some others are true. For instance, a project should not be flagged as "activating" is the anti fraud team has detected some anomalies in the order

anooparveti: Yes I think that is more readable. Can you make the change as you suggested above?

anooparveti (approved):

PR #10922, opened 2024-02-15
fix(ip.gamefirewall): check ports are not already used

Steffy29:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #UXCT-526

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

When adding new GAME firewall rule, check that the ports are not already used by other rules or are not into other rules ranges.

### DIFF
--- a/packages/manager/apps/dedicated/client/app/ip/components/list/firewall/game/ip-ip-firewall-game.controller.js
+++ b/packages/manager/apps/dedicated/client/app/ip/components/list/firewall/game/ip-ip-firewall-game.controller.js
@@ -294,6 +294,42 @@ export default /* @ngInject */ function IpGameFirewallCtrl(
     });
   }
 
+  function hasPortsAlreadyUsed() {
+    return self.table.rules.some((rule) => {
+      // Check if other rule has no range
+      if (rule.ports.from === rule.ports.to) {
+        // Check if new rule has no range
+        if (self.rule.ports.from === self.rule.ports.to) {
+          // Check if new rule is not already defined
+          if (self.rule.ports.from === rule.ports.from) {
+            return true;
+          }
+        } else if (IpGameFirewall.hasRuleIncludedInNewRule(self.rule, rule)) {
+          // Check if other rule is not included into new rule
+          return true;
+        }
+      } else if (self.rule.ports.from === self.rule.ports.to) {
+        // Check if new rule has no range
+        if (IpGameFirewall.hasNewRuleIncludedInRule(self.rule, rule)) {
+          // Check if new rule is not included into other rule
+          return true;
+        }
+      } else if (
+        IpGameFirewall.hasNewRuleIntoRule(self.rule, rule) ||
+        IpGameFirewall.hasNewRulePortToIntoRule(self.rule, rule) ||
+        IpGameFirewall.hasNewRulePortFromIntoRule(self.rule, rule) ||
+        IpGameFirewall.hasRuleIntoNewRule(self.rule, rule)
+      ) {
+        // Check if the new rule is not into other rules or
+        // Check if the new rule port to is not into other rules or
+        // Check if the new rule port from is not into other rules or
+        // Check if the rule is not into new rule
+        return true;
+      }
+      return false;
+    });
+  }
+
   $scope.$on('ips.gameFirewall.display.remove', (event, ruleId) => {
     changeStateRule(ruleId, self.constantes.DELETE_RULE_PENDING);
 
@@ -415,6 +451,18 @@ export default /* @ngInject */ function IpGameFirewallCtrl(
       Alerter.error(
         $translate.instant(
           'ip_game_mitigation_firewall_rule_add_invalid_parameters',
+        ),
+        alert,
+      );
+      self.loading = false;
+      return;
+    }
+
+    // Check if ports are not already used by other rules
+    if (hasPortsAlreadyUsed()) {
+      Alerter.error(
+        $translate.instant(
+          'ip_game_mitigation_firewall_rule_add_ports_already_used',
         ),
         alert,
       );
--- a/packages/manager/apps/dedicated/client/app/ip/components/list/firewall/game/ip-ip-firewall-game.service.js
+++ b/packages/manager/apps/dedicated/client/app/ip/components/list/firewall/game/ip-ip-firewall-game.service.js
@@ -166,4 +166,80 @@ export default /* @ngInject */ function IpGameFirewallService(
       (http) => $q.reject(http.data),
     );
   };
+
+  // Check if the new rule is not into other rules
+  // for example:
+  // newRule = {from: 4320, to: 4330},
+  // rule = {from: 4300, to: 4400}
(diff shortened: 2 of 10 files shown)


packages/manager/apps/dedicated/client/app/ip/components/list/firewall/game/ip-ip-firewall-game.controller.js
```diff
@@ -294,6 +294,47 @@ export default /* @ngInject */ function IpGameFirewallCtrl(
...
+          self.rule.ports.to > rule.ports.to) ||
+        (self.rule.ports.from < rule.ports.from &&
+          self.rule.ports.to > rule.ports.to) ||
+        (self.rule.ports.from < rule.ports.from &&
+          self.rule.ports.to > rule.ports.from)
+      ) {
+        isValid = false;
+      }
+    });
+    return isValid;
```
JacquesLarique, line 335:
```suggestion
   return !self.table.rules.some((rule) => {
      if (rule.ports.from === rule.ports.to) {
        if (self.rule.ports.from === self.rule.ports.to) {
          if (self.rule.ports.from === rule.ports.from) {
            return true;
          }
        } else if (
          self.rule.ports.from < self.rule.ports.to &&
          rule.ports.from > self.rule.ports.from
        ) {
          return true;
        }
      } else if (self.rule.ports.from === self.rule.ports.to) {
        if (
          self.rule.ports.from < rule.ports.to &&
          self.rule.ports.from > rule.ports.from
        ) {
          return true;
        }
      } else if (
        (self.rule.ports.from > rule.ports.from &&
          self.rule.ports.to < rule.ports.to) ||
        (self.rule.ports.from < rule.ports.from &&
          self.rule.ports.form < rule.ports.to &&
          self.rule.ports.to < rule.ports.to) ||
        (self.rule.ports.from > rule.ports.from &&
          self.rule.ports.from < rule.ports.to &&
          self.rule.ports.to > rule.ports.to) ||
        (self.rule.ports.from < rule.ports.from &&
          self.rule.ports.to > rule.ports.to) ||
        (self.rule.ports.from < rule.ports.from &&
          self.rule.ports.to > rule.ports.from)
      ) {
        return true;
      }
    });
```
This would be more performent (in some case) as it will stop looping at the first invalid rule

JacquesLarique (changes requested):

packages/manager/apps/dedicated/client/app/ip/components/list/firewall/game/ip-ip-firewall-game.controller.js
darsene: You should also split it in sub functions with explicit names to ease the comprehension of the algorithm. Quite difficult to understand currently

packages/manager/apps/dedicated/client/app/ip/components/list/firewall/game/ip-ip-firewall-game.controller.js
```diff
@@ -294,6 +294,47 @@ export default /* @ngInject */ function IpGameFirewallCtrl(
     });
   }
 
+  function checkPortsNotUsed() {
```
darsene, line 297:
```suggestion
  function hasPortsAlreadyUsed() {
```
With the current name, we don't understand if having `true` as a result is OK or not without looking into the code function.
You might need to change the return value regarding your name for sure

packages/manager/apps/dedicated/client/app/ip/components/list/firewall/game/ip-ip-firewall-game.controller.js
```diff
@@ -422,6 +463,19 @@ export default /* @ngInject */ function IpGameFirewallCtrl(
       return;
     }
 
+    // Check if ports are not already used by other rules
+    const isValid = checkPortsNotUsed();
+    if (!isValid) {
```
darsene, line 468:
```suggestion
    if (hasPortsAlreadyUsed()) {
```

darsene (changes requested):

packages/manager/apps/dedicated/client/app/ip/components/list/firewall/game/ip-ip-firewall-game.controller.js
```diff
@@ -294,6 +294,66 @@ export default /* @ngInject */ function IpGameFirewallCtrl(
...
+        }
+      } else if (
+        hasNewRuleIntoRule(self.rule, rule) ||
+        hasNewRulePortToIntoRule(self.rule, rule) ||
+        hasNewRulePortFromIntoRule(self.rule, rule) ||
+        hasRuleIntoNewRule(self.rule, rule)
+      ) {
+        return true;
+      }
+      return false;
```
anooparveti, line 353: **nitpick:** Can you add comments to explain what you're doing here. Complicated to understand the business logic.
Is it not possible to simplify the logic here?

anooparveti (changes requested):

packages/manager/apps/dedicated/client/app/ip/components/list/firewall/game/ip-ip-firewall-game.controller.js
Steffy29: I added the comments to explain, I should be difficult to simply the logic here.

packages/manager/apps/dedicated/client/app/ip/components/list/firewall/game/ip-ip-firewall-game.controller.js
```diff
@@ -294,6 +294,86 @@ export default /* @ngInject */ function IpGameFirewallCtrl(
...
+
+  // Check if the rule is not into new rule
+  // for example:
+  // newRule = {from: 4100, to: 4500},
+  // rule = {from: 4200, to: 4300}
+  function hasRuleIntoNewRule(newRule, rule) {
+    return (
+      newRule.ports.from < rule.ports.from && newRule.ports.to > rule.ports.to
+    );
+  }
```
darsene, line 339: should be static and moved in an utils service

packages/manager/apps/dedicated/client/app/ip/components/list/firewall/game/ip-ip-firewall-game.controller.js
```diff
@@ -294,6 +294,86 @@ export default /* @ngInject */ function IpGameFirewallCtrl(
...
+          rule.ports.from > self.rule.ports.from
+        ) {
+          return true;
+        }
+      } else if (self.rule.ports.from === self.rule.ports.to) {
+        if (
+          self.rule.ports.from < rule.ports.to &&
+          self.rule.ports.from > rule.ports.from
+        ) {
+          return true;
```
darsene, line 359: you can also split it in sub functions or at least comment what are the different cases

darsene (dismissed):

dectotam (dismissed):

ghyenne (approved):

PR #10936, opened 2024-02-16
feat(web): update dns dashboard

vmesnyovh:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | feat #MANAGER-13089

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Adds the ability to translate true/false DNS zones and change column titles

### Translations
- [x] CMT-9733

### DIFF
--- a/packages/manager/apps/web/client/app/dns-zone/zone.routing.js
+++ b/packages/manager/apps/web/client/app/dns-zone/zone.routing.js
@@ -9,11 +9,11 @@ export default /* @ngInject */ ($stateProvider) => {
       transition
         .injector()
         .getAsync('resources')
-        .then((resources) =>
-          resources.data.length === 0
+        .then((resources) => {
+          return resources.data.length === 0
             ? { state: 'app.zone.onboarding' }
-            : false,
-        ),
+            : false;
+        }),
     resolve: {
       ...ListLayoutHelper.stateResolves,
       schema: /* @ngInject */ ($http) =>
@@ -23,6 +23,52 @@ export default /* @ngInject */ ($stateProvider) => {
       defaultFilterColumn: () => 'name',
       header: /* @ngInject */ ($translate) => $translate.instant('zones_title'),
       customizableColumns: () => true,
+      columns: /* @ngInject */ ($translate) => [
+        {
+          property: 'name',
+          title: $translate.instant('zones_domain_name'),
+          template: `<a data-ng-href="{{ $ctrl.getServiceNameLink($row) }}" data-ng-bind="$row.name"></a>`,
+          searchable: true,
+          sortable: 'asc',
+          filterable: true,
+        },
+        {
+          property: 'hasDnsAnycast',
+          title: 'DNS Anycast',
+          template: `<span data-ng-bind="$row.hasDnsAnycast ? '${$translate.instant(
+            'zones_enabled',
+          )}' : '${$translate.instant(
+            'zones_disabled',
+          )}'" class="oui-badge" ng-class="{'oui-badge_success': $row.hasDnsAnycast, 'oui-badge_error': !$row.hasDnsAnycast}"></span>
+          `,
+          searchable: true,
+          sortable: true,
+          filterable: true,
+          type: 'boolean',
+          typeOptions: {
+            trueValue: $translate.instant('zones_enabled'),
+            falseValue: $translate.instant('zones_disabled'),
+          },
+        },
+        {
+          property: 'dnssecSupported',
+          title: 'DNSSEC',
+          template: `<span data-ng-bind="$row.dnssecSupported ? '${$translate.instant(
+            'zones_enabled',
+          )}' : '${$translate.instant(
+            'zones_disabled',
+          )}'" class="oui-badge" ng-class="{'oui-badge_success': $row.dnssecSupported, 'oui-badge_error': !$row.dnssecSupported}"></span>
+          `,
+          searchable: true,
+          sortable: true,
+          filterable: true,
+          type: 'boolean',
+          typeOptions: {
+            trueValue: $translate.instant('zones_enabled'),
+            falseValue: $translate.instant('zones_disabled'),
+          },
+        },
+      ],
       getServiceNameLink: /* @ngInject */ ($state) => ({ name: productId }) =>
         $state.href('app.zone.details', {
           productId,
--- a/packages/manager/apps/web/client/app/dns-zone/zone.routing.js
+++ b/packages/manager/apps/web/client/app/dns-zone/zone.routing.js
@@ -9,11 +9,9 @@ export default /* @ngInject */ ($stateProvider) => {
       transition
         .injector()
         .getAsync('resources')
-        .then((resources) => {
-          return resources.data.length === 0
-            ? { state: 'app.zone.onboarding' }
-            : false;
-        }),
+        .then(({ data }) =>
+          data.length === 0 ? { state: 'app.zone.onboarding' } : false,
+        ),
     resolve: {
       ...ListLayoutHelper.stateResolves,
       schema: /* @ngInject */ ($http) =>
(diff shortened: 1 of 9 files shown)


packages/manager/apps/web/client/app/dns-zone/zone.routing.js
```diff
@@ -9,11 +9,11 @@ export default /* @ngInject */ ($stateProvider) => {
       transition
         .injector()
         .getAsync('resources')
-        .then((resources) =>
-          resources.data.length === 0
+        .then((resources) => {
```
sachinrameshn, line 12: This change is not needed.

sachinrameshn (changes requested):

packages/manager/apps/web/client/app/dns-zone/zone.routing.js
```diff
@@ -23,6 +21,52 @@ export default /* @ngInject */ ($stateProvider) => {
...
+          property: 'name',
+          title: $translate.instant('zones_domain_name'),
+          template: `<a data-ng-href="{{ $ctrl.getServiceNameLink($row) }}" data-ng-bind="$row.name"></a>`,
+          searchable: true,
+          sortable: 'asc',
+          filterable: true,
+        },
+        {
+          property: 'hasDnsAnycast',
+          title: 'DNS Anycast',
```
anooparveti, line 35: **question:** Dont we need translation for this?

packages/manager/apps/web/client/app/dns-zone/zone.routing.js
```diff
@@ -23,6 +21,52 @@ export default /* @ngInject */ ($stateProvider) => {
       defaultFilterColumn: () => 'name',
       header: /* @ngInject */ ($translate) => $translate.instant('zones_title'),
       customizableColumns: () => true,
+      columns: /* @ngInject */ ($translate) => [
+        {
+          property: 'name',
+          title: $translate.instant('zones_domain_name'),
+          template: `<a data-ng-href="{{ $ctrl.getServiceNameLink($row) }}" data-ng-bind="$row.name"></a>`,
```
anooparveti, line 28: **nitpick:** `resolve` is executed before the controller is initialised, so using a function defined in the controller can be misleading. Can you define `getServiceNameLink` in the resolve and inject it here and use.

anooparveti (changes requested):

packages/manager/apps/web/client/app/dns-zone/zone.routing.js
vmesnyovh: Hello @anooparveti ,
No, `DNS Anycast` and `DNSSEC` are considered generic and therefore do not need to be translated.

packages/manager/apps/web/client/app/dns-zone/zone.routing.js
vmesnyovh: Hello @anooparveti 

Thank you for the review and the suggestion regarding the `getServiceNameLink` function. 
I understand the intention behind moving this function to the resolve section to ensure all dependencies are ready before the controller initializes. 
However, there are a couple of reasons why this approach might not be feasible in this context:

1. Execution Context: The getServiceNameLink function relies on the `$state` service to dynamically generate state-based links, which inherently requires access to the current execution context, including the controller's scope and potentially the current application state. Since the resolve block is executed before the controller is instantiated, it doesn't have access to the controller's context or any runtime state data that might influence link generation.

2. Service Usage: While we could inject the `$state` service into the resolve block, using it to dynamically generate links within resolve is somewhat unconventional. The primary purpose of resolve is to prepare and provide data or static dependencies necessary for the controller, rather than to serve dynamic functions like link generation based on the application's runtime state.

Given these considerations, `getServiceNameLink` might be more appropriately defined within the controller or as a method of a service that can be injected into the controller. 
This would ensure that it has access to the necessary context and services to function as intended.

I hope this clarifies why moving `getServiceNameLink` to the resolve section might not be the best approach in this scenario. If there are other concerns or alternative suggestions, I'm open to discussing them further.

vmesnyovh: Hello @anooparveti ,

Thank you for the follow-up and the suggestion to directly inject `getServiceNameLink` into the columns definition. 
I tried implementing the suggested change.

However, I encountered an issue with this approach. It seems that the function `getServiceNameLink` isn't directly accessible within the template string in this context. This might be due to the way AngularJS compiles and links templates, where the scope within `ng-href` might not have direct access to resolved dependencies in the same way the controller does.

Given this, while the function is indeed defined in the `resolve` section, it appears that its direct use within the template string doesn't work as expected. This could be related to AngularJS's scope hierarchy and how it resolves bindings in templates, especially when it comes to functions and their contexts.

To maintain the functionality and ensure that the links are correctly generated, it might be necessary to keep the `$ctrl.getServiceNameLink` reference in the template. This ensures the function is called within the correct scope and has access to all necessary dependencies.

If there's a workaround or a different approach that could make this work as intended, I'd be keen to try it out.

Thank you again for your review and guidance.

sachinrameshn (changes requested):
Hello,  Can you please add ticket number reference in your commit message ?

feat(universe): description
   ref: MANAGER-####

Signed-off-by:  XXXX <XXXX@corp.ovh.com>

dectotam (approved):

vmesnyovh:
> Hello, Can you please add ticket number reference in your commit message ?
> 
> feat(universe): description ref: MANAGER-####
> 
> Signed-off-by: XXXX [XXXX@corp.ovh.com](mailto:XXXX@corp.ovh.com)

Hello @sachinrameshn ,

Thanks for the reminder, it's done

packages/manager/apps/web/client/app/dns-zone/zone.routing.js
```diff
@@ -23,6 +21,52 @@ export default /* @ngInject */ ($stateProvider) => {
...
+        },
+        {
+          property: 'hasDnsAnycast',
+          title: 'DNS Anycast',
+          template: `<span data-ng-bind="$row.hasDnsAnycast ? '${$translate.instant(
+            'zones_enabled',
+          )}' : '${$translate.instant(
+            'zones_disabled',
+          )}'" class="oui-badge" ng-class="{'oui-badge_success': $row.hasDnsAnycast, 'oui-badge_error': !$row.hasDnsAnycast}"></span>
+          `,
```
JacquesLarique, line 41:
```suggestion
          template: `<span data-ng-bind="'${$translate.instant($row.hasDnsAnycast ? 'zones_enabled' : 'zones_disabled')}'" class="oui-badge" ng-class="{'oui-badge_success': $row.hasDnsAnycast, 'oui-badge_error': !$row.hasDnsAnycast}"></span>
          `,
```
I think this is easier to read, what do you think ?

packages/manager/apps/web/client/app/dns-zone/zone.routing.js
```diff
@@ -23,6 +21,52 @@ export default /* @ngInject */ ($stateProvider) => {
...
+        },
+        {
+          property: 'dnssecSupported',
+          title: 'DNSSEC',
+          template: `<span data-ng-bind="$row.dnssecSupported ? '${$translate.instant(
+            'zones_enabled',
+          )}' : '${$translate.instant(
+            'zones_disabled',
+          )}'" class="oui-badge" ng-class="{'oui-badge_success': $row.dnssecSupported, 'oui-badge_error': !$row.dnssecSupported}"></span>
+          `,
```
JacquesLarique, line 59: Same as https://github.com/ovh/manager/pull/10936#discussion_r1505653330

packages/manager/apps/web/client/app/dns-zone/zone.routing.js
vmesnyovh: Hello @JacquesLarique ,
It seems easier but I can't get the code to work with this method or via translate as below, but I'd have to change the rest to keep it consistent:

```js
template: `<span data-ng-bind="'{{ $row.hasDnsAnycast ? 'zones_enabled' : 'zones_disabled' | translate }}'" class="oui-badge" ng-class="{'oui-badge_success': $row.hasDnsAnycast, 'oui-badge_error': !$row.hasDnsAnycast}"></span>`

```

anooparveti (dismissed):

darsene (approved):

PR #11031, opened 2024-02-27
feat(network-security): add params into traffic url

Steffy29:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #UXCT-532

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Add parameters into traffic URL to use this URL from directly from email for example.


packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.controller.js
```diff
@@ -40,6 +41,14 @@ export default class TrafficController {
       this.model = this.selectedIp;
       this.getTraffic();
     }
+    this.ip = this.$stateParams.ip;
+    this.dateTime = this.$stateParams.dateTime
```
ghyenne, line 45:
```suggestion
```
No need because this.ip and this.dateTime is already defined by routing an binded in component so you can directly call this.ip and this.dateTime without get in $stateParams again.

packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.routing.js
```diff
@@ -1,14 +1,24 @@
...
+    })
+    .state('network-security.traffic.ip', {
+      url: '?ip&dateTime',
+      component: 'traffic',
+      resolve: {
+        ip: /* @ngInject */ ($transition$) => $transition$.params().ip,
+        dateTime: /* @ngInject */ ($transition$) =>
+          $transition$.params().dateTime,
+      },
+    });
```
anooparveti, line 23: You're using the same controller and template.Please use the existing state with optional state params.

anooparveti (dismissed):

packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.routing.js
Steffy29: I removed this resolve to keep the state params which work well.

packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.routing.js
```diff
@@ -1,14 +1,19 @@
...
+      resolve: {
+        breadcrumb: /* @ngInject */ ($translate) =>
+          $translate.instant('network_security_dashboard_breadcrumb'),
+        getSubnet: /* @ngInject */ ($state) => () => $state.params.subnet,
+      },
+    })
+    .state('network-security.traffic.ip', {
+      url: '?ip&dateTime',
+      component: 'traffic',
+    });
```
ghyenne, line 18: Why 2 states for same component ?
And in you component file you bind ip and dateTime that not resolve in routing but directly in you controller. I think we can simplify.

ghyenne (changes requested):

packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.routing.js
Steffy29: I'm ok to simplify but all I'm trying doesn't work but the way I took worked. Detailed advice are welcome to make a simple solution.

packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.component.js
```diff
@@ -4,6 +4,8 @@ import template from './traffic.html';
 export default {
   bindings: {
     getSubnet: '<',
+    ip: '<',
+    dateTime: '<',
```
ghyenne, line 8:
```suggestion
```
No need because you call directly $stateParams in your controller also ip and dataTime doesn't injected in the routing.

ghyenne (approved):

packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.controller.js
```diff
@@ -34,11 +35,30 @@ export default class TrafficController {
...
+      dateLimit.setDate(dateLimit.getDate() - 14);
+      if (this.dateTime >= dateLimit) {
+        const customPeriod = {
+          name: 'custom',
+          label: this.$translate.instant(
+            'network_security_dashboard_filter_custom',
+          ),
+        };
+        this.periods.push(customPeriod);
+        this.period = this.periods[this.periods.length - 1];
```
sachinrameshn, line 53: Performance of slice is better
```suggestion
        this.period = this.periods.slice(-1);
```

PR #11057, opened 2024-02-29
fix: ldp streams action menu: better wording and copy stream id

pfrayer:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Ticket | MANAGER-12517 / CMT-10872

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

In Logs Data Platform → Data Stream, this PR enhance the action menu by:

- having a better wording ("xxx" becomes "Manage xxx" everywhere, not for just one item)
- add a "Copy stream Id" option, which was requested by some users


qpavy (dismissed):

packages/manager/modules/dbaas-logs/src/logs/detail/streams/logs-streams.service.js
```diff
@@ -425,6 +425,30 @@ export default class LogsStreamsService {
...
   }
 
+  /**
+   * copies stream ID to clipboard.
+   *
+   * @param {any} stream
+   * @memberof LogsStreamsService
+   */
+  copyStreamId(stream) {
+    const error = this.CucControllerHelper.constructor.copyToClipboard(
```
dectotam, line 435: the copyToClipboard function doesn't work on safari (cf DTRSD-126045 issue). 

In the CucControllerHelper.copyToClipboard function, the function needs to be enriched with the navigator.clipboard API usage and keep the current one as a fallback if navigator api is not supported by the user's browser : 

```   
static copyToClipboard(messageToCopy) {
 // check if API available on browser
  if (navigator.clipboard && navigator.clipboard.writeText) {
    return navigator.clipboard.writeText(messageToCopy)
      .then(() => '')
      .catch(err => err);
  } else {
  // current usage
    try {
      const dummy = document.createElement('input');
      document.body.appendChild(dummy);
      dummy.setAttribute('id', 'dummy_id');
      dummy.setAttribute('value', messageToCopy);
      dummy.select();
      document.execCommand('copy');
      document.body.removeChild(dummy);
    } catch (err) {
      return err;
    }
    return '';
  }
}
```

pfrayer: I applied this change with a small diff, to always return an empty string in case of success and a not-empty string in case of failure (and not a promise). 
Otherwise it would have break current usage of this function

dectotam (approved):

qpavy (approved):

pfrayer:
Translations pulled

pfrayer:
Rebased, fixed conflicts, added translations ticket to PR description. QA validated

dectotam (approved):

qpavy (approved):

PR #11110, opened 2024-03-06
feat(pci-users): add unit tests for download actions

kqesar:
ref: DTCORE-1873

Add uni tests for download actions modals


sidlynx (approved):

packages/manager/apps/pci-users/src/__tests__/components/S3StorageRegions.spec.tsx
```diff
@@ -0,0 +1,68 @@
...
+        onS3StorageRegionChange={() => {}}
+      ></S3StorageRegions>
+      ,
+    </QueryClientProvider>,
+  );
+}
+
+describe('S3StorageRegions components', () => {
+  it('should call the S3StorageRegions component', async () => {
+    renderComponent();
```
frenautvh, line 66: it looks like you are doing a render but not checking any value, so this test will always pass
what do you want to test ?

packages/manager/apps/pci-users/src/__tests__/components/StorageRegions.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+        onStorageRegionChange={() => {}}
+      ></StorageRegions>
+      ,
+    </QueryClientProvider>,
+  );
+}
+
+describe('StorageRegions components', () => {
+  it('should call the StorageRegions component', async () => {
+    renderComponent();
```
frenautvh, line 65: same issue as in S3StorageRegions, this test will always pass since there are no checks performed

frenautvh (changes requested):

PR #11136, opened 2024-03-08
Feat/add user modal

sidlynx:
ref: DTCORE-1859

| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/pci-users`
| Bug fix?         | yes
| New feature?     | yes
| Breaking change? | no
| Tickets          | DTCORE-1859

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Add the user delete modal and update ui when a user is deleted

### DIFF
--- a/packages/manager/apps/pci-users/src/pages/add/Index.tsx
+++ b/packages/manager/apps/pci-users/src/pages/add/Index.tsx
@@ -1,0 +1,198 @@
+import { useNavigate, useParams } from 'react-router-dom';
+import { useTranslation } from 'react-i18next';
+import { useNotifications } from '@ovhcloud/manager-components';
+import { useEffect, useState } from 'react';
+import {
+  OsdsButton,
+  OsdsFormField,
+  OsdsInput,
+  OsdsModal,
+  OsdsSpinner,
+  OsdsText,
+} from '@ovhcloud/ods-components/react';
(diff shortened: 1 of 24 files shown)


qpavy:
The ticket reference seems not correct and you should put the ticket reference in the commit as well in this form: 
ref: MANAGER-XXXXX

Also, I see you have translation files commited but I don't see any translation commit and you shouldn't have push to translation before review / QA

packages/manager/apps/pci-users/src/pages/add/Index.tsx
```diff
@@ -0,0 +1,198 @@
...
+    <OsdsModal
+      headline={
+        !state.isNextStepReached
+          ? tCommon('pci_projects_project_users_add_title')
+          : tCommon('pci_projects_project_users_roles')
+      }
+      onOdsModalClose={modalActions.close}
+    >
+      <slot name="content">
+        {(() => {
```
frenautvh, line 103: why do you need to (() => {... your code ...})() ? do you need a closed scope ?

frenautvh (approved):

packages/manager/apps/pci-users/src/pages/add/Index.tsx
sidlynx: no, to avoid nested ternaries, that's all

kqesar (approved):

PR #11166, opened 2024-03-13
fix(pci-users): update tsconfig and use ParseInt instead of cast params

kqesar:
Update tsconfig and use parseInt instead of cast for searchParams

### DIFF
--- a/packages/manager/apps/pci-users/src/pages/EditRolesPage.tsx
+++ b/packages/manager/apps/pci-users/src/pages/EditRolesPage.tsx
@@ -12,7 +12,7 @@ export default function EditRolesPage() {
   const onClose = () => {
     navigate('..');
   };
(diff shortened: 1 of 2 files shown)


packages/manager/apps/pci-users/src/pages/EditRolesPage.tsx
```diff
@@ -12,7 +12,7 @@ export default function EditRolesPage() {
   const onClose = () => {
     navigate('..');
   };
-  const userId = searchParams.get('userId') as number | null;
+  const userId = parseInt(searchParams.get('userId'), 10) || null;
```
frenautvh, line 15: i think it won't matter in this case but keep in mind that "0" will be "null" with this

sidlynx (approved):

seven-amid (approved):

PR #11226, opened 2024-03-19
feat(web): cdn funnel for express order

antonymarion:
ref: MANAGER-13728

| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Feat#MANAGER-13728

- [x] Commits are signed-off
~~- [ ] Only FR translations have been updated~~
- [x] Lint has passed locally
~~- [ ] Breaking change is mentioned in relevant commits~~

### DIFF
--- a/packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
+++ b/packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
@@ -1,6 +1,7 @@
 export default class HostingCdnOrderCtrl {
   $onInit() {
     this.workflowOptions.getPlanCode = () => this.getPlanCode();
+    this.workflowOptions.expressOrder = true;
 
     // Preselect CDN Advanced
     this.cdnPlanCode = this.planToPreselect;
--- a/packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
+++ b/packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
@@ -1,7 +1,6 @@
 export default class HostingCdnOrderCtrl {
   $onInit() {
     this.workflowOptions.getPlanCode = () => this.getPlanCode();
-    this.workflowOptions.expressOrder = true;
 
     // Preselect CDN Advanced
     this.cdnPlanCode = this.planToPreselect;
--- a/packages/manager/modules/product-offers/src/product-offers.html
+++ b/packages/manager/modules/product-offers/src/product-offers.html
@@ -65,6 +65,15 @@
         ></p>
     </oui-step-form>
 
+    <button
+        type="button"
+        class="oui-button oui-button_l oui-button_primary"
+        data-ng-click="$ctrl.workflow.validateOffer()"
+        data-ng-if="$ctrl.workflowOptions.expressOrder && $ctrl.workflow.pricings.length === 1"
+    >
+        <span data-translate="ovh_product_offers_order_express_submit"></span>
+    </button>
+
     <!-- Payment -->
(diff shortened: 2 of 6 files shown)


JacquesLarique (approved):

packages/manager/modules/product-offers/src/workflows/product-offers-order-workflow.class.js
```diff
@@ -251,12 +251,21 @@ export default class OrderWorkflow extends Workflow {
...
       const checkoutObject = this.getCheckoutInformations();
+      const { configuration } = checkoutObject;
+      const serviceName =
+        typeof this.serviceNameToAddProduct === 'function'
+          ? this.serviceNameToAddProduct()
+          : this.serviceNameToAddProduct;
       const jsUrlToSend = {
         productId,
-        configuration: checkoutObject.configuration,
+        ...(configuration.length > 0 ? { configuration } : []),
```
darsene, line 266: `configuration.length > 0 ` means you always have an empty array at list (can't be null). Why creating a copy here ?
```suggestion
        ...{ configuration },
```

packages/manager/modules/product-offers/src/workflows/product-offers-order-workflow.class.js
```diff
@@ -251,12 +251,21 @@ export default class OrderWorkflow extends Workflow {
...
         ...checkoutObject.product,
+        ...(serviceName ? { serviceName } : []),
```
darsene, line 268: mixing object and array is a bit weird

packages/manager/apps/web/client/app/hosting/cdn/order/hosting-cdn-order.controller.js
```diff
@@ -1,6 +1,7 @@
 export default class HostingCdnOrderCtrl {
   $onInit() {
     this.workflowOptions.getPlanCode = () => this.getPlanCode();
+    this.workflowOptions.expressOrder = true;
```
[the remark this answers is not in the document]
antonymarion: Yes we can :)

image

packages/manager/modules/product-offers/src/workflows/product-offers-order-workflow.class.js
antonymarion: Yes we always have an array (never null)
image

antonymarion: I do not want to have in my generated JSURL a configuration property visible inside that's why I do this weird object or empty array with ternary condition before

packages/manager/modules/product-offers/src/workflows/product-offers-order-workflow.class.js
antonymarion: Same as previous

packages/manager/modules/product-offers/src/product-offers.html
```diff
@@ -65,6 +65,15 @@
...
     </oui-step-form>
 
+    <button
+        type="button"
+        class="oui-button oui-button_l oui-button_primary"
+        data-ng-click="$ctrl.workflow.validateOffer()"
+        data-ng-if="$ctrl.workflowOptions.expressOrder && $ctrl.workflow.pricings.length === 1"
+    >
+        <span data-translate="ovh_product_offers_order_express_submit"></span>
+    </button>
```
ghyenne, line 75: I think we don't need create new button but we can change the rule for don't automaticaly select the price if we have only one price in case where we are in expressOrder configuration

if (this.hasUniquePricing() && !this.expressOrder) {
in class order / service

PR #11256, opened 2024-03-21
fix(hub): make dashboard usable without rights on /services

rjamet-ovh:
ref: 
| Question         | Answer
| ---------------- | ---
| Branch?          | master
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-13620

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

fix users without right on /services being redirected on ErrorPage.


darsene (dismissed):

packages/manager/modules/hub/src/components/catalog-items/controller.js
```diff
@@ -11,7 +11,11 @@ export default class {
 
       this.itemsPromise
         .then((products) => {
-          this.items = products;
+          if (Object.keys(products).length > 0) {
+            this.items = products;
+          } else {
+            this.items = undefined;
+          }
```
dectotam, line 18:
```suggestion
          this.items = Object.keys(products).length > 0 ? products : undefined;   
```

dectotam (approved):

packages/manager/modules/hub/src/components/catalog-items/controller.js
```diff
@@ -11,7 +11,7 @@ export default class {
 
       this.itemsPromise
         .then((products) => {
-          this.items = products;
+          this.items = Object.keys(products).length > 0 ? products : undefined;
```
anooparveti, line 14: **nitpick:** use `null` instead.
```suggestion
          this.items = Object.keys(products).length > 0 ? products : null;
```

dectotam (approved):

anooparveti (changes requested):
**question:** Where is it mentioned to remove the display of catalog items if user does not have any product? I checked the ticket and could not find the trace of it.

packages/manager/apps/hub/src/routing.js
```diff
@@ -50,11 +50,8 @@ export default /* @ngInject */ ($stateProvider, $urlRouterProvider) => {
           ),
       numberOfServices: /* @ngInject */ ($http) =>
         $http
-          .get('/services', {
-            headers: {
-              'X-Pagination-Mode': 'CachedObjectList-Pages',
-              'X-Pagination-Size': 5,
-            },
+          .get('/hub/services', {
+            serviceType: 'aapi',
```
anooparveti, line 54: **nitpick:** This change can be costly because `hub` 2API fetches the entire list of expanded services and this can take more time when the user has a huge set of services. Moreover, we dont use the data, so we are over fetching. Instead still continue to use `/services` as shown below.
```suggestion
          .get('/services', {
                headers: {
                  'X-Pagination-Size': 5,
                },
            })
            .then((data) => return data.data.length)
            .catch(() => return null);
```

anooparveti (changes requested):

packages/manager/apps/hub/src/dashboard/routing.js
```diff
@@ -43,6 +30,19 @@ export default /* @ngInject */ ($stateProvider, $urlRouterProvider) => {
...
+            serviceType: 'aapi',
+          })
+          .then((data) => {
+            const { catalog } = data.data.data;
+            return groupBy(
+              filter(catalog.data, ({ highlight }) => highlight),
+              'universe',
+            );
+          });
+      },
```
anooparveti, line 45: **nitpick:** This results in over-fetching since we will be fetching catalog items for every entry to the dashboard and is not used when the user has ordered at least 1 product.

anooparveti (changes requested):

packages/manager/apps/hub/src/dashboard/routing.js
rjamet-ovh: You're right. I can move the call to `/hub/catalog` in the `catalog-items` component to prevent over-fetching.

darsene (approved):

anooparveti (approved):

anooparveti (approved):

qpavy (approved):

PR #11344, opened 2024-04-03
fix(dedicated): fix resiliated 3az node

qpavy:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #INC0025392

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Display detailled error to inform the user his service is expired when a AZ cluster is resiliated
Screenshot from 2024-04-03 10-33-10

### DIFF
--- a/packages/manager/modules/bm-server-components/src/network-interfaces/interfaces.service.js
+++ b/packages/manager/modules/bm-server-components/src/network-interfaces/interfaces.service.js
@@ -66,6 +66,12 @@ export default class DedicatedServerInterfacesService {
   }
 
   getInterfaces(serverName, specifications) {
+    if (
+      specifications.constructor === Object &&
+      Object.keys(specifications).length === 0
+    )
+      return [];
+
     let nics;
     this.typeVrack = specifications.vrack.type;
 


packages/manager/apps/dedicated/client/app/dedicated/dedicated-server/clusters/nodes/node/node.routing.js
```diff
@@ -20,10 +20,13 @@ export default /* @ngInject */ ($stateProvider) => {
         DedicatedServerInterfacesService,
         specifications,
       ) =>
-        DedicatedServerInterfacesService.getInterfaces(
-          serverName,
-          specifications,
-        ),
+        specifications.constructor === Object &&
+        Object.keys(specifications).length !== 0
```
darsene, line 24: What is the value of specifications which forces that check ?
Moreover, you should move the check directly in the service to avoid duplication your code if the service is called in different places

qpavy: the value is {} created by this line https://github.com/ovh/manager/blob/master/packages/manager/modules/bm-server-components/src/server/server.service.js#L1109 ok I move it in the service

anooparveti (approved):

PR #11361, opened 2024-04-03
fix(rancher): disable refetch pooling on create page

lionel95200x:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | yes/no
| Tickets          | Fix #MANAGER-14087

- [ ] Commits are signed-off
- [ ] Lint has passed locally

## Description

Disable hook that refetch rancher on Create Page

### DIFF
--- a/packages/manager/apps/pci-rancher/src/hooks/useRancher.tsx
+++ b/packages/manager/apps/pci-rancher/src/hooks/useRancher.tsx
@@ -21,20 +21,20 @@ export const useRancher = () => {
 export const ranchersQueryKey = (projectId: string) => ['project', projectId];
 
 export const useRanchers = ({
-  disablePolling,
+  shouldDisableRefetch = false,
 }: {
-  disablePolling?: boolean;
+  shouldDisableRefetch?: boolean;
 }) => {
   const { projectId } = useParams();
   return useQuery<RancherService[], ErrorResponse>({
     queryKey: ranchersQueryKey(projectId),
     queryFn: () => getRancherProjectById(projectId),
-    refetchInterval: disablePolling ? false : 5000,
+    refetchInterval: shouldDisableRefetch ? false : 5000,
   });
 };
 
 export const useDeleteRancher = () => {
-  const { refetch } = useRanchers({ disablePolling: true });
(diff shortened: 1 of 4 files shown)


packages/manager/apps/pci-rancher/src/pages/create/index.tsx
```diff
@@ -35,7 +35,9 @@ export default function Create() {
 
   const { data: project } = usePciProject();
 
-  const { data: ranchers, refetch: refetchRancherList } = useRanchers();
+  const { data: ranchers, refetch: refetchRancherList } = useRanchers({
+    shouldDisableRefetch: true,
+  });
```
anooparveti, line 40: **nitpick:** We should make an HTTP call just to decide where should we take user on cancel click. This can be tackled in 2 ways,
1. In `CreateRancher` component `onCancelClick` function,  you can do `navigate(-1)`
2. Otherwise, you must have check in the Rancher listing component and in the `useEffect({}, [])` if the Rancher list is empty, redirect user to onboarding page.

frenautvh: don't know if that help but navigate('..') would navigate to the parent route no matter where the user comes from

vovh (approved):

aboungnaseng-ovhcloud (approved):

packages/manager/apps/pci-rancher/src/hooks/useRancher.tsx
```diff
@@ -13,11 +13,15 @@ export const useRancher = () => {
 
 export const ranchersQueryKey = (projectId: string) => ['project', projectId];
 
-export const useRanchers = () => {
+export const useRanchers = ({
+  shouldDisableRefetch = false,
```
anooparveti, line 17: **question:** `shouldDisableRefetch` is not sent as true from anywhere no? Isn't it redundant?

anooparveti (approved):

vovh (approved):

anooparveti (approved):

aboungnaseng-ovhcloud (approved):

PR #11438, opened 2024-04-11
feat(manager-components): pci project context provider

anooparveti:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-13936

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

 Addition of Public Cloud project provider and hooks to store and fetch project info

### DIFF
--- a/packages/manager-components/package.json
+++ b/packages/manager-components/package.json
@@ -2,7 +2,7 @@
   "name": "@ovhcloud/manager-components",
   "private": true,
   "version": "1.12.0",
-  "main": "src/components/index.ts",
+  "main": "src/lib.ts",
   "repository": {
     "type": "git",
     "url": "git+https://github.com/ovh/manager.git",
@@ -90,6 +90,9 @@
     "react-i18next": "^11.18.1",
     "react-router-dom": "^6.21.0",
     "zustand": "^4.5.0"
+  },
+  "peerDependencies": {
+    "@tanstack/react-query": "^5.18.1"
   },
   "files": [
     "dist"
--- a/packages/manager-components/src/hooks/index.ts
+++ b/packages/manager-components/src/hooks/index.ts
@@ -1,0 +1,1 @@
+export * from './pci-project-provider';
--- a/packages/manager-components/src/hooks/pci-project-provider/constants.ts
+++ b/packages/manager-components/src/hooks/pci-project-provider/constants.ts
@@ -1,0 +1,1 @@
+export const DISCOVERY_PROJECT_PLANCODE = 'project.discovery';
--- a/packages/manager-components/src/hooks/pci-project-provider/index.ts
+++ b/packages/manager-components/src/hooks/pci-project-provider/index.ts
@@ -1,0 +1,3 @@
+import { useProject, useIsDiscoveryProject } from './useProject';
+
+export { useProject, useIsDiscoveryProject };
--- a/packages/manager-components/src/hooks/pci-project-provider/publicCloudProject.interface.ts
+++ b/packages/manager-components/src/hooks/pci-project-provider/publicCloudProject.interface.ts
@@ -1,0 +1,17 @@
+export interface PublicCloudProject {
+  access: string;
+  creationDate: string;
+  description: string;
+  expiration: string;
+  iam: {
+    id: string;
+    urn: string;
+  };
+  orderId: number;
+  planCode: string;
+  project_id: string;
+  projectName: string;
+  unleash: boolean;
+  status: string;
+  manualQuota: boolean;
+}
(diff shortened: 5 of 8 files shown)


packages/manager-components/src/components/public-cloud-project-provider/PublicCloudProjectProvider.tsx
```diff
@@ -0,0 +1,28 @@
+import { createContext, useContext } from 'react';
+
+import { getIsDiscoveryProject } from './util';
+import { PublicCloudProject } from './publicCloudProject';
+
+const PublicCloudProjectContext = createContext(null);
+
+export const PublicCloudProjectProvider = ({
```
frenautvh, line 8: there is a discussion regarding usage of Context & react-query :
https://github.com/TanStack/query/discussions/3859

i think it *could* make sense for project since it's used basically everywhere in PCI but still i'm not sure if it's worth it when using react query

also this comment is interesting : "One thing to keep in mind is to memoize the children to the Provider to avoid unnecessarily re-rendering the whole tree, and you also cannot do fine grained subscriptions like you can do with e.g. the select option on useQuery. Tradeoffs everywhere "

packages/manager-components/src/components/public-cloud-project-provider/PublicCloudProjectProvider.tsx
```diff
@@ -0,0 +1,28 @@
...
+  return (
+    <PublicCloudProjectContext.Provider value={project}>
+      {children}
+    </PublicCloudProjectContext.Provider>
+  );
+};
+
+export const usePublicCloudProject = () => {
+  return useContext(PublicCloudProjectContext);
+};
```
chipp972, line 24: Why create a hook to just call another hook ?

packages/manager-components/src/components/public-cloud-project-provider/PublicCloudProjectProvider.tsx
chipp972: I don't think it's worth it either in our case since they talk about having hundred or thousands of useQuery.
It would make more sense to abstract the useQuery and return utils directly :

```
const query = useQuery(projectQueryData);

return {
   ...query,
   isDiscoveryProject: someFn(data)
};
```

packages/manager-components/src/components/public-cloud-project-provider/PublicCloudProjectProvider.tsx
anooparveti: The idea behind creating this provider is to,
1. provide an opportunity to developer to set the project information.
2. get the project information or any related information like `isDiscoveryProject` from this provider.

So, this implementation is done to mask any react or other external library specific implementations and also provide an easy way of integration.
Without this utility, developer will have to import `useContext` from react + `PublicCloudProjectContext` from manager-components.

packages/manager-components/src/components/public-cloud-project-provider/PublicCloudProjectProvider.tsx
anooparveti: I still feel using context can reduce the effort to get the project information except the initial set-up part. Lets say we create the custom hook to get the project details. So whenever you need to get the details, you end up doing the following,

```
import { useParams } from 'react-router-dom';
import { customHook } from 'manager-components';

// inside the component

const { projectId } = useParams();
const { isDiscoveryProject } = customHook(projectId);
```

But with context and provider, the implementation is reduced to half,

```
import { useIsDiscoveryProject }  from 'manager-components';

// inside the component

const isDiscoveryProject = useIsDiscoveryProject();
```

This reduces code duplication.

chipp972: You could call useParams in the customHook from manager-components instead

packages/manager-components/src/components/public-cloud-project-provider/useProject.ts
```diff
@@ -0,0 +1,47 @@
...
+  },
+});
+
+export const useProject = (
+  projectId?: string,
+  opt?: QueryOptions<PublicCloudProject, ResponseAPIError>,
+) => {
+  const { projectId: deafaultProjectId } = useParams();
+  return useQuery({
+    ...getProjectQuery(projectId || deafaultProjectId),
```
chipp972, line 39:
```suggestion
    ...getProjectQuery(projectId || defaultProjectId),
```

chipp972 (dismissed):

chipp972 (approved):

anooparveti:
> Ok for me but i think it will be better to add on same PR the hook on a existing project to check if everything is ok Because now it's not used by any project ..

@lionel95200x as mentioned yesterday, the scope of the PR is to implement the capabilities and update the documentation on manager pages. And it is done.

chipp972 (approved):

qpavy (approved):

PR #11451, opened 2024-04-12
fix(iam): infinite loop on iam menu

dectotam:
| Question         | Answer
| ---------------- | ---
| Branch?          | master
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Feat/MANAGER-13653

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

If user isn't granted to access iam policies page, there is an infinite loop. This PR adds a banner displaying an info message to inform the user he can't access the page.

### Translations
- [ ] CMT-11163

### DIFF
--- a/packages/manager/modules/iam/src/dashboard/dashboard.routing.js
+++ b/packages/manager/modules/iam/src/dashboard/dashboard.routing.js
@@ -2,8 +2,27 @@ export default /* @ngInject */ ($stateProvider) => {
   $stateProvider.state('iam.dashboard', {
     url: '/dashboard',
     component: 'iamDashboard',
-    redirectTo: 'iam.dashboard.policies',
+    redirectTo: (transition) =>
+      transition
+        .injector()
+        .getAsync('iamPolicies')
+        .then((data) =>
+          !data?.error ? { state: 'iam.dashboard.policies' } : false,
+        )
+        .catch(() => null),
     resolve: {
+      iamPolicies: /* @ngInject */ (IAMService) =>
+        IAMService.getPolicies({})
+          .then((policies) => policies)
+          .catch((error) => {
+            if (error.status === 403) {
+              return {
+                error: error.data?.details?.unauthorizedActionsByIAM,
+              };
+            }
+
+            throw error;
+          }),
       breadcrumb: () => null,
     },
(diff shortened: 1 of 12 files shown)


packages/manager/modules/iam/src/dashboard/dashboard.routing.js
```diff
@@ -2,8 +2,29 @@ export default /* @ngInject */ ($stateProvider) => {
...
+              details: null,
+            };
+          })
+          .catch(({ data }) => {
+            return {
+              authorized: false,
+              details: data?.details?.unauthorizedActionsByIAM,
+            };
+          });
+      },
```
ghyenne, line 27:
```suggestion
      iam: /* @ngInject */ (IAMService) => IAMService.getPolicies({}).then(() => ({
              authorized: true,
              details: null,
            }))
          .catch(({ data }) => ({
              authorized: false,
              details: data?.details?.unauthorizedActionsByIAM,
            })),
```

ghyenne (dismissed):

packages/manager/modules/iam/src/dashboard/dashboard.routing.js
```diff
@@ -2,8 +2,24 @@ export default /* @ngInject */ ($stateProvider) => {
...
+      iam: /* @ngInject */ (IAMService) =>
+        IAMService.getPolicies({})
+          .then(() => ({
+            authorized: true,
+            details: null,
+          }))
+          .catch(({ data }) => ({
+            authorized: false,
+            details: data?.details?.unauthorizedActionsByIAM,
+          })),
```
anooparveti, line 22: **nitpick:** Error can be because of API unavailability in which case we have to display the Error page and not the unauthorised message.
Display the message only when the response is 403.

Also, I dont like the idea of changing the signature of the method to Object. May be you can return `null` and use that to differentiate.

anooparveti (changes requested):

anooparveti (dismissed):

ghyenne (approved):

anooparveti (approved):

qpavy (approved):

PR #11457, opened 2024-04-15
feat(manager-components): add ips input validator

stif59100:
| Question         | Answer
| ---------------- | ---
| Branch?          | `MANAGER-13852
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-13853

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

I created the input validator component

### DIFF
--- a/packages/manager-components/package.json
+++ b/packages/manager-components/package.json
@@ -28,6 +28,7 @@
     "@ovhcloud/ods-theme-blue-jeans": "17.2.1",
     "@tanstack/react-table": "^8.11.6",
     "babel-preset-react-app": "^10.0.1",
+    "ipaddr.js": "^2.1.0",
     "react": "18.2.0",
     "sass": "1.71.0",
     "tailwindcss": "3.3.3"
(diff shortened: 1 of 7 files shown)


packages/manager-components/package.json
```diff
@@ -54,6 +54,7 @@
     "@vitest/coverage-v8": "0.34.1",
     "autoprefixer": "10.4.14",
     "babel-loader": "9.1.3",
+    "cidr-tools": "9.1.1",
```
frenautvh, line 57: did you check for alternative libraries ? this one does not seems very popular

packages/manager-components/package.json
```diff
@@ -64,7 +65,11 @@
     "eslint-plugin-storybook": "0.6.13",
     "file-loader": "^6.2.0",
     "husky": "8.0.3",
+    "is-ip": "5.0.1",
+    "is-cidr": "5.0.5",
```
frenautvh, line 69: it shouldn't be in dev deps if you are importing the lib in your code

also could you use a single lib to handle all ip validation needs ?

packages/manager-components/package.json
```diff
@@ -64,7 +65,11 @@
...
+    "inquirer": "^7.0.6",
```
frenautvh, line 70: looks like it's not used

PR #11531, opened 2024-04-23
feat(web): 2api response is 200 when detach domain fails

dectotam:
| Question         | Answer
| ---------------- | ---
| Branch?          |  master
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Feat MANAGER-14246

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

2api response is 200 when detach domain fails : add a test on state and throw an exception if state is error

### DIFF
--- a/packages/manager/apps/web/client/app/hosting/multisite/hosting-multisite.service.js
+++ b/packages/manager/apps/web/client/app/hosting/multisite/hosting-multisite.service.js
@@ -59,7 +59,10 @@ angular.module('services').service(
               });
             },
           );
+        } else {
+          throw new Error(response.messages[0]?.message);
         }
+
         this.Hosting.resetDomains();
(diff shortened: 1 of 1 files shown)


ghyenne (approved):

packages/manager/apps/web/client/app/hosting/multisite/hosting-multisite.service.js
```diff
@@ -59,7 +59,10 @@ angular.module('services').service(
               });
             },
           );
+        } else {
+          throw new Error(response.messages[0]?.message);
         }
```
darsene, line 64: IMO, it should be better to do the fix on the 2API side and return error 500 when an error occurs

PR #11556, opened 2024-04-24
fix(container): not displayed when card is going to expire

MaximeBajeux:
Fix time calculation plus 30 days
ref: MANAGER-13918

| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/MANAGER-12494` 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-13918

- [x] Commits are signed-off
- [x] Lint has passed locally
- [] Breaking change is mentioned in relevant commits (n/a)

## Description

I change how to check if the card expiry date is between now and now plus 30 days.

### DIFF
--- a/packages/manager/apps/container/src/payment-modal/PaymentModal.tsx
+++ b/packages/manager/apps/container/src/payment-modal/PaymentModal.tsx
@@ -44,8 +44,8 @@ const computeAlert = (paymentMethods: IPaymentMethod[]): string => {
     if (creditCardExpirationDate.getTime() < Date.now()) {
       return PAYMENT_ALERTS.EXPIRED_CARD;
     }
-    const expirationDateMinus30Days = new Date();
-    expirationDateMinus30Days.setDate(expirationDateMinus30Days.getDate() - 30);
+    const expirationDateMinus30Days = new Date(creditCardExpirationDate);
+    expirationDateMinus30Days.setDate(creditCardExpirationDate.getDate() - 30);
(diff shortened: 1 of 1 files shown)


antonymarion (approved):

packages/manager/apps/container/src/payment-modal/PaymentModal.tsx
```diff
@@ -38,14 +38,15 @@ const computeAlert = (paymentMethods: IPaymentMethod[]): string => {
...
+  
   if (currentCreditCard) {
     const creditCardExpirationDate = new Date(currentCreditCard.expirationDate);
     if (creditCardExpirationDate.getTime() < Date.now()) {
       return PAYMENT_ALERTS.EXPIRED_CARD;
     }
-    const currentDateMinus30Days = new Date();
-    currentDateMinus30Days.setDate(currentDateMinus30Days.getDate() - 30);
-    const isSoonToBeExpireCreditCard = currentDateMinus30Days.getTime() > Date.now() && creditCardExpirationDate.getTime() > Date.now();
+    const currentDatePlus30Days = new Date();
```
JacquesLarique, line 47: As seen together, this seems to be working as intended, but when reading the code we can understand that we're taking the current date, adding it 30 days and compared it to the expiration date which feels like the reverse logic.
So either add a comment to explain that it is in the end the same as testing that the current date is less than 30 days before the expiration date, or use the expiration date, substract 30 days on it and compare it to the current date

JacquesLarique (approved):

PR #11576, opened 2024-04-29
feat(pci-private-network): add global regions data grid rowspan

seven-amid:
ref: DTCORE-2074

| Question         | Answer
| ---------------- | ---
| Branch?          | feat/pci-private-network
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | DTCORE-2074

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager/apps/pci-private-network/src/api/data/subnets.ts
+++ b/packages/manager/apps/pci-private-network/src/api/data/subnets.ts
@@ -1,0 +1,29 @@
+import { v6 } from '@ovh-ux/manager-core-api';
+
+type AllocationPools = {
+  start: string;
+  end: string;
+};
+
+export type TSubnet = {
+  allocationPools?: AllocationPools[];
+  allocatedIp?: string;
+  cidr: string;
+  dhcpEnabled: boolean;
+  gatewayIp: string | null;
+  gatewayName: string;
+  networkId: string;
+  region: string;
+};
+
+export const getSubnets = async (
+  projectId: string,
+  networkId: string,
+  region: string,
+): Promise<TSubnet> => {
+  const { data } = await v6.get<TSubnet>(
+    `/cloud/project/${projectId}/region/${region}/network/${networkId}/subnet`,
+  );
+
+  return data[0];
+};
--- a/packages/manager/apps/pci-private-network/src/api/hooks/useNetwork.ts
+++ b/packages/manager/apps/pci-private-network/src/api/hooks/useNetwork.ts
@@ -1,0 +1,130 @@
+import { PaginationState } from '@ovhcloud/manager-components';
+import { useQuery } from '@tanstack/react-query';
+import { useMemo } from 'react';
+import { getAggregatedNetwork, TAggregatedNetwork } from '@/api/data/network';
+import { TGateway, TRegion } from '@/api/data/project';
+import { getSubnets, TSubnet } from '@/api/data/subnets';
+import {
+  isLocalZoneRegion,
+  getLocalZoneRegions,
+  paginateResults,
+} from '@/api/utils/utils';
+
+const getDetailSubnets = async (
+  projectId: string,
+  network: TAggregatedNetwork,
+  gatewaySubnetObj: Record<string, string>,
+) => {
+  const subnetsData: TSubnet[] = await Promise.all(
+    network.subnets.map(({ region, networkId }) =>
+      getSubnets(projectId, networkId, region),
+    ),
+  );
+
+  const updatedSubnets = network.subnets.map((subnet, index) => {
+    const subnetData = subnetsData[index];
+
+    if (subnetData) {
+      return {
+        networkId: subnet.networkId,
+        region: subnet.region,
+        cidr: subnetData.cidr,
+        dhcpEnabled: subnetData.dhcpEnabled,
+        gatewayIp: subnetData.gatewayIp,
+        allocatedIp: subnetData.allocationPools
+          .map((ipPool) => `${ipPool.start} - ${ipPool.end}`)
+          .join(' ,'),
+        gatewayName: gatewaySubnetObj[subnet.networkId],
+      };
+    }
+
+    return subnet;
+  });
+
+  return updatedSubnets;
+};
+
+export const filterNonLocalPrivateNetworks = (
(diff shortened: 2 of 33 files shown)


packages/manager/apps/pci-private-network/src/api/data/subnets.ts
```diff
@@ -0,0 +1,29 @@
+import { v6 } from '@ovh-ux/manager-core-api';
+
+type AllocationPools = {
+  start: string;
+  end: string;
+};
+
+export type TSubnet = {
```
chipp972, line 8: why do you prefix the type with a "T" ?

packages/manager/apps/pci-private-network/src/api/hooks/useNetwork.ts
```diff
@@ -0,0 +1,137 @@
...
+      };
+    } else {
+      privateNetworks[vlanId].subnets.push({ region, networkId });
+    }
+  });
+
+  return Object.values(privateNetworks);
+};
+
+export const getTAggregatedNetworks = (projectId: string, regions: TRegion[]) =>
```
chipp972, line 75:
```suggestion
export const useTAggregatedNetworks = (projectId: string, regions: TRegion[]) =>
```

packages/manager/apps/pci-private-network/src/api/hooks/useNetwork.ts
```diff
@@ -0,0 +1,137 @@
...
+    rows: items.slice(
+      pagination.pageIndex * pagination.pageSize,
+      (pagination.pageIndex + 1) * pagination.pageSize,
+    ),
+    pageCount: Math.ceil(items.length / pagination.pageSize),
+    totalRows: items.length,
+  };
+};
+
+export const getNetworks = (
```
chipp972, line 97:
```suggestion
export const useNetworks = (
```

packages/manager/apps/pci-private-network/src/components/global-regions/GlobalRegionsDatagrid.tsx
```diff
@@ -0,0 +1,246 @@
...
+            name={icon}
+            size={ODS_ICON_SIZE.xs}
+            color={ODS_THEME_COLOR_INTENT.primary}
+          />
+        </OsdsButton>
+        <OsdsTooltipContent slot="tooltip-content">
+          {tooltipContent}
+        </OsdsTooltipContent>
+      </OsdsTooltip>
+    );
```
chipp972, line 95: what prevents you from creating components instead of functions here ?
They are not context specific so you can just create components and put them outside the component `GlobalRegionDatagrid` for better readability

packages/manager/apps/pci-private-network/src/components/global-regions/GlobalRegionsDatagrid.tsx
```diff
@@ -0,0 +1,246 @@
...
+        <th className="w-[8rem]">{renderText()}</th>
+      </tr>
+    );
+  };
+
+  return (
+    <div className="container contents overflow-x-auto">
+      <table className="w-full mb-2 border-collapse">
+        <thead>{renderHeader()}</thead>
+        <tbody>{items.map((network) => renderBodyRow(network))}</tbody>
```
chipp972, line 206:
```suggestion
        <thead><Header /></thead>
        <tbody>{items.map((network) => <BodyRow {...network} />)}</tbody>
```

would be more natural for example

packages/manager/apps/pci-private-network/src/api/data/subnets.ts
seven-amid: The main reason is visibility. When we use both interfaces and types for typing, it's a good practice to distinguish between them when we read the code ( `T` prefix for type and `I` prefix for Interface ) . This is also why we added suffixes for components last time, for example.

packages/manager/apps/pci-private-network/src/interface/index.ts
```diff
@@ -0,0 +1,15 @@
+export interface ResponseAPIError {
```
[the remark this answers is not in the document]
seven-amid: Could you please provide more information about the type name? I'm not sure if it's the same as `ResponseAPIError` because I couldn't find it. Alternatively, could you please provide the alias import from core if you don't mind?

packages/manager/apps/pci-private-network/src/api/data/subnets.ts
chipp972: Can you link a source explaining why it's a good practice ?

packages/manager/apps/pci-private-network/src/hooks/usePageTracking.ts
```diff
@@ -1,7 +1,7 @@
 import { useEffect } from 'react';
```
frenautvh, line 1: we could probably externalize this hook in a separated module and reuse it in each pci-app to avoid code duplication

packages/manager/apps/pci-private-network/src/interface/index.ts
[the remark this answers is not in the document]
chipp972: Its definition is located in `packages/manager/core/api/src/client.ts`

and you can import it like this : `import {ApiError} from '@ovh-ux/manager-core-api';`
It just misses the headers definition I think but you can improve the type and add them if you want.

Currently it's defined like this :
```
export type ApiError = AxiosError<{ message: string }>;
```

but we can add the headers like this for example : 
```
export type ApiError = AxiosError<{ message: string }> & {
  response: { headers: Record<string, string> & { 'x-ohv-queryid': string } };
};
```

chipp972 (approved):

packages/manager/apps/pci-private-network/src/hooks/usePageTracking.ts
seven-amid: I'll note that, but maybe in another PR, what do you think ?

frenautvh: yes sure, let's do this in a separated task

frenautvh (approved):

PR #11630, opened 2024-05-03
feat(key-management-service): add guides in header page template

darsene:
ref: MANAGER-13983

| Question         | Answer
| ---------------- | ---
| Branch?          | `key-management-service`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-13983

- [x] Commits are signed-off
- [x] Lint has passed locally


packages/manager/apps/key-management-service/src/hooks/useKmsGuides.ts
```diff
@@ -0,0 +1,42 @@
...
+  const { ovhSubsidiary } = useEnvironment().getUser();
+  const { t } = useTranslation('key-management-service/guide');
+  const [kmsGuides, setKmsGuides] = useState<GuideItem[]>([]);
+
+  const KMS_GUIDES: GuideLinks = {
+    FR: [
+      {
+        id: 1,
+        href:
+          'https://help.ovhcloud.com/csm/fr-documentation-manage-operate?id=kb_browse_cat&kb_id=3d4a8129a884a950f07829d7d5c75243',
```
vovh, line 19: Maybe we should put thoses links into a const file ?

packages/manager/apps/key-management-service/src/hooks/useKmsGuides.ts
```diff
@@ -0,0 +1,42 @@
...
+        target: OdsHTMLAnchorElementTarget._blank,
+        label: t('guides_header_find_out_more'),
+      },
+    ],
+    DEFAULT: [
+      {
+        id: 1,
+        href:
+          'https://help.ovhcloud.com/csm/en-gb-documentation-manage-operate?id=kb_browse_cat&kb_id=3d4a8129a884a950f07829d7d5c75243',
```
vovh, line 28: Same as previously mentionned.

vovh (approved):

packages/manager/apps/key-management-service/src/i18n.ts
```diff
@@ -0,0 +1,36 @@
...
+        loadPath: (lngs: string[], namespaces: string[]) =>
+          `${import.meta.env.BASE_URL}translations/${namespaces[0]}/Messages_${
+            lngs[0]
+          }.json`,
+      },
+      postProcess: 'normalize',
+    });
+
+  return i18n;
+}
```
chipp972, line 36: using initI18n from @ovh-ux/manager-react-shell-client doesn't work ?

packages/manager/apps/key-management-service/src/index.tsx
```diff
@@ -3,14 +3,17 @@ import ReactDOM from 'react-dom/client';
 import {
   ShellProvider,
   initShellContext,
-  initI18n,
 } from '@ovh-ux/manager-react-shell-client';
 import App from './App';
+import initI18n from './i18n';
```
chipp972, line 8: using initI18n from @ovh-ux/manager-react-shell-client doesn't work ?
Or did you just revert a change without noticing ?

packages/manager/apps/key-management-service/src/hooks/useKmsGuides.ts
```diff
@@ -0,0 +1,42 @@
...
+        target: OdsHTMLAnchorElementTarget._blank,
+        label: t('guides_header_find_out_more'),
+      },
+    ],
+  };
+
+  useEffect(() => {
+    if (ovhSubsidiary) {
```
chipp972, line 36: I don't think you need an `if` here

packages/manager/apps/key-management-service/src/pages/onboarding/index.tsx
```diff
@@ -1,13 +1,22 @@
...
+  const [onboardingGuide, setOnboardingGuide] = useState<GuideItem>();
   const { t } = useTranslation('key-management-service/onboarding');
   const navigate = useNavigate();
+  const { kmsOnboardingGuide } = useKmsGuides();
+
+  useEffect(() => {
+    if (kmsOnboardingGuide) {
+      setOnboardingGuide(kmsOnboardingGuide);
+    }
+  }, [kmsOnboardingGuide]);
```
chipp972, line 19: I don't think you need a state and an useEffect here
And if you need one, it should be in the hook `useKmsGuides` (but you shouldn't since kmsOnboardingGuide is dependant on a state so it should trigger a re-render anyways).

packages/manager/apps/key-management-service/src/index.tsx
darsene: It didn't work for translation on manager-components like `GuideButton`, in the other way translation wasn't apply so I took example with zimbra PR which is also using the `GuideButton`

packages/manager/apps/key-management-service/src/hooks/useKmsGuides.ts
darsene: I'll do it when we have the final links

packages/manager/apps/key-management-service/src/hooks/useKmsGuides.ts
darsene: fixed

packages/manager/apps/key-management-service/src/pages/onboarding/index.tsx
darsene: fixed

packages/manager/apps/key-management-service/src/index.tsx
chipp972: What is the difference between them ?
We have to fix the shared configuration instead of duplicating it on every project

packages/manager/apps/key-management-service/src/hooks/useKmsGuides.ts
chipp972: There is still an if so in some cases it doesn't work ?

JacquesLarique (approved):

chipp972 (changes requested):
Please rebase on develop to avoid duplicating i18n config in the project

chipp972 (approved):

PR #11655, opened 2024-05-06
fix(pci-public-ip): fix the i18n t interpolation config

seven-amid:
ref: DTCORE-2189
ref: DTCORE-2198

| Question         | Answer
| ---------------- | ---
| Branch?          | feat/pci-public-ip
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | DTCORE-2189 & DTCORE-2198

- [ ] Commits are signed-off
- [ ] Lint has passed locally

### DIFF
--- a/packages/manager/apps/pci-public-ip/src/pages/additional-ips/edit/EditAdditionalIP.page.tsx
+++ b/packages/manager/apps/pci-public-ip/src/pages/additional-ips/edit/EditAdditionalIP.page.tsx
@@ -48,6 +48,9 @@ export default function EditInstancePage() {
         {(t) =>
           t('pci_additional_ips_failoverips_edit_error', {
             error: error?.response?.data?.message,
+            interpolation: {
+              escapeValue: false,
+            },
           })
         }
       </Translation>,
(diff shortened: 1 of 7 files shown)


frenautvh (approved):

packages/manager/apps/pci-public-ip/src/pages/additional-ips/edit/EditAdditionalIP.page.tsx
```diff
@@ -48,6 +48,9 @@ export default function EditInstancePage() {
         {(t) =>
           t('pci_additional_ips_failoverips_edit_error', {
             error: error?.response?.data?.message,
+            interpolation: {
+              escapeValue: false,
+            },
```
chipp972, line 53: we have this configuration directly in the initI18n function from react-shell-client package
It's better to not duplicate it in each translation

seven-amid: The reason behind not make it in global config is the risk of XXS attack https://www.i18next.com/translation-function/interpolation#unescape

chipp972: Except we usually interpolate data that come from our APIs and display it to the user. We don't often interpolate data that come from user input and even less often send those to the APIs.

We have more cases where we don't want to escape value (because we use accents in our languages) than not doing it so it should be the default in my opinion

chipp972 (approved):

PR #11662, opened 2024-05-07
fix(hub): ticket appears even cancelled

MaximeBajeux:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-13862

- [x] Commits are signed-off
- [x] Lint has passed locally
- [] Breaking change is mentioned in relevant commits (n/a)

## Description

Remove cache for tickets in dashboard

### DIFF
--- a/packages/manager/modules/hub/src/components/support/controller.js
+++ b/packages/manager/modules/hub/src/components/support/controller.js
@@ -28,7 +28,39 @@ export default class ManagerHubSupportCtrl {
     this.guideURL = this.RedirectionService.getURL('guides.home', {
       ovhSubsidiary: this.me.ovhSubsidiary,
     });
-    this.fetchTickets()
+    this.handleTicketsResponse(this.fetchTickets());
+  }
+
+  fetchTickets(cache = true) {
+    const params = {
+      serviceType: 'aapi',
+    };
+
+    if (!cache) {
+      params.headers = {
+        Pragma: 'no-cache',
+      };
+    }
+
+    return this.$http
+      .get('/hub/support', params)
+      .then(({ data }) => data)
+      .then((result) => get(result, 'data.support.data'));
+  }
+
+  refreshTickets() {
+    this.handleTicketsResponse(this.fetchTickets(false));
+  }
+
+  onSeeMore() {
+    this.atInternet.trackClick({
+      name: `${this.trackingPrefix}::activity::assistance::show-all`,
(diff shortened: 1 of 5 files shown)


packages/manager/modules/hub/src/components/support/controller.js
anooparveti, line 1: **question:** Does the 2API supports or forwards the headers to the underlying APIv6 call?

MaximeBajeux: On the 2api side, cache is handle for all the hub and it seems we can't override the behaviour without refactoring for all the hub. I tested and it seems the fix works well. Do you have another solution in head ?

MaximeBajeux: Seems there is no cache between apiV6 and 2api, so normally I think we not need to forward the headers to api V6 because the cache exists only between the 2api and the manager. About the performance cost, obviously that costs one API call each time the service is call. To reduce the performance issue, we could get only the two latest tickets from API instead of getting all and truncate into the 2api, like in the bff. But this is something that will be resolved once the bff is in production.

oalkabouss (dismissed):

oalkabouss (approved):

packages/manager/modules/hub/src/components/tile/tile.html
```diff
@@ -7,6 +7,15 @@ <h3 class="oui-tile__title ovh-manager-hub-tile__title">
                 data-ng-bind="$ctrl.badgeText"
             ></span>
         </h3>
+        <button
+            class="oui-button hub-button oui-button_ghost"
+            type="button"
+            data-ng-if="$ctrl.onRefresh"
+            data-ng-click="$ctrl.onRefresh()"
+        >
+            <span class="oui-icon oui-icon-refresh"></span>
```
anooparveti, line 16:
```suggestion
            <span class="oui-icon oui-icon-refresh" aria-hidden="true"></span>
```

oalkabouss (dismissed):

JacquesLarique (dismissed):

oalkabouss (dismissed):

anooparveti (approved):

acavalloni (approved):

oalkabouss (approved):

PR #11707, opened 2024-05-14
feat(public-cloud): fix instances prices

qpavy:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-14309

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

- Fix step 1 instances creation prices when flavors have different hourly / monthly prices

## Related

- [x] CMT-11886

### DIFF
--- a/packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.controller.js
+++ b/packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.controller.js
@@ -169,4 +169,56 @@ export default class FlavorsListController {
   formatStorage(capacity) {
     return this.$filter('bytes')(capacity, 2, false, 'GB');
   }
+
+  /**
+   * Get the lowest hourly price and lowest monthly price for the flavor
+   * and return distinct prices for both hourly and monthly
+   * @param {*} flavor
+   * @returns Object with lowest hourly / monthly and all distinct hourly / monthly
+   */
+  static getLowestPriceFromFlavor(flavor) {
+    const allHourlyPrices = [];
+    const allMontlyPrices = [];
+
+    // get the lowest hourly price for the flavor
+    const lowestHourlyPrice = flavor.priceInformation.reduce(
+      (lowest, current) => {
+        if (!current.prices.hourly?.value) return lowest;
+        // add current hourly price to distinct prices array
+        if (!allHourlyPrices.includes(current.prices.hourly.value)) {
+          allHourlyPrices.push(current.prices.hourly.value);
+        }
+        if (!lowest?.value) return current.prices.hourly;
+
+        return lowest.value > current.prices.hourly.value
+          ? current.prices.hourly
+          : lowest;
+      },
+      null,
+    );
+
+    // get the lowest monthly price for the flavor
+    const lowestMonthlyPrice = flavor.priceInformation.reduce(
+      (lowest, current) => {
+        if (!current.prices.monthly?.value) return lowest;
+        // add current monthly price to distinct prices array
+        if (!allMontlyPrices.includes(current.prices.monthly.value)) {
+          allMontlyPrices.push(current.prices.monthly.value);
+        }
+        if (!lowest?.value) return current.prices.monthly;
+
+        return lowest.value > current.prices.monthly.value
+          ? current.prices.monthly
+          : lowest;
+      },
+      null,
+    );
+
+    return {
+      hourly: lowestHourlyPrice,
+      monthly: lowestMonthlyPrice,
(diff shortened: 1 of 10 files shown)


packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.controller.js
```diff
@@ -169,4 +169,56 @@ export default class FlavorsListController {
...
+    const allMontlyPrices = [];
+
+    // get the lowest hourly price for the flavor
+    const lowestHourlyPrice = flavor.priceInformation.reduce(
+      (lowest, current) => {
+        if (!current.prices.hourly?.value) return lowest;
+        // add current hourly price to distinct prices array
+        if (!allHourlyPrices.includes(current.prices.hourly.value)) {
+          allHourlyPrices.push(current.prices.hourly.value);
+        }
```
chipp972, line 190: You could just fill this array like this before hand :
```
const allHourlyPrices = Array.from(new Set(flavor.priceInformation.map(price => price.prices.hourly.value)))
```

chipp972 (approved):

packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.controller.js
```diff
@@ -169,4 +169,56 @@ export default class FlavorsListController {
...
     return this.$filter('bytes')(capacity, 2, false, 'GB');
   }
+
+  /**
+   * Get the lowest hourly price and lowest monthly price for the flavor
+   * and return distinct prices for both hourly and monthly
+   * @param {*} flavor
+   * @returns Object with lowest hourly / monthly and all distinct hourly / monthly
+   */
+  static getLowestPriceFromFlavor(flavor) {
```
anooparveti, line 179: **nitpick:** Since the function does not just return the lowest price. `getLowestPrice` does not make sense.

packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.controller.js
```diff
@@ -169,4 +169,56 @@ export default class FlavorsListController {
...
+    const allHourlyPrices = [];
+    const allMontlyPrices = [];
```
anooparveti, line 181: **suggestion:** Since we just the need the info whether to display "from", I think we can maintain a boolean rather than maintaining the array.

packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.controller.js
```diff
@@ -169,4 +169,56 @@ export default class FlavorsListController {
...
+          allHourlyPrices.push(current.prices.hourly.value);
+        }
+        if (!lowest?.value) return current.prices.hourly;
+
+        return lowest.value > current.prices.hourly.value
+          ? current.prices.hourly
+          : lowest;
+      },
+      null,
+    );
```
anooparveti, line 198: **nitpick:** Logic can be simplified.
```suggestion
    const lowestHourlyPrice = flavor.priceInformation.reduce(
      (lowest, current) => {
       if (!lowest?.value) return current.prices.hourly;
       if(lowest.value > current.prices.hourly.value) {
         lowest = current.prices.hourly
         // also update the boolean
       }
       return lowest;
      },
      null,
    );
```

anooparveti (changes requested):

packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.controller.js
qpavy: I did that to allow different display if needed. There is 3 different possible cases for each hourly and monthly: 

- case 1 : there is no hourly or monthly price
- case 2 : there is 1 unique hourly / monthly price 
- case 3 : there is multiple different hourly / monthly prices.

For now we use case 1 for houlry to display or not the block, case 2 and 3 to change the translation key depending of the case.

So I think the boolean isn't enought

packages/manager/modules/pci/src/components/project/flavors-list/flavors-list.controller.js
qpavy: This only work if we do not manage the case 1 : a flavor with no hourly or monthly price.

chipp972 (dismissed):

anooparveti (approved):

ghyenne (approved):

darsene (approved):

PR #11748, opened 2024-05-20
feat(2fa): set up invisible challenge

JacquesLarique:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/acquisition-storage-2fa`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-14063

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Setup of an "invisible challenge" to ensure no bot is requesting the non authenticated API methods.

### DIFF
--- a/packages/manager/apps/account-disable-2fa/src/utils/iframe.ts
+++ b/packages/manager/apps/account-disable-2fa/src/utils/iframe.ts
@@ -1,0 +1,53 @@
+export type IframeLoaderPayload = {
+  src: string;
+  id?: string;
+  title?: string;
+  style?: string;
+  width?: string;
+};
+
+const getNewIframeElement = (
+  payload: IframeLoaderPayload,
+): HTMLIFrameElement => {
+  if (!payload || !payload.src) {
+    throw new Error('[IframeLoader][getNewIframeElement()] missing payload.');
+  }
+
+  const iframe = document.createElement('iframe');
+  iframe.src = payload.src;
+
+  if (payload.title) {
+    iframe.title = payload.title;
+  }
+  if (payload.width) {
+    iframe.width = payload.width;
+  }
+  if (payload.id) {
+    iframe.id = payload.id;
+  }
+  if (payload.style) {
+    iframe.style.cssText = payload.style;
+  }
+
+  return iframe;
+};
+
+export const loadIframe = (payload: IframeLoaderPayload): Promise<unknown> => {
+  if (!payload || !payload.src) {
+    throw new Error('[IframeLoader][load()] missing payload.');
+  }
+
+  const iframe = getNewIframeElement(payload);
+  const promise = new Promise((resolve, reject) => {
+    iframe.onload = resolve;
+    // The following code is never called
+    // see: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/iframe#error_and_load_event_behavior
+    iframe.onerror = reject;
+  });
+
+  document.body.appendChild(iframe);
+  return promise;
+};
+
+export const removeIframe = (iframe: HTMLIFrameElement): HTMLIFrameElement =>
+  document.body.removeChild(iframe);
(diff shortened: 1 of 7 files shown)


packages/manager/apps/account-disable-2fa/src/utils/iframe.ts
```diff
@@ -0,0 +1,53 @@
...
+    iframe.style.cssText = payload.style;
+  }
+
+  return iframe;
+};
+
+export const loadIframe = (payload: IframeLoaderPayload): Promise<unknown> => {
+  if (!payload || !payload.src) {
+    throw new Error('[IframeLoader][load()] missing payload.');
+  }
```
anooparveti, line 38: **nitpick:** Since the check is also added on `getNewIframeElement`, we can remove it.

packages/manager/apps/account-disable-2fa/src/utils/invisible-challenge.ts
anooparveti, line 1: **nitpick:** Rename the file to `invisible-challenge.class.ts`.

packages/manager/apps/account-disable-2fa/src/utils/invisible-challenge.ts
```diff
@@ -0,0 +1,165 @@
...
+export enum ChallengeFailureReasons {
+  NoIframe,
+  Timeout,
+}
+
+export const CHALLENGE_IFRAME_ID = 'ovh_security_challenge';
+export const TTL_TO_WAIT_IFRAME_RESPONSE = 1000;
+export const INVISIBLE_CHALLENGE_ERROR_CLASS =
+  'Client::BadRequest::ChallengeRequired';
+export const IFRAME_STYLE = 'position: absolute; width:0; height:0; border:0;';
```
anooparveti, line 23: **nitpick:** Can we add it in constants file?

packages/manager/apps/account-disable-2fa/src/utils/invisible-challenge.ts
```diff
@@ -0,0 +1,165 @@
...
+    this.iframe = null;
+    this.deferWaitingChallenge = null;
+    this.abortTimeout = null;
+  }
+
+  public attachMessageListener() {
+    window.addEventListener(
+      'message',
+      (message: MessageEvent) => this.onIframeMessage(message),
+      false,
```
anooparveti, line 45: **nitpick:** Anyways defaults to `false` (ref: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener#usecapture) . So better to remove.
```suggestion
```

packages/manager/apps/account-disable-2fa/src/utils/invisible-challenge.ts
```diff
@@ -0,0 +1,165 @@
...
+        this.onIframeLoaded();
+        return this.deferWaitingChallenge.promise;
+      }
+      // Otherwise we remove the previous one to use a new one
+      removeIframe(this.iframe);
+      this.iframe = null;
+    }
+    const payload: IframeLoaderPayload = {
+      src: url,
+      id: CHALLENGE_IFRAME_ID,
```
anooparveti, line 80: **nitpick:** Is it possible to load more than 1 challenge at the same time? If yes, then we will have more than 1 frame with same IDs otherwise ignore my comment.

JacquesLarique: We can't have multiple iframe with the same ID as the code right before this handle the case of an iframe already present (and remove it if the url of the challenge isn't the same)
Moreover, in the case of 2fa, we should not have concurrent requests and as a result only one challenge request at a time (unless we incorrectly implement the UI and let the user click multiple times on "Send documents")

oalkabouss (approved):

anooparveti (approved):

PR #11843, opened 2024-05-30
feat(2fa): set up user session management

JacquesLarique:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/acquisition-storage-2fa`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-13716

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Added the user session management:
- extracting token from url
- decoding token
- redirect to authentication page if no token or outdated token are given
- share user's data from token in a context for future usage


packages/manager/apps/account-disable-2fa/src/components/User/provider.tsx
```diff
@@ -0,0 +1,29 @@
...
+    let loginUrl =
+      window.location.host === 'www.ovhtelecom.fr'
+        ? 'https://www.ovh.com/auth/'
+        : '/auth';
+
+    if (user?.subsidiary) {
+      loginUrl += `?ovhSubsidiary=${user.subsidiary}`;
+    }
+
+    window.location.assign(loginUrl);
```
chipp972, line 21:
```suggestion
    const loginUrl =
      window.location.host === 'www.ovhtelecom.fr'
        ? 'https://www.ovh.com/auth/'
        : '/auth';

    const subsidiaryParams = user?.subsidiary ? `?ovhSubsidiary=${user.subsidiary}` : '';

    window.location.assign(`${loginUrl}${subsidiaryParams}`);
```

packages/manager/apps/account-disable-2fa/src/components/User/useUser.ts
```diff
@@ -0,0 +1,8 @@
+import { useContext } from 'react';
+import userContext, { UserContext } from '@/components/User/context';
+
+const useUser = (): UserContext => {
+  return useContext(userContext);
+};
+
+export default useUser;
```
chipp972, line 8: Just inline this part

packages/manager/apps/account-disable-2fa/src/index.tsx
```diff
@@ -1,27 +1,24 @@
...
-  language: 'fr_FR',
-  subsidiary: 'FR',
-};
-const locale = user.language;
+const token = extractToken();
+const user = decodeToken(token);
+const { language: locale, subsidiary } = user || {};
 
-initI18n(locale, [locale], user.subsidiary);
+initI18n(locale, [locale], subsidiary);
```
chipp972, line 16: can you use the initI18n from react-shell-client lib ? To avoid duplication

packages/manager/apps/account-disable-2fa/src/index.tsx
```diff
@@ -1,27 +1,24 @@
...
+initAuthenticationInterceptor(token);
 initInterceptor();
 
 ReactDOM.createRoot(document.getElementById('root')!).render(
   <React.StrictMode>
-    <App />
+    <App user={user} />
```
chipp972, line 22: You could just wrap App inside UserContextProvider directly

packages/manager/apps/account-disable-2fa/src/components/User/context.ts
```diff
@@ -0,0 +1,21 @@
...
+  exp: number;
+  nbf: number;
+  iat: number;
+};
+
+export type UserContext = {
+  user: User | null;
+};
+
+const userContext = createContext<UserContext>({} as UserContext);
```
chipp972, line 19:
```suggestion
const userContext = createContext<UserContext>({ user: null });
```

packages/manager/apps/account-disable-2fa/src/index.tsx
chipp972: Which app did you look at ?
catalog, kms, vrack-services and zimbra use initI18n from react-shell-client

Only the pci apps don't use it and duplicate it on each app

And we probably can't use a lot of things from react-shell-client lib if you don't have a container but the initI18n script is not dependant on having a container so I would rather not duplicate it one more time.

chipp972 (approved):

oalkabouss (approved):

packages/manager/apps/account-disable-2fa/src/index.tsx
chipp972: Yes you are right, it's to retrieve the available locales

PR #11846, opened 2024-05-30
Fix/location regions input

sidlynx:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes/no
| New feature?     | yes/no
| Breaking change? | yes/no
| Tickets          | Fix #... 

- [ ] Commits are signed-off
- [ ] Lint has passed locally


packages/manager-components/src/utils/index.ts
```diff
@@ -9,3 +9,33 @@ export const uniqBy = function uniqBy<I, U>(arr: I[], cb: (item: I) => U) {
       .values(),
   ];
 };
+
+// eslint-disable-next-line
+export const hashCode = (param: any) => {
```
frenautvh, line 14: maybe you could make Item having a key attribute (or unique id) mandatory and let the consumer of the component provide those unique ids

it would make this component simpler by not having to compute unique keys

i also think it's not necessary to provide use cases for items such as boolean or numbers, that's why you can enforce a unique key attribute on items

sidlynx: the tabs component(that require unique keys) is used inside the tiles input component, there's no way a user can add unique keys by himself, and i wanted to keep both components generic

seven-amid (approved):

kqesar (approved):

PR #11862, opened 2024-05-31
fix(pci-private-network): fix remaining visual and functional issues

seven-amid:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes/no
| New feature?     | yes/no
| Breaking change? | yes/no
| Tickets          | Fix #... 

- [ ] Commits are signed-off
- [ ] Lint has passed locally


packages/manager-components/src/components/notifications/notifications.component.tsx
```diff
@@ -20,14 +19,17 @@ interface NotificationProps {
...
-  const location = useLocation();
-  const [originLocation] = useState(location);
-  const { notifications, clearNotifications } = useNotifications();
+  const { notifications, clearNotification } = useNotifications();
 
   useEffect(() => {
-    if (clearAfterRead && originLocation.pathname !== location.pathname)
-      clearNotifications();
-  }, [clearAfterRead, location.pathname]);
+    const timer = setTimeout(() => {
```
frenautvh, line 25: i think this will create a lot of unecessary timeout functions each time notifications array changes (when a new notification is pushed for example)

you probably don't want to create multiple timeouts (even if only the first one will do something), it could be avoided by adding an attribute in notifcation type to check if the clear is planned for example

packages/manager/apps/pci-private-network/src/hooks/catalog-price.ts
```diff
@@ -0,0 +1,70 @@
...
+  'IT',
+  'LT',
+  'MA',
+  'NL',
+  'PL',
+  'PT',
+  'TN',
+];
+
+export const useCatalogPrice = (maximumFractionDigits?: number) => {
```
frenautvh, line 22: hook could probably be externalized

frenautvh (approved):

PR #11970, opened 2024-06-11
fix(pci-private-network): fix data fetching, refactoring

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | feat/pci-private-network
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | -

## Description

fix data fetching issues, refactoring

### DIFF
--- a/packages/manager/apps/pci-private-network/src/api/hooks/useNetwork.ts
+++ b/packages/manager/apps/pci-private-network/src/api/hooks/useNetwork.ts
@@ -1,6 +1,6 @@
 import { applyFilters, Filter } from '@ovh-ux/manager-core-api';
 import { PaginationState } from '@ovhcloud/manager-components';
-import { useMutation, useQuery } from '@tanstack/react-query';
+import { useMutation, useQueries, useQuery } from '@tanstack/react-query';
 import { useMemo } from 'react';
 import {
   deleteNetwork,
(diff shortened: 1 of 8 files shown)


packages/manager/apps/pci-private-network/src/api/hooks/useNetwork.ts
```diff
@@ -205,41 +200,44 @@ export const useLocalZoneNetworks = (
...
+          allocatedIp: ips,
+          search: `${network.id} ${network.name} ${network.region} ${subnet.cidr} ${subnet.gatewayIp} ${ips}`,
+        };
+      },
+    })),
   });
+  const isPending = subnetQueries.some((q) => q.isPending);
+  const isFetching = subnetQueries.some((q) => q.isFetching);
+  const data = useMemo(() => {
+    if (isFetching || isPending) return undefined;
```
kqesar, line 225: You can't return a empty array as you are in useMemo or it's a specific case where you need to return undefined ?

frenautvh: i prefer to return undefined because by default useQuery returns undefined also when data is not available instead of an empty array

kqesar: Ok for me

PR #12054, opened 2024-06-17
Feat/acquisition storage 2fa adjustments time form

oalkabouss:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/acquisition-storage-2fa-adjustments-time-form`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #... 

- [x] Commits are signed-off
- [x] Lint has passed locally
- [ ] ~~Ticket reference is mentioned in linked commits (internal only)~~


packages/manager/apps/procedures/src/pages/create/form/Form.page.tsx
```diff
@@ -104,7 +104,7 @@ const FormCreateRequest = () => {
         </div>
       )}
 
-      {legalForm && (
+      {(legalForm || legalForm?.trim()?.length === 0) && (
```
JacquesLarique, line 107: I think it would be best to test:
`(!isOtherLegalFormForFR || legalForm)`
This way when we are in a nominal case (which doesn't require to let user select his/her legal form), even if the `legalForm` is empty we would display the `FormDocumentFieldList`.
And when we are in a non nominal case (we have to display the legalForm dropdown) we won't display the form until the user has selected one

JacquesLarique (changes requested):

JacquesLarique (approved):

MaximeBajeux (approved):

PR #12090, opened 2024-06-20
feat(manager-components): dashboard template update

chipp972:
ref: MANAGER-14379

| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | yes
| Tickets          | MANAGER-14379 

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager-components/src/components/content/headers/headers.component.tsx
+++ b/packages/manager-components/src/components/content/headers/headers.component.tsx
@@ -1,5 +1,4 @@
 import React from 'react';
-
 import { Description, Title, Subtitle } from '../../typography';
 
 export interface HeadersProps {
@@ -16,13 +15,15 @@ export const Headers: React.FC<HeadersProps> = ({
   headerButton,
 }) => {
   return (
-    <div className="pl-5 flex items-center justify-between">
+    <div className="flex items-center justify-between">
       <div>
-        {title && <Title>{title}</Title>}
+        {title && <Title className="block mb-8">{title}</Title>}
         {subtitle && <Subtitle>{subtitle}</Subtitle>}
-        {description && <Description>{description}</Description>}
+        {description && (
+          <Description className="mb-6">{description}</Description>
+        )}
       </div>
-      <div>{headerButton}</div>
+      {headerButton && <div>{headerButton}</div>}
     </div>
   );
 };
--- a/packages/manager-components/src/components/templates/dashboard/dashboard.component.tsx
+++ b/packages/manager-components/src/components/templates/dashboard/dashboard.component.tsx
@@ -1,47 +1,51 @@
 import React from 'react';
-
 import { Headers, HeadersProps } from '../../content';
-
-import { Links, LinksProps } from '../../typography';
+import { Description, LinkType, Links, Subtitle } from '../../typography';
 import { PageLayout } from '../layout/layout.component';
+
 export interface DashboardLayoutProps {
   breadcrumb?: React.ReactElement;
   content?: React.ReactElement;
   header?: HeadersProps;
-  linkProps?: LinksProps;
+  message?: React.ReactElement;
+  description?: string;
+  subtitle?: string;
+  subdescription?: string;
+  backLinkLabel?: string;
   tabs?: React.ReactElement;
   onClickReturn?: () => void;
 }
 
 export const DashboardLayout = ({
-  linkProps,
-  breadcrumb,
-  content,
+  backLinkLabel,
   onClickReturn,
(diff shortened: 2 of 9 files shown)


packages/manager-components/src/components/content/headers/headers.component.tsx
```diff
@@ -16,13 +15,13 @@ export const Headers: React.FC<HeadersProps> = ({
   headerButton,
 }) => {
   return (
-    <div className="pl-5 flex items-center justify-between">
+    <div className="flex items-center justify-between">
       <div>
-        {title && <Title>{title}</Title>}
+        {title && <Title className="block mb-8">{title}</Title>}
```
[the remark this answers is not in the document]
chipp972: Personally I am not a big fan of including margin inside typography component
It is easier to import 0 margin components and manage all the margins in the templates in my opinion.

And by the way, it will break pci-rancher design when merging the PR if I change the default margin of the component.

qpavy (dismissed):

packages/manager-components/src/components/content/headers/headers.component.tsx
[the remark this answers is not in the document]
aboungnaseng-ovhcloud: we can let title component without classes and add a class for title component in headers if you are sure the templates will be use <headers> component.

Title can stay with 'className' attributes to be less restictive as we didn't know all cases in the future templates

aboungnaseng-ovhcloud (dismissed):

packages/manager-components/src/components/content/headers/headers.component.tsx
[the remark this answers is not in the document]
rjamet-ovh: A general rule of thumb is to avoid hard coded margin on components, it's the responsibility of the parent component to specify spacing between its children.

packages/manager-components/src/components/typography/description/description.component.tsx
```diff
@@ -9,13 +9,15 @@ import { OsdsText } from '@ovhcloud/ods-components/react';
...
+  className?: string;
 }
 export const Description: React.FC<DescriptionProps> = ({
   children,
+  className,
 }: DescriptionProps) => {
   return (
     <OsdsText
-      className="block mb-6"
+      className={`block mb-6 ${className}`}
```
rjamet-ovh, line 20:
```suggestion
      className={`block ${className}`}
```
remove hard-coded margin on component

chipp972: The purpose of this PR is not to refactor existing components and create breaking changes in apps that already use them. It is to update the dashboard template component.

We should discuss whether we include margin or not in our typography components and have the same approach in all of them in a separate ticket.

I agree with you though, I would rather not have margin included in this component.

rjamet-ovh: I've checked, this component is only used in `headers` and `dashboard` manager-components, so it would be ok to fix this on this PR (just have to update the `header` instance with a `mb-6` class)

But yes, i agree, it should be discussed with everyone.

chipp972: done since it doesn't impact any app

aboungnaseng-ovhcloud (approved):

rjamet-ovh (approved):

packages/manager-components/src/components/templates/dashboard/dashboard.component.tsx
```diff
@@ -1,47 +1,51 @@
...
-import { Links, LinksProps } from '../../typography';
+import { Description, LinkType, Links, Subtitle } from '../../typography';
 import { PageLayout } from '../layout/layout.component';
+
 export interface DashboardLayoutProps {
   breadcrumb?: React.ReactElement;
   content?: React.ReactElement;
   header?: HeadersProps;
-  linkProps?: LinksProps;
+  message?: React.ReactElement;
```
anooparveti, line 10: We can have a requirement to display multiple messages and its type can be varied as well among info/warning/error. I think no need to include message instead we can provide a slot to fill it with children.

chipp972: Yes it's fine we can put multiple messages in this slot using fragment :

```
<>
  <OsdsMessage>{message1}</OsdsMessage>
  <OsdsMessage>{message2}</OsdsMessage>
</>
```

anooparveti (approved):

PR #12108, opened 2024-06-24
feat(container): bullet added on pnr lvl 2

oalkabouss:
ref: MANAGER-14442

| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-14442

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

PNR v2: bullet added on lvl 2 to indicate an existing product.

### DIFF
--- a/packages/manager/apps/container/src/container/nav-reshuffle/sidebar/index.tsx
+++ b/packages/manager/apps/container/src/container/nav-reshuffle/sidebar/index.tsx
@@ -19,6 +19,8 @@ import {
   findPathToNode,
   initFeatureNames,
   shouldHideElement,
+  debounce,
+  IServicesCount,
 } from './utils';
 import { Node } from './navigation-tree/node';
 import useProductNavReshuffle from '@/core/product-nav-reshuffle';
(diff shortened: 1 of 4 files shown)


packages/manager/apps/container/src/container/nav-reshuffle/sidebar/index.tsx
```diff
@@ -191,22 +192,37 @@ const Sidebar = (): JSX.Element => {
...
-   */
+ * Initialize menu items based on currentNavigationNode
+ */
   useEffect(() => {
-    const count: ServicesCount = {
-      total: servicesCount?.total,
-      serviceTypes: {
-        ...servicesCount?.serviceTypes,
-      } as Record<string, number>,
+    if (!currentNavigationNode?.children || !servicesCount) return;
```
JacquesLarique, line 213: Won't this result in the display of previously displayed menu items when `!currentNavigationNode?.children || !servicesCount` is true ?

It seems to me that we should continue to call setMenuItems with undefined in such case.

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/style.module.scss
```diff
@@ -303,5 +303,6 @@
 .sidebarLinkTag {
   width: 0.8rem;
   margin-right: .25rem;
+  float: right;
```
MaximeBajeux, line 306: Not other solution than using a float here ?

oalkabouss: Yes, there are other solutions, like flex for example, but when I apply the flex display the tags appear oddly so I don't like to touch the case of other components to just align the dot icon right

MaximeBajeux: The issue with float is not to add css, but to use a float that not works correctly on this context. It's not vertically aligned with the text. You can try to wrap the icon like in sidebarLink to align vertically and use a justify content space between on the link the keep the dot on the right.

MaximeBajeux (approved):

JacquesLarique (approved):

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/index.tsx
```diff
@@ -191,22 +192,35 @@ const Sidebar = (): JSX.Element => {
     }
   }, [isNavigationSidebarOpened]);
 
+
+  const computeNodeCount = (count: IServicesCount, node: Node): number | boolean => {
```
anooparveti, line 196:
```suggestion
  const computeNodeCount = (count: IServicesCount, node: Node): number => {
```

chipp972 (approved):

MaximeBajeux (approved):

PR #12177, opened 2024-06-28
fix(web.exchange): no need redirection to order when is ovhtel

ghyenne:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | DTRSD-123276 

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
I removed no needed redirection to order when is ovhtel (linked with packxdsl)


darsene (approved):

antonymarion (approved):

packages/manager/modules/exchange/src/account/order/account-order.html
```diff
@@ -2,9 +2,10 @@
     <div
         data-wizard
         data-wizard-bread-crumb
-        data-wizard-confirm-button-text="'wizard_pay' | translate"
+        data-wizard-confirm-button-text="ctrl.isOvhtel ? 'wizard_close' : 'wizard_pay' | translate"
```
anooparveti, line 5: I dont think we can display `Close` as a primary button. We may need a validation from the design team.

anooparveti (changes requested):

packages/manager/modules/exchange/src/account/order/account-order.html
ghyenne: No it's problem ticket. In fact is close not cancel just for close the modal because no need redirect to order url and as it's wizard it isn't easy to do data-ng-if on step (issue with scope) so i've tried to keep it as simple as possible with a minimum of change / duplication

sachinrameshn (approved):

darsene (approved):

anooparveti (approved):

PR #12191, opened 2024-07-01
feat(vrack-services): automatic validation of order

chipp972:
ref: MANAGER-14440

| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-14440 

- [x] Commits are signed-off
- [x] Lint has passed locally


packages/manager/apps/vrack-services/src/utils/cart.ts
```diff
@@ -0,0 +1,43 @@
...
+        options: {
+          duration: 'P1M',
+          planCode: 'vrack-services',
+          pricingMode: 'default',
+          quantity: 1,
+        },
+        configurations: [{ label: 'region', value: region }],
+      },
+      hasVrack && vrackItem,
+    ].filter(Boolean),
```
qpavy, line 36: Why is filter needed ?

chipp972: If I don't filter it will keep a false (in case there is no region or no vrack) and do a call that will fail

qpavy (approved):

sachinrameshn (approved):

PR #12376, opened 2024-07-16
feat(container): added tooltip to reduced left pannel

oalkabouss:
ref: MANAGER-14710

| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-14710

- [x] Commits are signed-off
- [x] Lint has passed locally


packages/manager/apps/container/src/container/nav-reshuffle/sidebar/tooltip/SidebarLinkTooltip.tsx
```diff
@@ -0,0 +1,37 @@
...
+    children: React.ReactNode, 
+    content: string 
+};
+
+export const ShortSidebarLinkTooltip: FunctionComponent<Props> = ({
+    content,
+    children
+}) => {
+
+    const id = 'SidebarLinkTooltip';
```
JacquesLarique, line 14: Id should be unique

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/tooltip/SidebarLinkTooltip.tsx
```diff
@@ -0,0 +1,37 @@
...
+    return (
+        <>
+            <span data-tooltip-id={id}
+                data-tooltip-position-strategy={'fixed'}
+                data-tooltip-content={content}
+                data-tooltip-place="right"
+            >
+                {children}
+            </span>
+            <Tooltip
```
JacquesLarique, line 25: Why not using ODS tooltip ?
https://ovh.github.io/design-system/latest/?path=/story/tooltip--default

JacquesLarique (changes requested):

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/tooltip/SidebarLinkTooltip.tsx
oalkabouss: There are two ways to use Tooltip with: anchor a unique id , and a version with a non-unique id see the documentation. i had used non unique id version. to avoid confusion, I switched to anchor. https://react-tooltip.com/docs/examples/basic-examples

oalkabouss: There are two ways to use Tooltip with: anchor a unique id , and a version with a non-unique id see the documentation https://react-tooltip.com/docs/examples/basic-examples. i had used non unique id version. to avoid confusion, I switched to anchor.

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/tooltip/SidebarLinkTooltip.tsx
oalkabouss: i was expecting this question. i initially went with ODS but I switched to react tooltip because it has a hover anomaly with the navbar, it blocks the hover of the element below, we cannot manage the direction of the tooltip, and ODS does not manage well the fixed position because the sidebar is in hidden overflow.

JacquesLarique: Then we probably need to report a bug (because it seems this should be working: https://zeroheight.com/6fc8a63f7/p/719b66-tooltip/t/page-65d34c-74844005-40a887-11)

anooparveti (changes requested):

anooparveti (approved):

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/tooltip/SidebarLinkTooltip.tsx
oalkabouss: I created a bug ticket with the code: ODS-1367

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/SidebarLink.tsx
```diff
@@ -45,7 +46,15 @@ const SidebarLink: React.FC<ComponentProps<SidebarLinkProps>> = ({
...
-      <span> {t(isShortText ? node.shortTranslation : node.translation)}</span>
+      {isShortText ?
+        <ShortSidebarLinkTooltip content={t(node.translation)}>
+          {t(node.shortTranslation)}
+        </ShortSidebarLinkTooltip>
+        :
+        <span>
+          {t(node.translation)}
+        </span>
+      }
```
darsene, line 57: as you've created a component named ShortSidebarLinkTooltip, it means it's only for the sidebar (not a generic one). 
In that case, why not creating an entire component to deal with ?

```
{ isShortText ? <SidebarTooltipLink tooltip={content} text={text} /> : <span>
          {t(node.translation)}
        </span>
}
```

oalkabouss: I hadn't thought of that, I've just created it. Thank you for your feedback.

darsene (approved):

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/tooltip/SidebarTooltipLink.tsx
anooparveti, line 1: For just a `span` with title, we dont need a Component. Can you just add it inline?

anooparveti (changes requested):

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/tooltip/SidebarTooltipLink.tsx
oalkabouss: I can add it inline. Initially, we planned to use the ODS component, but I found a bug. Resolving this bug and deploying a new version will take time. Therefore, we switched to Title. I kept the component in case we go back to using ODS, it will be easy to update.

anooparveti (approved):

PR #12382, opened 2024-07-16
Fix(pci-databases analytics): check user inactivity for polling

jperchoc:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/pci-databases-analytics-epic`
| Bug fix?         | yes
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #DATATR-1236 DATATR-1237 DATATR-1238

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

- Remove test dependencies from package.json dependencies
- Check user inactivity to disable polling accordingly
- Add request-tagger


abullet33 (approved):

anooparveti (changes requested):
**question:** What is the need for creating `useImmediateRefetch` wrapper hook for `useQuery`?
If it is just enable/disable the polling upon user activity, no need for this wrapper as just using `refetchInterval: isUserActive && ...` will do the required job.

jperchoc:
Hi @anooparveti ,

> **question:** What is the need for creating `useImmediateRefetch` wrapper hook for `useQuery`? If it is just enable/disable the polling upon user activity, no need for this wrapper as just using `refetchInterval: isUserActive && ...` will do the required job.

The `useImmediateRefetch` check for changes in the polling time, and refetch immediately when it is changed. Without it, the request is not done when the user is active again, but 30s after that.

Here's an example with the hook and a poll interval of 5s. As you can see, as soon as the user is active, the polling starts again: 
093111 user is inactive, stopping polling


Without the hook, the user is active, and the fetch is done on the next iteration of the polling:
093244 user is inactive, stopping polling

anooparveti:
> Hi @anooparveti ,
> 
> > **question:** What is the need for creating `useImmediateRefetch` wrapper hook for `useQuery`? If it is just enable/disable the polling upon user activity, no need for this wrapper as just using `refetchInterval: isUserActive && ...` will do the required job.
> 
> The `useImmediateRefetch` check for changes in the polling time, and refetch immediately when it is changed. Without it, the request is not done when the user is active again, but 30s after that.
> 
> Here's an example with the hook and a poll interval of 5s. As you can see, as soon as the user is active, the polling starts again: 093111 user is inactive, stopping polling
> 
> Without the hook, the user is active, and the fetch is done on the next iteration of the polling: 093244 user is inactive, stopping polling

I think we can live with this behavior. Even with polling, if we fetch data at `X` and something updates on this data at `X + 1s`, the user only sees the new data at `X + 30s` when the data is refetched. So the same behavior will continue to exist on user activity as well and sees the updated data 30s after he interacts with tab. Considering it is the corner case of user inactivity, i dont see much advantages in adding the wrapper hook.

jperchoc:
I strongly disagree. We should always try our best to provide a satisfying experience for our customers. Even if it not "much advantages", it is still an improvement in quality, and we should never prioritize code quality over user experience.

abullet33:
Hey @anooparveti, we modified our code so that the signature matches the hook from the external library. Additionally, the wrapper is now used in all our TanStack hooks to avoid any confusion.

Is that ok for you?

anooparveti (approved):

PR #12578, opened 2024-07-30
fix(dedicated.hpc): fix offset query on operation datagrid

vovh:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-14943 INC0060568

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Hosted Private Cloud - Operation - Next page doesn't work


mhelhali-soufien (approved):

anooparveti (changes requested):
**question:** Since the removal is on the service, can you also check in all the places where this function is used to make sure there is no regression?

darsene:
> **question:** Since the removal is on the service, can you also check in all the places where this function is used to make sure there is no regression?

@anooparveti , yes all has been tested and furthermore, it's the default configuration used in the manager

PR #12582, opened 2024-07-31
feat(dedicated): reload boot details after the netboot

sachinrameshn:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-14866 

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Reload the boot details in general information section after changing the netboot.

### DIFF
--- a/packages/manager/modules/bm-server-components/src/netboot/controller.js
+++ b/packages/manager/modules/bm-server-components/src/netboot/controller.js
@@ -168,17 +168,15 @@ export default class BmServerComponentsNetbootCtrl {
           this.$translate.instant('server_configuration_netboot_success'),
         );
       })
-      .catch((error) =>
-        this.handleError(
+      .catch((error) => {
+        this.loading.setNetboot = false;
+        return this.handleError(
           error,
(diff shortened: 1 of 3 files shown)


qpavy (approved):

packages/manager/modules/bm-server-components/src/netboot/controller.js
```diff
@@ -168,17 +168,15 @@ export default class BmServerComponentsNetbootCtrl {
           this.$translate.instant('server_configuration_netboot_success'),
         );
```
mhelhali-soufien, line 169: mabe should add this.loading.setNetboot = false; in the end of the block ?

sachinrameshn: For this scenario I still need to retain loader for success case thats why I removed finally block and moved this.loading.setNetboot = false; into catch block.

mhelhali-soufien (approved):

PR #12590, opened 2024-07-31
fix(kms): fix breadcrumb

rjamet-ovh:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/key-management-service-B2`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-14767

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

fix breadcrumb:
- use react-router navigation instead of `href` to avoid all app refresh
- handle custom labels (translation or API value)
- handle more than 2 level deep navigation


vovh (approved):

packages/manager/apps/key-management-service/src/pages/dashboard/index.tsx
```diff
@@ -39,9 +43,22 @@ export default function DashboardPage() {
     },
   ];
 
+  const breadcrumbItems: BreadcrumbItem[] = [
+    {
+      id: okmsId,
+      label: displayName || okmsId,
+      navigateTo: `/${okmsId}`,
+    },
```
chipp972, line 51: This item could be integrated directly in the useBreadcrumb hook since it is specific to the kms project and it is duplicated everytime

chipp972 (approved):

packages/manager/apps/key-management-service/src/pages/dashboard/index.tsx
rjamet-ovh: `okmsId` comes from `useParam` and `displayName` comes from an API call, so i don't think it's a good idea to integrate it into the hook.

PR #12592, opened 2024-07-31
feat(*): move vite in devDependencies

aboungnaseng-ovhcloud:
ref: MANAGER-14678

| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix MANAGER-14678

- [x] Commits are signed-off
- [x] Lint has passed locally


packages/manager/core/application/package.json
```diff
@@ -27,19 +27,19 @@
...
-    "react-i18next": "^14.x",
-    "react-router-dom": "^6.2x"
+    "@tanstack/react-query": "^5.37.1",
+    "@tanstack/react-query-devtools": "^5.13.3",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "react-i18next": "^14.0.5",
+    "react-router-dom": "^6.22.0",
+    "vite": "^5.2.13",
+    "vitest": "^1.2.0"
```
anooparveti, line 43: **nitpick:** This must be added at the root `package.json` https://github.com/ovh/manager/blob/master/package.json

anooparveti (changes requested):

packages/manager/apps/vrack-services/package.json
```diff
@@ -52,15 +52,15 @@
     "msw": "2.1.7",
     "ts-jest": "27.x",
     "ts-loader": "^9.5.1",
-    "ts-node": "^10.9.1"
+    "ts-node": "^10.9.1",
+    "vite": "*",
```
frenautvh, line 56: ^5.2.13 ?

aboungnaseng-ovhcloud: 👍

packages/manager/core/application/package.json
```diff
@@ -27,19 +27,17 @@
...
-    "react": "18.x",
-    "react-dom": "18.x",
-    "react-i18next": "^14.x",
-    "react-router-dom": "^6.2x"
+    "@tanstack/react-query": "^5.37.1",
+    "@tanstack/react-query-devtools": "^5.13.3",
+    "react": "^18.2.0",
+    "react-dom": "^18.2.0",
+    "react-i18next": "^14.0.5",
+    "react-router-dom": "^6.22.0"
```
frenautvh, line 41: you have react-router-dom in both peer and deps

also i don't understand because react is in "peer dependency" in every package.json so which module is supposed to bring "react" as a dependency ? shouldn't core application have react in dependency instead of peer dependency (same for others deps declared) ?

packages/manager/core/application/package.json
frenautvh: i disagree, it shouldn't be added at the root of the monorepo's package.json

packages/manager/core/application/package.json
aboungnaseng-ovhcloud: I will delete all the peerDependencies unecessary 👍

anooparveti (approved):

chipp972 (approved):

PR #12715, opened 2024-08-12
fix(account): fix kyc document upload issue

MaximeBajeux:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-14991

- [x] Commits are signed-off
- [x] Lint has passed locally
- [] Breaking change is mentioned in relevant commits

## Description

catch s3 level at the upper level to avoid finish procedure


JacquesLarique (approved):

anooparveti (commented):
**question:** The ticket asks to implement retry mechanism when there is a failure. How does the current change tackle this?

MaximeBajeux:
It's my fault, I wrote the original ticket that was an analysis next to a meeting about incident with s3. I change the status of the ticket to transform it into a bug resolution but the retry loop implementation should be in another ticket to not delayed the bug resolution.

PR #12901, opened 2024-09-02
fix(container): fix popover size user widget

MaximeBajeux:
ref: MANAGER-14840

| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/MANAGER-14238` 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          |  MANAGER-14840

- [x] Commits are signed-off
- [x] Lint has passed locally

## Related

### DIFF
--- a/packages/manager/apps/container/src/container/nav-reshuffle/onboarding/walkMe/index.tsx
+++ b/packages/manager/apps/container/src/container/nav-reshuffle/onboarding/walkMe/index.tsx
@@ -147,10 +147,10 @@ export const OnboardingWalkMe = () => {
   };
 
   const resizeObserver = new ResizeObserver((entries) => {
-    entries.forEach((entry) => {
-      const el: HTMLElement = stepElement.current;
-      el.style.height = `${entry.borderBoxSize[0].blockSize + ELEMENT_OFFSET}px`;
-    })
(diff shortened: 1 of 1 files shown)


packages/manager/apps/container/src/container/nav-reshuffle/onboarding/walkMe/index.tsx
```diff
@@ -146,6 +146,13 @@ export const OnboardingWalkMe = () => {
     }
   };
 
+  const resizeObserver = new ResizeObserver((entries) => {
+    entries.forEach((entry) => {
+      const el: HTMLElement = stepElement.current;
+      el.style.height = `${entry.borderBoxSize[0].blockSize + ELEMENT_OFFSET}px`;
+    })
```
JacquesLarique, line 153: Maybe setup a debounce so that we do not loop over the `entries` too much

MaximeBajeux: I update the function with a find in place of the forEach, a little more efficient. Also I check and mainly the function Is called just once and there is in the entries only the observed elements. So it should be ok in term of performance.

JacquesLarique (approved):

oalkabouss (approved):

PR #12904, opened 2024-09-02
feat(dedicated.vrack): add new eligibility services for vrack

Steffy29:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #UXCT-539

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Use new endpoint call API on vRack eligibility product instead of endpoint call API available services.

### DIFF
--- a/packages/manager/modules/vrack/src/dashboard/vrack.controller.js
+++ b/packages/manager/modules/vrack/src/dashboard/vrack.controller.js
@@ -27,9 +27,12 @@ import without from 'lodash/without';
 import angular from 'angular';
 
 import {
+  API_STATUS,
+  DEDICATED_CLOUD_PRODUCT,
   FEATURE_NAMES,
   POLLING_INTERVAL,
-  STATUS,
+  SERVICES,
+  TYPE_SERVICE,
   VRACK_DASHBOARD_TRACKING_PREFIX,
   VRACK_ACTIONS_SUFFIX,
   VRACK_URLS,
@@ -98,6 +101,7 @@ export default class VrackMoveDialogCtrl {
     this.ipRegions = {};
     this.addIpv6ModalState = null;
     this.hasIpv6 = false;
+    this.TYPE_SERVICE = TYPE_SERVICE;
 
     if (this.coreConfig.getRegion() === 'US') {
       this.surveyUrl = US_SURVEY_LINK;
@@ -157,11 +161,10 @@ export default class VrackMoveDialogCtrl {
       adding: false,
       deleting: false,
       moving: false,
+      eligibleServices: false,
     };
 
     this.data = {
-      cloudProjects: [],
-      allowedServices: {},
       vrackServices: {},
       pendingTasks: [],
     };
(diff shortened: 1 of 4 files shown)


packages/manager/modules/vrack/src/dashboard/vrack.controller.js
```diff
@@ -264,82 +265,371 @@ export default class VrackMoveDialogCtrl {
...
+
+  getEligibleServices() {
+    return this.vrackService
+      .getEligibleServices(this.serviceName)
+      .then(({ data }) => {
+        if (!this.data.eligibleServices) {
+          this.data.eligibleServices = data.result;
+
+          // Update eligible services infos for each service not empty
+          if (
```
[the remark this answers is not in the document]
Steffy29: Yes it could be, but in the same time we could have many service should be update and if I use a switch case, I could update only one service. it's not a "if... else if..." but a "if... if"

packages/manager/modules/vrack/src/dashboard/vrack.controller.js
```diff
@@ -264,82 +266,374 @@ export default class VrackMoveDialogCtrl {
...
+        }),
+      )
+      .then((dedicatedServerInterface) => {
+        this.data.eligibleServices.dedicatedServerInterface = dedicatedServerInterface;
 
         // We need to append dedicatedServerInterfaces list to dedicatedServers list.
         if (
-          has(allServices, 'dedicatedServerInterface') &&
-          allServices.dedicatedServerInterface.length > 0
+          has(
```
antonymarion, line 367: we try to avoid lodash or underscore usage (even if it is true that our legacy is using it)

packages/manager/modules/vrack/src/dashboard/vrack.controller.js
```diff
@@ -264,82 +266,374 @@ export default class VrackMoveDialogCtrl {
...
+          niceName: service,
+          trueServiceType: TYPE_SERVICE.legacyVrack,
+        };
+        return formattedService;
+      },
     );
+    this.data.eligibleServices.legacyVrack = legacyVrackUpdated;
+  }
+
+  updateEligibleServices(result) {
```
antonymarion, line 446: This is quite a big function. it won't ease maintanability

packages/manager/modules/vrack/src/dashboard/vrack.controller.js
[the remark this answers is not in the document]
antonymarion: getEligibleServices and updateDedicatedCloudServiceInfo are not easy to read since too big.

Maybe a trick would be first to separate the logic of the updateDedicatedCloudServiceInfo  into several dedicated updateDedicatedCloudServiceInfo{Suffix} functions where you implement a logic for the given suffix.

Then a refactor on the common keys to avoid the `if if if if if` since you will have dedicated function, so it will fit the switch/case proposed above

packages/manager/modules/vrack/src/dashboard/vrack.controller.js
Steffy29: I just move some code lines.

packages/manager/modules/vrack/src/dashboard/vrack.controller.js
Steffy29: Yes we need to do many things.

antonymarion (approved):

PR #12907, opened 2024-09-02
feat(hub): created order tracking component

oalkabouss:
ref:MANAGER-15066

| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/hub-revamp-batch-1`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-15066

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager/apps/hub/src/components/hub-order-tracking/HubOrderTracking.component.tsx
+++ b/packages/manager/apps/hub/src/components/hub-order-tracking/HubOrderTracking.component.tsx
@@ -1,0 +1,223 @@
+import React, { FunctionComponent, Suspense, useContext, useMemo } from 'react';
+import {
+  OsdsChip,
+  OsdsIcon,
+  OsdsLink,
+  OsdsSkeleton,
+  OsdsText,
+  OsdsTile,
+} from '@ovhcloud/ods-components/react';
+import {
+  ODS_TILE_VARIANT,
+  ODS_CHIP_SIZE,
+  ODS_ICON_NAME,
+  ODS_TEXT_SIZE,
+  ODS_ICON_SIZE,
+} from '@ovhcloud/ods-components';
+import {
+  ODS_THEME_COLOR_INTENT,
+  ODS_THEME_TYPOGRAPHY_LEVEL,
+} from '@ovhcloud/ods-common-theming';
+import { useTranslation } from 'react-i18next';
+import {
+  OdsHTMLAnchorElementRel,
+  OdsHTMLAnchorElementTarget,
+} from '@ovhcloud/ods-common-core';
+import {
+  ButtonType,
+  ShellContext,
+  useOvhTracking,
+} from '@ovh-ux/manager-react-shell-client';
+import { Await } from 'react-router-dom';
+import { useFetchLastOrder } from '@/data/hooks/apiOrder/useLastOrder';
+import {
+  ERROR_STATUS,
+  WAITING_PAYMENT_LABEL,
+} from '@/data/api/apiOrder/apiOrder.constants';
+import useDateFormat from '@/hooks/dateFormat/useDateFormat';
+import { Skeletons } from '../skeletons/Skeletons.component';
+import { LastOrderTrackingResponse, OrderHistory } from '@/types/order.type';
+import { TileError } from '../tile-error/TileError.component';
+
+export const HubOrderTracking: FunctionComponent = () => {
+  const { t } = useTranslation('hub/order');
+  const {
+    data: orderDataResponse,
+    isLoading,
+    error,
+    refetch,
+  } = useFetchLastOrder();
+  const context = useContext(ShellContext);
+  const { navigation } = context.shell;
+  const { trackClick } = useOvhTracking();
+
+  const orderTrackingLinkAsync = useMemo(
+    () =>
+      navigation.getURL(
+        'dedicated',
+        `#/billing/order/${orderDataResponse?.orderId}`,
+        {},
+      ),
+    [orderDataResponse?.orderId],
+  );
+
+  const ordersTrackingLinkAsync = useMemo(
+    () => navigation.getURL('dedicated', `#/billing/orders`, {}),
+    [],
+  );
+
+  const getInitialStatus = (orderData: LastOrderTrackingResponse) => ({
+    date: orderData.date,
+    label:
+      orderData.status === 'delivered'
+        ? 'INVOICE_IN_PROGRESS'
+        : 'custom_creation',
+  });
+
+  const getLatestStatus = (history: OrderHistory[]) => {
+    return history.reduce((latest, item) => {
+      return new Date(item.date).getTime() > new Date(latest.date).getTime()
+        ? item
+        : latest;
+    });
+  };
+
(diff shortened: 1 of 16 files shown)


packages/manager/apps/hub/src/data/api/apiOrder/apiOrder.constants.ts
JacquesLarique, line 1: These constant are available in '@ovh-ux/ng-ovh-order-tracking', but as a matter of fact this module is an Angular JS module (which should not be integrated in a React app)

So IMO, we should keep this constants file and remove '@ovh-ux/ng-ovh-order-tracking' from the package.json, and in the future split the '@ovh-ux/ng-ovh-order-tracking' module into '@ovh-ux/ovh-order-tracking' (keeping pure JS/TS code) and '@ovh-ux/ng-ovh-order-tracking' (with the Angular JS specific code)

packages/manager/apps/hub/src/data/api/apiOrder/apiOrder.ts
```diff
@@ -0,0 +1,85 @@
...
+export const getLastOrder: () => Promise<OrderResponseData> = async () =>
+  aapi.get('/hub/lastOrder').then(({ data }) => {
+    return data.data.lastOrder;
+  });
+
+export const getOrderDetails = async (
+  orderId: number,
+): Promise<OrderDetailsResponse> => {
+  const response = await v6.get(`/me/order/${orderId}/details`);
+  return response.data;
```
JacquesLarique, line 21:
```suggestion
  const { data } = await v6.get(`/me/order/${orderId}/details`);
  return data;
```

packages/manager/apps/hub/src/data/api/apiOrder/apiOrder.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+export const getOrderStatus = async (orderId: number): Promise<OrderStatus> => {
+  const response = await v6.get(`/me/order/${orderId}/status`);
+  return response.data as OrderStatus;
```
JacquesLarique, line 26:
```suggestion
  const { data } = await v6.get(`/me/order/${orderId}/status`);
  return data as OrderStatus;
```

packages/manager/apps/hub/src/data/api/apiOrder/apiOrder.ts
```diff
@@ -0,0 +1,85 @@
...
+};
+
+export const getOrderFollowUp = async (
+  orderId: number,
+): Promise<OrderFollowUpResponse> => {
+  const response = await v6.get(`/me/order/${orderId}/followUp`);
+  return response.data as OrderFollowUpResponse;
```
JacquesLarique, line 33:
```suggestion
  const { data } = await v6.get(`/me/order/${orderId}/followUp`);
  return data as OrderFollowUpResponse;
```

packages/manager/apps/hub/src/data/api/apiOrder/apiOrder.ts
```diff
@@ -0,0 +1,85 @@
...
+export const getCompleteHistory = async (
+  orderId: number,
+  orderStatus: string,
+  orderDate: string,
+): Promise<OrderHistory[]> => {
+  try {
+    const followUp = await getOrderFollowUp(orderId);
+    const history = followUp.flatMap((follow) => follow.history).reverse();
+
+    if (orderStatus === 'notPaid' && history.length === 0) {
```
JacquesLarique, line 45: Why not use the constant here ?
This may require you to redo the definition of ERROR_STATUS

JacquesLarique (changes requested):

packages/manager/apps/hub/src/components/hub-order-tracking/style.module.scss
```diff
@@ -0,0 +1,7 @@
+.manager_hub_order_tracking {
+  @import '@ovh-ux/manager-hub/src/variables.scss';
+  @import '@ovh-ux/ui-kit/dist/scss/_tokens';
```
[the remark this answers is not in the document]
oalkabouss: It was just to import the background color of the card.

packages/manager/apps/hub/src/components/tile-skeleton/TileSkeleton.component.tsx
```diff
@@ -0,0 +1,11 @@
+import React, { FunctionComponent } from 'react';
+import { OsdsSkeleton } from '@ovhcloud/ods-components/react';
+
+export const TileSkeleton: FunctionComponent = () => (
```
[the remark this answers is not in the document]
oalkabouss: You are right about the Skeletons. It doesn't make sense to call it TileSkeleton since it is not wrapped in a tile. i named it that way becaus i use it inside a tile, which is why i have 4 Skeleton. i propose creating a Skeleton component with a prop for the number of Skeletons to generate, for example <Skeleton count={4} />. What do you think?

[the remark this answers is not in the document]
JacquesLarique: Not sure about that (as I'm not sure I understand the usage guidelines of the skeleton) but in my opinion your usage of skeletons depends on the layout of the component where you're using skeleton and as such I'm not sure we can have a "generic" component for it

packages/manager/apps/hub/src/data/api/apiOrder/apiOrder.ts
```diff
@@ -0,0 +1,89 @@
...
+import {
+  OrderFollowUpResponse,
+  OrderResponseData,
+  OrderDetailsResponse,
+  OrderStatus,
+  OrderHistory,
+  LastOrderTrackingResponse,
+} from '@/types/order.type';
+import {
+  ERROR_STATUS,
```
JacquesLarique, line 11: ERROR_STATUS doesn't seem to be used in this file

packages/manager/apps/hub/src/components/hub-order-tracking/HubOrderTracking.component.tsx
```diff
@@ -0,0 +1,220 @@
...
+      const urlOrders = (await navigation.getURL(
+        'dedicated',
+        `#/billing/orders`,
+        {},
+      )) as string;
+
+      setOrderTrackingLink(urlOrder);
+      setOrdersTrackingLink(urlOrders);
+    })();
+  }, [orderDataResponse]);
```
JacquesLarique, line 79: I feel like it is clearer to use an <Await> component to handle the link creation / retrieval, as you don't have this superfluous code to handle the fact that `getURL` is asynchronous.

Moreover, using `<Await>` allow you to encapsulate it in a `<Suspense>` and display a Skeleton while URL is being retrieved.

You can find an example here: https://github.com/ovh/manager/pull/12951/files#diff-94fad87b597e8757a7f97eb5939de7aed32d15f34c52ca50a3c1a7f746925982R113

JacquesLarique (approved):

MaximeBajeux (approved):

Issue #12978, opened 2024-09-06
[manager-components]: hook for service details API


chipp972:
### Have you already checked if a similar item is present on manager-components?

- [ ] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Hook

### Description

We already have hooks to centralize the implementation of /services API
We have :
- get /services
- put /services/:id
- post /services/:id/terminate

It would be nice to also have the get /services/:id to query the details of a service and to not have to duplicate the logic and types in multiple project (it is usually used in the dashboards to display subscription information of a service).

### Where do you expect to use this?

veeam-backup

### Do you have mock-up?

https://www.figma.com/design/n7Qg4YVkpI8Qa5H8B4m009/VCD-Backup-experience?node-id=1-86&node-type=CANVAS&t=uFqZTwRFrhYkxQaV-0

PR #12979, opened 2024-09-06
feat(manager-react-components): hook for service details api

chipp972:
ref: #12978

| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | ISSUE-12978 

- [x] Commits are signed-off
- [x] Lint has passed locally
- [ ] ~~Standalone app was ran and tested locally~~

### DIFF
--- a/packages/manager-react-components/src/hooks/services/hooks/useServiceDetails.ts
+++ b/packages/manager-react-components/src/hooks/services/hooks/useServiceDetails.ts
@@ -1,0 +1,38 @@
+import { ApiError, ApiResponse } from '@ovh-ux/manager-core-api';
+import { useQuery, useQueryClient } from '@tanstack/react-query';
+import {
+  getResourceServiceId,
+  getResourceServiceIdQueryKey,
+  getServiceDetails,
+} from '../api';
+import { ServiceDetails } from '../services.type';
+
+export const getServiceDetailsQueryKey = (resourceName: string) => [
+  'service-details',
+  resourceName,
+];
+
+export type UseServiceDetailsParams = {
+  queryKey?: string[];
+  resourceName: string;
+};
+
+export const useServiceDetails = ({
+  queryKey,
+  resourceName,
+}: UseServiceDetailsParams) => {
+  const queryClient = useQueryClient();
+  return useQuery<ApiResponse<ServiceDetails>, ApiError>({
+    queryKey: queryKey ?? getServiceDetailsQueryKey(resourceName),
+    queryFn: async () => {
+      const { data } = await queryClient.fetchQuery<
+        ApiResponse<number[]>,
+        ApiError
+      >({
+        queryKey: getResourceServiceIdQueryKey({ resourceName }),
+        queryFn: () => getResourceServiceId({ resourceName }),
+      });
+      return getServiceDetails(data[0]);
+    },
+  });
+};
--- a/packages/manager-react-components/src/hooks/services/services.type.ts
+++ b/packages/manager-react-components/src/hooks/services/services.type.ts
@@ -1,0 +1,148 @@
+import { CurrencyCode } from '../../enumTypes';
+
+export type EndRuleStrategy =
+  | 'CANCEL_SERVICE'
+  | 'REACTIVATE_ENGAGEMENT'
+  | 'STOP_ENGAGEMENT_FALLBACK_DEFAULT_PRICE'
+  | 'STOP_ENGAGEMENT_KEEP_PRICE';
+
+export type LifecycleAction =
+  | 'earlyRenewal'
+  | 'terminate'
+  | 'terminateAtEngagementDate'
+  | 'terminateAtExpirationDate';
+
+export type LifecycleState =
+  | 'active'
+  | 'error'
+  | 'rupture'
+  | 'terminated'
+  | 'toRenew'
+  | 'unpaid'
+  | 'unrenewed';
+
+export type PricingCapacity =
+  | 'consumption'
+  | 'detach'
+  | 'downgrade'
+  | 'dynamic'
+  | 'installation'
+  | 'renew'
+  | 'upgrade';
+
+export type EndAction =
+  | 'CANCEL_SERVICE'
+  | 'REACTIVATE_ENGAGEMENT'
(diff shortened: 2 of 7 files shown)


darsene (dismissed):

packages/manager-react-components/src/hooks/services/hooks/useServiceDetails.ts
```diff
@@ -0,0 +1,35 @@
+import { ApiError, ApiResponse } from '@ovh-ux/manager-core-api';
+import { useQuery, useQueryClient } from '@tanstack/react-query';
+import {
+  getResourceServiceId,
+  getResourceServiceIdQueryKey,
+  getServiceDetails,
+} from '../api';
+import { ServiceDetails } from '../services.type';
+
+export const serviceDetailsQueryKey = ['service-details'];
```
anooparveti, line 10: **nitpick:** Query key should also include serviceId otherwise can have a wrong result is query is made with a different service due to caching.
```suggestion
export const serviceDetailsQueryKey = ['service-details'];
```

packages/manager-react-components/src/hooks/services/services.type.ts
```diff
@@ -0,0 +1,151 @@
...
+        createDate: string;
+        pendingActions: LifecycleAction[];
+        state:
+          | 'active'
+          | 'error'
+          | 'rupture'
+          | 'terminated'
+          | 'toRenew'
+          | 'unpaid'
+          | 'unrenewed';
```
rjamet-ovh, line 62: why not extract the remaning enums as their own types like your did for `ResourceStatus` or `LifecycleAction` for example ?

packages/manager-react-components/src/hooks/services/hooks/useServiceDetails.ts
tibs245:
```suggestion
export const serviceDetailsQueryKey = (resourceName) => ['service-details', resourceName];
```

packages/manager-react-components/src/hooks/services/hooks/useServiceDetails.ts
```diff
@@ -0,0 +1,35 @@
...
+        ApiResponse<number[]>,
+        ApiError
+      >({
+        queryKey: getResourceServiceIdQueryKey({ resourceName }),
+        queryFn: () => getResourceServiceId({ resourceName }),
+      });
+      return getServiceDetails(data[0]);
+    },
+  });
+};
```
tibs245, line 35: For readability, may be it's better to use 2 useQuery with this second query a disabled key for wait data[0] ?

chipp972: There are 2 problems :
- you double the code to maintain for the same result
- if I return the result of the second useQuery, it won't be loading until the first query returns. Or I could return the isLoading of the first query. Same thing for the error management. Why bother when I could just use one useQuery ?

packages/manager-react-components/src/hooks/services/hooks/useServiceDetails.ts
chipp972: Done

packages/manager-react-components/src/hooks/services/services.type.ts
chipp972: Done

packages/manager-react-components/src/hooks/services/hooks/useServiceDetails.ts
tibs245: I think is more readable and more reusable. This avoids rewrite useResource if we needs and recreate different `queryKey`. 

We can update serviceDetails query key too

```ts
export const useResource = ({
  resourceName,
}: UseServiceDetailsParams) => {
  return useQuery<ApiResponse<Resource>, ApiError>({
    queryKey: getResourceServiceIdQueryKey({ resourceName }),
    queryFn: getResourceServiceId({ resourceName }),
  });
};

export const useServiceDetails = ({
  resourceName,
}: UseServiceDetailsParams) => {
  const {resourceData, isSuccess} = useResource(resourceName)

  return useQuery<ApiResponse<ServiceDetails>, ApiError>({
    queryKey: generateMyKey(resourceName, resourceData[0]),
    queryFn: getServiceDetails(resourceData[0]),
    enabled: isSuccess && resourceData,
  });

  return queryClient
};

```

tibs245: The code is in two function but less complex. So easier to maintain
And for isFetching and isError

 ```ts
export const useServiceDetails = ({
  resourceName,
}: UseServiceDetailsParams) => {
  const {resourceData, isSuccess} = useResource(resourceName)

  const { isError, error, isSuccess, isFetching, isPending, ...rest} =  useQuery<ApiResponse<ServiceDetails>, ApiError>({
    queryKey: generateMyKey(resourceName, resourceData[0]),
    queryFn: getServiceDetails(resourceData[0]),
    enabled: isSuccess && resourceData,
  });

  return { isError: isError || resourceData.isError, error: error || resourceData.error, isSuccess: isSuccess && resourceData.isSuccess, isFetching: isFetching || resourceData.isFetching, isPending: isPending: resourceData.isPending, ...rest }
};
```

I think we can separate this merge mechanism in a useful utils to simplify this function

anooparveti (approved):

rjamet-ovh (approved):

packages/manager-react-components/src/hooks/services/hooks/useServiceDetails.ts
tibs245: It's the principle of Single responsibility of SOLID to improve maintainable code and avoid duplicate code.

I'm agree on the first view it's not very complex and easier to write this function with the good finality.

But I think in this case this principle can be good in the time.

chipp972: "Single responsibility" != "Single operation"
The hook has only one responsibility : querying the service details data from the resource name.

The only benefit I see from extracting the part that query the service id from the resource name would be reusability (since it doesn't improve maintainability or performance as stated previously) but I don't see any use-case where we use the service id separately.

Do you see any other benefit ?

PR #12983, opened 2024-09-06
feat(manager-react-components): useFormattedDate hook

chipp972:
ref: ISSUE-12982

| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | ISSUE-12982 

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager-react-components/src/hooks/date/useFormattedDate.spec.tsx
+++ b/packages/manager-react-components/src/hooks/date/useFormattedDate.spec.tsx
@@ -1,0 +1,77 @@
+import { renderHook } from '@testing-library/react';
+import {
+  useFormattedDate,
+  defaultUnknownDateLabel,
+  DateFormat,
+} from './useFormattedDate';
+
+jest.mock('react-i18next', () => ({
+  useTranslation: () => ({
+    t: (key: string) => key,
+    i18n: {
+      language: 'fr-FR',
+    },
+  }),
+}));
+
+describe('useFormattedDate', () => {
+  it.each([
+    {
+      case: 'label for no date',
+      input: 'invalid',
+      dateString: undefined,
+      format: undefined,
+      expected: defaultUnknownDateLabel,
+    },
+    {
+      case: 'label for no date',
+      input: 'empty',
+      dateString: '',
+      format: undefined,
+      expected: defaultUnknownDateLabel,
+    },
+    {
+      case: 'a valid date',
+      input: 'null',
+      dateString: null,
+      format: undefined,
+      expected: '1 janv. 1970',
+    },
+    {
+      case: 'a valid date with abbreviated month',
+      input: 'valid',
+      dateString: '2024-09-14T09:21:21.943Z',
+      format: undefined,
+      expected: '14 sept. 2024',
+    },
+    {
+      case: 'a valid date with abbreviated month',
+      input: 'valid',
+      dateString: '2024-10-14T09:21:21.943Z',
+      format: DateFormat.display,
+      expected: '14 oct. 2024',
+    },
+    {
+      case: 'a valid date with non-abbreviated month',
+      input: 'valid',
+      dateString: '2024-09-14T09:21:21.943Z',
+      format: DateFormat.fullDisplay,
+      expected: '14 septembre 2024',
+    },
+    {
+      case: 'a valid date with compact format',
+      input: 'valid and format is compact',
+      dateString: '2024-06-14T09:21:21.943Z',
+      format: DateFormat.compact,
+      expected: '14/06/2024',
+    },
+  ])(
+    'displays %s if the date is %s',
+    async ({ dateString, format, expected }) => {
+      const { result } = renderHook(() =>
+        useFormattedDate({ dateString, format }),
+      );
+      expect(result.current).toBe(expected);
+    },
+  );
+});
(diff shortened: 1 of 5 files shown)


darsene (approved):

darsene (approved):

packages/manager-react-components/src/hooks/date/useFormattedDate.ts
```diff
@@ -0,0 +1,46 @@
...
+  unknownDateLabel?: string;
+  defaultLocale?: string;
+  format?: DateFormat;
+}) => {
+  const { i18n } = useTranslation('date-utils');
+  const date = new Date(dateString);
+  const locale = i18n?.language?.replace('_', '-') || defaultLocale;
+
+  if (date.toString() === 'Invalid Date') {
+    return unknownDateLabel;
```
tibs245, line 36: As `unknownDateLabel` is optionnal maybe add : 

I don't known if it's relevant

```suggestion
    return unknownDateLabel ?? DEFAULT_INVALID_DATE_LABEL;
```

packages/manager-react-components/src/hooks/date/useFormattedDate.ts
```diff
@@ -0,0 +1,46 @@
...
+  compact = 'compact',
+  /**
+   * dd abbreviated month YYYY
+   */
+  display = 'display',
+  /**
+   * dd month YYYY
+   */
+  fullDisplay = 'fullDisplay',
+}
```
tibs245, line 18: It's just for my information. I imagine that you have already had a debate on this.

Why don't use : 

```suggestion
export type DateFormat = "COMPACT" | "DISPLAY" | "FULL_DISPLAY";
```

packages/manager-react-components/src/hooks/date/useFormattedDate.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+
+describe('useFormattedDate', () => {
+  it.each([
+    [
+      'label for no date',
+      'invalid',
+      undefined,
+      undefined,
+      defaultUnknownDateLabel,
+    ],
```
rjamet-ovh, line 27: it would be easier to read as an object

```
{
  label: 'a label',
  ...
}
```

packages/manager-react-components/src/hooks/date/useFormattedDate.ts
anooparveti, line 1: **nitpick:** Date display also depends on the locale. Can you try to integrate locale along with the format as well.

anooparveti (changes requested):

packages/manager-react-components/src/hooks/date/useFormattedDate.spec.tsx
```diff
@@ -0,0 +1,67 @@
...
+jest.mock('react-i18next', () => ({
+  useTranslation: () => {
+    return {
+      t: (key: string) => key,
+      i18n: {
+        language: 'fr-FR',
+      },
+    };
+  },
+}));
```
tibs245, line 17: May be relevant to do test in english too ?

packages/manager-react-components/src/hooks/date/useFormattedDate.ts
```diff
@@ -0,0 +1,46 @@
+import { useTranslation } from 'react-i18next';
+
+export const defaultUnknownDateLabel = 'N/A';
```
rjamet-ovh, line 3: shouldn't  that label be translated?

packages/manager-react-components/src/hooks/date/useFormattedDate.ts
chipp972: We didn't have a debate on this but imo :
If you have to use the variable it's clearer to have an enum.

If you have only to compare to a property, having an union type is lighter.

But in this case, you will have to type :

`useFormattedDate({ dateString: 'something', format: DateFormat.xxx })

packages/manager-react-components/src/hooks/date/useFormattedDate.ts
chipp972: The locale is detected with react-i18n

packages/manager-react-components/src/hooks/date/useFormattedDate.ts
chipp972: Well actually I put this to have a default value without translation and we can provide a translated one via the parameters.

But maybe we should return undefined in this case and let the app manage the absence of value

Or have a value that works for every project, I don't know.

packages/manager-react-components/src/hooks/date/useFormattedDate.spec.tsx
chipp972: Yes that's better for sure

packages/manager-react-components/src/hooks/date/useFormattedDate.spec.tsx
chipp972: I did them in a different file since we can only mock a module once and I can't change the value dynamically

packages/manager-react-components/src/hooks/date/useFormattedDate.ts
tibs245: Ah yes ! My bad

rjamet-ovh (approved):

tibs245 (approved):

PR #12987, opened 2024-09-06
fix(manager-react-components): fix onboarding template image

chipp972:
ref: ISSUE-12986

| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | ISSUE-12986 

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager-react-components/src/components/templates/onboarding/onboarding.stories.tsx
+++ b/packages/manager-react-components/src/components/templates/onboarding/onboarding.stories.tsx
@@ -5,6 +5,7 @@ import {
   OnboardingLayoutProps,
 } from './onboarding.component';
 import { Card } from '../../navigation/card/card.component';
+import placeholderSrc from './assets/placeholder.png';
 
 export const defaultProps: OnboardingLayoutProps = {
   title: 'Welcome to Onboarding',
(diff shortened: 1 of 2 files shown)


darsene (approved):

packages/manager-react-components/src/components/templates/onboarding/onboarding.stories.tsx
```diff
@@ -5,6 +5,7 @@ import {
   OnboardingLayoutProps,
 } from './onboarding.component';
 import { Card } from '../../navigation/card/card.component';
+import placeholderSrc from './assets/placeholder.png';
```
tibs245, line 8: I known it's done in app with this nomenclature.

But why don't use constante nomenclature for external image imported ? 
```suggestion
import PLACE_HOLDER_SRC from './assets/placeholder.png';
```

anooparveti (approved):

packages/manager-react-components/src/components/templates/onboarding/onboarding.stories.tsx
chipp972: Why use "constante nomenclature" actually ?
We rarely mutate variable in React apps

tibs245: To known it's external value declared on top of file. Not a params or a local variable (As declared as const)

It's for differentiate scope directly when we read the code.

PR #13134, opened 2024-09-17
feat(onboarding): improve onboarding template component

tibs245:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | None

- [x] Commits are signed-off
- [x] Lint has passed locally
- [ ] ~Ticket reference is mentioned in linked commits (internal only)~

## Description

Improve img props type to add more flexibility on image rules and be more readable

### DIFF
--- a/packages/manager-react-components/src/components/templates/onboarding/onboarding.component.tsx
+++ b/packages/manager-react-components/src/components/templates/onboarding/onboarding.component.tsx
@@ -32,11 +32,7 @@ export type OnboardingLayoutProps = OnboardingLayoutButtonProps &
     hideHeadingSection?: boolean;
     title: string;
     description?: React.ReactNode;
-    img?: {
-      src: string;
-      width?: number;
-      height?: number;
-    };
+    img?: React.ComponentProps<'img'>;
   }>;
 
 const OnboardingLayoutButton: React.FC<OnboardingLayoutButtonProps> = ({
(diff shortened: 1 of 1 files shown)


fredericvilcot (dismissed):
Great improvement 👍🏻

packages/manager-react-components/src/components/templates/onboarding/onboarding.component.tsx
```diff
@@ -106,18 +102,8 @@ export const OnboardingLayout: React.FC<OnboardingLayoutProps> = ({
...
-                className="max-h-[150px]"
-                src={img?.src ?? placeholderSrc}
-                alt=""
-                width={img?.width}
-                height={img?.height}
-              />
-            </div>
-          )}
+        <section className="flex flex-col items-center mt-8">
+          <img className="max-h-28" src={placeholderSrc} {...img} />
```
anooparveti, line 106: **nitpick(blocking):**`placeHolderSrc` is just an example image. We have to consider the `src` from `img` tag if it's present. This causes confusion if it display the `img.src` or `placeholderSrc`.

anooparveti (changes requested):

chipp972 (dismissed):

packages/manager-react-components/src/components/templates/onboarding/onboarding.component.tsx
chipp972: Maybe we should concat img.className with the current className instead of overwriting it completely though

tibs245: It's a choice. 
- No do that allow to have more flexibility.
- Do that add behavior by default but can add adverse reaction on specific case.

Example : On MANAGER-14496. We have image with text title. So I want set `max-h-36`

It's easier it className is overwrite. And if we add a className. I think you can be responsible of all display rules.

After. May be export "DEFAULT_MAX_HEIGHT_CLASS" if we want add possibility to extends on parent component ?

I try to keep simple. That's how I did it. But not hesitate if you think it's better to do otherwise

anooparveti: I agree with @chipp972 for concatenating the classname.

tibs245: Done. Not hesitate to say me if change is good for you ;)

packages/manager-react-components/src/components/templates/onboarding/onboarding.component.tsx
```diff
@@ -103,21 +99,16 @@ export const OnboardingLayout: React.FC<OnboardingLayoutProps> = ({
   img,
```
chipp972, line 99:
```suggestion
  img = {},
```

To be able to destructure `img` in all cases (solves the breaking test)

darsene (changes requested):
Unit test to fix

chipp972 (dismissed):

packages/manager-react-components/src/components/templates/onboarding/onboarding.component.tsx
```diff
@@ -103,24 +99,19 @@ export const OnboardingLayout: React.FC<OnboardingLayoutProps> = ({
   children,
   onOrderButtonClick,
   onmoreInfoButtonClick,
-  img,
+  img = {},
   isActionDisabled,
 }) => {
+  const { className: brandImageClassName, ...brandImageProps } = img;
```
anooparveti, line 105:
```suggestion
  const { className: imgClassName, ...imgProps } = img;
```

chipp972 (dismissed):

darsene (approved):

anooparveti (approved):

PR #13159, opened 2024-09-18
fix(pci-instances): improve cache management for refresh function

fredericvilcot:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/pci-instances`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #1724

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

This Pr ensures that cache is well handled when user refreshes the list, notably with active pagination.

### DIFF
--- a/packages/manager/apps/pci-instances/src/pages/instances/Instances.page.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/Instances.page.tsx
@@ -43,7 +43,7 @@ import { Spinner } from '@/components/spinner/Spinner.component';
 import { TInstance, useInstances } from '@/data/hooks/instances/useInstances';
 import StatusChip from '@/components/statusChip/StatusChip.component';
 
-const initialSort = {
+const initialSorting = {
   id: 'name',
   desc: false,
 };
@@ -53,7 +53,7 @@ const Instances: FC = () => {
   const { projectId } = useParams() as { projectId: string }; // safe because projectId has already been handled by async route loader
   const project = useRouteLoaderData('root') as PublicCloudProject;
   const navigate = useNavigate();
-  const [sorting, setSorting] = useState(initialSort);
+  const [sorting, setSorting] = useState(initialSorting);
   const [searchField, setSearchField] = useState('');
   const { filters, addFilter, removeFilter } = useColumnFilters();
   const {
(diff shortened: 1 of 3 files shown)


Tsiorifamonjena (approved):

packages/manager/apps/pci-instances/src/pages/instances/Instances.page.tsx
```diff
@@ -211,9 +211,10 @@ const Instances: FC = () => {
   );
 
   const resetSortAndFilters = useCallback(() => {
-    setSorting(initialSort);
-    filters.forEach(removeFilter);
-  }, [filters]);
+    if (filters.length) filters.forEach(removeFilter);
+    if (JSON.stringify(sorting) !== JSON.stringify(initialSorting))
```
SimonChaumet, line 215: It seems to be a downfall on performance, could we use something else to make a deep compare ?

SimonChaumet (approved):

packages/manager/apps/pci-instances/src/pages/instances/Instances.page.tsx
fredericvilcot: @SimonChaumet Yep, agree on that but I see no other way to achieve that in a better way.
Maybe using [_isEqual](https://lodash.com/docs/4.17.15#isEqual) from Lodash, but I'm not sure about better performance, what do you think?

SimonChaumet: You can check the tests here : https://jsperf.app/lo-dash-vs-underscore-vs-json-stringify-isequal/2

image

fredericvilcot: Thanks @SimonChaumet for the link:)
I also ran the test, that shows that `lodash.isEqual` is the slowest, and `underscore.isEqual` is a bit better: 
Capture d’écran 2024-09-18 à 18 18 32
`JSON.stringify` seems to be in the middle, so, if you don't mind, I'll suggest to keep it like that for just such a tinny usage, instead of downloading the whole `underscore.js` library that is not declared yet in the manager.

PR #13188, opened 2024-09-19
feat(account): set up retry on error for kyc india procedure creation

JacquesLarique:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-14992

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Split upload documents call into two separates "tasks" as to be able to avoid calling the first part after it is successful, since it would end up failing everytime.
Set up a retry mecanism, so in the case of a failure, we'll retry to upload the document automatically after 3s.
Also "fixed" the CTA, so that if the user want to retry, he should be able (thanks to the first split)

### DIFF
--- a/packages/manager/apps/dedicated/client/app/account/identity-documents/user-identity-documents.controller.js
+++ b/packages/manager/apps/dedicated/client/app/account/identity-documents/user-identity-documents.controller.js
@@ -7,6 +7,8 @@ import {
   LEGAL_LINK3,
   KYC_STATUS,
   KYC_ALLOWED_FILE_EXTENSIONS,
+  DELAY_BETWEEN_RETRY,
+  MAX_RETRIES,
 } from './user-identity-documents.constant';
 
 export default class AccountUserIdentityDocumentsController {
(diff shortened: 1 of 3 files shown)


packages/manager/apps/dedicated/client/app/account/identity-documents/user-identity-documents.controller.js
```diff
@@ -72,35 +90,53 @@ export default class AccountUserIdentityDocumentsController {
...
+            try {
+              /* eslint-disable no-await-in-loop */
+              await this.finalizeProcedure(uploadLinks);
+              success = true;
+            } catch (error) {
+              remainingRetries -= 1;
+            }
+          }
+          return success ? this.$q.resolve() : this.$q.reject();
+        })
```
anooparveti, line 123: **question:** Does not this have any side-effect with respect to digest-cycle. Because AngularJS does not pick changes in `setTimeout` and `async await` and update the UI which would require to run the digest cycle manually.

JacquesLarique: It should not as in this function we do not make changes on UI, the UI changes are done in `then` or `catch` triggered by this Promise

anooparveti (approved):

PR #13319, opened 2024-09-27
fix(manager-component): fix and refactor useResourcesIcebergV6

tibs245:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-14495 #MANAGER-15573

- [x] Commits are signed-off
- [x] Lint has passed locally
- [ ] ~Standalone app was ran and tested locally~

## Description

Fix `useResoucesIceberg` because when I read the code we can see if 2 components call this `useResoucesIceberg` or we go back on page with the same app we trigger more one time fetchNextPage.

So we have multiple same data in result. Multiple same pages on data.

So : I proposal use `fetchNextPage` only on `onFetchNextPage` callback call.

I refactor too variable only managed with `useEffect`. I think `useMemo` is more efficient and more readable on this case.

### DIFF
--- a/packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
+++ b/packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState } from 'react';
 import { IcebergFetchParamsV6, fetchIcebergV6 } from '@ovh-ux/manager-core-api';
 import { useInfiniteQuery } from '@tanstack/react-query';
 import { ColumnSort } from '../../components';
@@ -6,11 +6,6 @@ import { ColumnSort } from '../../components';
 interface IcebergV6Hook {
   queryKey: string[];
   defaultSorting?: ColumnSort;
-}
-
-interface QueryParams {
-  pageIndex: number;
-  sorting?: ColumnSort;
 }
 
 /**
@@ -24,77 +19,41 @@ export function useResourcesIcebergV6<T = unknown>({
   queryKey,
   defaultSorting = undefined,
 }: IcebergFetchParamsV6 & IcebergV6Hook) {
-  const [totalCount, setTotalCount] = useState(0);
-  const [flattenData, setFlattenData] = useState<T[]>([]);
-  const [queryParams, setQueryParams] = useState<QueryParams>({
-    pageIndex: 1,
-    sorting: { desc: false, id: '' },
-  });
   const [sorting, setSorting] = useState<ColumnSort>(defaultSorting);
 
-  const {
-    data,
-    fetchNextPage,
-    isError,
-    isLoading,
-    error,
-    status,
-  } = useInfiniteQuery({
-    initialPageParam: null,
-    queryKey: [...queryKey, sorting],
-    queryFn: ({ pageParam }) =>
+  const { data: dataSelected, ...rest } = useInfiniteQuery({
+    initialPageParam: 1,
+    queryKey: [...queryKey, pageSize, sorting],
+    staleTime: Infinity,
+    retry: false,
+    queryFn: ({ pageParam: pageIndex }) =>
       fetchIcebergV6<T>({
         route,
         pageSize,
-        page: pageParam,
+        page: pageIndex,
         sortBy: sorting?.id || null,
         sortReverse: sorting?.desc,
       }),
-    staleTime: Infinity,
-    retry: false,
-    getNextPageParam: () => queryParams.pageIndex,
+    getNextPageParam: (lastPage, _allPages, lastPageIndex) => {
+      if (lastPage.totalCount / pageSize > lastPageIndex) return null;
+      return lastPageIndex + 1;
+    },
+    select: (data) => {
+      const pageIndex = data.pageParams[data.pageParams.length - 1];
+      const { totalCount } = data.pages[0];
+      return {
+        data,
+        pageIndex,
+        totalCount,
+        flattenData: data.pages.flatMap((page) => page.data),
+        hasNextPage: totalCount / pageSize > pageIndex,
+      };
+    },
   });
 
-  useEffect(() => {
-    const flatten = data?.pages.map((page) => page.data).flat();
-    setFlattenData(flatten);
-  }, [data]);
-
-  useEffect(() => {
-    if (totalCount === 0 && data?.pages) {
-      setTotalCount(data.pages[0].totalCount);
-    }
-  }, [data]);
-
-  useEffect(() => {
-    if (sorting) {
-      setQueryParams(() => ({ pageIndex: 1, sorting }));
-    }
-  }, [sorting]);
-
-  useEffect(() => {
-    fetchNextPage();
-  }, [queryParams]);
-
-  const onFetchNextPage = () => {
-    setQueryParams((previousParams: QueryParams) => ({
-      ...previousParams,
-      pageIndex: previousParams.pageIndex + 1,
-    }));
-  };
-
   return {
-    data,
-    fetchNextPage: onFetchNextPage,
-    setFlattenData,
-    flattenData,
-    totalCount,
-    pageIndex: queryParams.pageIndex,
-    hasNextPage: totalCount / pageSize > queryParams.pageIndex,
-    isError,
-    error,
-    isLoading,
-    status,
+    ...(dataSelected ?? { totalCount: 0, hasNextPage: false }),
+    ...rest,
     sorting,
     setSorting,
   };


chipp972 (approved):

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
```diff
@@ -24,46 +24,35 @@ export function useResourcesIcebergV6<T = unknown>({
...
+  const flattenData = useMemo(() => {
+    return data?.pages.map((page) => page.data).flat() ?? [];
   }, [data]);
 
-  useEffect(() => {
-    if (totalCount === 0 && data?.pages) {
-      setTotalCount(data.pages[0].totalCount);
-    }
+  const totalCount = useMemo(() => {
+    return data?.pages[0].totalCount ?? 0;
```
chipp972, line 55: Nice optimization 👍🏾

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
```diff
@@ -24,46 +24,35 @@ export function useResourcesIcebergV6<T = unknown>({
...
+      staleTime: Infinity,
+      retry: false,
+      getNextPageParam: () => queryParams.pageIndex,
+    });
 
-  useEffect(() => {
-    const flatten = data?.pages.map((page) => page.data).flat();
-    setFlattenData(flatten);
+  const flattenData = useMemo(() => {
+    return data?.pages.map((page) => page.data).flat() ?? [];
```
fredericvilcot, line 51: Why not using the `select` built-in function to compute data?

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
fredericvilcot: Maybe, as a suggestion, we can consider returning a new `data` object that contains both computed data and page count through a native memoized selector.

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
tristanwagner: indeed it will avoid rerenders if for example the cursor changes but the data remains the same

tristanwagner: (while using a refetchInterval)

tristanwagner: and btw it would be nice to be able to pass some options to the useInfiniteQuery

tibs245: I understand for the select. Is a good idea. 

But for the possibility to pass options in `useInfiniteQuery` :  I think it's another subject and this implicate breaking change. And I don't think it's good to include this debate on a bug fix. 

If I can include fast optimization I do it with pleasure. But I scare this change block the validation.

tibs245: But for information @tristanwagner If I'm not mistaken : The rerender is not avoid because the object is regenerated with another reference. 

So the object content is not compared only the reference. And React see a change same it's exactly same data.

tibs245: Oh good to known ! Thank you for this information 😁 I will do it !

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
```diff
@@ -24,46 +24,35 @@ export function useResourcesIcebergV6<T = unknown>({
...
+          route,
+          pageSize,
+          page: pageParam,
+          sortBy: sorting?.id || null,
+          sortReverse: sorting?.desc,
+        }),
+      staleTime: Infinity,
+      retry: false,
+      getNextPageParam: () => queryParams.pageIndex,
+    });
```
ghyenne, line 48:
```suggestion
  const { data, fetchNextPage, isError, isLoading, error, status } =
    useInfiniteQuery({
      initialPageParam: null,
      queryKey: [...queryKey, queryParams.pageIndex, sorting],
      queryFn: ({ pageParam }) =>
        fetchIcebergV6<T>({
          route,
          pageSize,
          page: pageParam,
          sortBy: sorting?.id || null,
          sortReverse: sorting?.desc,
        }),
      staleTime: Infinity,
      retry: false,
      getNextPageParam: () => queryParams.pageIndex,
      select: (data) =>
      data?.pages.flatMap(
        (page) => page.data,
      ),
    });
```

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
ghyenne, line 51: I think we can remove the useEffect and use the function "select" from tanstack query for flat the result.

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
```diff
@@ -24,73 +19,48 @@ export function useResourcesIcebergV6<T = unknown>({
...
   return {
-    data,
-    fetchNextPage: onFetchNextPage,
-    setFlattenData,
-    flattenData,
-    totalCount,
-    pageIndex: queryParams.pageIndex,
-    hasNextPage: totalCount / pageSize > queryParams.pageIndex,
+    ...dataSelected,
+    fetchNextPage,
```
fredericvilcot, line 63: IMO, this is useless to return `isError, isLoading, isError ...` unless we want to constraint the properties returned by `useInfiniteQuery` to the returned object.
If not, I'll suggest to use rest parameter: 

```typescript
const {
    data,
   ...rest
  } = useInfiniteQuery(...)

return {
 pageIndex,
...,
...rest
}

```

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
```diff
@@ -24,73 +19,48 @@ export function useResourcesIcebergV6<T = unknown>({
...
-    getNextPageParam: () => queryParams.pageIndex,
+    getNextPageParam: (lastPage, _allPages, lastPageIndex) => {
+      if (lastPage.totalCount / pageSize > lastPageIndex) return null;
+      return lastPageIndex + 1;
+    },
+    select: (data) => {
+      const pageIndex = data.pageParams?.[data.pageParams.length - 1] ?? 0;
+      const totalCount = data?.pages[0].totalCount ?? 0;
+      return {
+        data,
```
fredericvilcot, line 52: **Question**: what's the point to return raw `data` as the aim of a selector is to compute `data` for the view?

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
```diff
@@ -24,73 +19,48 @@ export function useResourcesIcebergV6<T = unknown>({
...
     isLoading,
     error,
     status,
   } = useInfiniteQuery({
-    initialPageParam: null,
+    initialPageParam: 1,
     queryKey: [...queryKey, sorting],
-    queryFn: ({ pageParam }) =>
+    staleTime: Infinity,
+    retry: false,
```
fredericvilcot, line 35: These properties should be given through parameters of the hook, to let the caller use it at its convenience and to prevent from being too specific for a general purpose.

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
tibs245: It's to avoid break changes. I don't want create a bugs on a previous or actual developpement. 

If we can be sure at 100% it's ok. We can remove it. But I am not.

Maybe create a type for deprecated data on a first time and remove later ?

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
tibs245: I approve. 

But It's a Breaking change. 

And I don't want to do it on this PR. The first priority is fix bug. And I take this opportunity for do a refacto.

But I avoid create bug or blocked this PR for that.

May be another ticket ?

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
fredericvilcot: I see the point 👍🏻
But I think it's kind of anti-pattern with the selector's definition so maybe we can consider to remove it in a next feature, after being sure that there is no breaking change :)

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
fredericvilcot: Yep, agree 👍🏻

fredericvilcot: I will add that we need a deep refactor of this hook, because, as it is, it's not generic at all, I mean a lot of parameters are hard coded and the `select` function itself should be a hook parameter so the caller could do whatever he wants with the raw data.

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
tibs245: Yes you're right ! 👌

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
```diff
@@ -24,77 +19,41 @@ export function useResourcesIcebergV6<T = unknown>({
...
+      return lastPageIndex + 1;
+    },
+    select: (data) => {
+      const pageIndex = data.pageParams?.[data.pageParams.length - 1] ?? 0;
+      const totalCount = data?.pages[0].totalCount ?? 0;
+      return {
+        data,
+        pageIndex,
+        totalCount,
+        flattenData: data?.pages.flatMap((page) => page.data) ?? [],
```
fredericvilcot, line 48: `data` select parameter is never `undefined` according to its type and purpose so you can safely remove optional chaining and nullish coalescing operator

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
```diff
@@ -1,4 +1,4 @@
-import { useState, useEffect } from 'react';
+import { useState } from 'react';
```
fredericvilcot, line 1: looks like it's not used anymore?

fredericvilcot (changes requested):

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
fredericvilcot: My bad !

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
fredericvilcot: @tibs245 
You also can remove `?? []` because data is never `undefined` 😉

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
```diff
@@ -24,77 +19,41 @@ export function useResourcesIcebergV6<T = unknown>({
...
-    flattenData,
-    totalCount,
-    pageIndex: queryParams.pageIndex,
-    hasNextPage: totalCount / pageSize > queryParams.pageIndex,
-    isError,
-    error,
-    isLoading,
-    status,
+    ...dataSelected,
+    ...rest,
```
fredericvilcot, line 56: IMO, spred operator on `dataSelected` has no effect.
```suggestion
    dataSelected,
    ...rest,
```

tibs245: The problem is with second implementation I change output signature. And I risk cause bug where pages already use this hook.

Bug yes I can do the next implementation : 

```suggestion
    ...(dataSelected ?? {}),
    ...rest,
```

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
```diff
@@ -24,77 +19,41 @@ export function useResourcesIcebergV6<T = unknown>({
...
+      return lastPageIndex + 1;
+    },
+    select: (data) => {
+      const pageIndex = data.pageParams[data.pageParams.length - 1];
+      const { totalCount } = data.pages[0];
+      return {
+        data,
+        pageIndex,
+        totalCount,
+        flattenData: data.pages.flatMap((page) => page.data) ?? [],
```
fredericvilcot, line 48: I'm not sure to understand the purpose of the nullish operator

tibs245: Sorry it's the problem when on quick fix 🙃

fredericvilcot (approved):
Looks good, thanks for your work on it :)

ghyenne (approved):

anooparveti (approved):

PR #13383, opened 2024-10-01
feat(account): add kyc digital process tracking

MaximeBajeux:
ref: MANAGER-15213

| Question         | Answer
| ---------------- | ---
| Branch?          | `MANAGER-15213`
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-15213

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager/apps/dedicated/client/app/account/identity-documents/user-identity-documents.constant.js
+++ b/packages/manager/apps/dedicated/client/app/account/identity-documents/user-identity-documents.constant.js
@@ -275,7 +275,7 @@ export const TRACKING_PREFIX = 'Hub::account::identity-files';
 
 export const TRACKING_TASK_TAG = {
   dashboard: `${TRACKING_PREFIX}::identity-files::dashboard`,
-  openDetailView: `${TRACKING_PREFIX}::tile::button::{{name_click}}_identity-files::{{identity-file}}`,
+  openDetailView: `${TRACKING_PREFIX}::tile::button::{{name_click}}_identity-files::{{identity-files}}`,
   clickSendMyDocuments: `${TRACKING_PREFIX}::page::button::submit_identity-files`,
   displayPopUpSendMyDocuments: `${TRACKING_PREFIX}::identity-files::pop-up::identity-files-sent`,
   confirmSendMyDocuments: `${TRACKING_PREFIX}::identity-files::pop-up::button::identity-files-sent::confirm`,
--- a/packages/manager/apps/dedicated/client/app/account/identity-documents/user-identity-documents.constant.js
+++ b/packages/manager/apps/dedicated/client/app/account/identity-documents/user-identity-documents.constant.js
@@ -271,13 +271,19 @@ export const LEGAL_LINK3 = {
 
 export const MAX_SIZE = 10000000;
 
-export const TRACKING_PREFIX = 'dedicated::account::identity-files';
+export const TRACKING_PREFIX = 'Hub::account::identity-files';
 
 export const TRACKING_TASK_TAG = {
-  upload: `${TRACKING_PREFIX}::upload`,
-  uploadSuccess: `${TRACKING_PREFIX}::upload-success`,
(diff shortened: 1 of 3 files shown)


packages/manager/apps/dedicated/client/app/account/identity-documents/user-identity-documents.controller.js
```diff
@@ -2,13 +2,25 @@ import {
...
+  if (!params) return hit;
+  let formatted = hit;
+  params.forEach((param) => {
+    const { key, value } = Object.entries(param);
+    const replaced = hit.replace(`{${key}}`, value);
+    if (replaced) formatted = replaced;
+  });
+  return formatted;
+};
+
```
JacquesLarique, line 23: Do we still need this ?
It seems it's used in the two tracking methods (`trackClick` and `trackPage`) which are never called with the `params` parameter meaning this method always exit on line 14 (`if (!params) return hit;`)

packages/manager/apps/dedicated/client/app/account/identity-documents/user-identity-documents.constant.js
```diff
@@ -271,13 +271,19 @@ export const LEGAL_LINK3 = {
 
 export const MAX_SIZE = 10000000;
 
-export const TRACKING_PREFIX = 'dedicated::account::identity-files';
+export const TRACKING_PREFIX = 'Hub::account::identity-files';
```
JacquesLarique, line 274: I guess it was asked this way (so this comment is not blocking), but I think this prefix should be removed entirely, as 'Account' module will be extracted from dedicated (and is in no way part of the hub module as suggested by this prefix)

JacquesLarique (approved):

tristanwagner (dismissed):

packages/manager/apps/dedicated/client/app/account/identity-documents/user-identity-documents.controller.js
```diff
@@ -2,13 +2,25 @@ import {
...
   KYC_STATUS,
   DOCUMENTS_MATRIX,
 } from './user-identity-documents.constant';
 
+const replaceTrackingParams = (hit, params) => {
+  if (!params) return hit;
+  let formatted = hit;
+  params.forEach((param) => {
+    const { key, value } = Object.entries(param);
+    const replaced = hit.replace(`{${key}}`, value);
```
tristanwagner, line 18: is the final output of the replaced part supposed to be like `{to_add}_identity-files::{something}` or `{{to_add}}_identity-files::{{something}}` or `to_add_identity-files::something` because now I think it will be this case : `{to_add}_identity-files::{something}`

packages/manager/apps/dedicated/client/app/account/identity-documents/user-identity-documents.controller.js
```diff
@@ -2,13 +2,25 @@ import {
...
   LEGAL_LINK3,
   KYC_STATUS,
   DOCUMENTS_MATRIX,
 } from './user-identity-documents.constant';
 
+const replaceTrackingParams = (hit, params) => {
+  if (!params) return hit;
+  let formatted = hit;
+  params.forEach((param) => {
+    const { key, value } = Object.entries(param);
```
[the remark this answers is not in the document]
tristanwagner: also what is the type of params supposed to be ? an array ? if it's an object I think it would be more like

`
Object.entries(params).forEach(([key, value]) => {
// do your stuff here
})
`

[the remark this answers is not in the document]
tristanwagner: see example here https://developer.mozilla.org/fr/docs/Web/JavaScript/Reference/Global_Objects/Object/entries#parcourir_un_objet

tristanwagner (approved):

oalkabouss (approved):

PR #13401, opened 2024-10-02
feat(pci-load-balancer): create a pool

sidlynx:
ref: DTCORE-2648

| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/pci-load-balancer`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | DTCORE-2648

- [x] Commits are signed-off
- [x] Lint has passed locally


packages/manager/apps/pci-load-balancer/package.json
```diff
@@ -31,9 +31,11 @@
     "@tanstack/react-query": "^5.50.1",
     "@tanstack/react-query-devtools": "^5.50.1",
     "@tanstack/react-table": "^8.17.3",
+    "axios": "^1.7.7",
     "date-fns": "^4.1.0",
     "i18next": "^23.11.5",
     "i18next-http-backend": "^2.5.2",
+    "nanoid": "^5.0.7",
```
frenautvh, line 38: is it used ?

packages/manager/apps/pci-load-balancer/src/api/data/api-specs.ts
```diff
@@ -0,0 +1,20 @@
...
+    'cloud.loadbalancing.PoolSessionPersistenceTypeEnum': {
+      enum: string[];
+    };
+  };
+};
+
+export const getApiSpecs = async (): Promise<TApiSpecsResponse> => {
+  const { data } = await v6.get<TApiSpecsResponse>(`/cloud.json`);
+  return data;
+};
```
frenautvh, line 20: this could be put in pci-common as it is global to pci

also i would suggest naming it api schema instead of api specs

packages/manager/apps/pci-load-balancer/package.json
```diff
@@ -31,9 +31,11 @@
     "@tanstack/react-query": "^5.50.1",
     "@tanstack/react-query-devtools": "^5.50.1",
     "@tanstack/react-table": "^8.17.3",
+    "axios": "^1.7.7",
```
seven-amid, line 34: The only use of Axios exists on the PR is for `AxiosError` (correct me if am wrong) , and there are already objects available from the common internal module that you can use:

`import { ApiError } from '@ovh-ux/manager-core-api';`
Or ( depend to u're use case )
`import { ResponseAPIError } from '@ovh-ux/manager-pci-common';`

packages/manager/apps/pci-load-balancer/src/api/hook/pools.hook.ts
```diff
@@ -0,0 +1,54 @@
+import { useMutation } from '@tanstack/react-query';
```
frenautvh, line 1: i see the team already named hook files such as "usePools" etc so i think it should be better to stick with just one naming convention

packages/manager/apps/pci-load-balancer/package.json
sidlynx: yes, to generate a key for a select that doesn't take into account the initial value(124 bytes), but may replaced by Date.now or equivalent

packages/manager/apps/pci-load-balancer/src/api/data/api-specs.ts
sidlynx: was the actual name, but okey for the move and the rename

frenautvh (approved):

seven-amid (approved):

kqesar (approved):

PR #13620, opened 2024-10-16
Fix/manager-15697 handle multiple can correct

jvaumoroovh:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     |no
| Breaking change? no
| Tickets          | Fix # MANAGER-15697

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
Typically on .es domains, multiple canCorrect generate multiple simultaneous API calls which overwrite one another.

In the end, we lose data in the arguments due to this race condition.

Way to fix the bug: make all front-end calls sequential (wait for one to finish to call the next one)


packages/manager/apps/web/client/app/domain-operation/update/domain-operation-update.controller.js
```diff
@@ -192,115 +192,128 @@ angular.module('App').controller(
...
-              this.Alerter.success(
-                this.$translate.instant('domain_tab_OPERATION_update_success'),
-                ALERTER_ID,
-              );
-              return null;
-          }
-        })
-        .catch((err) =>
+      // eslint-disable-next-line no-restricted-syntax
+      for (const promise of promises) {
```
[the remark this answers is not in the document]
tristanwagner: if not you could also use maybe use an like this helper function to run your promises sequentially

```
/*
 * serial executes Promises sequentially.
 * @param {funcs} An array of funcs that return promises.
 * @example
 * const urls = ['/url1', '/url2', '/url3']
 * serial(urls.map(url => () => $.ajax(url)))
 *     .then(console.log.bind(console))
 */
const serial = funcs =>
    funcs.reduce((promise, func) =>
        promise.then(result => func().then(Array.prototype.concat.bind(result))), Promise.resolve([]))
```

packages/manager/apps/web/client/app/domain-operation/update/domain-operation-update.controller.js
```diff
@@ -192,115 +192,128 @@ angular.module('App').controller(
...
+                ALERTER_ID,
+              ),
+            )
+            .catch((err) =>
+              this.Alerter.alertFromSWS(
+                this.$translate.instant('domain_operations_cancel_error'),
+                err,
+                ALERTER_ID,
+              ),
+            );
```
ghyenne, line 285: There is some time Alerter.success and Alerter.alert
Maybe it's possible to avoid the duplication by creating dedicated function.

tristanwagner:
commit description is missing jira ticket ref

tristanwagner (approved):

ghyenne (approved):

packages/manager/apps/web/client/app/domain-operation/update/domain-operation-update.controller.js
```diff
@@ -192,115 +196,97 @@ angular.module('App').controller(
...
+          this.displayAlert(
             this.$translate.instant('domain_tab_OPERATION_update_error'),
-            err,
+            ALERTER_ID,
+            error,
+          );
+        }
+      }
+
+      switch (this.todoOperation) {
```
antonymarion, line 232: nitpick: switch case could be replaced here by dictionnary and compact the code with one call to this.processOperation instead

ghyenne (approved):

Issue #13649, opened 2024-10-17
[manager-components]: Add hook for navigation getUrl


tibs245:
### Have you already checked if a similar item is present on manager-components?

- [X] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

Today when we need get Url from navigation it's not easy :

Here a extract code of KMS:

`packages/manager/apps/key-management-service/src/components/layout-helpers/Dashboard/GeneralInformationsTiles/BillingInformationsTile.tsx`

```ts
const [contactUrl, setContactUrl] = useState('');
  const {
    shell: { navigation },
  } = useContext(ShellContext);
  
  
    useEffect(() => {
    const fetchUrl = async () => {
      try {
        const response = await navigation.getURL(
          'dedicated',
          '#/contacts/services',
          {},
        );
        setContactUrl(response as string);
      } catch (error) {
        setContactUrl('#');
      }
    };
    fetchUrl();
  }, [navigation]);
```

And after we can use contactUrl without informations on the loading and error state.

So we have TanStack Query to facilitate promise management. And in finallity possibility just use : 

```ts
 const { data: contactUrl } = useNavigationGetUrl(['dedicated', '#/contacts/services', {}];
```

### Where do you expect to use this?

hycu

### Do you have mock-up?

No

### When do you expect this to be delivered?

31-10-2024

anooparveti:
The request will be tackled with MANAGER-15760

PR #13656, opened 2024-10-17
Feat/navigation get url

tibs245:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Feat #13649 

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

See issue to understand the needs


packages/manager-react-components/src/hooks/navigation/useNavigationGetUrl.spec.tsx
```diff
@@ -0,0 +1,45 @@
+// eslint-disable-next-line prettier/prettier
+import type { ParamValueType } from '@ovh-ux/url-builder';
```
chipp972, line 2: maybe remove this line since the package is not in the right format

anooparveti (changes requested):
I think `react-shell-client` will be the right place to add this hook.

PR #13733, opened 2024-10-22
Feat/manager 15766 improuve can correct modal

jvaumoroovh:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? |no
| Tickets          | Fix # MANAFER-15766

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
Now the `accelerate` operation is selected first. then the  `relaunch` operation and finally the `cancel`
Also move the `loading` in the .finally to be sure is called a the real end of the promise

### DIFF
--- a/packages/manager/apps/web/client/app/domain-operation/operation-table/operation-table.html
+++ b/packages/manager/apps/web/client/app/domain-operation/operation-table/operation-table.html
@@ -102,8 +102,7 @@
         data-compact
         data-disabled="!$row.canCancel && !$row.canRelaunch && !$row.canAccelerate && $row.function !== 'DomainIncomingTransfer'"
         data-placement="end"
-        data-oui-tooltip="{{:: 'domain_operations_no_action' | translate }}"
-        data-oui-tooltip-enable="{{!$row.canCancel && !$row.canRelaunch && !$row.canAccelerate}}"
+        data-oui-tooltip="{{ :: !$row.canCancel && !$row.canRelaunch && !$row.canAccelerate ? 'domain_operations_no_action' : '' | translate }}"
         data-oui-tooltip-placement="left"
     >
         <oui-action-menu-item
--- a/packages/manager/apps/web/client/app/domain-operation/update/domain-operation-update.controller.js
+++ b/packages/manager/apps/web/client/app/domain-operation/update/domain-operation-update.controller.js
@@ -47,12 +47,12 @@ angular.module('App').controller(
       this.uploadMode = {};
       this.viewMode = {};
 
-      if (this.operation.canCancel) {
-        this.todoOperation = 'cancel';
+      if (this.operation.canAccelerate) {
+        this.todoOperation = 'accelerate';
       } else if (this.operation.canRelaunch) {
         this.todoOperation = 'relaunch';
-      } else if (this.operation.canAccelerate) {
-        this.todoOperation = 'accelerate';
+      } else if (this.operation.canCancel) {
+        this.todoOperation = 'cancel';
       }
 
       this.contactUrl = this.coreURLBuilder.buildURL('dedicated', '#/contact');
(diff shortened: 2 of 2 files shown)


packages/manager/apps/web/client/app/domain-operation/update/domain-operation-update.controller.js
```diff
@@ -192,115 +196,97 @@ angular.module('App').controller(
...
-                    ),
-                    err,
-                    ALERTER_ID,
-                  ),
-                );
-            default:
-              this.Alerter.success(
-                this.$translate.instant('domain_tab_OPERATION_update_success'),
-                ALERTER_ID,
+      for (const arg of this.args) {
```
ghyenne, line 202:
```suggestion
      this.args.forEach ({key, value, type, readOnly}) {
```
I think it's better to use foreach and destructured object

jvaumoroovh: with for...of, we can still use destructured objects.

In my opinion, for...of is more readable, and from a performance point of view, it's also better (not a big performance improvement, but an improvement nonetheless).

packages/manager/apps/web/client/app/domain-operation/update/domain-operation-update.controller.js
```diff
@@ -192,115 +196,97 @@ angular.module('App').controller(
...
+          this.displayAlert(
             this.$translate.instant('domain_tab_OPERATION_update_error'),
-            err,
+            ALERTER_ID,
+            error,
+          );
+        }
+      }
+
+      switch (this.todoOperation) {
```
antonymarion, line 232: Maybe replace switch case by dictionnary here

packages/manager/apps/web/client/app/domain-operation/update/domain-operation-update.controller.js
```diff
@@ -1,3 +1,5 @@
+/* eslint-disable no-await-in-loop */
```
antonymarion, line 1: Maybe we can avoir disabling eslint here ?

packages/manager/apps/web/client/app/domain-operation/operation-table/operation-table.html
```diff
@@ -102,8 +102,7 @@
         data-compact
         data-disabled="!$row.canCancel && !$row.canRelaunch && !$row.canAccelerate && $row.function !== 'DomainIncomingTransfer'"
         data-placement="end"
-        data-oui-tooltip="{{:: 'domain_operations_no_action' | translate }}"
-        data-oui-tooltip-enable="{{!$row.canCancel && !$row.canRelaunch && !$row.canAccelerate}}"
+        data-oui-tooltip="{{ :: !$row.canCancel && !$row.canRelaunch && !$row.canAccelerate ? 'domain_operations_no_action' : '' | translate }}"
```
sachinrameshn, line 105: Is this working without any common bracket ?

sachinrameshn: data-oui-tooltip="{{ :: (!$row.canCancel && !$row.canRelaunch && !$row.canAccelerate ? 'domain_operations_no_action' : '') | translate }}"

jvaumoroovh: It's working

image

packages/manager/apps/web/client/app/domain-operation/update/domain-operation-update.controller.js
jvaumoroovh: no, we can't avoid disabling eslint here, I need to have await in a for loop to make my http calls sequential (or i have to add an eslint disable for each ligne that have await in the for loop

PR #13808, opened 2024-10-28
feat(manager-components): disable update name button if no change

chipp972:
ref: ISSUE-13807

| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | ISSUE-13807 

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager-react-components/src/components/templates/update-name-modal/update-name-modal.component.tsx
+++ b/packages/manager-react-components/src/components/templates/update-name-modal/update-name-modal.component.tsx
@@ -108,14 +108,13 @@ export const UpdateNameModal: React.FC<UpdateNameModalProps> = ({
         )}
       </OdsFormField>
       <OdsButton
-        isDisabled={isLoading}
         slot="actions"
         variant={ODS_BUTTON_VARIANT.ghost}
         {...handleClick(closeModal)}
         label={cancelButtonLabel || t('updateModalCancelButton')}
       />
       <OdsButton
-        isDisabled={isPatternError}
+        isDisabled={isPatternError || defaultValue === displayName || undefined}
         slot="actions"
         isLoading={isLoading}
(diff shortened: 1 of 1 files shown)


pauldkn (approved):

tristanwagner (approved):

packages/manager-react-components/src/components/templates/update-name-modal/update-name-modal.component.tsx
```diff
@@ -139,7 +139,12 @@ export const UpdateNameModal: React.FC<UpdateNameModalProps> = ({
         {cancelButtonLabel || t('updateModalCancelButton')}
       </OsdsButton>
       <OsdsButton
-        disabled={isLoading || isPatternError || undefined}
+        disabled={
+          isLoading ||
+          isPatternError ||
+          defaultValue === displayName ||
```
anooparveti, line 145: **question:** What if user wants to proceed with the displayName itself? i would prefer to accept a function or expression as prop and based on that disable the button.

anooparveti (dismissed):

packages/manager-react-components/src/components/templates/update-name-modal/update-name-modal.component.tsx
chipp972: If there is no change, we should not let the user trigger the API
He should cancel the action instead of triggering the API since he doesn't want to change the displayName

tibs245 (approved):

tristanwagner (approved):

anooparveti (approved):

SimonChaumet (commented):
Maybe the PR name should be changed, it is weird to have 3 PR with the same name

anooparveti (approved):

PR #13861, opened 2024-10-30
build(manager-react-components): exclude ODS from bundle

aboungnaseng-ovhcloud:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix 14822

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Fix bundle size external ods and dedupe mrc lib


tristanwagner:
@aboungnaseng-ovhcloud I think we should specify the required comptatible dependencies as peerDependencies if we don't include them in the bundle to avoid version mismatching

PR #14081, opened 2024-11-13
test(kms): add create kms integration tests

rjamet-ovh:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-15111

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager/apps/key-management-service/src/components/layout-helpers/Create/RegionSelector.tsx
+++ b/packages/manager/apps/key-management-service/src/components/layout-helpers/Create/RegionSelector.tsx
@@ -90,6 +90,7 @@ const RegionSelector = ({
           {orderCatalogOKMS && !isLoading && (
             <OsdsSelect
               inline
+              data-testid="select-region"
               size={ODS_SELECT_SIZE.md}
               onOdsValueChange={(v) => {
                 const value = v.detail.value.toString();
@@ -110,7 +111,11 @@ const RegionSelector = ({
               {orderCatalogOKMS.plans[0].configurations[0].values.map(
                 (region) => {
                   return (
-                    <OsdsSelectOption value={region} key={region}>
+                    <OsdsSelectOption
+                      value={region}
+                      key={region}
+                      data-testid={`select-region-option-${region}`}
+                    >
                       <Region
                         mode={'region'}
                         name={region.toLowerCase().replaceAll('_', '-')}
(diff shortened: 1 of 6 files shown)


anooparveti (changes requested):
Since it's addition of test-cases, can you change the type of commit to `test` instead?

pauldkn (approved):

packages/manager/apps/key-management-service/src/mocks/catalog/catalog.handler.ts
```diff
@@ -0,0 +1,23 @@
+import { Handler } from '../../../../../../../playwright-helpers';
```
anooparveti, line 1: **question:** We will soon remove playwright from the repo.
Can you eliminate the dependency on it?

rjamet-ovh: This type is cross product so it should be set on MRC or another shared dependency to be available for all apps.

tibs245: For delete this dependencies we need to do have alternative of mock. I think can change the mock with Axios Mock : https://www.npmjs.com/package/axios-mock-adapter

And with this the test can be faster.

But to not create another new implementation. I agree with Romain. All units tests have this dependancies. 

It's possible to planned on control tower a example of implementation with the method of mock recommended ?

Maybe implementation on generator ?

packages/manager/apps/key-management-service/src/components/layout-helpers/Create/RegionSelector.tsx
```diff
@@ -90,6 +90,7 @@ const RegionSelector = ({
           {orderCatalogOKMS && !isLoading && (
             <OsdsSelect
               inline
+              data-testid={'select-region'}
```
tibs245, line 93:
```suggestion
              data-testid='select-region'
```

tibs245 (approved):

packages/manager/apps/key-management-service/src/mocks/catalog/catalog.handler.ts
rjamet-ovh: we will clean our dependencies to the playwright folder on a specific development.

darsene (dismissed):

pauldkn (dismissed):

tibs245 (approved):

anooparveti (approved):

pauldkn (approved):

chipp972 (approved):

tibs245 (approved):

PR #14164, opened 2024-11-19
fix(dedicated-server): change route name of installation-ovh

tibs245:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-16095

- [x] Commits are signed-off
- [x] Lint has passed locally


anooparveti (changes requested):
**question:** Have you verified the fix in LABEU? Because i dont think changing the state name will resolve the issue rather you should change the url of state if your analysis is correct.

tibs245:
> **question:** Have you verified the fix in LABEU? Because i dont think changing the state name will resolve the issue rather you should change the url of state if your analysis is correct.

Yes I tested and It's work. I don't known why this work well with this configuration but for that I need known what the CI/CD do really and how.

darsene (approved):

JayBeeDe (approved):

darsene (approved):

PR #14176, opened 2024-11-19
fix(mrc): set icon actionmenu compact ellipsisvertical

### LINKED ISSUE #14172: [manager-components]: ActionMenu design

ActionMenu in Datagrid (compact)

### Description

On the last version of the MRC with ODS v18, I've been told that the design of the ActionMenu button in a Datagrid row should be with variant `ghost` and `ellipsis-vertical` 

Actual design:

image

Should be (no border)

image

### Steps to reproduce

just browse [component in storybook](https://ovh.github.io/manager/storybook-static/index.html?path=/story/components-table--default-props)

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

tristanwagner:
ref: ISSUE-14172

| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | fix #14172  

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager-react-components/src/components/navigation/menus/action/action.component.tsx
+++ b/packages/manager-react-components/src/components/navigation/menus/action/action.component.tsx
@@ -98,9 +98,10 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
           onClick={() => setIsTrigger(true)}
           {...(!isCompact && { label: t('common_actions') })}
           icon={
-            icon ?? isCompact
-              ? ODS_ICON_NAME.ellipsisHorizontal
-              : ODS_ICON_NAME.chevronDown
+            icon ||
+            (isCompact
+              ? ODS_ICON_NAME.ellipsisVertical
+              : ODS_ICON_NAME.chevronDown)
           }
         />
       </div>
--- a/packages/manager-react-components/src/components/navigation/menus/action/action.spec.tsx
+++ b/packages/manager-react-components/src/components/navigation/menus/action/action.spec.tsx
@@ -64,12 +64,12 @@ describe('ActionMenu', () => {
     const actionMenuIcon = screen.getByTestId(
       'navigation-action-trigger-action',
     );
-    expect(actionMenuIcon.getAttribute('icon')).toBe('ellipsis-horizontal');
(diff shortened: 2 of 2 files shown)


anooparveti (changes requested):
Since, action menu is different when it's displayed on dashboard and listing page. This component should accept a new prop called 'mode' which can be with button or ghost and decide what to display.

packages/manager-react-components/src/components/navigation/menus/action/action.component.tsx
```diff
@@ -23,11 +23,15 @@ export interface ActionMenuItem {
   className?: string;
 }
 
+export enum ActionMenuMode {
+  STANDARD = 'standard',
+  DATAGRID = 'datagrid',
+  CARD = 'card',
```
anooparveti, line 29: This component need not know the parent components which are using it. So can we make these mode more generic? 
For example, this configuration will not be scalable if we want to use the same display as Datagrid in other component like form or so on.

anooparveti (changes requested):

packages/manager-react-components/src/components/navigation/menus/action/action.component.tsx
aboungnaseng-ovhcloud: Using the property 'variant' is not enough ? Or we want to be more restrictive ?

tristanwagner: I think this component has 3 use cases and the current modes make it obvious what mode you need to choose for which use case, but standard, which I didn't find a better name

please provide me mode names if you have an idea

tristanwagner: would "STANDARD", "COMPACT_OUTLINE" and "COMPACT_GHOST" be better ?

aboungnaseng-ovhcloud (approved):

anooparveti (approved):

PR #14188, opened 2024-11-20
feat(manager-pci-common): add get instances by region

SimonChaumet:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #TAPC-2218 

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Add get instances by region with regionalize version for performance and resilience issue

## Related

For #14150

### DIFF
--- a/packages/manager/modules/manager-pci-common/src/api/data/instance.ts
+++ b/packages/manager/modules/manager-pci-common/src/api/data/instance.ts
@@ -35,6 +35,9 @@ export const getInstance = async (
   return data;
 };
 
+/**
+ * If possible use getInstancesByRegion instead
+ */
 export const getInstances = async (
   projectId: string,
   region?: string,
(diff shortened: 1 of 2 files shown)


Tsiorifamonjena (approved):

packages/manager/modules/manager-pci-common/src/api/data/instance.ts
```diff
@@ -35,6 +35,9 @@ export const getInstance = async (
   return data;
 };
 
+/**
+ * If possible use getInstancesByRegion instead
+ */
```
fredericvilcot, line 40: Why not deprecating this call and replacing the clients impacted?

SimonChaumet: I am not sure if all non-regionalized routes should be deprecated, if so yes I should mark it.

Replacing all the clients impacted would take a lot of time since in the usages they do not all uses the filter on region.

SimonChaumet: I can also remove the region parameter if that makes sense but would lead to more development if region is not required

fredericvilcot (approved):

anooparveti (approved):

PR #14234, opened 2024-11-22
fix(manager-react-components): useNotifications prevent early clears

tristanwagner:
this is a cherry pick of [this commit](https://github.com/ovh/manager/commit/881645356019834987bdd43af05cb25b0b5a638a) on mrc v1 maintenance branch

| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #... 

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager-react-components/src/components/notifications/notifications.spec.tsx
+++ b/packages/manager-react-components/src/components/notifications/notifications.spec.tsx
@@ -1,4 +1,4 @@
-import { vitest } from 'vitest';
+import { vi, vitest } from 'vitest';
 import React, { useEffect } from 'react';
 import { render } from '@testing-library/react';
 import { useNotifications, NotificationType } from './useNotifications';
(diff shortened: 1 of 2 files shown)


packages/manager-react-components/src/components/notifications/notifications.spec.tsx
```diff
@@ -35,11 +35,27 @@ describe('notifications component', () => {
...
+    expect(container.children.length).toBe(2);
+
+    await new Promise((resolve) => setTimeout(resolve, 2000));
+
+    render(<ClearNotifications />);
+    container = render(<Notifications />).container;
+
+    await waitFor(() => {
+      expect(container.children.length).toBe(0);
+    });
```
tibs245, line 59: I think it's better mock Date. 

- It's a unit test we don't test Data.now() method
- This allow to do the test faster

packages/manager-react-components/src/components/notifications/useNotifications.ts
```diff
@@ -61,7 +68,12 @@ export const useNotifications = create<NotificationState>((set, get) => ({
         ({ uid }) => uid !== toRemoveUid,
       ),
     })),
-  clearNotifications: () => set(() => ({ notifications: [] })),
+  clearNotifications: () =>
+    set((state) => ({
+      notifications: state.notifications.filter(
+        (notification) => Date.now() - notification.creationTimestamp < 1000,
```
tibs245, line 74:
```suggestion
        (notification) => Date.now() - notification.creationTimestamp < DEFAULT_MINIMUN_LIFETIME,
```

Optionnal suggestion : Possible to add constant for default minimum lifetime and maybe add parameters if we want more or less on the future ?

packages/manager-react-components/src/components/notifications/notifications.spec.tsx
tristanwagner: enhanced the tests according to your suggestions

packages/manager-react-components/src/components/notifications/useNotifications.ts
tristanwagner: To me the whole api should be reworked, for example we should be able to set "clearAfterRead" per notification instead of for all of them, but I think it should be another subject, right now I'm just tried to be up to date with MRC v1 on this issue

frenautvh: FYI we opened this PR to sync the changes we introduced in MRC 1.x : https://github.com/ovh/manager/pull/14170

tristanwagner: Oh ok wasn't aware of that I'll just close this one then

tristanwagner:
should be handled by https://github.com/ovh/manager/pull/14170

PR #14285, opened 2024-11-26
build(pci-common): add vite config to publish library

fredericvilcot:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #TAPC-2321

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

This PR adds configuration to enable the publishing of the library.

## Nota:

- The build phase only targets `esm` file, because of the uselessness of building `cjs` or `umd` files that are not compatible anymore with certain libraries like `@tanstak/react-query`.
Furthermore, all browsers are now compatible with `ES` modules so there is no need to weigh down the produced `dist` folder.
As a consequence, there is no need to add `output.globals` instructions for rollup because this is only needed for `umd` files.

- I added hidden `sourcemap` files to ensure a better debugging with tools like `Sentry` without exposing any reference to the source code in the browser.

## Related

Will be treated with #14344


packages/manager/modules/manager-pci-common/package.json
```diff
@@ -1,70 +1,90 @@
...
+    "prettier": "prettier --write \"src/**/*.{ts,tsx,js,mdx}\"",
     "test": "vitest run",
     "test:coverage": "vitest run --coverage",
-    "test:coverage:watch": "vitest --coverage",
-    "test:watch": "vitest"
+    "test:watch": "vitest --watch",
+    "type:check": "tsc --noEmit"
   },
   "dependencies": {
+    "@ovh-ux/manager-tailwind-config": "^0.2.1",
```
ppprevost, line 37: devDEp ?

ppprevost (changes requested):

packages/manager/modules/manager-pci-common/package.json
fredericvilcot: No. We need to bundle this package as it's a private one, so it has to embedded in the minified code.

ppprevost (dismissed):

packages/manager/modules/manager-pci-common/vite.config.mjs
```diff
@@ -0,0 +1,76 @@
...
+  plugins: [
+    react(),
+    dts({
+      root: __dirname,
+      insertTypesEntry: true,
+      outDir: 'dist/types',
+    }),
+    visualizer({
+      filename: 'stats/bundle-analysis.html',
+    }),
```
frenautvh, line 29: to remove ?

fredericvilcot: IMO, we should keep it to have some stats each time the bundle is built.  
If somebody adds or remove dependencies, we should keep an eye on the produces stats to see if the package has to be external or not.

It only generates a `stats` folder that is not embedded in the `dist`,  so there is no impact on the performance and the bundle size.

packages/manager/modules/manager-pci-common/package.json
```diff
@@ -1,70 +1,90 @@
...
+  "main": "dist/index.js",
+  "types": "dist/types/index.d.ts",
+  "files": [
+    "dist"
+  ],
   "scripts": {
-    "build": "tsc --project tsconfig.build.json",
-    "dev": "tsc",
+    "build": "vite build",
+    "clean": "rimraf node_modules && rimraf dist && yarn cache clean",
```
anooparveti, line 25: **question:** It is anyways removed with the global `yarn clean` command. Is this command to handle the clean only for the lib?

packages/manager/modules/manager-pci-common/package.json
```diff
@@ -1,70 +1,90 @@
...
+    "dev": "vite build",
```
anooparveti, line 26: **question:** `vite ` itself isn't enough? I am not sure if HMR works in case of `vite build`.

packages/manager/modules/manager-pci-common/vite.config.mjs
```diff
@@ -0,0 +1,57 @@
...
+      entry: path.resolve(pathSrc, 'index.ts'),
+      name: 'ManagerPciCommonLib',
+      fileName: 'index',
+      formats: ['es'],
+    },
+    rollupOptions: {
+      external: (id) => {
+        const deps = [...Object.keys(packageJson.peerDependencies || {})];
+        return deps.some((dep) => id === dep || id.startsWith(`${dep}/`));
+      },
```
anooparveti, line 53: Since `rollup` considers exact path of imports, i think you should also addd `@ovhcloud/ods-components/react` to not include ODS components in the bundle.

anooparveti (changes requested):

packages/manager/modules/manager-pci-common/vite.config.mjs
anooparveti: I think we dont use it on a day-to-day basis. So, to make the global node_modules light, i think we can install it in local when it's required and do the analysis.

packages/manager/modules/manager-pci-common/package.json
fredericvilcot: @anooparveti 
Yes, this script only concerns the lib.
Before building, if we want to clean the package's node_modules and the dist folder, it's better to not remove the whole monorepo elements and this ensures that lib has been correctly cleaned.

packages/manager/modules/manager-pci-common/package.json
fredericvilcot: @anooparveti 
IMO, there is no point of having a dev server with HMR as this is a library, not an application.

The good way to develop a library in local is to associate a demo playground like StoryBook, as it has been done for MRC, to be the entry point of new changes.

So, meanwhile, we can keep like this or remove it as it's not manadatory. (StoryBook Will need a specific script)

What do you think?

SimonChaumet (dismissed):

frenautvh (dismissed):

packages/manager/modules/manager-pci-common/vite.config.mjs
fredericvilcot: Ok, I comment the plugin usage and remove the devDependency, so it could be easily used if needed

frenautvh (dismissed):

packages/manager/modules/manager-pci-common/package.json
anooparveti: Are you planning to have a storybook or such documentation for components and other utils?

fredericvilcot: @anooparveti Yes, I think this is a pre-requisite for developing and maintaining a library so we will need to implement it in a next future

packages/manager/modules/manager-pci-common/.gitignore
anooparveti, line 1: Since `visualiser` is not included, this change is obsolete.

anooparveti (changes requested):

anooparveti (approved):

aboungnaseng-ovhcloud (approved):

frenautvh (dismissed):

anooparveti:
Rebasing with release branch.

PR #14322, opened 2024-11-28
test(mrc): ensure useAuthorizationIam query option enabled is a boolean

### LINKED ISSUE #13791: [manager-components]: useAuthorizationIam is not working as expected

src/hooks/iam/useOvhIam.tsx

### Description

When you pass an undefined string to useAuthorizationIam for urn, a request is still sent with the urn as undefined

This is because the following line will be evaluated as undefined (instead of false):
`urn && urn.length > 0 && actions && actions.length > 0 && isTrigger,`
And as a result, tanstack consider that enabled isn't given to the useQuery.

A similar issue happens when you pass an empty string (but it will throw an error on tanstack side)

To fix this issue, we should replace the previously mentioned line by this one:
`Boolean(urn) && urn.length > 0 && Boolean(actions) && actions.length > 0 && isTrigger`
This way it would ensure that if we pass either an empty string or undefined to useAuthorizationIam, the enabled property given to useQuery will always be a boolean

### Steps to reproduce

Call useAuthorizationIam with either an empty string or undefined

### Browser

_No response_

### Operating System

- [X] macOS
- [X] Windows
- [X] Linux

### Additional Information

_No response_

JacquesLarique:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #13791 

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Added tests around sending the request of authorization when passing different values to `useAuthorizationIam`

### DIFF
--- a/packages/manager-react-components/src/hooks/iam/useOvhIam.spec.ts
+++ b/packages/manager-react-components/src/hooks/iam/useOvhIam.spec.ts
@@ -1,9 +1,0 @@
-import { getAuthorizationCheckUrl } from './useOvhIam';
-
-describe('getAuthorizationCheckUrl', () => {
-  it('encodes the urn if it contains /', () => {
-    expect(getAuthorizationCheckUrl('test/urn')).toBe(
-      '/iam/resource/test%2Furn/authorization/check',
-    );
-  });
-});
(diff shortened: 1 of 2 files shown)


anooparveti (changes requested):
Since this is just a test case movement. can you update the type of the commit to `test`

anooparveti (approved):

anooparveti (approved):

PR #14381, opened 2024-12-02
feat(pci-object-storage): create user modal

sidlynx:
| Question         | Answer
| ---------------- | ---
| Branch?          | `feat/pci-object-storage`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | DTCORE-2877

- [x] Commits are signed-off
- [x] Lint has passed locally


packages/manager/apps/pci-object-storage/src/utils.ts
```diff
@@ -1,8 +1,31 @@
...
+  onSuccess?: ({ value, iteration }: { value: T; iteration: number }) => void;
+  onFail?: (iteration: number) => void;
+};
+
+export const poll = async <T>(param: TParam<T>, iteration = 0) => {
+  const [maxTries, interval] = [param.maxTries || 10, param.interval || 5000];
+  if (iteration < maxTries) {
+    const value: T = await param.fn();
+    if (param.ruleFn(value)) {
+      param?.onSuccess({ value, iteration });
```
frenautvh, line 15: shouldn't those be wrapped in a try catch in order to call onFail is the 'fn' or 'ruleFn' function throws an error ?

sidlynx: i thought about it, now i have an extra opinion, yeah it makes sense, and maybe it should be moved to a common package so that it can be reused

kqesar (approved):

PR #14430, opened 2024-12-05
fix(dedicated): fix Problem with timezone on graph displays

sepiropht:
ref: UXCT-668

| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #UXCT-668

- [x] Commits are signed-off
- [ ] ~Only FR translations have been updated~ [n/a]
- [x] Lint has passed locally
- [ ] ~Breaking change is mentioned in relevant commits~ [n/a]

## Description

Fix graph timezone

### DIFF
--- a/packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.controller.js
+++ b/packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.controller.js
@@ -1,8 +1,11 @@
+import { addMilliseconds } from 'date-fns';
 import NetworkSecurityService from '../network-security.service';
 import {
   PAGE_SIZE,
   TRAFFIC_PERIODS,
   TRAFFIC_PERIOD_LIST,
+  MILLISECONDS_PER_MINUTE,
+  TIMEZONE_MULTIPLIER,
   CHART,
 } from './traffic.constant';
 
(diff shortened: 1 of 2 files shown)


packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.controller.js
```diff
@@ -154,7 +154,10 @@ export default class TrafficController {
         after.setDate(after.getDate() - 14);
         break;
       case 'custom':
-        after.setTime(this.dateTime);
+        after = new Date(this.dateTime);
+        after = new Date(
+          after.getTime() + after.getTimezoneOffset() * 4 * 60000,
```
SimonChaumet, line 159: I would recommend using date-fns to do this computation, it would be more understandable

You can also separate variables into constant

Personnaly I don't understand what 4 and 60 000 stand for

SimonChaumet (commented):
name of commit/PR isn't understandable, could you specify what was fixed

packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.controller.js
sepiropht: When accessing the traffic dashboard via URL with a dateTime parameter, 
alerts were often not visible due to timezone misalignment. 
i just added proper timezone offset handling to ensure alerts appear within the 
displayed time window.

qpavy (approved):

packages/manager/apps/dedicated/client/app/network-security/traffic/traffic.controller.js
SimonChaumet: I still don't understand what this `4` stands for

Steffy29: This part of code is outdate, if you look to the new version of code you could see a constant variable which explain what is this `4`.

PR #14462, opened 2024-12-06
feat(dedicated.vrack): restrict single service selection for vs

sachinrameshn:
ref: MANAGER-16093

| Question         | Answer
| ---------------- | ---
| Branch?          | `master` 
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-16093 

- [x] Commits are signed-off
- [x] Lint has passed locally


packages/manager/modules/vrack/src/select-dialog/template.html
```diff
@@ -11,7 +11,7 @@
                     class="ng-isolate-scope oui-modal oui-modal_shadow oui-modal_warning"
                 >
                     <p
-                        data-translate="vrack_select_dialog_modal_description"
+                        data-ng-bind="($ctrl.isIpv6 ? 'vrack_select_dialog_modal_description_for_ipv6': 'vrack_select_dialog_modal_description_for_vs') | translate"
```
qpavy, line 14: I think we should use a variable for vrack services as well (isVrackServices)

sachinrameshn: Yes I taught of same design. In that case I need to add another <p> tag with the content for each case.

qpavy (approved):

mhelhali-soufien (approved):

PR #14518, opened 2024-12-10
feat(pci.workflow): open workflow creation to LZ regions

y4nnL:
| Question         | Answer
| ---------------- | ---
| Branch?          | `TAPC-1429`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #TAPC-1338

- [X] Commits are signed-off
- [X] Lint has passed locally

### DIFF
--- a/packages/manager/apps/pci-workflow/src/api/hooks/order.ts
+++ b/packages/manager/apps/pci-workflow/src/api/hooks/order.ts
@@ -27,11 +27,11 @@ export const useProjectAddons = (projectId: string, ovhSubsidiary: string) => {
   });
 };
 
-export const findInstanceSnapshotPricing = (
+const useRegionInstanceSnapshotPricing = (
   projectId: string,
   instanceRegion: string,
   ovhSubsidiary: string,
-) => {
+) => () => {
   const query = useProjectAddons(projectId, ovhSubsidiary);
   const addons = query.data;
   const snapshotAddon =
@@ -51,11 +51,11 @@ export const findInstanceSnapshotPricing = (
   };
 };
 
-export const findLocalZoneInstanceSnapshotPricing = (
+const useLocalZoneInstanceSnapshotPricing = (
   projectId: string,
   instanceRegion: string,
   ovhSubsidiary: string,
-) => {
+) => () => {
   const queries = useQueries({
     queries: [
       { ...getCatalogQuery(ovhSubsidiary), enabled: Boolean(ovhSubsidiary) },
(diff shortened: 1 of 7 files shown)


SimonChaumet (approved):

packages/manager/apps/pci-workflow/src/api/hooks/order.ts
```diff
@@ -42,3 +50,56 @@ export const useInstanceSnapshotPricing = (
     data: snapshotPricing,
   };
 };
+
+export const findLocalZoneInstanceSnapshotPricing = (
```
fredericvilcot, line 54: Same here, as you are calling `useQueries` hook inside, so this function should be a hook.

packages/manager/apps/pci-workflow/src/api/hooks/order.ts
```diff
@@ -42,3 +50,56 @@ export const useInstanceSnapshotPricing = (
...
+      code.startsWith('snapshot.consumption') &&
+      regions.find(({ name }) => name === instanceRegion),
+  )?.code;
+  const snapshotAddon =
+    catalog.data?.addons?.find(
+      (addon) => addon.planCode === snapshotPlanCode,
+    ) ||
+    catalog.data?.addons?.find(
+      (addon) => addon.planCode === 'snapshot.consumption.LZ',
+    );
```
fredericvilcot, line 80: Could be simplified:
```suggestion
  const snapshotAddon = catalog.data?.addons?.find(
    (addon) =>
      addon.planCode === snapshotPlanCode ||
      addon.planCode === 'snapshot.consumption.LZ',
  );
```

packages/manager/apps/pci-workflow/src/api/hooks/order.ts
```diff
@@ -19,7 +27,7 @@ export const useProjectAddons = (projectId: string, ovhSubsidiary: string) => {
   });
 };
 
-export const useInstanceSnapshotPricing = (
+export const findInstanceSnapshotPricing = (
```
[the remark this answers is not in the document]
y4nnL: The `useInstanceSnapshotPricing` still exists at the bottom of the file, but its content has been split into two sub functions `findInstanceSnapshotPricing` & `findLocalZoneInstanceSnapshotPricing`, because there are two ways of finding this price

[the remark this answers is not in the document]
fredericvilcot: @y4nnL 
Yes, I see the point 👍🏻
But you cannot invoke any hook inside a regular JS function, that violates the rules of hooks:
https://legacy.reactjs.org/docs/hooks-rules.html#only-call-hooks-from-react-functions

packages/manager/apps/pci-workflow/src/api/hooks/order.ts
y4nnL: This is not the same algorithm
The first one acts as a fallback if the `snapshotPlanCode` it not found in the first place, witch is what we want to achieve. With your suggestion, we could find the `snapshot.consumption.LZ` before the `snapshotPlanCode` depending on the order of the addons and we could end up with an incorrect price

fredericvilcot: Yes, it's another algorithm, thanks for the clarification :)

fredericvilcot (approved):

PR #14579, opened 2024-12-13
fix(mrc): headers props accept ReactNode

### LINKED ISSUE #14570: [manager-components]: Change props title on headers component

Component

### Description

For the office365 react App, we need to add a button on the header title for the dashboard.

Actually we're on props string for title but it's necessary to update from string to React.reactNode

web-office

### Do you have mock-up?

yes

_No response_

### Additional Information

image

tristanwagner:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #14570  

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager-react-components/src/components/content/headers/headers.component.tsx
+++ b/packages/manager-react-components/src/components/content/headers/headers.component.tsx
@@ -1,12 +1,12 @@
-import React from 'react';
+import React, { ReactNode } from 'react';
 import { OdsText } from '@ovhcloud/ods-components/react';
 import { Title, Subtitle } from '../../typography';
 
 export interface HeadersProps {
-  title?: string;
-  subtitle?: string;
-  description?: string;
-  headerButton?: React.ReactElement;
+  title?: ReactNode;
+  subtitle?: ReactNode;
+  description?: ReactNode;
+  headerButton?: ReactNode;
 }
 
 export const Headers: React.FC<HeadersProps> = ({


SimonChaumet (approved):

sachinrameshn (approved):

packages/manager-react-components/src/components/content/headers/headers.component.tsx
```diff
@@ -1,12 +1,12 @@
...
 import { Title, Subtitle } from '../../typography';
 
 export interface HeadersProps {
-  title?: string;
-  subtitle?: string;
-  description?: string;
-  headerButton?: React.ReactElement;
+  title?: ReactNode;
+  subtitle?: ReactNode;
+  description?: ReactNode;
```
anooparveti, line 8: Lets discuss with DEsign team if we need to make `ReactNode` for title, subtitle & decsription.

anooparveti (changes requested):

packages/manager-react-components/src/components/content/headers/headers.component.tsx
tristanwagner: any update about this @anooparveti ?

anooparveti (changes requested):
> So after speaking about this with Anoop, a "← Back" Button left of the label should make more sense in this situation, as a single Tag does the same thing as going back to a higher navigation.

As per @feoche's confirmation, we cannot approve this design. @tristanwagner i let you manage this case with suggested behavior or you can discuss with Design team.

PR #14716, opened 2024-12-24
feat(pci-common): add optional tooltip content props for 1AZ chip

Tsiorifamonjena:
| Question         | Answer
| ---------------- | ---
| Branch?          | `maintenant/manager-pci-common-v0.x`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #TAPC-2269, #TAPC-2270, #TAPC-2309, #TAPC-2413

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Add an optional props to control the tooltip content of a 1AZ Chip for networks µapp only. It will be removed after beta.
We use context to avoid props drilling: the component responsible for the display is in the 4d level from parent


packages/manager/modules/manager-pci-common/src/components/region-selector/RegionSelector.component.tsx
```diff
@@ -62,12 +65,14 @@ export function RegionSelector({
             selectedRegion={selectedMacroRegion}
             onClick={selectMacroRegion}
             render={(region: TLocalisation, isSelected) => (
-              <RegionTile
-                key={region.name}
-                region={region}
-                isSelected={isSelected}
-                isCompact={compactMode}
-              />
+              <GlobalRegionTooltipContext.Provider value={globalRegionTooltip}>
```
SimonChaumet, line 68: Should be above TabsComponent to avoid reproduction

packages/manager/modules/manager-pci-common/src/components/region-selector/RegionGlobalzoneChip.component.tsx
```diff
@@ -26,6 +26,10 @@ import { URL_INFO } from './constants';
 
 export const FEATURE_REGION_1AZ = 'public-cloud:region-1AZ';
 
+export const GlobalRegionTooltipContext = createContext<undefined | string>(
```
SimonChaumet, line 29: I think this should be renamed and move outside

packages/manager/modules/manager-pci-common/src/components/region-selector/RegionSelector.component.tsx
SimonChaumet: Should be in micro app

packages/manager/modules/manager-pci-common/src/translations/region-selector/Messages_fr_FR.json
```diff
@@ -12,6 +12,7 @@
   "pci_project_flavors_zone_localzone_tooltip": "Les Local Zones sont un nouveau type de localisation, qui prennent en charge une partie de notre portefeuille de produits Public Cloud. Nous allons progressivement augmenter le nombre total de Local Zones dans le monde au cours des prochaines années.",
   "pci_project_flavors_zone_localzone_1AZ_tooltip": "Une Local Zone correspond à un mode de déploiement des produits Public Cloud d’OVHcloud dans des datacenters situés au plus près des utilisateurs finaux.",
   "pci_project_flavors_zone_globalregions_tooltip": "Les Régions sont supportées par un ou plusieurs datacenters gérés par OVHCloud. Chaque région fournit une ou plusieurs Availability Zone avec le portefeuille complet de services OVHCloud.",
+  "pci_project_flavors_zone_1AZ_with_3AZ_tooltip": "Avec l'arrivée prochaine des régions 3-AZ, le type de déploiement 'Régions' change de nom ! Une région 1-AZ (zone de disponibilité) regroupe un ou plusieurs datacenters localisés sur un même site. ",
```
SimonChaumet, line 15:
```suggestion
  "pci_project_flavors_zone_1AZ_with_3AZ_tooltip": "Avec l'arrivée prochaine des régions 3-AZ, le type de déploiement 'Régions' change de nom ! Une région 1-AZ (zone de disponibilité) regroupe un ou plusieurs datacenters localisés sur un même site.",
```

SimonChaumet (approved):

packages/manager/modules/manager-pci-common/src/contexts/MetaContext.ts
```diff
@@ -0,0 +1,9 @@
+import { createContext } from 'react';
+
+interface InternalMeta {
+  has3AZ?: boolean;
+}
+
+export type MetaType = InternalMeta | Record<string, unknown> | undefined;
+
+export const MetaContext = createContext<MetaType>(undefined);
```
fredericvilcot, line 9: As the meta context provider is localized in the `RegionSelector` component, I will be more precise: 
```suggestion
export const RegionMetaContext = createContext<RegionMetaType>(undefined);
```

SimonChaumet: @fredericvilcot as meta could be shared between multiple components I think we should keep it not localized

Or else we should rename meta prop to regionMeta in component and put relative files inside the region selector folder

Both solutions are good to me

Tsiorifamonjena: I agree to rename `RegionContext` as we put it in the folder `contexts`. If we need another context we will create another file. And I think also it's more clear to rename meta prop to regionMeta

fredericvilcot: @SimonChaumet 
If we want to handle an agnostic object as `meta`, we could imagine removing the provider from the lib and then just consume the abstraction of the whole object thanks to context consumers

packages/manager/modules/manager-pci-common/src/contexts/PCICommonContext/PCICommonContext.ts
```diff
@@ -0,0 +1,31 @@
...
+ */
+export const PCICommonContext = createContext<PCICommonMetaType>(undefined);
+
+/**
+ * Use this to merge meta properties with previously set properties (merge is only on level 1)
+ *
+ * To override all properties use the provider directly
+ * @param meta
+ */
+export function usePCICommonContextFactory(meta?: PCICommonMetaType) {
```
fredericvilcot, line 27: I will move this hook to the `hook` folder, to ensure global coherence of the library architecture.

PR #14720, opened 2024-12-24
fix(nutanix): disable link in node list when hidden

ThibaudCrespin:
ref: MANAGER-16321

| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-16321 

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager/modules/nutanix/src/dashboard/nodes/list/template.html
+++ b/packages/manager/modules/nutanix/src/dashboard/nodes/list/template.html
@@ -11,8 +11,13 @@
                 class="oui-link"
                 data-ui-sref="{{:: $ctrl.constructor.getNodeDetailsState($row.name) }}"
                 data-ng-bind="$row.iam.displayName"
+                data-ng-if="!$row.isHiddenNode"
             >
             </a>
+            <span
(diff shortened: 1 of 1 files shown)


packages/manager/modules/nutanix/src/dashboard/nodes/list/template.html
```diff
@@ -11,8 +11,13 @@
                 class="oui-link"
                 data-ui-sref="{{:: $ctrl.constructor.getNodeDetailsState($row.name) }}"
                 data-ng-bind="$row.iam.displayName"
+                data-ng-if="!$row.isHiddenNode"
             >
             </a>
+            <span
+                data-ng-if="$row.isHiddenNode"
+                data-ng-bind="$row.iam.displayName"
+            ></span>
```
tibs245, line 20: It's duplicated with `iam-protected-data` code.

- May be see for fix `iam-protected-data` if it's the root cause
- Or anythings else without duplicate logic

tibs245 (changes requested):

PR #14731, opened 2024-12-27
Feat/pci kubernetes multiple fix tapc 2080

ppprevost:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | TAPC-2080 TAPC-2083 TAPC-2441

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager/apps/pci-kubernetes/src/pages/reset/ResetCluster.page.tsx
+++ b/packages/manager/apps/pci-kubernetes/src/pages/reset/ResetCluster.page.tsx
@@ -94,7 +94,7 @@ export default function ResetClusterPage() {
   const [formState, setFormState] = useState({
     workerNodesPolicy: WORKER_NODE_POLICIES.DELETE,
     selectedVersion: getFormatedKubeVersion(kubernetesCluster?.version || ''),
-    privateNetworkId: kubernetesCluster?.privateNetworkId,
+    privateNetworkId: kubernetesCluster?.privateNetworkId ?? undefined,
     subnet: null,
     loadBalancersSubnet: null,
     gateway: {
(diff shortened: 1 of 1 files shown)


packages/manager/apps/pci-kubernetes/src/pages/reset/ResetCluster.page.tsx
```diff
@@ -94,7 +94,7 @@ export default function ResetClusterPage() {
   const [formState, setFormState] = useState({
     workerNodesPolicy: WORKER_NODE_POLICIES.DELETE,
     selectedVersion: getFormatedKubeVersion(kubernetesCluster?.version || ''),
-    privateNetworkId: kubernetesCluster?.privateNetworkId,
+    privateNetworkId: kubernetesCluster?.privateNetworkId ?? undefined,
```
ppprevost, line 97: i know it is weird, but i need to send undefined to the server and not null

Eric-ciccotti (approved):

SimonChaumet (dismissed):

packages/manager/apps/pci-kubernetes/src/pages/reset/ResetCluster.page.tsx
fredericvilcot: @ppprevost 
Do you have any documentation on it?
In the API Schema, `privateNetworkId` is specified as below: 
```txt
privateNetworkId
string
OpenStack private network ID that the cluster will use. Optional, can only be set on cluster creation or reset. If unspecified, the cluster will use the public network.
```

ppprevost: Yes it's optional so not null. 
Today i receive from the server null and send back null if i don't change the value which is normally ok but the server does not accept null but undefined. See with Nicolas Lacroux

fredericvilcot: OK, but why not making the property optional in the object instead?
If the API property is optional, it should not be provided if undefined, it's clearer than sending `undefined value`: 
```suggestion
   ...( !!kubernetesCluster?.privateNetworkId && { privateNetworkId: kubernetesCluster?.privateNetworkId},
```

tibs245 (dismissed):

fredericvilcot:
@ppprevost 
As some of your commits are not relevant anymore, could you please squash them into one?

fredericvilcot (approved):

tibs245 (approved):

PR #14752, opened 2025-01-02
feat(common-translations): init common trans module

anooparveti:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-14580

- [x] Commits are signed-off
- [x] Lint has passed locally
- [ ] ~~Breaking change is mentioned in relevant commits~~ [n/a]

## Description

Initialise new module to maintain common translations used across the µ-apps.

### DIFF
--- a/packages/manager/modules/common-translations/package.json
+++ b/packages/manager/modules/common-translations/package.json
@@ -1,0 +1,30 @@
+{
+  "name": "@ovh-ux/manager-common-translations",
+  "version": "0.1.0",
+  "private": true,
+  "description": "Common translations for Manager µ-apps",
+  "repository": {
+    "type": "git",
+    "url": "git+https://github.com/ovh/manager.git",
+    "directory": "packages/manager/modules/common-translations"
+  },
+  "license": "BSD-3-Clause",
+  "author": "OVH SAS",
+  "sideEffects": false,
+  "main": "./dist/index.cjs.js",
+  "module": "./dist/index.esm.js",
+  "types": "./dist/types/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsc && vite build",
+    "dev": "tsc && vite build",
+    "start:watch": "tsc -w"
+  },
+  "devDependencies": {
+    "vite": "^5.2.13",
+    "vite-plugin-dts": "3.5.1",
+    "vite-plugin-static-copy": "^2.0.0"
+  }
+}
--- a/packages/manager/modules/common-translations/tsconfig.json
+++ b/packages/manager/modules/common-translations/tsconfig.json
@@ -1,0 +1,19 @@
+{
+  "compilerOptions": {
+    "lib": ["dom", "es2021"],
+    "target": "es2021",
+    "types": ["vite/client", "node"],
+    "module": "ES2020",
+    "outDir": "dist",
+    "allowJs": false,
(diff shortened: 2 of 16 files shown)


packages/manager/modules/common-translations/package.json
```diff
@@ -0,0 +1,30 @@
...
+  "author": "OVH SAS",
+  "sideEffects": false,
+  "main": "./dist/index.cjs.js",
+  "module": "./dist/index.esm.js",
+  "types": "./dist/types/index.d.ts",
+  "files": [
+    "dist"
+  ],
+  "scripts": {
+    "build": "tsc && vite build",
```
fredericvilcot, line 21: Question: As we are only handling here `json` files, I am wondering if it's mandatory to have tsc in the scope?

I mean, there is no type checking and no declaration files generated as we just copy static elements into the dist folder.

By the way, if needed, `typescript` should be in scope to run tsc scripts.

packages/manager/modules/common-translations/tsconfig.json
```diff
@@ -0,0 +1,19 @@
...
+    "module": "ES2020",
+    "outDir": "dist",
+    "allowJs": false,
+    "skipLibCheck": true,
+    "esModuleInterop": false,
+    "allowSyntheticDefaultImports": true,
+    "moduleResolution": "Node",
+    "isolatedModules": true,
+    "declaration": true,
+    "declarationDir": "./dist/types",
```
fredericvilcot, line 15: Are we emitting some `*.d.ts` files in this library?

packages/manager/modules/common-translations/vite.config.mjs
```diff
@@ -0,0 +1,40 @@
...
+export default defineConfig({
+  root: path.resolve(process.cwd(), 'src'),
+  resolve: {
+    alias: {
+      '@/': pathSrc,
+    },
+  },
+  plugins:[
+    dts({
+      outDir: '../dist/types',
```
fredericvilcot, line 18: IMO, it's safer to use absolute path: 
```suggestion
      root: __dirname,
      outDir: 'dist/types',
```

packages/manager/modules/common-translations/package.json
anooparveti: There's a TS file which exports the namespaces https://github.com/ovh/manager/pull/14752/files#diff-1185723c5f7c95552f7c4b7f4020358e42d516c528f68949c24147533f15ce0c. This way it will be easy to consume with i18n.

packages/manager/modules/common-translations/tsconfig.json
anooparveti: Yes.

packages/manager/modules/common-translations/package.json
fredericvilcot: Yes, but is it mandatory to use tsc and vite-dts (that have a cost on build phase) to produce a simple raw object?

anooparveti: It is to assist the developers with the available namespaces from the package with the auto-completion.

fredericvilcot (approved):
👍🏻

frenautvh (approved):

ghyenne (approved):

JacquesLarique (approved):

PR #14800, opened 2025-01-07
test(container): add test for tracking init in cookie policy modal

aottr:
ref: MANAGER-16559

| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | MANAGER-16559

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Testing the tracking plugins behavior (and it's initialization) depending on the customers region and tracking cookie value.

### DIFF
--- a/packages/manager/apps/container/src/cookie-policy/CookiePolicy.spec.tsx
+++ b/packages/manager/apps/container/src/cookie-policy/CookiePolicy.spec.tsx
@@ -1,0 +1,92 @@
+import { render, waitFor } from '@testing-library/react';
+import { it, vi, describe, expect } from 'vitest';
+import { initShell } from '@ovh-ux/shell';
+import CookiePolicy from './CookiePolicy';
+import { Environment, User } from '@ovh-ux/manager-config';
+import ApplicationContext from '@/context/application.context';
+import { useCookies } from 'react-cookie';
+
+const onValidate = vi.fn();
+const trackingInit = vi.fn().mockResolvedValue(undefined);
+const trackingSetEnabled = vi.fn().mockResolvedValue(undefined);
+const trackingOnConsentModalDisplay = vi.fn().mockResolvedValue(undefined);
+
+const renderCookiePolicy = async () => {
+  const shell = await initShell();
+  const environment = shell.getPlugin('environment').getEnvironment();
+  render(
+    <ApplicationContext.Provider value={{ shell, environment }}>
+      <CookiePolicy onValidate={onValidate} shell={shell} />
+    </ApplicationContext.Provider>
+  );
+};
+
+const mockedShell = (region: string) => ({
+  getPlugin: (plugin: string) => ({
+    environment: {
+      getEnvironment: () => ({
+        user: {
+          ovhSubsidiary: region,
+        } as User,
+        getRegion: () => region,
+      } as Environment),
+    },
+    tracking: {
+      setRegion: vi.fn(),
+      init: trackingInit,
+      setEnabled: trackingSetEnabled,
+      onConsentModalDisplay: trackingOnConsentModalDisplay,
+    },
+  }[plugin]),
+});
+
+vi.mock('@ovh-ux/shell');
+vi.mock('react-cookie');
+
+
+describe('CookiePolicy.component', () => {
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it.each([
+    ['EU', 'valid'],
+    ['US', 'invalid'],
+  ])('should init tracking if region is %s and cookie is %s', async (region, cookieValidity) => {
+
+    const cookieValue = cookieValidity === 'valid' ? '1' : '0';
+    vi.mocked(useCookies).mockReturnValue([{ MANAGER_TRACKING: cookieValue, }, vi.fn(), vi.fn()]);
+    (await import('@ovh-ux/shell')).initShell = vi.fn().mockResolvedValue(mockedShell(region));
+    void renderCookiePolicy();
+    await waitFor(() => {
+      expect(trackingInit).toHaveBeenCalledWith(true);
+      expect(trackingSetEnabled).not.toHaveBeenCalled();
+      expect(trackingOnConsentModalDisplay).not.toHaveBeenCalled();
+    }, { timeout: 2000 });
+  });
+
+  it('should show consent modal if cookie is null and region is not US', async () => {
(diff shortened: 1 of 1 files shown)


packages/manager/apps/container/src/cookie-policy/CookiePolicy.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+    (await import('@ovh-ux/shell')).initShell = vi.fn().mockResolvedValue(mockedShell('EU'));
+    void renderCookiePolicy();
+    await waitFor(() => {
+      expect(trackingInit).not.toHaveBeenCalled();
+      expect(trackingSetEnabled).not.toHaveBeenCalled();
+      expect(trackingOnConsentModalDisplay).toHaveBeenCalled();
+    }, { timeout: 2000 });
+  });
+
+  it('should disable tracking plugin if cookie is not valid or null and region is not US', async () => {
```
JacquesLarique, line 93: The description of this test is incorrect, `trackingSetEnabled` will only be called if MANAGER_TRACKING is not null
```suggestion
  it('should disable tracking plugin if cookie is not valid and region is not US', async () => {
```

packages/manager/apps/container/src/cookie-policy/CookiePolicy.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+
+    (await import('react-cookie')).useCookies = vi.fn().mockReturnValue([{ MANAGER_TRACKING: '0', }, vi.fn()]);
+    (await import('@ovh-ux/shell')).initShell = vi.fn().mockResolvedValue(mockedShell('EU'));
+    void renderCookiePolicy();
+    await waitFor(() => {
+      expect(trackingInit).not.toHaveBeenCalled();
+      expect(trackingSetEnabled).toHaveBeenCalled();
```
JacquesLarique, line 100: I think we can test that `trackingSetEnabled` is called with `false` parameter:
```suggestion
      expect(trackingSetEnabled).toHaveBeenCalledWith(false);
```

packages/manager/apps/container/src/cookie-policy/CookiePolicy.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+  }[plugin]),
+});
+
+vi.mock('@ovh-ux/shell', () => ({
+  initShell: vi.fn()
+}));
+
+vi.mock('react-cookie', () => ({
+  useCookies: vi.fn(),
+}));
```
SimonChaumet, line 48:
```suggestion
vi.mock('@ovh-ux/shell');

vi.mock('react-cookie');
```
Returning specific function isn't necessary

packages/manager/apps/container/src/cookie-policy/CookiePolicy.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+describe('CookiePolicy.component', () => {
+
+  afterEach(() => {
+    vi.restoreAllMocks();
+  });
+
+  it('should init tracking if cookie is valid and region is not US', async () => {
+
+    (await import('react-cookie')).useCookies = vi.fn().mockReturnValue([{ MANAGER_TRACKING: '1', }, vi.fn()]);
+    (await import('@ovh-ux/shell')).initShell = vi.fn().mockResolvedValue(mockedShell('EU'));
```
SimonChaumet, line 60: You can import them directly and mock the functions with `vi.mocked(useCookies).mockReturnValue(...)`

packages/manager/apps/container/src/cookie-policy/CookiePolicy.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+const trackingOnConsentModalDisplay = vi.fn().mockResolvedValue(undefined);
+
+const renderCookiePolicy = async () => {
+  const shell = await initShell();
+  const environment = shell.getPlugin('environment').getEnvironment();
+  render(
+    <ApplicationContext.Provider value={{ shell, environment }}>
+      <CookiePolicy onValidate={onValidate} shell={shell} />
+    </ApplicationContext.Provider>
+  );
```
SimonChaumet, line 20: I think using a wrapper is more adapted in this situation

packages/manager/apps/container/src/cookie-policy/CookiePolicy.spec.tsx
```diff
@@ -0,0 +1,104 @@
...
+    (await import('@ovh-ux/shell')).initShell = vi.fn().mockResolvedValue(mockedShell('EU'));
+    void renderCookiePolicy();
+    await waitFor(() => {
+      expect(trackingInit).toHaveBeenCalled();
+      expect(trackingSetEnabled).not.toHaveBeenCalled();
+      expect(trackingOnConsentModalDisplay).not.toHaveBeenCalled();
+    }, { timeout: 2000 });
+  });
+
+  it('should init tracking if cookie is invalid and region is US', async () => {
```
SimonChaumet, line 69: You can use it.each to avoid duplication in tests

packages/manager/apps/container/src/cookie-policy/CookiePolicy.spec.tsx
aottr: This works on the `useCookies` hook but is more difficult on the `initShell` because it expects a type-safe `Shell` object. Good idea though. I switched to a "dynamic" import on some cases because it sometimes did not reset the mocked function despite calling a restore after each test run.

packages/manager/apps/container/src/cookie-policy/CookiePolicy.spec.tsx
aottr: I thought about that but the only...really duplicated code is the render call. Otherwise I set the hook return values to different values in each test case. Since this happens before the render, I can not even put the render in the each-statement.

packages/manager/apps/container/src/cookie-policy/CookiePolicy.spec.tsx
aottr: I know, maybe the description was not distinct enough. The condition is:

Cookie is not valid and Cookie is not `null` and the region is not US.

anooparveti (approved):

PR #14853, opened 2025-01-09
feat(container): replace lodash with built-in functions and helpers

aottr:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-15629

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description
We only used a few methods of lodash which could be replaced by simple helper functions.
These include:
- `groupBy[Func]`
- `truncate`
- `capitalize`
- `useDebounce` from `react-use` was used instead of `debounce` from lodash

Behavior was tested against the implementation in lodash, function signatures share a high similarity.


JacquesLarique (approved):

packages/manager/apps/container/src/helpers/collectionHelper/collectionHelper.ts
```diff
@@ -0,0 +1,20 @@
+/**
+ * Creates an object composed of keys generated from the results of running each element of an array through mapper.
+ * @param array array iterate over
+ * @param mapper The function to transform keys.
+ * @returns Returns the composed aggregate object.
+ */
+export function groupByFunc<
```
anooparveti, line 7:
```suggestion
export function groupBy<
```

aottr: I chose the name `groupByFunc` since the method is much simpler than the implementation of lodash. Since TS doesn't support overloading, I wanted the name a bit more distinct if we, for example, need a `groupBy` that works with the key-name in the future.

If a rename to `groupBy` is still preferred, I can change it.

packages/manager/apps/container/src/helpers/functionHelper/functionHelper.ts
anooparveti, line 1: **nitpick:** No helper needed for debounce, you can use [useDebounce](https://streamich.github.io/react-use/?path=/story/side-effects-usedebounce--docs) from [react-use](https://github.com/ovh/manager/blob/87fd370b0f20b830cb2b88c10779a0bd2aebbd66/packages/manager/apps/container/package.json#L72)

anooparveti (changes requested):

JacquesLarique (dismissed):

packages/manager/apps/container/src/helpers/functionHelper/functionHelper.ts
aottr: fair point, I gonna look into it since we have that package already in the container.

aottr: I'm not sure the usage here would be beneficial. As far as I can see, the hook expects a state/value for it to trigger the change and we can't just use the returned function like before.

A possible implementation would be creating another state for the window-size (since we care about the resize-event, or a trigger increment that gets called on each resize.

Both would lead to another state and callback function, just to trigger the debounce hook. The main use-case seems to be for debouncing controlled input fields etc.

```typescript
const [windowSize, setWindowSize] = useState({
  width: window.innerWidth,
  height: window.innerHeight,
});

const debouncedWindowSize = useDebounce(() => {
  setIsPopoverVisible(false);
  calculateTargetBound();
  updatePopper();
}, 100);

const onWindowResize = useCallback(() => {
  setWindowSize({
    width: window.innerWidth,
    height: window.innerHeight,
  });
}, []);
```
and inside the `useEffect`:
```typescript
window.addEventListener('resize', onWindowResize);

return () => window.removeEventListener('resize', onWindowResize);
```

something like this (untested)

packages/manager/apps/container/src/container/nav-reshuffle/onboarding/walkMe/index.tsx
```diff
@@ -258,11 +283,12 @@ export const OnboardingWalkMe = () => {
     [currentStepIndex],
   );
 
-  const onWindowResize = useCallback(() => {
+  useDebounce(() => {
     setIsPopoverVisible(false);
     calculateTargetBound();
     updatePopper();
-  }, [currentStepIndex]);
+    console.log('windowSize', windowSize);
```
JacquesLarique, line 290: I think you forgot to remove this `console.log`

aottr: ahh.. true. Thanks !

JacquesLarique (approved):

anooparveti (approved):

PR #14889, opened 2025-01-13
feat(mrc): add changeling component

MaximeBajeux:
| Question         | Answer
| ---------------- | ---
| Branch?          | `master`
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #... 

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager-react-components/src/components/content/headers/headers.component.tsx
+++ b/packages/manager-react-components/src/components/content/headers/headers.component.tsx
@@ -22,7 +22,9 @@ export const Headers: React.FC<HeadersProps> = ({
         {subtitle && <Subtitle className="block">{subtitle}</Subtitle>}
         {description && <OdsText preset="span">{description}</OdsText>}
       </div>
-      {headerButton && <div className="flex flex-wrap">{headerButton}</div>}
+      {headerButton && (
+        <div className="flex flex-wrap justify-end">{headerButton}</div>
+      )}
     </div>
   );
 };
--- a/packages/manager-react-components/src/components/content/headers/headers.component.tsx
+++ b/packages/manager-react-components/src/components/content/headers/headers.component.tsx
@@ -22,7 +22,7 @@ export const Headers: React.FC<HeadersProps> = ({
         {subtitle && <Subtitle className="block">{subtitle}</Subtitle>}
         {description && <OdsText preset="span">{description}</OdsText>}
       </div>
-      {headerButton && <div className="flex flex-nowrap">{headerButton}</div>}
+      {headerButton && <div className="flex flex-wrap">{headerButton}</div>}
     </div>
   );
 };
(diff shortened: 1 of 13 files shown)


packages/manager-react-components/src/components/navigation/menus/changelog/changelog.component.tsx
JacquesLarique, line 1: Can't we reuse `GuideButton` ? (and maybe rename it to something like `DropdownMenu`)
It seems the code are the same

packages/manager-react-components/src/components/content/headers/headers.component.tsx
```diff
@@ -26,7 +26,9 @@ export const Headers: React.FC<HeadersProps> = ({
           </OdsText>
         )}
       </div>
-      {headerButton && <div>{headerButton}</div>}
+      {headerButton && (
+        <div className="flex flex-wrap justify-end">{headerButton}</div>
+      )}
```
anooparveti, line 31: **question:** What about the layout for Mobile devices? Do we have the specifications for it and is it cosidered?

anooparveti (changes requested):

packages/manager-react-components/src/components/content/headers/headers.component.tsx
MaximeBajeux: Yes I handled the behaviour on mobile

Capture d’écran 2025-02-05 à 16 12 34

qpavy (approved):

JacquesLarique (dismissed):

ghyenne (dismissed):

packages/manager-react-components/src/components/content/headers/headers.component.tsx
anooparveti, line 31: **nitpick:** Instead of accepting all the buttons as one single react node, i think it can be beneficial to accept it as a separate prop called `changelogButton`. Also, since the current prop is just `headerButton` (singular button) it does not convey implicitly that it could accept and well adjust the layout for more buttons.

anooparveti (changes requested):

packages/manager-react-components/src/components/content/headers/headers.component.tsx
MaximeBajeux: I not changed the name of the prop to not cause a breaking change with the existing implementation. But if you want I can handle the dump in the changelog implementation's PR.

For me add a new prop "changelogButton" without changing headerButton into guideButton is a little bit confusing, as changelogButton is a header button too.

I propose to change headerButton to headerButtons, also to avoid to add a potential third prop in the future

JacquesLarique (approved):

PR #15056, opened 2025-01-22
feat(mrc): add order config/summary component

darsene:
ref: MANAGER-16647

| Question         | Answer
| ---------------- | ---
| Branch?          |`develop` 
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #MANAGER-16647

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Add a component to wrap the order of a new product when it uses express order

image

Firt of all, the order configuration part should be provided as a children of `<Order.Configuration>`

image

Once order configuration is OK, by clicking on order button, a summary is displayed while opening a new tab with the express order (provided by the given link)

image

`Product label` can be overridden otherwise it will be `service` by default

Tracking and navigation routing could be provided  by all the event button props 

## Related

Github issue : https://github.com/ovh/manager/issues/15068

### DIFF
--- a/packages/manager-react-components/src/components/order/Order.component.spec.tsx
+++ b/packages/manager-react-components/src/components/order/Order.component.spec.tsx
@@ -1,0 +1,130 @@
+import React, { fireEvent } from '@testing-library/react';
+import { vi, vitest } from 'vitest';
+import { Order } from './Order.component';
+import { render } from '../../utils/test.provider';
+
+describe('<Order> tests suite', () => {
+  // Mock global window.open
+  vi.stubGlobal('open', vi.fn());
+
+  const onCancelSpy = vi.fn();
+  const onValidateSpy = vi.fn();
+  const onFinishSpy = vi.fn();
+  const onClickLinkSpy = vi.fn();
+  const orderLink = 'https://order-link';
+
+  afterEach(() => {
+    vitest.resetAllMocks();
+  });
+
+  const renderComponent = (
+    isValid: boolean,
+    link: string,
+    productName: string,
+  ) =>
+    render(
+      <Order>
+        <Order.Configuration
+          isValid={isValid}
+          onCancel={onCancelSpy}
+          onConfirm={onValidateSpy}
+        >
+          <p>Order steps</p>
+        </Order.Configuration>
+        <Order.Summary
+          onFinish={onFinishSpy}
+          orderLink={link}
+          onClickLink={onClickLinkSpy}
+          productName={productName}
+        ></Order.Summary>
+      </Order>,
+    );
+
+  it.each([{ valid: true }, { valid: false }])(
+    'when order configuration validity is $valid confirm button disabled attribute should be $valid',
+    ({ valid }) => {
+      const { getByTestId, getByText } = renderComponent(valid, null, null);
+
+      expect(getByText('Order steps')).toBeVisible();
+
+      const orderButton = getByTestId('cta-order-configuration-order');
+      expect(orderButton).toHaveAttribute('label', 'Commander');
+      expect(orderButton).toHaveAttribute('is-disabled', `${!valid}`);
+    },
+  );
+
+  it('confirm button should be enabled and clickable when order configuration is valid', () => {
+    const { getByTestId } = renderComponent(true, null, null);
+
+    fireEvent.click(getByTestId('cta-order-configuration-order'));
+
+    expect(onValidateSpy).toHaveBeenCalled();
+  });
+
+  it('should cancel order configuration when cancel button is clicked', () => {
+    const { getByTestId } = renderComponent(false, null, null);
+
+    fireEvent.click(getByTestId('cta-order-configuration-cancel'));
+
+    expect(onCancelSpy).toHaveBeenCalled();
+  });
+
+  it('should open order link and display order summary when order configuration is confirmed ', () => {
+    vi.spyOn(window, 'open');
+    const { getByTestId, queryByText } = renderComponent(true, orderLink, null);
+
+    fireEvent.click(getByTestId('cta-order-configuration-order'));
+
+    // order configuration is hidden
+    expect(queryByText('Order steps')).not.toBeInTheDocument();
+
+    expect(getByTestId('order-summary-title')).toBeVisible();
+    expect(getByTestId('order-summary-link')).toBeVisible();
+
+    expect(window.open).toHaveBeenCalledTimes(1);
+    expect(window.open).toHaveBeenCalledWith(
+      orderLink,
+      '_blank',
+      'noopener,noreferrer',
+    );
+  });
+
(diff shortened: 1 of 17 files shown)


packages/manager-react-components/src/components/order/Order.component.spec.tsx
```diff
@@ -0,0 +1,131 @@
...
+      const orderButton = getByTestId('cta-order-configuration-order');
+      expect(orderButton).toHaveAttribute('label', 'Commander');
+      expect(orderButton).toHaveAttribute('is-disabled', `${!valid}`);
+    },
+  );
+
+  it('confirm button should be enabled and clickable when order configuration is valid', () => {
+    const { getByTestId } = renderComponent(true, null, null);
+
+    fireEvent.click(getByTestId('cta-order-configuration-order'));
```
tibs245, line 59: I known it's work but the docs say to use userEvent to do this. So when you prefer

https://testing-library.com/docs/user-event/intro/#writing-tests-with-userevent

packages/manager-react-components/src/components/order/Order.component.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { OrderContextProvider } from './Order.context';
+import { OrderConfiguration } from './OrderConfiguration.component';
+import { OrderSummary } from './OrderSummary.component';
+import './translations';
+
+export const Order = ({ children }: { children: JSX.Element[] }) => {
```
tibs245, line 7: It's exactly same but it's own name. I let's you juge. No impact 

```suggestion
export const Order = ({ children }: React.PropsWithChildren) => {
```

packages/manager-react-components/src/components/order/Order.context.tsx
```diff
@@ -0,0 +1,31 @@
...
+export type TOrderContext = {
+  setIsOrderInitialized: (isOrderInitialized: boolean) => void;
+  isOrderInitialized: boolean;
+};
+
+const OrderContext = createContext<TOrderContext>({} as TOrderContext);
+
+export const OrderContextProvider: React.FC<React.PropsWithChildren> = ({
+  children,
+}) => {
```
tibs245, line 12:
```suggestion
export const OrderContextProvider = ({
  children,
}: React.PropsWithChildren) => {
```

If we want uniformize

tibs245 (dismissed):

packages/manager-react-components/src/components/order/Order.component.spec.tsx
darsene: The library is not used in MRC so for now, I let as is

tibs245 (dismissed):

hbelkhir (approved):

packages/manager-react-components/src/components/order/Order.context.tsx
```diff
@@ -0,0 +1,32 @@
...
+  setIsOrderInitialized: (isOrderInitialized: boolean) => void;
+  isOrderInitialized: boolean;
+};
+
+const OrderContext = createContext<TOrderContext>({} as TOrderContext);
+
+export const OrderContextProvider = ({ children }: React.PropsWithChildren) => {
+  const [isOrderInitialized, setIsOrderInitialized] = useState<boolean>(false);
+
+  const context = useMemo<TOrderContext>(
```
ppprevost, line 13: why using useMemo here if you don't use a loop ?

packages/manager-react-components/src/components/order/translations/index.ts
```diff
@@ -0,0 +1,27 @@
...
+// import en_GB from './Messages_en_GB.json';
+// import es_ES from './Messages_es_ES.json';
+// import fr_CA from './Messages_fr_CA.json';
+import fr_FR from './Messages_fr_FR.json';
+// import it_IT from './Messages_it_IT.json';
+// import pl_PL from './Messages_pl_PL.json';
+// import pt_PT from './Messages_pt_PT.json';
+
+function addTranslations() {
+  // i18next.addResources('de_DE', 'order', de_DE);
```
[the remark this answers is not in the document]
darsene: Once logic approved, I'll generate translation with systran and I'll uncomment those lines

packages/manager-react-components/src/components/order/Order.context.tsx
darsene: I agree but it was required by a sonar issue

packages/manager-react-components/src/components/order/Order.context.tsx
```diff
@@ -0,0 +1,32 @@
...
+  );
+};
+
+export const useOrderContext = (): TOrderContext => {
+  const context = useContext(OrderContext);
+  if (context === undefined) {
+    throw new Error('Order-related components must be used within <Order>');
+  }
+  return context;
+};
```
anooparveti, line 32: **nitpick:** If the goal of the context is to get the information in the children components down the hierarchy of OrderConfiguration or OrderSummary, you will have to export it from index.ts.

anooparveti (changes requested):

packages/manager-react-components/src/components/order/Order.context.tsx
tibs245: I think it's because the object change reference on each call. And this trigger a re-render to verify if nothing has changed.

With `useMemo` the ref of object change only is the value of state change

rjamet-ovh (dismissed):

packages/manager-react-components/src/components/order/index.ts
```diff
@@ -0,0 +1,2 @@
+export * from './Order.component';
+export * from './Order.context';
```
anooparveti, line 2: **nitpick:** In packages/manager-react-components/src/components/order/Order.context.tsx, you're also exporting ContextProvider to use it in Order component. And this will export it from the library as well and it's not desired. Can you only export `useOrderContext` specifically.

anooparveti (changes requested):

packages/manager-react-components/src/components/order/OrderSummary.component.tsx
```diff
@@ -0,0 +1,83 @@
...
+    <div className="flex flex-col gap-8">
+      <div className="flex flex-col gap-4">
+        <OdsText
+          preset={ODS_TEXT_PRESET.heading2}
+          data-testid="order-summary-title"
+        >
+          {t('order_summary_order_initiated_title', { product })}
+        </OdsText>
+        <OdsText preset={ODS_TEXT_PRESET.paragraph}>
+          <Trans
```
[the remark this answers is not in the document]
darsene: @seven-amid  As the page is reloaded when the language is changed it will work

anooparveti (approved):

tibs245 (approved):

PR #15058, opened 2025-01-22
feat(ips): parking order

chipp972:
ref: MANAGER-15843

| Question         | Answer
| ---------------- | ---
| Branch?          | `project/ip-revamp` 
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | MANAGER-15843 

- [x] Commits are signed-off
- [x] Lint has passed locally


packages/manager/apps/ips/src/data/hooks/catalog/catalog.utils.ts
```diff
@@ -1,5 +1,16 @@
...
+} from '@/components/RegionSelector/region-selector.utils';
 import { CatalogIpPlan } from '../../api/catalog';
 
+export const getContinentKeyFromRegion = (region: string) => {
+  if (isRegionInEu(region)) {
+    return 'EU';
+  }
+  return isRegionInUs(region) ? 'US' : 'CA';
+};
+
```
sachinrameshn, line 13: For better readability can we use switch ?
```suggestion
export const getContinentKeyFromRegion = (region: string) => {
   switch (region) {
    case isRegionInEu(region):
      return 'EU';
    case isRegionInUs(region):
      return 'US';
    default:
      return 'CA';
  }
};

```

sachinrameshn (changes requested):

packages/manager/apps/ips/src/data/hooks/catalog/useAdditionalIpsRegions.ts
```diff
@@ -13,7 +18,11 @@ export const useAdditionalIpsRegions = () => {
       ? Array.from(
           new Set(
             data.data.plans
-              .filter((plan) => plan.planCode.includes('v4'))
+              .filter((plan) =>
+                serviceType === ServiceType.ipParking
+                  ? plan.planCode.includes('failover')
+                  : plan.planCode.includes('v4'),
+              )
```
sachinrameshn, line 25:
```suggestion
              .filter(({ planCode }) =>
                serviceType === ServiceType.ipParking
                  ? planCode.includes('failover')
                  : planCode.includes('v4'),
              )
```

packages/manager/apps/ips/src/data/hooks/useServiceList.ts
```diff
@@ -0,0 +1,58 @@
...
+      if (queries[1]?.data?.includes(serviceId)) {
+        return ServiceType.server;
+      }
+      if (queries[2]?.data?.includes(serviceId)) {
+        return ServiceType.vps;
+      }
+      return queries[3]?.data?.data?.includes(serviceId)
+        ? ServiceType.vrack
+        : ServiceType.unknown;
+    },
```
sachinrameshn, line 49: Possible to use switch here ?

packages/manager/apps/ips/src/pages/layout.tsx
```diff
@@ -17,6 +17,7 @@ export default function Layout() {
   useEffect(() => {
     const match = matches.slice(-1);
     defineCurrentPage(`app.ips-${match[0]?.id}`);
+    console.log(`app.ips-${match[0]?.id}`);
```
sachinrameshn, line 20: To remove console statement.
```suggestion
   
```

packages/manager/apps/ips/src/data/hooks/catalog/catalog.utils.ts
chipp972: Well the `switch(region)` is not usefull since we have to call a function in each case
It would do the same with `switch(true)`

I can just use a second `if` instead of a ternary if you find it clearer to read
```
if (isRegionInEu(region))
  return 'EU';
if (isRegionInUs(region))
  return 'US';
return 'CA';
```

packages/manager/apps/ips/src/data/hooks/useServiceList.ts
chipp972: I don't think we can do this with the `switch` syntax
It's made to test a variable against multiple values (like the first case) but the other cases test the inclusion inside an array of values

sachinrameshn (approved):

packages/manager/apps/ips/src/pages/order/order.utils.ts
qpavy, line 1: Can we add some comments to explain what utils functions are doing ? (for most complex ones ?)

chipp972: Yes
Is it good like this ?

PR #15102, opened 2025-01-24
fix(pci-object-storage): add object banner

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | feat/pci-object-storage
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | DTCORE-3005

## Description

fix add object banner conditions


kqesar (approved):

seven-amid (approved):

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add/AddObject.page.tsx
```diff
@@ -103,14 +98,28 @@ export default function AddObjectPage() {
...
+  );
+
+  const showStorageClass = useMemo(
+    () =>
+      containerRegion?.services?.length > 1 &&
+      containerRegion?.services?.some(
+        (item) => item.name === 'storage-s3-high-perf' && item.status === 'UP',
+      ),
+    [containerRegion?.services],
+  );
```
tibs245, line 121: For informations/improvement if services is a empty array the result is false. So we can remove this line

```suggestion
  const showStorageClass = useMemo(
    () =>
      containerRegion?.services?.some(
        (item) => item.name === 'storage-s3-high-perf' && item.status === 'UP',
      ),
    [containerRegion?.services],
  );
```

tibs245: After re-read. 

If `containerRegion?.services?.length > 1` is not a error and not should be `containerRegion?.services?.length >= 1`

In the first case (not a error), your code is well

In the second case. Be careful if you have only one result the result will be false. So you can just remove this line

tibs245 (approved):

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add/AddObject.page.tsx
frenautvh: hello, this part is copy pasted straight from the angularjs code, but i agree with you that after re-reading it, it looks that it could fail i'm going to remove the length check (i think this is an error on actual production as it doesn't make sense to me to check for the lenght greater than 1) thanks for your feedback

PR #15105, opened 2025-01-24
feat(manager-react-component): add billing informations tile component

### LINKED ISSUE #15104: [billing-react-components]: Add generic component for BillingInformationTile

Component

### Description

On all dashboard of product we have a BillingInformation Tile with different use for each product without reason.

This may be useful have unique standard and use it for all product

KMS and HYCU

### Do you have mock-up?

https://xd.adobe.com/view/ab63e2f1-e9ff-4b25-b769-d4efc0997896-fe30/screen/0c83e032-d756-45a3-8c61-8b70317b684f/

30-01-2025

### Additional Information

_No response_

tibs245:
| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | yes
| Breaking change? | no
| Tickets          | Fix #15104

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

See `documentation.mdx`

Screenshot 2025-01-20 at 16 15 39
Screenshot 2025-01-23 at 13 50 42


packages/manager-react-components/src/components/tiles/BillingInformationsTile/BillingDetails.class.ts
```diff
@@ -0,0 +1,235 @@
+import { isAfter, isBefore } from 'date-fns';
+import { ServiceDetails } from '../../../hooks/services/services.type';
+
+export default class BillingDetails {
```
chipp972, line 4: Why use class syntax ?
It just looks like it makes it harder to maintain and read

packages/manager-react-components/src/components/tiles/BillingInformationsTile/BillingInformationsTile.stories.tsx
```diff
@@ -0,0 +1,44 @@
...
+    <BillingInformationsTile.Divider />
+    <BillingInformationsTile.CreationDate />
+    <BillingInformationsTile.Divider />
+    <BillingInformationsTile.State />
+    <BillingInformationsTile.Divider />
+    <BillingInformationsTile.NextBillingDate />
+    <BillingInformationsTile.Divider />
+    <BillingInformationsTile.ResiliateLink />
+  </BillingInformationsTile>
+);
```
chipp972, line 18: With this syntax you have to repeat the Divider component every time.
Do we really need the flexibility that comes with this syntax ?
Most of the Billing tiles look alike (and that's a good thing for UX I think) and the elements are ordered the same.

packages/manager-react-components/src/components/tiles/BillingInformationsTile/BillingDetails.class.ts
tibs245: The main reason it's because this is the code in AngularJS. And I tried to keep the maximum of this logic. And I tried to do the better with the time allocated and keep the maximum same method to keep comparison possible

anooparveti (changes requested):
https://github.com/ovh/manager/issues/15104#issuecomment-2619058227

PR #15121, opened 2025-01-24
feat(ovh-at-internet): remove lodash dependency

tristanwagner:
ref: MANAGER-16864

| Question         | Answer
| ---------------- | ---
| Bug fix?         | no
| New feature?     | no
| Breaking change? | hopefully not
| Tickets          | MANAGER-16864 

- [x] Commits are signed-off
- [x] Lint has passed locally

## Description

Any app that use this package will include the full lodash dependency, so we can remove it from here to save a lot of bundle size in a lot of apps


packages/components/ovh-at-internet/src/utils.ts
SimonChaumet, line 1: Could you group functions by there usage or type to avoid having a utils file that could be filled with anything ? Maybe a folder utils with array.ts, object.ts etc

ghyenne (dismissed):

packages/components/ovh-at-internet/src/utils.ts
```diff
@@ -1,6 +1,44 @@
...
+  ) {
     console.debug(...args);
   }
 }
+
+export const isEmpty = (obj: Record<string, unknown>) => {
+  if (obj?.length || obj?.size) return false;
+  if (typeof obj !== 'object') return true;
+  return !Object.values(obj).length;
+};
```
SimonChaumet, line 14: I think we should specialize this function for each type that is used because this function is too generic

packages/components/ovh-at-internet/src/utils.ts
```diff
@@ -1,6 +1,44 @@
...
+
+/**
+ * Returns an unique ID (used by orders and carts)
+ * To ensure that the ID is unique, we generate it from current milliseconds
+ * and concat a random string to avoid any collisions.
+ */
+export const getUniqueId = (): string => {
+  // unique key : current milliseconds in base 36 concatenated with 8 random chars
+  return new Date().valueOf().toString(36) + getRandomString(8);
```
SimonChaumet, line 43: why not used uuid instead ?

tristanwagner: I did not changed this function, only deleted the file from `utils/index.ts` and moved it all into this `utils.ts` file

packages/components/ovh-at-internet/src/utils.ts
tristanwagner: it is only using on object in this project, actually the type of `obj` should be `unknown`

packages/components/ovh-at-internet/src/utils.ts
tristanwagner: I would prefer to keep it simple, I don't really see the point to have x files with one function in each file

sachinrameshn (dismissed):

packages/components/ovh-at-internet/src/utils.ts
```diff
@@ -1,6 +1,44 @@
...
+  ) {
     console.debug(...args);
   }
 }
+
+export const isEmpty = (obj: any) => {
+  if (obj?.length || obj?.size) return false;
+  if (typeof obj !== 'object') return true;
+  return !Object.values(obj).length;
+};
```
sachinrameshn, line 14: Can you please add comments regarding the goal of this function ?

packages/components/ovh-at-internet/src/utils.ts
SimonChaumet: The goal is to keep it simple : by grouping them inside only one file we often result in a big file that can't be refactored easily

packages/components/ovh-at-internet/src/utils.ts
SimonChaumet: I think we should keep a function for each : isSetEmpty, isObjectEmpty... because if we can only have a map we wouldn't recreate a function and it will be worst to refactor later

tristanwagner: I get your point and you are right, but to me it's kinda out of scope, the goal of this PR is to remove `lodash` so I want to just replace the functions used from `lodash` to avoid any breaking changes, unexpected side effects. If we want to improve the code of this package I'd suggest refactoring it in another PR

SimonChaumet: What is the point to remove lodash if we copy the functions inside the project ? To me we need to remove lodash to avoid using functions that already exists in native javascript

Or maybe it is to reduce bundle size ? But isn't lodash tree-shaked ?

tristanwagner: it is indeed to reduce bundle size, and while `lodash` supports tree-shaking, it is not used in this project as it imports the full library

so instead of just moving to treeshaked dependency declarations I made the decision to replace the function instead to remove any dependence to this package (I think we should aim to this goal for the whole manager project)

packages/components/ovh-at-internet/src/utils.ts
tristanwagner: I commented the functions that were added

tristanwagner:
I had to update devDependencies of container app to add lodash types because it would not build without it and was relying on the declaration in devDeps of `ovh-at-internet` (bad)

ghyenne (approved):

PR #15244, opened 2025-01-31
Feat/pci common dtcore

frenautvh:
| Question         | Answer
| ---------------- | ---
| Branch?          | master
| Bug fix?         | yes
| New feature?     | yes
| Breaking change? | no
| Tickets          | DTCORE-2868, DTCORE-2951

## Description

Add features & bugfixes related to migration of quota & regions and object storage uapps


seven-amid (approved):

kqesar (approved):

sachinrameshn (approved):

frenautvh:
rebased on master cab098fd8528a96a2ed0223ec439514c4f340c70

packages/manager/modules/manager-pci-common/src/api/data/regions.ts
```diff
@@ -21,3 +25,27 @@ export const getProjectRegions = async (
   });
   return data;
 };
+
+export const getS3StorageRegions = async (
+  regions: TRegion[],
+): Promise<TRegion[]> =>
+  regions.filter(({ services }) =>
+    services.find(({ name }) => S3_REGION_CAPACITY.includes(name)),
```
fredericvilcot, line 33: To respect SRP and inversion of dependencies principles, I'll suggest to limit API functions to retrieve only DTO and let hooks that contains business logic to handle computations.
By doing this, we make code reusable and not dependent from the UI

packages/manager/modules/manager-pci-common/src/api/data/regions.ts
```diff
@@ -21,3 +25,27 @@ export const getProjectRegions = async (
...
+  );
+
+export const getStorageRegions = async (
+  regions: TRegion[],
+): Promise<TRegion[]> =>
+  regions.filter(({ services }) =>
+    services.find(({ name }) => name === REGION_CAPACITY),
+  );
```
fredericvilcot, line 41: Same here :)

packages/manager/modules/manager-pci-common/src/api/hook/useRegions.ts
```diff
@@ -10,3 +15,53 @@ export const useGetProjectRegions = (projectId: string) =>
...
+export const useStorageRegions = (projectId: string) => {
+  const { data: regions } = useGetProjectRegions(projectId);
+
+  return useQuery({
+    queryKey: ['project', projectId, 'storages-regions'],
+    queryFn: () => getStorageRegions(regions || []),
+    retry: false,
+    enabled: !!regions,
+  });
+};
```
fredericvilcot, line 39: Could be a single parametrized function to handle the same code :)

packages/manager/modules/manager-pci-common/src/components/accordion/Accordion.component.tsx
```diff
@@ -13,46 +13,29 @@ export const AccordionComponent = ({
   title: ReactNode;
   children: ReactNode;
   isOpen?: boolean;
-  onToggle?: (isOpen: boolean) => void;
+  onToggle: (isOpen: boolean) => void;
```
fredericvilcot, line 16: As discussed, commit should be marked as a breaking change :)

packages/manager/modules/manager-pci-common/src/index.ts
```diff
@@ -8,8 +8,11 @@ export * from './components/region-selector';
 export * from './components/logs';
 export * from './components/modal';
 export * from './components/flavor-selector';
+export * from './components/tabs-panel';
```
fredericvilcot, line 11: I think you are missing changes from master ;)

packages/manager/modules/manager-pci-common/src/components/shape-input/ShapeInput.component.tsx
```diff
@@ -355,7 +358,13 @@ export const ShapesInputComponent = function ShapesInputComponent<T>({
...
                 />
               }
-              isOpen={!groupName}
+              isOpen={isAccordionOpen[groupName || 'none']}
+              onToggle={() => {
+                const name = groupName || 'none';
+                setIsAccordionOpen({
+                  [name]: !isAccordionOpen[name],
+                });
+              }}
```
fredericvilcot, line 367: Could be isolated into a `useCallbak` to prevent from re-computing

packages/manager/modules/manager-pci-common/src/components/shape-input/ShapeInput.component.tsx
```diff
@@ -117,6 +117,9 @@ export const ShapesInputComponent = function ShapesInputComponent<T>({
     value,
     group: undefined,
   });
+  const [isAccordionOpen, setIsAccordionOpen] = useState<{
+    [groupName: string]: boolean;
```
fredericvilcot, line 121: Could we use  instead an object with known properties to enhance typing ?
Because here I can add anything as a key

packages/manager/modules/manager-pci-common/src/components/rclone-download/StorageRegions.tsx
```diff
@@ -0,0 +1,58 @@
...
+    <>
+      <OdsSelect
+        value={currentRegion}
+        name="currentRegion"
+        data-testid="storageRegions_select"
+        onOdsChange={(event) => {
+          const { value } = event.detail;
+          setCurrentRegion(`${value}`);
+          onStorageRegionChange(`${value}`);
+        }}
```
fredericvilcot, line 45: Could be moved into a handler function wrapped into a `useCallback`

packages/manager/modules/manager-pci-common/src/components/rclone-download/S3StorageRegions.tsx
```diff
@@ -0,0 +1,49 @@
...
+  const { data: s3StorageRegions, isLoading } = useS3StorageRegions(projectId);
+
+  useEffect(() => {
+    if (s3StorageRegions?.length) {
+      setCurrentRegion(s3StorageRegions[0].name);
+      onS3StorageRegionChange(s3StorageRegions[0].name);
+    }
+  }, [s3StorageRegions]);
+
+  if (isLoading || !currentRegion) {
```
SimonChaumet, line 28: Testing on currentRegion will set a spinner if we don't have any region, is it possible ?

packages/manager/modules/manager-pci-common/src/components/rclone-download/StorageRegions.tsx
```diff
@@ -0,0 +1,58 @@
...
+import { OdsSelect, OdsSpinner, OdsText } from '@ovhcloud/ods-components/react';
+import { useStorageRegions } from '../../api/hook/useRegions';
+import { TRegion } from '../../api/data/regions';
+
+interface S3StorageRegionsProps {
+  projectId: string;
+  onStorageRegionChange: (value: string) => void;
+}
+
+export default function StorageRegions({
```
SimonChaumet, line 13: This seems a duplicate of S3StorageRegions

packages/manager/modules/manager-pci-common/src/components/tabs-panel/TabsPanel.component.tsx
```diff
@@ -0,0 +1,62 @@
...
+    } else {
+      clearNotifications();
+      setActivePanel(tabs[0].name);
+    }
+  }, [location.pathname]);
+
+  return (
+    <OdsTabs id="osds-tab-bar">
+      {tabs.map((tab: TabItemProps) => (
+        <React.Fragment key={`osds-tab-bar-item-${tab.name}`}>
```
SimonChaumet, line 41:
```suggestion
        <React.Fragment key={tab.name}>
```

packages/manager/modules/manager-pci-common/src/components/tiles-input-choice/TileChoice.component.tsx
```diff
@@ -0,0 +1,31 @@
...
+  selectedTileClass,
+  tileBorderClass,
+  tileDisabledClass,
+} from '../../constants/style';
+
+export interface TileChoiceProps {
+  isSelected?: boolean;
+  isDisabled?: boolean;
+}
+export function TileChoice({
```
SimonChaumet, line 14: I don't understand the purpose of this, could you rename it ?

packages/manager/modules/manager-pci-common/src/components/tiles-input-choice/TilesInputChoice.component.tsx
```diff
@@ -0,0 +1,59 @@
...
+export interface TileInputChoiceProps<T extends Item> {
+  items: T[];
+  selectedItem?: T;
+  onSelectItem?: (T: Item) => void;
+  columnsCount?: 1 | 2 | 3 | 4 | 5 | 6;
+  isSubmitted?: boolean;
+  children: (item: T, isSelected: boolean) => JSX.Element;
+}
+
+export function TileInputChoice<T extends Item>({
```
SimonChaumet, line 18: This is not specific to pci and should be moved to MRC

packages/manager/modules/manager-pci-common/src/components/tiles-input-choice/TilesInputChoice.component.tsx
```diff
@@ -0,0 +1,59 @@
...
+  isDisabled?: boolean;
+}
+
+export interface TileInputChoiceProps<T extends Item> {
+  items: T[];
+  selectedItem?: T;
+  onSelectItem?: (T: Item) => void;
+  columnsCount?: 1 | 2 | 3 | 4 | 5 | 6;
+  isSubmitted?: boolean;
+  children: (item: T, isSelected: boolean) => JSX.Element;
```
SimonChaumet, line 15: We should avoid children function when possible because it is not memoized

packages/manager/modules/manager-pci-common/src/translations/rclone-download/Messages_fr_CA.json
```diff
@@ -0,0 +1,14 @@
...
+  "pci_projects_project_users_download-rclone_cancel_label": "Annuler",
+  "pci_projects_project_users_download-rclone_content": "Vous allez télécharger un fichier de configuration permettant de vous connecter avec rclone.",
+  "pci_projects_project_users_download-rclone_more_link": "En savoir plus",
+  "pci_projects_project_users_download-rclone_region_label": "Sélectionner la région dans laquelle ce fichier s'appliquera :",
+  "pci_projects_project_users_download-rclone_region_label_global": "Sélectionner la région dans laquelle ce fichier s'appliquera (seules les régions proposant les services Object Storage et Cloud Archive sont listées) :",
+  "pci_projects_project_users_download-rclone_region_help": "Pour des raisons de sécurité le fichier de configuration ne contient pas votre mot de passe. Il est donc nécessaire d'éditer le fichier de configuration pour y ajouter manuellement votre mot de passe.",
+  "pci_projects_project_users_download-rclone_success_message": "Le téléchargement du fichier de configuration rclone a commencé, si ce n'est pas le cas ",
+  "pci_projects_project_users_download-rclone_success_message_link": "cliquez-ici",
```
SimonChaumet, line 11:
```suggestion
  "pci_projects_project_users_download-rclone_success_message_link": "cliquez ici",
```

packages/manager/modules/manager-pci-common/src/index.ts
frenautvh: i'm not sure to understand, this component has been added in this PR, what do you see missing ?

packages/manager/modules/manager-pci-common/src/components/shape-input/ShapeInput.component.tsx
frenautvh: since "name" comes from a map in the template i'm not sure it wouid be easiy doable, since it still need to be passed to the function dynamically, do you see another way ?

packages/manager/modules/manager-pci-common/src/components/shape-input/ShapeInput.component.tsx
frenautvh: the groupName can be any string so i'm not sure that it's possible to improve it, the component is pretty generic

packages/manager/modules/manager-pci-common/src/components/rclone-download/RCloneDownload.tsx
```diff
@@ -0,0 +1,183 @@
...
+  const rCloneGuideURL = RCLONE_GUIDE[ovhSubsidiary];
+
+  const { download, isLoading } = useDownloadRCloneConfig({
+    projectId: `${projectId}`,
+    userId,
+    onError: (error) => {
+      navigate(`..`);
+      addError(
+        <Translation ns="pci-rclone-download">
+          {(_t) =>
```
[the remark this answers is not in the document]
frenautvh: because the message is stored in notifications array until displayed, using t directly won't work lf the user changes his language, we tested it in uapps

packages/manager/modules/manager-pci-common/src/components/tiles-input-choice/TileChoice.component.tsx
frenautvh: Capture d’écran 2025-02-10 à 13 21 01

the purpose is a tile selector like the one in the screenshot (with generic tile content), which better name would you suggest ?

packages/manager/modules/manager-pci-common/src/components/tiles-input-choice/TilesInputChoice.component.tsx
frenautvh: i think it's specific to pci ui, other universes do not use this type of selector

packages/manager/modules/manager-pci-common/src/translations/rclone-download/Messages_fr_CA.json
frenautvh: it comes from angulars js : https://github.com/ovh/manager/blob/master/packages/manager/modules/pci/src/components/users/download-rclone/translations/Messages_fr_FR.json

we won't submit a new translation request here, if there is an issue with those translations used in production it can be handled in the scope of run

packages/manager/modules/manager-pci-common/src/components/rclone-download/StorageRegions.tsx
frenautvh: it looks like it is not strictly the same, the data source is different, one of the component is also displaying a specific text caption

packages/manager/modules/manager-pci-common/src/components/rclone-download/S3StorageRegions.tsx
frenautvh: i don't believe it's a case that should happen since storages have at least one region

packages/manager/modules/manager-pci-common/src/index.ts
fredericvilcot: I was referring to the export updates recently pushed to `master` (https://github.com/ovh/manager/commit/e3aaf63e420a096055c4ff2a457aa48380d93e18) but I see you are now up to date, everything's fine :)

packages/manager/modules/manager-pci-common/src/components/shape-input/ShapeInput.component.tsx
fredericvilcot: We can use currification to achieve that, but I see that it does not work with `useCallback`  as we can lose reference of the curried function at re-render :/
Maybe using `useMemo`, but to be tested :)

```ts
  const handleToggle = useMemo(
    () => (groupName: string) => () => {
      setIsAccordionOpen((prev) => ({
        [groupName || 'none']: !prev[groupName || 'none'],
      }));
    },
    [setIsAccordionOpen],
  );
```

And it's simply usage in jsx: 

```ts
 onToggle={handleToggle(groupName)}
```

packages/manager/modules/manager-pci-common/src/components/shape-input/ShapeInput.component.tsx
fredericvilcot: @frenautvh 
Why not creating an intersection of types ? 
I mean we can declare known keys as an union type and extend the object with opaque keys, what do you think?

frenautvh: i'm not sure of the exact behavior of this component as i'm not the author of it, from what i understand the ShapeInput acceps items which can be grouped or stacked using key of type string

these keys are unknown and can be any string, so i don't see how you can use a more specific type than just "string" unless you have an example proving me wrong :)

fredericvilcot: AFAIS, as we are using `none` property, it's a known one :)
So could be something like that :
```ts
type TFoo = { none: string } & { [key: string]: string };
```

packages/manager/modules/manager-pci-common/src/components/shape-input/ShapeInput.component.tsx
frenautvh: just added

```typescript
type TGroupName = 'none' | string;
```

please let me know if it's fine for you

packages/manager/modules/manager-pci-common/src/components/shape-input/ShapeInput.component.tsx
frenautvh: just added

```typescript
type TGroupName = 'none' | string;
```

please let me know if it's fine for you

fredericvilcot: Mmh I will keep the original answer as it's not the same ;)

```ts
type TGroupName = 'none' | string;
```

This means that `TGroupName` can be any string, but TypeScript does not treat 'none' as a special case. In fact, 'none' | string is equivalent to just string because string already includes all possible string values.

Doing this: 

```ts
type Foo = { none: boolean } & { [key: string]: boolean };
```

Ensures that 'none' is required while still allowing other keys. :)

frenautvh: fair enough i just updated this part

packages/manager/modules/manager-pci-common/src/components/tiles-input-choice/TileChoice.component.tsx
frenautvh: it contains all the specific behavior of pci tile selection such as a specific single tile display once the choice is submitted, it also handle the disability of tiles (both of thoses are currently not handled in mrc)

i think it will be nice to not have to style a mrc component each time you want to display a tile choice in pci 

this component is currently used in object storage uapp so removing it from this PR will also imply to having to refactor this uapp

fredericvilcot (approved):
LTGM, thanks :)
Please, @frenautvh may you squash the commits into a single one with breaking change notation ? (to be sure that we will not forget to mention it when merging on master and that we will tell Lerna how to bump the lib)

frenautvh:
> LTGM, thanks :) Please, @frenautvh may you squash the commits into a single one with breaking change notation ? (to be sure that we will not forget to mention it when merging on master and that we will tell Lerna how to bump the lib)

@fredericvilcot just squashed into a single commit as requested

packages/manager/modules/manager-pci-common/src/components/rclone-download/StorageRegions.tsx
SimonChaumet: Could you factorize these ?

SimonChaumet (approved):

kqesar (approved):

seven-amid (approved):

PR #15261, opened 2025-02-03
fix(mrc): dont render TileBlock optional label

### LINKED ISSUE #15260: [manager-components]: <TileBlock /> label should not render if empty

TileBlock

### Description

Due to recent changes, right now the `TileBlock` label props is optional but it still renders if it is empty (at least on Firefox)

### Steps to reproduce

Declare a `<DashboardTile />` with some `items` without a `label` property

Image

We want to be able to not render the `label` element if it is empty

Image

### Browser

_No response_

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

### Additional Information

_No response_

tristanwagner:
ref: ISSUE-15260

| Question         | Answer
| ---------------- | ---
| Bug fix?         | yes
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #15260  

- [x] Commits are signed-off
- [x] Lint has passed locally

### DIFF
--- a/packages/manager-react-components/src/components/content/dashboard-tile/tile-block.component.tsx
+++ b/packages/manager-react-components/src/components/content/dashboard-tile/tile-block.component.tsx
@@ -12,7 +12,7 @@ export const TileBlock: React.FC<React.PropsWithChildren<TileBlockProps>> = ({
 }) => (
   <dl className="flex flex-col gap-1 m-0">
     <dt>
-      <OdsText preset={ODS_TEXT_PRESET.heading6}>{label}</OdsText>
+      {!!label && <OdsText preset={ODS_TEXT_PRESET.heading6}>{label}</OdsText>}
     </dt>
     <dd className="m-0">{children}</dd>
   </dl>
--- a/packages/manager-react-components/src/components/content/dashboard-tile/dashboard-tile.stories.tsx
+++ b/packages/manager-react-components/src/components/content/dashboard-tile/dashboard-tile.stories.tsx
@@ -59,6 +59,10 @@ const items: DashboardTileBlockItem[] = [
       </div>
     ),
   },
+  {
+    id: 'without-label',
+    value: <OdsText preset="span">Without label</OdsText>,
+  },
 ];
 
 export const WithTitle = () => <DashboardTile title="Title" items={items} />;


tibs245 (dismissed):

ghyenne (dismissed):

packages/manager-react-components/src/components/content/dashboard-tile/tile-block.component.tsx
```diff
@@ -11,9 +11,11 @@ export const TileBlock: React.FC<React.PropsWithChildren<TileBlockProps>> = ({
...
-    <dt>
-      <OdsText preset={ODS_TEXT_PRESET.heading6}>{label}</OdsText>
-    </dt>
+    {label && (
+      <dt>
+        <OdsText preset={ODS_TEXT_PRESET.heading6}>{label}</OdsText>
+      </dt>
+    )}
     <dd className="m-0">{children}</dd>
   </dl>
```
anooparveti, line 20: **nitpick:** By definition, `dd` tag provides the description for the preciding term `dt`.
If you have a requirement to not display `dt`, then imo, we have to think of adding modes for this where you display datalist/text/link.

anooparveti (changes requested):

packages/manager-react-components/src/components/content/dashboard-tile/tile-block.component.tsx
```diff
@@ -11,9 +11,11 @@ export const TileBlock: React.FC<React.PropsWithChildren<TileBlockProps>> = ({
   children,
 }) => (
   <dl className="flex flex-col gap-1 m-0">
-    <dt>
-      <OdsText preset={ODS_TEXT_PRESET.heading6}>{label}</OdsText>
-    </dt>
+    {label && (
```
SimonChaumet, line 14:
```suggestion
    {!!label && (
```
If label is a string it will display both

tristanwagner: I agree it's better that the boolean check is explicit, however I'm pretty sure it won't have this behavior you describe. Anyway I updated it !

packages/manager-react-components/src/components/content/dashboard-tile/tile-block.component.tsx
tristanwagner: I updated as we said earlier, keeping the `dt` anyway but not rendering the `OdsText` since it is the part that cause the unwanted space

rjamet-ovh: For actions items, it will be better to not use dl/dt/dd, and render something specific. (probably a div with margin children)

anooparveti (approved):

aboungnaseng-ovhcloud (approved):

PR #15475, opened 2025-02-14
feat: add product names translation for LDP integrations

pfrayer:
## Description

With a feature named "Logs to Customers", we are integrating Logs Data Platform inside each OVHcloud product. Somewhere in the LDP Manager, we display products for which customers have subscribed to logs forwarding (between a product & LDP).

We need to nicely display product names, and some are missing. This PR add the needed translations.

(I don't know if this needs to go to the translation team or if I can update all languages files, as these trads are about products names)


pfrayer:
Hello, any news here ?

PR #15478, opened 2025-02-14
feat(container): add zimbra to web side bar order

tristanwagner:
ref: MANAGER-17092

## Description

Adding the "Zimbra Mail" entry to the sidebar order of the WebCloud universe

Screenshot 2025-02-14 at 14 44 50

Ticket Reference: #MANAGER-17092


packages/manager/apps/container/src/container/legacy/server-sidebar/order/shop-config/web.ts
```diff
@@ -49,6 +50,15 @@ const webShopConfig = (
         tracking: 'web::orders::cloud-web::order',
       }
     : null,
+  features['zimbra'] && ORDER_URLS[region].zimbra
```
JacquesLarique, line 53: `ORDER_URLS[region].zimbra` this will crash for any subsidiary not listed in the constant file (for ex: US)

I see that FF is available only for EU, but if we open it more we will need to synchronize the FF deployment with the Manager deployment
Changing to `ORDER_URLS[region]?.zimbra` would give us more leeway

tristanwagner: I think it will crash for any region not listed in the `ORDER_URLS` constant which is not very likely to happened, but better safe than sorry :)

JacquesLarique (approved):

SimonChaumet (approved):

JacquesLarique (approved):

SimonChaumet (approved):

Issue #14962, opened 2025-01-16
[manager-components]: Datagrid - Accordion rows


qpavy:
### Have you already checked if a similar item is present on manager-components?

- [x] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

We have the need in Ip project to display child ips of ip block within the datagrid. To do that, we should click on an expand button at the beginning of the row to load children of the ip block (with an asynchronous call) and then display it as "sub rows" of the selected ip block:

example:

![Image](https://github.com/user-attachments/assets/e1cc6612-a6fa-48e7-a311-095dc9c9247a)

![Image](https://github.com/user-attachments/assets/e618de24-43ac-464f-9004-26e71b62c81f)

tanstack table expanding feature seems to correspond to this need.

### Where do you expect to use this?

ips uapp

### Additional Information

For now, the child rows have the same columns as the parent row.

aboungnaseng-ovhcloud:
We have found an exmeple in tanstack query https://tanstack.com/table/latest/docs/framework/react/examples/expanding

aboungnaseng-ovhcloud:
We are working on the technical and functional specs for this

aboungnaseng-ovhcloud:
Aurore from ods team will write the specs
Go from Francois to start the dev 👍

aboungnaseng-ovhcloud:
Atfer a meeting with quentin and Anoop, we take the decision to use subComponent

https://tanstack.com/table/latest/docs/framework/react/examples/sub-components

So you need to update add an attribute in the datagrid

renderSubComponent: (props: { row: Row<TData> }) => React.ReactElement

maybe this one too =
getRowCanExpand={() => true}

```
    <Datagrid
      data={data}
      columns={columns}

      getRowCanExpand={() => true}
      renderSubComponent={renderSubComponent}
    />
```

PR #15494, opened 2025-02-17
feat(mrc): datagrid - add sub component feature

qpavy:
## Description
Add props on MRC Datagrid to use tanstack table sub components.

Ticket Reference: #[14962](https://github.com/ovh/manager/issues/14962)    

## Additional Information

If nether getRowCanExpand and renderSubComponent is set, do not display the expand action column.
If getRowCanExpand and renderSubComponent are set and getRowCanExpand return true, an OdsButton is displayed to toggle the expandable row.
renderSubComponent can accept any React component and get row as param.

image
image

### DIFF
--- a/packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
+++ b/packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
@@ -1,4 +1,4 @@
-import React, { useMemo } from 'react';
+import React, { Fragment, useMemo } from 'react';
 import { useTranslation } from 'react-i18next';
 import {
   ColumnDef,
@@ -8,8 +8,13 @@ import {
   getCoreRowModel,
   useReactTable,
   getSortedRowModel,
+  Row,
 } from '@tanstack/react-table';
-import { ODS_ICON_NAME, ODS_BUTTON_VARIANT } from '@ovhcloud/ods-components';
+import {
+  ODS_ICON_NAME,
+  ODS_BUTTON_VARIANT,
+  ODS_BUTTON_SIZE,
+} from '@ovhcloud/ods-components';
 import {
   OdsButton,
   OdsIcon,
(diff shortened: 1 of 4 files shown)


packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
```diff
@@ -33,6 +34,7 @@ import { FilterWithLabel } from '../filters/interface';
 import { DataGridTextCell } from './text-cell.component';
 import { defaultNumberOfLoadingRows } from './datagrid.contants';
 import './translations';
+import cn from 'classnames';
```
aboungnaseng-ovhcloud, line 37:
```suggestion
import cn from 'classnames';
```

Can you use 'clsx' as Its already installed ?

aboungnaseng-ovhcloud (dismissed):

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
```diff
@@ -273,24 +310,38 @@ export const Datagrid = <T,>({
...
+                  </tr>
+                  {row.getIsExpanded() && (
+                    <tr>
+                      {/* 2nd row is a custom 1 cell row */}
+                      <td colSpan={row.getVisibleCells().length}>
+                        {renderSubComponent?.(row) || (
+                          <DataGridTextCell>
+                            {noResultLabel ?? t('common_pagination_no_results')}
+                          </DataGridTextCell>
+                        )}
```
anooparveti, line 340: **question:** Are you sure we can use `common_pagination_no_results` to display if `renderSubComponent` does not return anything?

qpavy: indeed we don't even need this because we should have renderSubcomponent define to be able to expand row. It is just in case

anooparveti: I would prefer having an empty cell `<></>` over this as we cannot be sure of what kind of items can be displayed.

qpavy: Ok I will refactor this part

anooparveti (commented):
I think it's beneficial to add other APIs like `getIsExpanded`, `getCanExpand`,  `onExpandedChange`.

anooparveti (approved):

aboungnaseng-ovhcloud (approved):

PR #15508, opened 2025-02-17
fix(manager-pci-common): rclone download modal named export

frenautvh:
## Description

fix missing named export for rclone download modal 

Ticket Reference: DTCORE-3108

### DIFF
--- a/packages/manager/modules/manager-pci-common/src/components/rclone-download/RCloneDownloadModal.tsx
+++ b/packages/manager/modules/manager-pci-common/src/components/rclone-download/RCloneDownloadModal.tsx
@@ -35,7 +35,7 @@ import '../../index.css';
 
 type RCloneDownloadModalProps = { userId: string };
 
-export default function RCloneDownloadModal({
+export function RCloneDownloadModal({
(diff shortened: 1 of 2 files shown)


kqesar (dismissed):

kqesar (approved):

darsene (approved):

SimonChaumet (commented):
Why not keep both exports ? This would create a breaking change that might not be wanted

kb-ovh (approved):

frenautvh:
> Why not keep both exports ? This would create a breaking change that might not be wanted

keeping the default export won't change anything from the perspective of a pci-common user because you will only be able to import it using a named import :

```
import { RCloneDownloadModal } from "@ovh-ux/manager-pci-common";
```

SimonChaumet (approved):

anooparveti (approved):

anooparveti:
In the absence of @fredericvilcot, considering the approvals from TAPC as valid `CODEOWNERS` and merging the PR with bypass.

PR #15530, opened 2025-02-18
feat(hub): refactor order api hooks to optimize call counts

aottr:
## Description
The request was to optimize the api calls to `/lastOrder`. The old hook for the order tracking aggregated multiple awaited api calls into one query function. 

I therefore split the hook into dedicated hooks to have each sub-result in the query cache. This made it possible to fetch the `/lastOrder` only once and re-use the cached result.

Ticket Reference: MANAGER-16593


packages/manager/apps/hub-react/src/components/hub-support/HubSupport.component.tsx
```diff
@@ -64,7 +64,7 @@ export default function HubSupport() {
               refetch={refetch}
             />
           )}
-          {!error && data.count <= 0 && <HubSupportHelp />}
+          {!error && (!data.count || data.count <= 0) && <HubSupportHelp />}
           {!error && data.count > 0 && (
```
JacquesLarique, line 68: Shouldn't we do the same for line 68: `{!error && data?.count > 0 && (` ?

aottr: good point, it didn't trigger again since the error was set in all tests after (which is still weird) But I'll set it just to be sure (and since it's not an else)

JacquesLarique (approved):

MaximeBajeux (approved):

PR #15547, opened 2025-02-19
fix(svp): add exact search on product code

lionel95200x:
## Description

- Add exact product search on commercial catalog

Ticket Reference: #TAPC-2970


darsene (changes requested):
Ref is missing in the commit message and it's not signed

darsene (approved):

SimonChaumet (approved):

PR #15558, opened 2025-02-20
Feat/pci instances hybrid

SimonChaumet:
## Description

Ticket Reference: #...


anooparveti (changes requested):
You need to add the new prop `hashes` in the type definition. https://github.com/ovh/manager/blob/master/packages/manager/modules/config/src/application/index.ts#L3-L9

SimonChaumet:
Code still needs to be spitted but PR is open for comments for pci

PR #15612, opened 2025-02-21
feat(container): replace legacy navbar spinner with skeletons

JacquesLarique:
## Description

Replace the spinner displayed while we fecth necessary data to display the menu by a few skeletons
Fixed the retrieval of KYC India and Fraud procedure state to display their respective menu entry

Ticket Reference: MANAGER-15634

### DIFF
--- a/packages/manager/apps/container/src/api/procedure/procedure.constants.ts
+++ b/packages/manager/apps/container/src/api/procedure/procedure.constants.ts
@@ -1,0 +1,6 @@
+import { Procedures } from '@/types/procedure';
+
+export const PROCEDURE_API_PATH_MAP = {
+  [Procedures.INDIA]: 'identity',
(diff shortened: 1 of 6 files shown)


MaximeBajeux (dismissed):

anooparveti (dismissed):

anooparveti (commented):
**perf:** On the initial load, there are layout shifts with respect to the display of icons. So, I believe there is scope for improvement on the skeletons. Below techniques could be more handy,
1. Set the height of the skeleton same as the height of the icons displayed.
2. Evolve the skeleton structure to display skeleton for icons + skeleton for text.

In terms of CLS scoring, both might have the same impact but 2/ is more close towards end layout.

oalkabouss (approved):

aottr (approved):

Issue #13543, opened 2024-10-10
[manager-components]: refactor useIceberg hooks


tristanwagner:
### Have you already checked if a similar item is present on manager-components?

- [X] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Hooks `useIcebergV2`, `useIcebergV6`, `useResourcesV6`..

### Description

Right now I feel like these hooks that extends the `useInfiniteQuery` calls provide not enough value and also are very restraining for the end user of the hook, I propose that we update the way of exposing this kind of hooks in a more generic and versatile way, that will allow developers to be able to use more of the `react-query` features for their usecases.

To illustrate by an example, to me an updated version of the hook would look like that:

```
export function useResourcesIcebergV2<T = unknown>({
  route,
  pageSize = defaultPageSize,
  ...options
}: IcebergFetchParamsV2 & IcebergV2Hook) {
  return useInfiniteQuery({
    initialPageParam: null,
    queryFn: ({ pageParam }) =>
      fetchIcebergV2<T>({ route, pageSize, cursor: pageParam }),
    getNextPageParam: (lastPage) => lastPage.cursorNext,
    select: (data) => { ...data, flattenData},
    ...options,
  });
}
```

If you really want to flatten the data in there (to me that should not be the place where you should do this but maybe I'm wrong) then you should do it using the `select` option from `useInfiniteQuery` and avoid at all cost using a `useState` in this place.

I want to add that the way the cursor are designed in the APIs, if you use `refetch` or like `refetchInterval` feature, when refetching the data, a new cursor will be provided by the API even tho your actual response payload didn't changed so your `data` object returned will have changed since it contains `pageParams` that will contain the current cursor, causing unwanted rerenders in your UI, you can avoid this by omitting the cursor in the `data.pageParams` using `select`.

To me it also applies to any other hook that will wrap a `useQuery` or `useInfiniteQuery` call: you should be able to pass additional options and also retrieve the returns from the call.

### Where do you expect to use this?

any

### When do you expect this to be delivered?

ASAP

tristanwagner:
linked to #13513 #13331

anooparveti:
> The change is no longer needed

From the requester. @pauldkn

pauldkn:
> > The change is no longer needed
> 
> From the requester. @pauldkn

@anooparveti the #13513 is no longer needed because this one exist and cover a more complete scope on this hook. 
I closed the smallest one but this one should still be open

[...]

PR #15613, opened 2025-02-21
feat(mrc): add useIcebergV2InfiniteQuery

tristanwagner:
ref: ISSUE-13543

## Description

Ticket Reference: #13543

### DIFF
--- a/packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx
+++ b/packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx
@@ -1,8 +1,19 @@
 import { useEffect, useState } from 'react';
-import { IcebergFetchParamsV2, fetchIcebergV2 } from '@ovh-ux/manager-core-api';
-import { useInfiniteQuery } from '@tanstack/react-query';
+import {
+  IcebergFetchParamsV2,
+  IcebergFetchResultV2,
+  fetchIcebergV2,
+} from '@ovh-ux/manager-core-api';
+import {
+  InfiniteData,
+  useInfiniteQuery,
+  UseInfiniteQueryOptions,
+  UseInfiniteQueryResult,
+} from '@tanstack/react-query';
 import { defaultPageSize } from './index';
 import { ColumnSort } from '../../components';
+
+export const API_V2_MAX_PAGE_SIZE = 9999;
 
 interface IcebergV2Hook<T> {
   queryKey: string[];
@@ -11,13 +22,14 @@ interface IcebergV2Hook<T> {
   shouldFetchAll?: boolean;
 }
 
-export const API_V2_MAX_PAGE_SIZE = 9999;
-
 /**
  * @deprecated use fetchIcebergV2 from @ovh-ux/manager-core-api
  */
 export const getResourcesIcebergV2 = fetchIcebergV2;
 
+/**
+ * @deprecated use useIcebergV2InfiniteQuery
+ */
 export function useResourcesIcebergV2<T = unknown>({
   route,
   pageSize = defaultPageSize,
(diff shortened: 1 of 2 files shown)


packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx
```diff
@@ -72,3 +84,53 @@ export function useResourcesIcebergV2<T = unknown>({
...
+    queryKey: [
+      ...queryKey,
+      shouldFetchAll ? 'all' : '',
+      search ? `${search.key}=${search.value}` : '',
+    ].filter(Boolean),
+    queryFn: ({ pageParam }) =>
+      fetchIcebergV2<T[]>({
+        pageSize:
+          pageSize || (shouldFetchAll ? API_V2_MAX_PAGE_SIZE : defaultPageSize),
+        cursor: pageParam as string,
```
SimonChaumet, line 120: if this isn't a string it should be called with String(pageParam)

packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx
```diff
@@ -72,3 +84,53 @@ export function useResourcesIcebergV2<T = unknown>({
...
+  const query = useInfiniteQuery({
+    // return only flattened data without cursor by default to avoid
+    // unwanted rerenders on refetch if data is the same
+    select: (data) =>
+      data?.pages?.flatMap((page: UseInfiniteQueryResult<T[]>) => page?.data),
+    ...queryOptions,
+    initialPageParam: null,
+    queryKey: [
+      ...queryKey,
+      shouldFetchAll ? 'all' : '',
```
SimonChaumet, line 113:
```suggestion
      shouldFetchAll,
```

packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx
```diff
@@ -72,3 +84,53 @@ export function useResourcesIcebergV2<T = unknown>({
...
+  > & {
+    select?: (data: InfiniteData<T, unknown>) => T[];
+  };
+  shouldFetchAll?: boolean;
+};
+
+export function useIcebergV2InfiniteQuery<T>({
+  iceberg: { pageSize, search, ...icebergOptions },
+  query: { queryKey, ...queryOptions },
+  shouldFetchAll = false,
```
SimonChaumet, line 102: I think this should be a new function since it is not used the same way : on previous function you needed to fetch next page by yourselves but with this it isn't required

packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx
```diff
@@ -72,3 +84,53 @@ export function useResourcesIcebergV2<T = unknown>({
...
+    queryFn: ({ pageParam }) =>
+      fetchIcebergV2<T[]>({
+        pageSize:
+          pageSize || (shouldFetchAll ? API_V2_MAX_PAGE_SIZE : defaultPageSize),
+        cursor: pageParam as string,
+        search,
+        ...icebergOptions,
+      }),
+    getNextPageParam: (lastPage) =>
+      (lastPage as IcebergFetchResultV2<T[]>).cursorNext,
```
SimonChaumet, line 125: Why cast ?

packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx
```diff
@@ -72,3 +84,53 @@ export function useResourcesIcebergV2<T = unknown>({
...
+  });
+
+  useEffect(() => {
+    // fetch next page if needed
+    if (shouldFetchAll && query.hasNextPage && !query.isFetchingNextPage) {
+      query.fetchNextPage();
+    }
+  }, [query.data]);
```
SimonChaumet, line 133: On first call you already know the size of your request so you could request them all instead of fetching page by page

packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx
```diff
@@ -72,3 +84,53 @@ export function useResourcesIcebergV2<T = unknown>({
...
+
+export function useIcebergV2InfiniteQuery<T>({
+  iceberg: { pageSize, search, ...icebergOptions },
+  query: { queryKey, ...queryOptions },
+  shouldFetchAll = false,
+}: UseIcebergV2InfiniteQueryParams<T>) {
+  const query = useInfiniteQuery({
+    // return only flattened data without cursor by default to avoid
+    // unwanted rerenders on refetch if data is the same
+    select: (data) =>
```
aboungnaseng-ovhcloud, line 107: can we just put this 'select' attribute on 'useResourcesIcebergV2' without duplicating this ?

And adding the params 'icebergOptions' and 'queryOptions' as optional ?

packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx
tristanwagner: for APIV2 we don't know the size

packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx
tristanwagner: in my opinion it's better to just deprecate the old one and have this new one, for a few reasons, the main one being that the params changed a lot, some were moved and some removed,  i'd rather not mix them, for example `route` would be outside of `iceberg` object which would be kinda confusing, `flattenData` was completely removed since it is returned by default already flatten in `data`, etc..

packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx
tristanwagner: I'm not sure to get the point here, this same functionality was added a few weeks back on the previous hook, it's just an helper to make it easy to fetch all the pages, of course now you will be able to handle it yourself but it's practical to have this imo

packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx
tristanwagner: thinking about it, to me this hook should not be in datagrid folder because you can have some use cases where you want to fetch data from api using useInfiniteQuery without displaying it in a datagrid, I would name it `packages/manager-react-components/src/hooks/icebergV2/useIcebergV2.tsx`

packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx
```diff
@@ -1,23 +1,35 @@
...
-export const API_V2_MAX_PAGE_SIZE = 9999;
-
 /**
  * @deprecated use fetchIcebergV2 from @ovh-ux/manager-core-api
  */
 export const getResourcesIcebergV2 = fetchIcebergV2;
 
+/**
+ * @deprecated use useIcebergV2InfiniteQuery
+ */
```
anooparveti, line 32: **nitpick:** I dont think adding `Infinite` is required here. And moreover, this hook is not based on Iceberg but using PU's pagination. So, in future, this hook will be replaced with 2 new hooks `useResourcesV2` and `useIcebergV2`.

tristanwagner:
moved to #16212

PR #15640, opened 2025-02-24
refactor(pci): move unused calls to children

SimonChaumet:
## Description

Ticket Reference: #...


SimonChaumet:
Will not be treated

PR #15641, opened 2025-02-24
test(container): add test for pnr header component(s)

aottr:
## Description

Add tests for header component in PNR (beta) navigation.
It consists of many hook-mocks, since the whole component needed to be rendered for the first test (menu) and the sub-components consist of many dependencies.

Tests consist of (non-exhaustive):
- display of hamburger menu
- format of the username in navbar, depending on individual or corp-user
- available links depending on feature availability
- link-rewrites depending on feature availability
- available links depending on region
- aria-labels for links (non-Chip)
- aria-labels for user menu button
- format of username (and email) in user-menu
- visibility of payment method badge (enterprise/individual)

Ticket Reference: MANAGER-16558

## Additional Information
Changed bdd test workflow to build the the `@ovh-ux/ovh-payment-method` package, which is used in the `UserAccountMenu` component

### DIFF
--- a/packages/manager/apps/container/src/container/nav-reshuffle/header/Header.spec.tsx
+++ b/packages/manager/apps/container/src/container/nav-reshuffle/header/Header.spec.tsx
@@ -1,0 +1,369 @@
+import { render, screen, waitFor, act } from '@testing-library/react';
+import HeaderComponent from './index';
+import UserAccountMenu from './user-account-menu/Content';
+import '@testing-library/jest-dom';
+import { vi } from 'vitest';
+import { KeyPairName } from '@ovh-ux/manager-config';
+import { LEGAL_FORMS } from '@/container/common/constants';
+import { fetchFeatureAvailabilityData } from '@ovh-ux/manager-react-components';
+import { useReket } from '@ovh-ux/ovh-reket';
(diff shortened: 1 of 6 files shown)


packages/manager/apps/container/src/container/nav-reshuffle/header/Header.spec.tsx
```diff
@@ -0,0 +1,362 @@
...
+    });
+
+    // check for logo
+    expect(
+      screen.queryByTitle('OVHcloud_master_logo_white_RGB'),
+    ).toBeInTheDocument();
+  });
+
+  it.each([
+    ['n individual', LEGAL_FORMS.INDIVIDUAL],
```
MaximeBajeux, line 149: That not break the test but that should not do just 'individual' ?

MaximeBajeux (approved):

packages/manager/apps/container/src/container/nav-reshuffle/header/Header.spec.tsx
aottr: the first item is just for the name of the test case. Since individual starts with an (i) it is an individual :)

PR #15651, opened 2025-02-25
fix: display success mesage after zerto activation

sefiani-ovh:
ref : MANAGER-17096
**Description:** : Display a success message & PO after activation of Zerto


tibs245 (changes requested):
Commit message must be `fix(dedicated): ...`

(dedicated) is the context

darsene (approved):

tibs245 (approved):

Issue #15421, opened 2025-02-12
[manager-components]: Unable to translate a link with <Trans> component


darsene:
### On which entity, you would like to report a bug?

packages/manager-react-components/src/components/order

### Description

According to https://react.i18next.com/latest/trans-component, a link label can be set by using trans component.
However, the OdsLink component (ODS 18) doesn't accept the label value as children prop and the translation tool doesn't allow to translate prop of a component (https://github.com/ovh/manager/pull/15414/files, see detail in CMT-21344).
Therefore, which workaround can be used ?

### Steps to reproduce

https://github.com/ovh/manager/pull/15414

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

tristanwagner:
As I suggested I would try

Translation file
```
"your_key": "Some text with a <Link /> that can move depending on the language",
"your_key_link_label": "labeled link",
```

In your component
```
<Trans
  i18nKey="myKey"
  components={{ 
    Link: <OdsLink href={href} label={t('your_key_link_label)}/> 
    }}
/>
```

This way the translation team will be able to move the link depending on the language, the downside is that it is hard for them to have the full context of the translation since the label of the link is separated from the "main" text

darsene:
Fixes in #15660

@tristanwagner @anooparveti

PR #15414, opened 2025-02-11
fix(i18n): add missing translations [CDS 3713]

darsene:
| Question         | Answer
| ---------------- | ---
| Branch?          |  `develop` 
| Bug fix?         | no
| New feature?     | no
| Breaking change? | no
| Tickets          | Fix #MANAGER-16647
| License          | BSD 3-Clause

- [x] Try to keep pull requests small so they can be easily reviewed.
- [ ] Commits are signed-off
- [ ] Only FR translations have been updated
- [x] Branch is up-to-date with target branch
- [x] Lint has passed locally
- [ ] Standalone app was ran and tested locally
- [ ] Ticket reference is mentioned in linked commits (internal only)
- [ ] Breaking change is mentioned in relevant commits

PR #15660, opened 2025-02-25
fix(mrc): fix ignore translation in component property

darsene:
fixes: gh-15421

## Description

Ticket Reference: #gh-15421

## Additional Information

Fixes #15414 where translations have been ignored

:flags: Translation
- [x] CMT-22293

### DIFF
--- a/packages/manager-react-components/src/components/typography/links/links.component.tsx
+++ b/packages/manager-react-components/src/components/typography/links/links.component.tsx
@@ -21,6 +21,7 @@ export interface LinksProps {
   color?: ODS_LINK_COLOR;
   download?: string;
   label?: string;
+  children?: string;
   href?: string;
   rel?: string;
   target?: string;
@@ -31,6 +32,7 @@ export interface LinksProps {
 }
 
 export const Links: React.FC<LinksProps> = ({
+  children,
   label,
   onClickReturn,
   type,
@@ -39,7 +41,7 @@ export const Links: React.FC<LinksProps> = ({
   iconAlignment,
   className = '',
   ...props
-}: LinksProps) => (
+}) => (
   <OdsLink
     className={className}
     href={href}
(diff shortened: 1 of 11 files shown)


tristanwagner (approved):

anooparveti (dismissed):

tristanwagner (dismissed):

packages/manager-react-components/src/components/typography/links/TransExternalLink.component.tsx
anooparveti, line 1: **question:** Why add a new component when there is a link component already?
It can be enhanced to accept children and display `label || children`
And moreover, we should not create components like `ExternalLink` and `TransExternalLink` for with just added capability to handle trans. That's not good to maintain so many components.

anooparveti: I mean you can directly enhance Links component and use it in OrderSummary component. Creating these wrappers does not add much value imo.

packages/manager-react-components/src/components/typography/links/links.component.tsx
```diff
@@ -55,7 +56,7 @@ export const Links: React.FC<LinksProps> = ({
     })}
     {...(type === LinkType.next && { icon: ODS_ICON_NAME.arrowRight })}
     {...(type === LinkType.external && { icon: ODS_ICON_NAME.externalLink })}
-    label={label}
+    label={label ?? children?.toString()}
```
fredericvilcot, line 59: Applying `toString()` method to an object (React Element) will return [object, object]

darsene: fixed

packages/manager-react-components/src/components/typography/links/TransExternalLink.component.tsx
darsene: Fixed

fredericvilcot (dismissed):

SimonChaumet (dismissed):

anooparveti (approved):

anooparveti (approved):

PR #15665, opened 2025-02-26
feat(ips): remove 2api calls

chipp972:
ref: MANAGER-15789

## Description

Ticket Reference: #...


packages/manager/apps/ips/src/data/api/dedicated-server.ts
```diff
@@ -30,6 +36,21 @@ export type DedicatedServer = {
   iam: IamObject;
 };
 
+export const getDedicatedServerList = async (): Promise<IcebergFetchResultV6<{
+  serviceName: string;
+  displayName: string;
+}>> => {
+  const response = await fetchIcebergV6<DedicatedServer>({
+    route: '/dedicated/server',
+    pageSize: 1000,
```
qpavy, line 45: We will have issue for account with more than 1000 services

qpavy (approved):
I approve for now but we will have to check after ODS combo box release if we can change the service selection to make it more performant

packages/manager/apps/ips/src/data/api/dedicated-server.ts
chipp972: Yes but I think the select will be unusable if there is 4000 services to scroll to also ^^'

PR #15675, opened 2025-02-26
feat(mrc): improve datagrid in datagrid with col size

qpavy:
## Description

add some improvement to handle datagrid inside datagrid with sub component by giving size of parent datagrid headers to child datagrid in order to define sub-datagrid columns sizes:

image

Why not use expanding instead of subcomponents ?

-> To be able to use expanding feature from tanstack table. The cell definition from sub rows should be the same as for main rows. But in some cases, even if column names are same, the way the data is retrieve is different between parent and sub row. For that use case, we are not able to use expanding feature from tanstack table.

Ticket Reference: #...

### DIFF
--- a/packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
+++ b/packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
@@ -1,4 +1,4 @@
-import React, { Fragment, useMemo } from 'react';
+import React, { Fragment, useEffect, useMemo, useRef } from 'react';
 import { useTranslation } from 'react-i18next';
 import {
   ColumnDef,
@@ -59,6 +59,8 @@ export interface DatagridColumn<T> {
   isFilterable?: boolean;
   /** Trigger the column search */
   isSearchable?: boolean;
+  /** Set default column size */
+  size?: number;
 }
 
 type ColumnFilterProps = {
@@ -122,9 +124,18 @@ export interface DatagridProps<T> {
   /** Add react element at left in the datagrid topbar */
   topbar?: React.ReactNode;
   /** Function to render sub component as row child */
-  renderSubComponent?: (row: Row<T>) => JSX.Element;
+  renderSubComponent?: (
+    row: Row<T>,
+    headerRefs?: React.MutableRefObject<Record<string, HTMLTableCellElement>>,
+  ) => JSX.Element;
   /** function to define if row can be expanded or not */
   getRowCanExpand?: (row: Row<T>) => boolean;
+  /** Hide datagrid header if true */
+  hideHeader?: boolean;
+  /** Resets the expanded rows when data is updated */
+  resetExpandedRowsOnItemsChange?: boolean;
+  /** When true, will fix the columns size by column definition size */
+  tableLayoutFixed?: boolean;
 }
 
 export const Datagrid = <T,>({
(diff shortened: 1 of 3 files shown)


packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
```diff
@@ -122,9 +124,18 @@ export interface DatagridProps<T> {
...
-  renderSubComponent?: (row: Row<T>) => JSX.Element;
+  renderSubComponent?: (
+    row: Row<T>,
+    headerRefs?: React.MutableRefObject<Record<string, unknown>>,
+  ) => JSX.Element;
   /** function to define if row can be expanded or not */
   getRowCanExpand?: (row: Row<T>) => boolean;
+  /** Hide datagrid header if true */
+  hideHeader?: boolean;
+  /** Function datagrid will perform when items change */
```
anooparveti, line 135:
```suggestion
  /** Resets the expanded rows when data is updated */
```

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
```diff
@@ -220,6 +237,10 @@ export const Datagrid = <T,>({
     }),
   });
 
+  useEffect(() => {
+    resetExpandedRowsOnItemsChange && table.resetExpanded();
```
anooparveti, line 241: > Expected an assignment or function call and instead saw an expression

Can you fix this lint error?

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
```diff
@@ -254,60 +275,69 @@ export const Datagrid = <T,>({
...
+            }
+          >
+            {!hideHeader && (
+              <thead>
+                {table.getHeaderGroups().map((headerGroup) => (
+                  <tr key={headerGroup.id}>
+                    {headerGroup.headers.map((header) => (
+                      <th
+                        key={header.id}
+                        ref={(el) => (headerRefs.current[header.id] = el)}
```
anooparveti, line 291: > Arrow function should not return assignment

Can you fix the lint error?

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
```diff
@@ -254,60 +275,69 @@ export const Datagrid = <T,>({
...
-                                header.getContext(),
-                              )}
-                            </>
-                          </span>
-                          <span
-                            className={`align-middle inline-block h-4 -mt-6`}
+          <table
+            className="w-full border-collapse"
+            style={
+              { '--expander-column-width': '2.5rem' } as React.CSSProperties
```
anooparveti, line 281: **question:** Do we have this design guideline from design team?

anooparveti (changes requested):

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
qpavy: Nop I just check what was the size from the icon to have the smallest column possible

anooparveti: We have to get the confirmation then.

qpavy: Just had a call with ux and 2.5rem has been validated as a size for the expander column

aboungnaseng-ovhcloud (approved):

PR #15677, opened 2025-02-26
fix(pci-object-storage): add ods modal overflow hack

frenautvh:
## Description

add ods modal overflow hack in order to temporarily fix the ods modal overflow issues
when displaying long ods select lists

Ticket Reference: DTCORE-3001

### DIFF
--- a/packages/manager/apps/pci-object-storage/src/hooks/useOdsModalOverflowHack.ts
+++ b/packages/manager/apps/pci-object-storage/src/hooks/useOdsModalOverflowHack.ts
@@ -1,0 +1,24 @@
+import { RefObject, useEffect } from 'react';
+
+// @TODO delete this hook when ods modal overflow is fixed
+export function useOdsModalOverflowHack(
+  modalRef: RefObject<HTMLOdsModalElement>,
+) {
+  /* The following useEffect is a hack to allow the OdsSelect overflowing
+   * in the OdsModal thus allowing the user to select an item even for long lists.
+   * This hack can be removed once this issue is fixed on ods side, and this
+   * version of ods is available in pci-common & manager-react-components.
+   */
+  useEffect(() => {
+    if (!modalRef.current) return;
+    const { shadowRoot } = modalRef.current;
+    if (!shadowRoot.querySelector('style')) {
+      const style = document.createElement('style');
+      style.innerHTML = `
+          .ods-modal__dialog { overflow: visible !important; }
+          .ods-modal__dialog__content { overflow: visible !important; }
+        `;
+      shadowRoot.appendChild(style);
+    }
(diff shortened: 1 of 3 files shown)


kqesar (approved):

packages/manager/apps/pci-object-storage/src/hooks/useOdsModalOverflowHack.ts
```diff
@@ -0,0 +1,24 @@
...
+    const { shadowRoot } = modalRef.current;
+    if (!shadowRoot.querySelector('style')) {
+      const style = document.createElement('style');
+      style.innerHTML = `
+          .ods-modal__dialog { overflow: visible !important; }
+          .ods-modal__dialog__content { overflow: visible !important; }
+        `;
+      shadowRoot.appendChild(style);
+    }
+  }, [modalRef.current]);
```
SimonChaumet, line 23: if you want to listed to changes on a ref please use useCallback refs or useState instead

https://react.dev/reference/react-dom/components/common#ref-callback
https://stackoverflow.com/questions/55838351/how-do-we-know-when-a-react-ref-current-value-has-changed

packages/manager/apps/pci-object-storage/src/hooks/useOdsModalOverflowHack.ts
```diff
@@ -0,0 +1,24 @@
...
+) {
+  /* The following useEffect is a hack to allow the OdsSelect overflowing
+   * in the OdsModal thus allowing the user to select an item even for long lists.
+   * This hack can be removed once this issue is fixed on ods side, and this
+   * version of ods is available in pci-common & manager-react-components.
+   */
+  useEffect(() => {
+    if (!modalRef.current) return;
+    const { shadowRoot } = modalRef.current;
+    if (!shadowRoot.querySelector('style')) {
```
SimonChaumet, line 15: You might want to check first if shadowRoot isn't null and opened

fredericvilcot (approved):

packages/manager/apps/pci-object-storage/src/hooks/useOdsModalOverflowHack.ts
frenautvh: hello Simon, i'll keep your suggestion in mind as i think your proposal is indeed interesting and a better solution than relying on a useEffect

for this PR since this part of code is meant to be thrown off as soon as the fix is provided by ODS (the latest version have the fix but it need to be synced with pci-common & MRC) and since this this fix has been validated by QA i prefer to not invest more dev time + QA time on this, thanks for your understanding

kqesar (approved):

PR #15686, opened 2025-02-26
ci: fix missing code ownerships

darsene:
## Description

Ticket Reference: #...

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -15,6 +15,12 @@
 /packages/manager/apps/container @ovh/dev-manager-gcj-all
 /packages/manager/apps/dbaas-logs @ovh/dev-manager-enabler-all
 /packages/manager/apps/dedicated @ovh/dev-manager-infra-all
+/packages/manager/apps/dedicated/client/*.js* @ovh/dev-manager-infra-all @ovh/dev-manager-hpc-all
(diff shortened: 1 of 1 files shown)


acavalloni (approved):

.github/CODEOWNERS
```diff
@@ -15,6 +15,12 @@
 /packages/manager/apps/container @ovh/dev-manager-gcj-all
 /packages/manager/apps/dbaas-logs @ovh/dev-manager-enabler-all
 /packages/manager/apps/dedicated @ovh/dev-manager-infra-all
+/packages/manager/apps/dedicated/client/*.js* @ovh/dev-manager-infra-all @ovh/dev-manager-hpc-all
```
anooparveti, line 18: This would require HPC's approval mandatory for every work on Infra side. So, i think we can keep Infra as the code-owners of the dedicated app by default and not combination.

darsene: Infra team is defined by default (l.17) so I don't understand your point @anooparveti

darsene: `/packages/manager/apps/dedicated/client/*.js` should just check at first level.
It's different from /packages/manager/apps/dedicated/client/**/*.js or maybe I'm wrong

anooparveti (approved):

anooparveti (approved):

PR #15724, opened 2025-02-28
feat(instance): soft reboot

lionel95200x:
ref: TAPC-2155

## Description

- Add soft reboot action

Ticket Reference: #TAPC-2155


packages/manager/apps/pci-instances/src/data/api/instance.ts
```diff
@@ -4,14 +4,23 @@ import {
...
+const getBasePathname = (projectId: string, instanceId: string) =>
+  `/cloud/project/${projectId}/instance/${instanceId}`;
 
 const instanceActionUrl = (
   projectId: string,
   instanceId: string,
   action: TInstanceAction,
 ): string => {
-  const basePathname = `/cloud/project/${projectId}/instance/${instanceId}`;
+  const basePathname = getBasePathname(projectId, instanceId);
```
SimonChaumet, line 23: Since we have a specific case for delete it seems weird to separate them

packages/manager/apps/pci-instances/src/pages/instances/action/InstanceAction.page.tsx
```diff
@@ -27,6 +35,8 @@ const InstanceAction: FC = () => {
   const section = useUrlLastSection<TSectionType>(
     actionSectionRegex.test.bind(actionSectionRegex),
   );
+
+  const snakeCaseSection = toSnakeCase(section ?? '');
```
SimonChaumet, line 39: should be wrapped in useMemo

packages/manager/apps/pci-instances/src/utils/index.ts
```diff
@@ -42,3 +42,5 @@ export const instancesQueryKey = (
   'instances',
   ...(rest && rest.length > 0 ? rest : []),
 ];
+
+export const toSnakeCase = (str: string): string => str.replace(/-/g, '_');
```
SimonChaumet, line 46:
```suggestion
export const toSnakeCase = (str: string): string => str.replaceAll('-', '_');
```

packages/manager/apps/pci-instances/src/utils/utils.spec.tsx
```diff
@@ -126,4 +130,15 @@ describe('Utility functions', () => {
       },
     );
   });
+
+  describe('Considering the toSnakeCase function', () => {
+    describe.each`
+      input            | expectedOutput
+      ${'soft-reboot'} | ${'soft_reboot'}
```
SimonChaumet, line 137: I think we should test globally with multiple dash or things like that

packages/manager/apps/pci-instances/src/data/api/instance.ts
```diff
@@ -65,3 +74,11 @@ export const unshelveInstance = (
   instanceId: string,
 ): Promise<null> =>
   v6.post(instanceActionUrl(projectId, instanceId, 'unshelve'));
+
+export const softRebootInstance = (
+  projectId: string,
+  instanceId: string,
+): Promise<null> =>
+  v6.post(`${getBasePathname(projectId, instanceId)}/reboot`, {
```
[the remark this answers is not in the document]
fredericvilcot: So, if the endpoint is the same, why not having a single `reboot` const in `TInstanceAction`? 
```ts
type TInstanceAction =
   | 'delete'
   | 'stop'
   | 'start'
   | 'shelve'
   | 'unshelve'
   | 'reboot';
```

fredericvilcot (approved):

SimonChaumet (approved):

PR #15746, opened 2025-02-28
fix(svp): prevent multi click on create button

lionel95200x:
## Description

- Prevent multi click on create button

Ticket Reference: #TAPC-3213


packages/manager/apps/pci-savings-plan/src/components/CreatePlanForm/CreatePlanForm.tsx
```diff
@@ -104,6 +104,7 @@ export type CreatePlanFormProps = {
     size: number;
   }) => void;
   isDiscoveryProject: boolean;
+  isPendingCreatePlan: boolean;
```
fredericvilcot, line 107: I would have rather call it `isCreatePlanPending` because it sounds better to me, but it's up to you 😌

fredericvilcot (approved):

fredericvilcot (dismissed):

fredericvilcot (dismissed):

fredericvilcot (approved):

PR #15750, opened 2025-03-03
fix(i18n): add missing translations [CDS 3756]

sachinrameshn:
## Description

Ticket Reference: # MANAGER-17132


sachinrameshn:
Closed this as its not needed because of fast pass trans

PR #15760, opened 2025-03-03
fix(manager-pci-common): rclone download select issue & refactoring

frenautvh:
## Description

fix region selection + refactor

Ticket Reference: DTCORE-3163


packages/manager/modules/manager-pci-common/src/components/rclone-download/StorageRegionSelector.tsx
```diff
@@ -0,0 +1,45 @@
...
+  const { translateMicroRegion } = useTranslatedMicroRegions();
+
+  // pre-select first region
+  useEffect(() => {
+    if (!currentRegion && regions?.length) {
+      const [region] = regions;
+      setCurrentRegion(region.name);
+      onRegionChange(region.name);
+    }
+  }, [currentRegion, regions, onRegionChange]);
```
qpavy, line 25:
```suggestion
  }, [regions, onRegionChange]);
```
I don't think you need to put currentRegion as dependency here

frenautvh: it is used in the condition inside the useEffect so it seems it should be listed, do you have something else in mind ?

qpavy: You will enter in the useEffect if the currentRegion changed and was not undefined but the useEffect will do nothing

qpavy (dismissed):

packages/manager/modules/manager-pci-common/src/components/rclone-download/StorageRegionSelector.tsx
```diff
@@ -0,0 +1,34 @@
...
+  onRegionChange,
+}: Readonly<StorageRegionSelectorProps>) {
+  const { translateMicroRegion } = useTranslatedMicroRegions();
+
+  return (
+    <OdsSelect
+      defaultValue={regions?.[0]?.name}
+      name="currentRegion"
+      data-testid="regions_select"
+      onOdsChange={(event) => {
```
fredericvilcot, line 22: If I inspect the Event Type from ODS: 
```ts
interface OdsSelectChangeEventDetail {
    name: string;
    previousValue?: string | null;
    validity?: ValidityState;
    value: string | null;
}
```
There is no need IMO to stringify the `event.detail.value`

packages/manager/modules/manager-pci-common/src/components/rclone-download/StorageRegionSelector.tsx
```diff
@@ -0,0 +1,34 @@
...
+        if (!event?.detail?.value) return;
+        onRegionChange(`${event.detail.value}`);
+      }}
+    >
+      {regions?.map((region: TRegion, index: number) => (
+        <option key={`${region.name}-${index}`} value={region.name}>
```
fredericvilcot, line 28: `region.name` isn't it already unique? So concatenation with the index is not required.

packages/manager/modules/manager-pci-common/src/components/rclone-download/StorageRegionSelector.tsx
frenautvh: done thanks

packages/manager/modules/manager-pci-common/src/components/rclone-download/StorageRegionSelector.tsx
frenautvh: done thanks

fredericvilcot (approved):
LGTM, thx :)

kqesar (approved):

PR #15793, opened 2025-03-04
fix(nutanix): remove column of pack type in nutanix

sefiani-ovh:
## Description

ref: MANAGER-16213

### DIFF
--- a/packages/manager/modules/nutanix/src/routing.js
+++ b/packages/manager/modules/nutanix/src/routing.js
@@ -3,7 +3,7 @@ import statusTemplate from './templates/status.html';
 import prismUrl from './templates/prismUrl.html';
 import serviceLink from './templates/serviceLink.html';
 import localizationTemplate from './templates/localization.html';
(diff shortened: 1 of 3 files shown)


tibs245 (changes requested):
After search on all files nutanix with key word : `isPackTypeAvailable`

I see 4 files concerned : 


`packages/manager/modules/nutanix/src/dashboard/general-info/routing.js` use constant `FEATURES.PACK_TYPE` . You have not error on our side ?

packages/manager/modules/nutanix/src/routing.js
darsene, line 1: You should also delete `nutanix_cluster_list_pack_type_details` from fr_FR translation file

darsene (changes requested):
Keep in mind to add `#` in your commit message so : 

```
ref: #MANAGER-16213
```

darsene (changes requested):
Can you only keep one commit ? The good one with ref: #...

darsene (approved):

PR #15810, opened 2025-03-05
feat(pci-object-storage): same name and region issue

selm3n:
ref: feat/TAPC-3003

## Description

Ticket Reference: #...


SimonChaumet (commented):
I think the new page should also be fixed ?

packages/manager/apps/pci-object-storage/src/constants.ts
```diff
@@ -434,3 +434,5 @@ export const X_CONTAINER_META_WEB_LISTINGS = 'x-container-meta-web-listings';
 export const X_CONTAINER_READ_PUBLIC_VALUE = '.r:*,.rlistings';
 
 export const BACKUP_KEY = 'ovh:backup';
+
+export const MUMBAI_REGION_NAME = 'AP-SOUTH-MUM';
```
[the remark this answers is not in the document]
selm3n: encryption is not yet available in the Mumbai region for legal reasons i.e. Indian law seems to require that the storage of encryption keys must be done on Indian soil for data stored in the country.
the test is temporary because as soon as we have clarified this subject, the encryption will be available on mumbai and the code will go in the trash

SimonChaumet (dismissed):

packages/manager/apps/pci-object-storage/src/pages/objects/container/Actions.tsx
```diff
@@ -117,6 +117,7 @@ export function Actions({
           search: `?${createSearchParams({
             containerId: storage.name,
             region: storage.region,
+            containerType: storage.containerType || '',
```
fredericvilcot, line 120: If `storage.containerType` is not defined or null, it should not been given to `createSearchParams`. Because, you will have an empty query parameter.

So please, make the property optional instead.

selm3n: I did it on purpose because i test on containerType when it contains a value and when it's ''

fredericvilcot: From what I see, if `storage.containerType` is false, you give an empty string as default value.
So query parameter exists with an empty string, is it that you want to achieve, or `containerType ` query parameter could be optional ?

darsene (dismissed):

darsene (dismissed):

packages/manager/apps/pci-object-storage/src/pages/objects/container/Actions.tsx
selm3n: Done

SimonChaumet (approved):

fredericvilcot (approved):

darsene (approved):

SimonChaumet (approved):

Issue #15101, opened 2025-01-24
[manager-components]: Generic modal component


ghyenne:
### Have you already checked if a similar item is present on manager-components?

- [x] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

Hello,

We need a generic modal with some prop for title, type info, warning, children (generic), button / loading for example.
For zimbra we created one and we reuse the same on office 365 maybe we can put directly in manager component
https://github.com/ovh/manager/blob/master/packages/manager/apps/zimbra/src/components/Modals/Modal.tsx

Thanks

### Where do you expect to use this?

where we need a modal

### Do you have mock-up?

![Image](https://github.com/user-attachments/assets/be3cc882-8bd8-441c-ad33-fb2dd3e988d7)

![Image](https://github.com/user-attachments/assets/31f903cd-ea7a-48cb-9345-6dd980d148ae)

![Image](https://github.com/user-attachments/assets/74aa4c3c-a363-435f-a0bd-4578fe356514)

### Additional Information

- [x] Component Approved.

1. Component to be created in the folder `packages/manager-react-components/src/components`
2. When `isLoading` evaluates to true, action buttons must not be displayed.
3. With `isLoading` only spinner will be displayed, and if the skeletons are to be displayed, that needs to be handled on the µ-app side.

Component structure,
```
<Modal
     heading=string - optional
     type=info|help|success|warning|error
     isLoading=boolean - optional
     primaryLabel=string - optional
     isPrimaryButtonLoading=Boolean - optional
     isPrimaryButtonDisabled=Boolean - optional
     onPrimaryButtonClick=function - optional
     secondaryLabel=string - optional
     isSecondaryButtonDisabled=Boolean - optional
     onSecondaryButtonClick=function - optional
     onDismiss=function - optional
/>
```

PR: https://github.com/ovh/manager/pull/15816

anooparveti:
@ghyenne to attend open office on 31/01.

anooparveti:
@ghyenne guidelines has been added on the "Additional Information" section.

PR #15816, opened 2025-03-05
feat(mrc): add generic modal component

ghyenne:
## Description
It's a new component generic modal

Ticket Reference: #MANAGER-16942 #15101 

## Additional Information
Screenshot 2025-03-05 at 16-27-12 components _ Modal - Loading ⋅ Storybook
Screenshot 2025-03-05 at 16-25-54 components _ Modal - Full ⋅ Storybook
Screenshot 2025-03-05 at 16-25-34 components _ Modal - Full ⋅ Storybook
Screenshot 2025-03-05 at 16-25-17 components _ Modal - Full ⋅ Storybook
Screenshot 2025-03-05 at 16-24-55 components _ Modal - Full ⋅ Storybook

### DIFF
--- a/packages/manager-react-components/src/components/Modal/Modal.component.tsx
+++ b/packages/manager-react-components/src/components/Modal/Modal.component.tsx
@@ -1,0 +1,132 @@
+import React, { Ref } from 'react';
+import {
+  OdsButton,
+  OdsModal,
+  OdsText,
+  OdsSpinner,
+} from '@ovhcloud/ods-components/react';
+import {
+  ODS_BUTTON_VARIANT,
+  ODS_MODAL_COLOR,
+  ODS_BUTTON_COLOR,
+  ODS_TEXT_PRESET,
+  ODS_SPINNER_SIZE,
+} from '@ovhcloud/ods-components';
+
+export interface ModalProps {
+  /** Title of modal */
+  heading?: string;
+  /** Type of modal. It can be any of `ODS_MODAL_COLOR` */
+  type?: ODS_MODAL_COLOR;
+  /** Is loading state for display a spinner */
+  isLoading?: boolean;
+  /** Label of primary button */
+  primaryLabel?: string;
+  /** Is loading state for primary button */
+  isPrimaryButtonLoading?: boolean;
+  /** Is disabled state for primary button */
+  isPrimaryButtonDisabled?: boolean;
+  /** Action of primary button */
+  onPrimaryButtonClick?: () => void;
+  /** Label of secondary button */
+  secondaryLabel?: string;
+  /** Is loading state for secondary button */
+  isSecondaryButtonDisabled?: boolean;
+  /** Is loading state for secondary button */
+  isSecondaryButtonLoading?: boolean;
+  /** Is disabled state for secondary button */
+  onSecondaryButtonClick?: () => void;
+  /** Is dismissible action */
+  onDismiss?: () => void;
+  /** Is modal open state */
+  isOpen?: boolean;
+  /** Children of modal */
+  children?: React.ReactNode;
+}
+
+const getPrimaryButtonColor = (type: ODS_MODAL_COLOR) =>
+  type === ODS_MODAL_COLOR.critical
+    ? ODS_BUTTON_COLOR.critical
+    : ODS_BUTTON_COLOR.primary;
+
+export const Modal = React.forwardRef<HTMLOdsModalElement, ModalProps>(
+  (
+    {
+      heading,
+      type = ODS_MODAL_COLOR.information,
+      isLoading,
+      primaryLabel,
+      isPrimaryButtonLoading,
+      isPrimaryButtonDisabled,
+      onPrimaryButtonClick,
+      secondaryLabel,
+      isSecondaryButtonDisabled,
+      isSecondaryButtonLoading,
+      onSecondaryButtonClick,
+      onDismiss,
+      isOpen = true,
+      children,
+    },
+    ref: Ref<HTMLOdsModalElement>,
+  ) => {
+    const buttonColor = getPrimaryButtonColor(type);
+
+    return (
+      <OdsModal
+        data-testid="modal"
+        color={ODS_MODAL_COLOR[type]}
+        isDismissible={!!onDismiss}
+        className="text-left"
+        onOdsClose={onDismiss}
+        isOpen={isOpen}
+        ref={ref}
+      >
+        {heading && (
+          <OdsText className="mb-4" preset={ODS_TEXT_PRESET.heading4}>
+            {heading}
+          </OdsText>
+        )}
+        {isLoading && (
+          <div data-testid="spinner" className="flex justify-center my-5">
+            <OdsSpinner size={ODS_SPINNER_SIZE.md} inline-block></OdsSpinner>
+          </div>
+        )}
+        {!isLoading && (
+          <>
+            <div className="flex text-left">{children}</div>
+            {secondaryLabel && (
+              <OdsButton
+                data-testid="secondary-button"
+                slot="actions"
+                color={buttonColor}
+                onClick={
+                  !isSecondaryButtonDisabled ? onSecondaryButtonClick : null // need for don't break the test
+                }
+                isDisabled={isSecondaryButtonDisabled}
+                isLoading={isSecondaryButtonLoading}
+                variant={ODS_BUTTON_VARIANT.ghost}
+                label={secondaryLabel}
+                className="mt-4"
+                type="button"
+              />
+            )}
+            {primaryLabel && (
+              <OdsButton
+                data-testid="primary-button"
+                slot="actions"
+                color={buttonColor}
+                onClick={!isPrimaryButtonDisabled ? onPrimaryButtonClick : null} // need for don't break the test
+                isDisabled={isPrimaryButtonDisabled}
+                isLoading={isPrimaryButtonLoading}
+                variant={ODS_BUTTON_VARIANT.default}
+                label={primaryLabel}
+                className="mt-4"
+                type="button"
+              />
+            )}
+          </>
+        )}
+      </OdsModal>
+    );
+  },
+);
--- a/packages/manager-react-components/src/components/Modal/Modal.mock.tsx
+++ b/packages/manager-react-components/src/components/Modal/Modal.mock.tsx
@@ -1,0 +1,28 @@
+export const basic = {
+  heading: 'Example of modal',
+  children: <div>Example of content</div>,
+};
+
+export const actions = {
+  primaryLabel: 'Confirm',
+  isPrimaryButtonLoading: false,
+  isPrimaryButtonDisabled: false,
+  onPrimaryButtonClick: () => 'onPrimaryButtonClick',
+  secondaryLabel: 'Cancel',
+  isSecondaryButtonDisabled: false,
+  isSecondaryButtonLoading: false,
+  onSecondaryButtonClick: () => 'onSecondaryButtonClick',
+  onDismiss: () => 'onDismiss',
+};
+
+export const type = {
+  type: 'warning',
+};
+
+export const loading = {
+  isLoading: true,
+};
+
+export const other = {
+  isOpen: true,
+};
--- a/packages/manager-react-components/src/components/Modal/Modal.spec.tsx
+++ b/packages/manager-react-components/src/components/Modal/Modal.spec.tsx
@@ -1,0 +1,104 @@
+import { vi } from 'vitest';
+import { ODS_MODAL_COLOR } from '@ovhcloud/ods-components';
+import { act, fireEvent, render, screen } from '@testing-library/react';
+import { Modal } from './Modal.component';
+import {
+  basic as basicMock,
+  actions as actionsMock,
+  loading as loadingMock,
+} from './Modal.mock';
+
+it('should display the basic modal', () => {
+  const { getByTestId } = render(<Modal {...basicMock} />);
+  expect(getByTestId('modal')).not.toBeNull();
+  expect(screen.queryByText('Example of modal')).not.toBeNull();
+  expect(screen.queryByText('Example of content')).not.toBeNull();
+});
+
+it('should display the modal with action', async () => {
+  const onPrimaryButtonClick = vi.fn(actionsMock.onPrimaryButtonClick);
+  const onSecondaryButtonClick = vi.fn(actionsMock.onSecondaryButtonClick);
+  const { getByTestId } = render(
+    <Modal
+      {...basicMock}
+      {...actionsMock}
+      onPrimaryButtonClick={onPrimaryButtonClick}
+      onSecondaryButtonClick={onSecondaryButtonClick}
+    />,
+  );
+  const primaryButton = getByTestId('primary-button');
+  const secondaryButton = getByTestId('secondary-button');
+
+  expect(primaryButton).toHaveAttribute('label', 'Confirm');
+  expect(primaryButton).toHaveAttribute('color', 'primary');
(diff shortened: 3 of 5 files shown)


packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,142 @@
...
+  warning: ODS_MODAL_COLOR.warning,
+  error: ODS_MODAL_COLOR.critical,
+};
+
+export const Modal: React.FC<ModalProps> = ({
+  heading,
+  type = ModalTypeEnum.info,
+  isLoading,
+  primaryLabel,
+  isPrimaryButtonLoading,
```
tristanwagner, line 75: is there any good reason to not have 2 props, like `primaryButton` and `secondaryButton` that are of type 

```
Omit<
  JSX.OdsButton & React.HTMLAttributes<HTMLOdsButtonElement>,
  'slot' | 'color' | 'variant'
>
``` 
for example and allow to pass any other props that are not set  by this component It will be more flexible, generic, and less verbose. 
Or just

 ```
Pick<
  JSX.OdsButton & React.HTMLAttributes<HTMLOdsButtonElement>,
  'isLoading' | 'isDisabled' | 'label'
>
``` 

if we really want to be restrictive ?

File: packages/manager-react-components/src/components/Modal/Modal.component.tsx [hunk 2]
```diff
@@ -0,0 +1,142 @@
...
+  /** Is loading state for secondary button */
+  isSecondaryButtonLoading?: boolean;
+  /** Is disabled state for secondary button */
+  onSecondaryButtonClick?: () => void;
+  /** Is dismissible action */
+  onDismiss?: () => void;
+  /** Is modal open state */
+  isOpen?: boolean;
+  /** Children of modal */
+  children?: React.ReactElement;
```
tristanwagner, line 52: Prefer use `React.PropsWithChildren<YourProps>` type instead of declaring it like this, or put a `React.ReactNode` type

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,142 @@
...
+import {
+  ODS_BUTTON_VARIANT,
+  ODS_MODAL_COLOR,
+  ODS_BUTTON_COLOR,
+  ODS_TEXT_PRESET,
+  ODS_SPINNER_SIZE,
+} from '@ovhcloud/ods-components';
+
+export enum ModalTypeEnum {
+  'info' = 'info',
```
tristanwagner, line 17: let's keep the enum items UPPERCASED to match code style ?

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,142 @@
...
+    <OdsModal
+      data-testid="modal"
+      color={mapModalColor[type]}
+      isDismissible={!!onDismiss}
+      className="text-left"
+      onOdsClose={onDismiss}
+      isOpen={isOpen}
+    >
+      <OdsText className="mb-4" preset={ODS_TEXT_PRESET.heading4}>
+        {heading}
```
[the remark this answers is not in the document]
ghyenne: It's fixed

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,142 @@
...
+              className="mt-4"
+              role="button"
+            />
+          )}
+          {primaryLabel && (
+            <OdsButton
+              data-testid="primary-button"
+              slot="actions"
+              color={buttonColor}
+              onClick={!isPrimaryButtonDisabled ? onPrimaryButtonClick : null}
```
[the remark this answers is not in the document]
ghyenne: Can't be fixed, because break the vitest testing. I added a comment.

packages/manager-react-components/src/components/Modal/Modal.component.tsx
ghyenne: I put React.ReactNode, so it's fixed.

packages/manager-react-components/src/components/Modal/Modal.component.tsx
ghyenne: It's fixed

packages/manager-react-components/package.json
```diff
@@ -77,7 +77,7 @@
     "@types/react-dom": "18.2.7",
     "@typescript-eslint/eslint-plugin": "^4.7.4",
     "@vitejs/plugin-react": "4.0.4",
-    "@vitest/coverage-v8": "0.34.1",
+    "@vitest/coverage-v8": "3.0.7",
```
aboungnaseng-ovhcloud, line 80:
```suggestion
    "@vitest/coverage-v8": "^2.1.4"
```

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,150 @@
...
+  onSecondaryButtonClick?: () => void;
+  /** Is dismissible action */
+  onDismiss?: () => void;
+  /** Is modal open state */
+  isOpen?: boolean;
+  /** Children of modal */
+  children?: React.ReactNode;
+}
+
+const mapModalColorToButtonColor = (type: ModalTypeEnum) => {
```
aboungnaseng-ovhcloud, line 55:
```suggestion
const mapModalColorToButtonColor = (type: ModalTypeEnum) =>
  type === ModalTypeEnum.ERROR
    ? ODS_BUTTON_COLOR.critical
    : ODS_BUTTON_COLOR.primary;
```

packages/manager-react-components/package.json
ghyenne: We have to have the same version for @vitest/coverage-v8 and vitest. today vitest is ^2.1.9. Why we don't want upgrade to last version available ?

packages/manager-react-components/src/components/Modal/Modal.component.tsx
ghyenne: it's fixed

packages/manager-react-components/package.json
```diff
@@ -77,7 +77,7 @@
     "@types/react-dom": "18.2.7",
     "@typescript-eslint/eslint-plugin": "^4.7.4",
     "@vitejs/plugin-react": "4.0.4",
-    "@vitest/coverage-v8": "0.34.1",
+    "@vitest/coverage-v8": "^2.1.9",
```
anooparveti, line 80: The bump is being handled as part of a dedicated ticket - https://github.com/ovh/manager/pull/15922/files#diff-67b8a223598086ec961e7875d48f79f30a64cd6c42551a033132f51fd6ed55aa
Can you remove this change?

anooparveti (changes requested):

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,148 @@
...
+} from '@ovhcloud/ods-components/react';
+import {
+  ODS_BUTTON_VARIANT,
+  ODS_MODAL_COLOR,
+  ODS_BUTTON_COLOR,
+  ODS_TEXT_PRESET,
+  ODS_SPINNER_SIZE,
+} from '@ovhcloud/ods-components';
+
+export enum ModalTypeEnum {
```
anooparveti, line 16:
```suggestion
export enum MODAL_TYPE {
```

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,148 @@
...
+  onSecondaryButtonClick?: () => void;
+  /** Is dismissible action */
+  onDismiss?: () => void;
+  /** Is modal open state */
+  isOpen?: boolean;
+  /** Children of modal */
+  children?: React.ReactNode;
+}
+
+const mapModalColorToButtonColor = (type: ModalTypeEnum) =>
```
anooparveti, line 55:
```suggestion
const getPrimaryButtonColor = (type: ModalTypeEnum) =>
```

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,148 @@
...
+    ? ODS_BUTTON_COLOR.critical
+    : ODS_BUTTON_COLOR.primary;
+
+const mapModalColor = {
+  info: ODS_MODAL_COLOR.information,
+  help: ODS_MODAL_COLOR.neutral,
+  success: ODS_MODAL_COLOR.success,
+  warning: ODS_MODAL_COLOR.warning,
+  error: ODS_MODAL_COLOR.critical,
+};
```
anooparveti, line 66: **nitpick:** Can we avoid this mapping and directly use ODS Modal color.

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,148 @@
...
+          </OdsText>
+        )}
+        {isLoading && (
+          <div data-testid="spinner" className="flex justify-center my-5">
+            <OdsSpinner size={ODS_SPINNER_SIZE.md} inline-block></OdsSpinner>
+          </div>
+        )}
+        {!isLoading && (
+          <>
+            <div className="flex flex-col text-left">{children}</div>
```
anooparveti, line 112: **question:** Are you sure about `flex-col`? Because if we add 2 divs, 

```
<div></div>
<div></div>
```
It would be stacked as columns instead of rows.

anooparveti (changes requested):

packages/manager-react-components/src/components/Modal/Modal.component.tsx
ghyenne: Enum name `MODAL_TYPE` must match one of the following formats: PascalCaseeslint[@typescript-eslint/naming-convention](https://github.com/typescript-eslint/typescript-eslint/blob/v4.33.0/packages/eslint-plugin/docs/rules/naming-convention.md)

ghyenne: But we can use directly ODS_MODAL_COLOR like suggest your second comment for avoid mapping

packages/manager-react-components/src/components/Modal/Modal.component.tsx
ghyenne: You right the behaviour is strange with and without flex-col, but if we want row we don't put flex-col but for make a column we have to put 2 div in parent div

```
<>
   <div>
      <div>Example of content div 1</div>
      <div>Example of content div 2</div>
   </div>
   <div>Example of content div 3</div>
</>
 ```
Screenshot 2025-03-12 at 15-43-11 components _ Modal - Actions ⋅ Storybook

File: packages/manager-react-components/package.json (hunk 8 above)
Comment (ghyenne) [reply packages/manager-react-components/package.json]: it's fixed but the yarn.lock have still change even if no change in package.json

File: packages/manager-react-components/src/components/Modal/Modal.component.tsx (hunk 9 above)
Comment (ghyenne) [reply packages/manager-react-components/src/components/Modal/Modal.component.tsx]: it's fixed

File: packages/manager-react-components/src/components/Modal/Modal.component.tsx (hunk 10 above)
Comment (ghyenne) [reply packages/manager-react-components/src/components/Modal/Modal.component.tsx]: it's fixed

File: packages/manager-react-components/src/components/Modal/Modal.component.tsx (hunk 11 above)
Comment (ghyenne) [reply packages/manager-react-components/src/components/Modal/Modal.component.tsx]: it's fixed

File: packages/manager-react-components/src/components/Modal/Modal.component.tsx (hunk 12 above)
Comment (ghyenne) [reply packages/manager-react-components/src/components/Modal/Modal.component.tsx]: it's fixed

File: packages/manager-react-components/src/components/Modal/Modal.component.tsx [hunk 13]
```diff
@@ -0,0 +1,132 @@
...
+  ODS_MODAL_COLOR,
+  ODS_BUTTON_COLOR,
+  ODS_TEXT_PRESET,
+  ODS_SPINNER_SIZE,
+} from '@ovhcloud/ods-components';
+
+export interface ModalProps {
+  /** Title of modal */
+  heading?: string;
+  /** Type of modal */
```
anooparveti, line 19:
```suggestion
  /** Type of modal. It can be any of `ODS_MODAL_COLOR`  */
```

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,132 @@
...
+          </div>
+        )}
+        {!isLoading && (
+          <>
+            <div className="flex text-left">{children}</div>
+            {secondaryLabel && (
+              <OdsButton
+                data-testid="secondary-button"
+                slot="actions"
+                color={buttonColor}
```
anooparveti, line 101: **question:** Secondary button always have same color no?

anooparveti (changes requested):

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,132 @@
...
+                className="mt-4"
+                type="button"
+              />
+            )}
+            {primaryLabel && (
+              <OdsButton
+                data-testid="primary-button"
+                slot="actions"
+                color={buttonColor}
+                onClick={!isPrimaryButtonDisabled ? onPrimaryButtonClick : null} // need for don't break the test
```
anooparveti, line 118:
```suggestion
                onClick={!isPrimaryButtonDisabled && onPrimaryButtonClick} // need for don't break the test
```

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -0,0 +1,132 @@
...
+        {!isLoading && (
+          <>
+            <div className="flex text-left">{children}</div>
+            {secondaryLabel && (
+              <OdsButton
+                data-testid="secondary-button"
+                slot="actions"
+                color={buttonColor}
+                onClick={
+                  !isSecondaryButtonDisabled ? onSecondaryButtonClick : null // need for don't break the test
```
anooparveti, line 103:
```suggestion
                  !isSecondaryButtonDisabled && onSecondaryButtonClick
```

yarn.lock
anooparveti, line 1: You dont need to include `yarn.lock` changes since you removed the dependency update.

anooparveti (changes requested):

packages/manager-react-components/src/components/Modal/Modal.component.tsx
ghyenne: The testing fail if i put other thing that null if disabled. It's issue with react testing, because the function it isn't really triggered if the button is disabled only during the test

packages/manager-react-components/src/components/Modal/Modal.component.tsx
ghyenne: The testing fail if i put other thing that null if disabled. It's issue with react testing, because the function it isn't really triggered if the button is disabled only during the test

yarn.lock
ghyenne: It's fixed

packages/manager-react-components/src/components/Modal/Modal.component.tsx
ghyenne: it's fixed

anooparveti (approved):

aboungnaseng-ovhcloud (approved):

PR #15821, opened 2025-03-05
fixup! feat(vcd): bump ods version to 18

pauldkn:
## Description
Fix QA feedbacks (referenced below)
Fix demo feedback :
- modal buttons (gap, cancel variant)

Ticket Reference:
- MANAGER-17284
- MANAGER-17293
- MANAGER-17301


packages/manager/apps/hpc-vmware-managed-vcd/src/components/message/Message.component.tsx
```diff
@@ -14,7 +14,7 @@ export const Message: React.FC<MessageProps> = ({ message }) => {
     content,
     uid,
     type = 'information',
-    isDismissible,
+    isDismissible = true,
```
tibs245, line 17: I'm not fan props true by default. Possible rename it to have false by default ?

`isPersistent`?

pauldkn: It was like this before the migration but `isDismissible` is the same name as the modal props so I renamed it to be more consistent

PR #15826, opened 2025-03-06
fix(pci-cold-archive): add missing user secret

frenautvh:
## Description

add missing user secret

Ticket Reference: DTCORE-3171, DTCORE-3176

### DIFF
--- a/packages/manager/apps/pci-cold-archive/src/pages/containers/new/New.page.tsx
+++ b/packages/manager/apps/pci-cold-archive/src/pages/containers/new/New.page.tsx
@@ -28,6 +28,7 @@ import { ContainerNameStep } from './steps/ContainerNameStep.component';
 import { LinkUserStep } from './steps/LinkUserStep.component';
 import { useContainerCreationStore } from './useContainerCreationStore';
 import GuideMenu from '@/components/GuideMenu.component';
+import { usePostS3Secret } from '@/api/hooks/useUsers';
 
 export default function ContainerNewPage() {
   const { t } = useTranslation([
@@ -59,7 +60,40 @@ export default function ContainerNewPage() {
     clearNotifications();
   }, []);
 
-  const { createContainer, isPending } = useCreateContainer({
+  const {
+    mutateAsync: postSecretAsync,
+    isPending: isSecretPending,
+  } = usePostS3Secret({
+    projectId: project?.project_id,
+    userId: form.selectedUser?.id,
+    userAccess: form.selectedUser?.s3Credentials?.access,
+    onSuccess({ secret }) {
+      if (form.selectedUser?.s3Credentials) {
+        form.selectedUser.s3Credentials.secret = secret;
+      }
+    },
+    onError: (error: ApiError) => {
+      addError(
+        <Translation ns="cold-archive/new">
+          {(_t) =>
+            _t(
+              'pci_projects_project_storages_cold_archive_add_action_create_archive_create_request_failed',
+              {
+                containerName: form.containerName,
+                message:
+                  error?.response?.data?.message || error?.message || null,
+              },
+            )
+          }
+        </Translation>,
+        true,
+      );
+      trackErrorPage();
+      window.scrollTo(0, 0);
+    },
+  });
+
+  const { createContainer, isPending: isCreationPending } = useCreateContainer({
     projectId: project.project_id,
     onSuccess: (container: TArchiveContainer) => {
       clearNotifications();
@@ -116,8 +150,13 @@ export default function ContainerNewPage() {
     },
   });
 
-  const onCreateContainer = () => {
+  const isPending = isSecretPending || isCreationPending;
+
+  const onCreateContainer = async () => {
     trackConfirmAction();
+    if (!form.selectedUser?.s3Credentials?.secret) {
+      await postSecretAsync();
+    }
     createContainer({
       ownerId: form.ownerId,
       name: form.containerName,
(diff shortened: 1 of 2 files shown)


packages/manager/apps/pci-cold-archive/src/pages/containers/new/New.page.tsx
```diff
@@ -55,7 +56,38 @@ export default function ContainerNewPage() {
...
-  const { createContainer, isPending } = useCreateContainer({
+  const {
+    mutateAsync: postSecretAsync,
+    isPending: isSecretPending,
+  } = usePostS3Secret({
+    projectId: project.project_id,
+    userId: form.selectedUser?.id,
+    userAccess: form.selectedUser?.s3Credentials?.access,
+    onSuccess({ secret }) {
+      form.selectedUser.s3Credentials.secret = secret;
```
fredericvilcot, line 67: `s3Credentials` is possibly `undefined`

packages/manager/apps/pci-cold-archive/src/pages/containers/new/New.page.tsx
```diff
@@ -110,8 +142,13 @@ export default function ContainerNewPage() {
     },
   });
 
-  const onCreateContainer = () => {
+  const isPending = isSecretPending || isCreationPending;
+
+  const onCreateContainer = async () => {
     trackConfirmAction();
+    if (!form.selectedUser.s3Credentials?.secret) {
```
fredericvilcot, line 149: `form.selectedUser` is possibly `undefined`

packages/manager/apps/pci-cold-archive/src/pages/containers/new/New.page.tsx
```diff
@@ -55,7 +56,38 @@ export default function ContainerNewPage() {
 
   const goBack = () => navigate('..');
 
-  const { createContainer, isPending } = useCreateContainer({
+  const {
+    mutateAsync: postSecretAsync,
+    isPending: isSecretPending,
+  } = usePostS3Secret({
+    projectId: project.project_id,
```
fredericvilcot, line 63: `project` is possibly `undefined`

packages/manager/apps/pci-cold-archive/src/pages/containers/new/New.page.tsx
```diff
@@ -55,7 +56,38 @@ export default function ContainerNewPage() {
...
+    userId: form.selectedUser?.id,
+    userAccess: form.selectedUser?.s3Credentials?.access,
```
fredericvilcot, line 65: By contract, `userId` & `userAccess` should always be defined

packages/manager/apps/pci-cold-archive/src/pages/containers/new/steps/LinkUserCreation.component.tsx
```diff
@@ -74,20 +74,17 @@ export default function LinkUserCreation({
...
-              s3Credentials: credentials,
-            });
-          })
-          .finally(() => {
-            invalidateGetUsersCache(projectId);
-            onCreateUser(value);
-            setIsLoading(false);
-          });
+      onSuccess: async ({ value }) => {
+        const credentials = await generateS3Credentials(projectId, value.id);
```
fredericvilcot, line 78: You cannot invoke `generateS3Credentials` with `projectId` that could be `undefined`

packages/manager/apps/pci-cold-archive/src/pages/containers/new/steps/LinkUserCreation.component.tsx
```diff
@@ -74,20 +74,17 @@ export default function LinkUserCreation({
...
+        const userWithCredentials = {
+          ...value,
+          access: credentials.access,
+          s3Credentials: credentials,
+        };
+        setNewUser(userWithCredentials);
+        onCreateUser(userWithCredentials);
+        invalidateGetUsersCache(projectId);
```
fredericvilcot, line 86: Same here ;)

packages/manager/apps/pci-cold-archive/src/pages/containers/new/New.page.tsx
frenautvh: yes and the api will return an error that will be displayed if some mandatory parameters are missing, that's the behavior we expect when we perform the api call

i don't think checking manually that userId & userAccess is defined then throw and display an error manually is better since it would mean that you don't trust the api behavior, it will also add more unecessary logic to the view

packages/manager/apps/pci-cold-archive/src/pages/containers/new/steps/LinkUserCreation.component.tsx
fredericvilcot: yes, but in terms of TS typing, it should not be the case ;)

packages/manager/apps/pci-cold-archive/src/pages/containers/new/New.page.tsx
fredericvilcot: But here, you are assigning non defined values that must be always be defined, by contrat (typing).
So, it come to do some plain JS

frenautvh: the ts compiler is not reporting issues so i wouldn't call this "not compliant with ts engine"

fredericvilcot: ts engine is not reporting issues because we don't use the `strict` mode, that should be enabled by default :)
But anyway, we can keep like this so.

kqesar (approved):

frenautvh:
rebased

Issue #15846, opened 2025-03-06
[manager-components]: Date filtering


FabienHenon:
### Have you already checked if a similar item is present on manager-components?

- [x] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

Filtering columns with fields of type date does not work properly.
This uses a string field and comparison which makes it very hard to filter correctly.
Other issues have also been found while solving this one.

Among these issues:
1. Filtering a Date field with comparisons like Is Equal and Is Different does not work properly
2. The DatePicker component is not used in `FilterAdd` for Date fields
3. The comparator `OdsSelect` field of `FilterAdd` also has some issues (value disappearing)
4. The `FilterList` component does not show date values correctly (date is not formatted)
5. There are also issues with some translations for the comparators

## Issue 1. Date filtering not working

As you can see, I have components with a date:
<img width="860" alt="Image" src="https://github.com/user-attachments/assets/fd714394-3efb-42e2-ac36-e36858372d36" />

When I try to filter some of them with a date (as a string set manually), I have no result:
<img width="843" alt="Image" src="https://github.com/user-attachments/assets/50662287-38d9-40ca-bd8b-dd6692eb4e1c" />

I should get some results for this date.

## Issue 2. No datepicker for date fields

There is no datepicker for date fields in filtering:
<img width="263" alt="Image" src="https://github.com/user-attachments/assets/623b7441-0773-4d33-a465-aa02b378d12c" />

I'm forced to set a date as string manually, which causes bad filtering.

## Issue 3. `OdsSelect`

Sometimes the value of the select becomes blank after selecting a valid value:
<img width="231" alt="Image" src="https://github.com/user-attachments/assets/0fcad6bf-2edd-479b-9002-f655c30660a9" />

The issues seems to come from rendering several `OdsSelect` components (when only one for comparison is rendered the issue is gone)

## Issue 4. `FilterList` date format

Date is not formatted correctly:

<img width="837" alt="Image" src="https://github.com/user-attachments/assets/67fd84b4-565d-450e-b3eb-68616c4a356a" />

## Issue 5. translations

Here is an example for english (the same error applies for all languages):

```
"common_criteria_adder_operator_is_before": "is",
"common_criteria_adder_operator_is_after": "is before",
```

### Where do you expect to use this?

manager

### When do you expect this to be delivered?

ASAP

PR #15847, opened 2025-03-06
fix(mrc): date filtering

FabienHenon:
ref: #15846

## Description

Ticket Reference: #15846


packages/manager/core/api/src/filters.ts
```diff
@@ -67,10 +68,24 @@ export function applyFilters<T>(items: T[] = [], filters: Filter[] = []) {
           keep = keep && `${value}`.toLowerCase().endsWith(comp.toLowerCase());
           break;
         case FilterComparator.IsEqual:
-          keep = keep && `${value}`.toLowerCase() === comp.toLowerCase();
+          if (filter.type === FilterTypeCategories.Date) {
+            keep =
+              keep &&
+              new Date(`${value}`).setHours(0, 0, 0, 0) ===
+                new Date(comp).setHours(0, 0, 0, 0);
```
anooparveti, line 75: As your example also displays time, i think this can contradict the time part and maybe we need to accept time as well in the filtering.

FabienHenon: That's a good point. I assumed filtering on dates would only be done on dates without time.
I'll ask QA for more information about this point

FabienHenon: @anooparveti I just had confirmation that it's only date filtering (ISO with prod)

packages/manager-react-components/src/components/filters/filter-list.component.tsx
```diff
@@ -31,7 +33,11 @@ export function FilterList({
           label={`${
             filter.label ? `${filter.label} ${tComp(filter.comparator)} ` : ''
           }
-          ${filter.value}`}
+          ${
+            filter.type === FilterTypeCategories.Date
```
aboungnaseng-ovhcloud, line 37: I found an issue with filter `type`, It seems to miss a typing in `FilterWithLabel`

packages/manager-react-components/src/components/filters/filter-add.component.tsx
```diff
@@ -116,26 +122,36 @@ export function FilterAdd({ columns, onAddFilter }: Readonly<FilterAddProps>) {
...
-                submitAddFilter();
-              }
-            }}
-          />
+          {selectedColumn?.type === FilterTypeCategories.Date ? (
+            <OdsDatepicker
+              name="filter-add_value-input"
+              className="border"
+              value={dateValue}
+              data-testid="filter-add_value-input"
```
anooparveti, line 130:
```suggestion
              data-testid="filter-add_value-date"
```

packages/manager-react-components/src/components/filters/filter-add.component.tsx
```diff
@@ -82,31 +96,23 @@ export function FilterAdd({ columns, onAddFilter }: Readonly<FilterAddProps>) {
...
-                      </option>
-                    ))}
-                  </OdsSelect>
-                </div>
-              );
-            })}
+          {selectedColumn && (
+            <div key={`filter-condition-select-${selectedColumn.id}`}>
+              <OdsSelect
+                name={`add-operator-${selectedColumn?.id}`}
```
SimonChaumet, line 102:
```suggestion
                name={`add-operator-${selectedColumn.id}`}
```
cannot be undefined, or else there would be an error on line 100

packages/manager-react-components/src/components/filters/filter-add.component.tsx
```diff
@@ -41,15 +48,22 @@ export function FilterAdd({ columns, onAddFilter }: Readonly<FilterAddProps>) {
       {
         key: selectedId,
         comparator: selectedComparator,
-        value,
+        value:
+          selectedColumn?.type === FilterTypeCategories.Date
+            ? dateValue?.toISOString()
+            : value,
+        type: selectedColumn?.type,
```
SimonChaumet, line 55: Is it possible that selectedColumn and dateValue could be undefined ? If no there is no need for `?`

packages/manager-react-components/src/components/filters/filter-add.component.tsx
```diff
@@ -30,6 +36,7 @@ export function FilterAdd({ columns, onAddFilter }: Readonly<FilterAddProps>) {
     columns?.[0]?.comparators?.[0] || FilterComparator.IsEqual,
   );
   const [value, setValue] = useState('');
+  const [dateValue, setDateValue] = useState<Date>(null);
```
SimonChaumet, line 39:
```suggestion
  const [dateValue, setDateValue] = useState<Date | null>(null);
```

packages/manager-react-components/src/components/filters/filter-add.component.tsx
```diff
@@ -82,31 +96,23 @@ export function FilterAdd({ columns, onAddFilter }: Readonly<FilterAddProps>) {
...
+            <div key={`filter-condition-select-${selectedColumn.id}`}>
+              <OdsSelect
+                name={`add-operator-${selectedColumn?.id}`}
+                value={selectedComparator}
+                onOdsChange={(event) => {
+                  setSelectedComparator(event.detail.value as FilterComparator);
+                }}
+              >
+                {selectedColumn?.comparators?.map((comp) => (
+                  <option key={`${selectedColumn.id}-${comp}`} value={comp}>
```
SimonChaumet, line 109:
```suggestion
                  <option key={comp} value={comp}>
```
Key is not an id, you only need the key of your element inside the list

packages/manager-react-components/src/components/filters/filter-add.component.tsx
SimonChaumet, line 39: I think we should have only one value and other should have computed values or else we will duplicate code only for different types

Or we should think of another way to handle multiple types

packages/manager-react-components/src/components/filters/filter-list.component.tsx
```diff
@@ -31,7 +33,11 @@ export function FilterList({
...
+              ? new Date(`${filter.value}`).toLocaleDateString(locale)
```
SimonChaumet, line 38: why convert your value with template string ? You can use String or something else, it seems weird to have any type and use toString

packages/manager/apps/pci-object-storage/src/pages/objects/container/__snapshots__/Listing.page.spec.tsx.snap
SimonChaumet, line 1: is there an issue with these files ?

packages/manager/core/api/src/filters.ts
```diff
@@ -67,10 +68,24 @@ export function applyFilters<T>(items: T[] = [], filters: Filter[] = []) {
           keep = keep && `${value}`.toLowerCase().endsWith(comp.toLowerCase());
           break;
         case FilterComparator.IsEqual:
-          keep = keep && `${value}`.toLowerCase() === comp.toLowerCase();
+          if (filter.type === FilterTypeCategories.Date) {
+            keep =
+              keep &&
+              new Date(`${value}`).setHours(0, 0, 0, 0) ===
```
SimonChaumet, line 74: you can use startOfDay from date fns instead

packages/manager-react-components/src/components/filters/filter-add.component.tsx
FabienHenon: I tried using only one value but it was make the code harder to write and to read but several parsing/formatting of the date.
So I figured out it was better to separate the state by type in this case.
2 types is still easy to maintain in such a component.
But if one day more types come up, we should, indeed, think about refactoring it

packages/manager-react-components/src/components/filters/filter-list.component.tsx
FabienHenon: I used the same way of converting to string that the one already in place because I didn't want to break anything. 
According to type spec, the value can be an array of strings.

packages/manager/apps/pci-object-storage/src/pages/objects/container/__snapshots__/Listing.page.spec.tsx.snap
FabienHenon: I changed the way filter add was rendering. The snapshot failed

SimonChaumet: yes but it seems that the column were removed, is it wanted ?

packages/manager-react-components/src/components/filters/filter-add.component.tsx
FabienHenon: That should not happen

packages/manager/apps/pci-object-storage/src/pages/objects/container/__snapshots__/Listing.page.spec.tsx.snap
FabienHenon: What's removed is actually the comparison selects.
Initially, comparison selects for every column were rendered and only the active one was not hidden, causing issues with OdsSelect (a ticket has been filed at ODS about this).
I changed the code to only render the active column.
It solves the issue with OdsSelect and is, I think, cleaner.

aboungnaseng-ovhcloud: Good fix ! Its normal if the snapshot of listing is updated 👍

aboungnaseng-ovhcloud:
Some tests are failing, but it seems good to me, I have test on my side

packages/manager-react-components/src/components/filters/filter-add.component.tsx
FabienHenon: The value should be a Date, I set null to keep it consistent with the expected type

FabienHenon:
> Some tests are failing, I have test on my side, we have a PR with @kb-ovh, he update the same filter file and add other optional attribute. Lets synchronise before merging it

I think tests are failing because of a change in the styles of datagrid. This has no conflict with this PR. I think tests should be broken on develop. Once this is fixed in develop it should pass here

frenautvh (approved):

packages/manager/core/api/src/filters.ts
frenautvh: since this dependency is built and bundled, adding dependencies have an additional cost on the bundle size so i would advocate to do so only if it is mandatory

SimonChaumet: Or maybe we could add tree shaking since date fns greatly supports it to avoid reinventing the wheel

frenautvh: as you can see here, tree shaking have already been considered by specifying the sideEffect attribute in the package json of this package : https://github.com/ovh/manager/blob/master/packages/manager/core/api/package.json#L7

i don't think the wheel have been reinvented in this particular case, but it should be interesting to see the cost of importing the startOfDay function and see the impact on the whole bundle size

FabienHenon: Guys, it's not like I was coding a complex function with a complex algorithm... It's just the use of `setHours` which has the same effect as `startOfDay` without the need to add a dependency

packages/manager-react-components/src/components/filters/filter-list.component.tsx
aboungnaseng-ovhcloud: Its ok for the rendering, we can see, how we can improve it after

aboungnaseng-ovhcloud (approved):

kqesar (approved):

frenautvh (dismissed):

aboungnaseng-ovhcloud (approved):

aboungnaseng-ovhcloud (dismissed):

anooparveti (approved):

PR #15869, opened 2025-03-07
fix(pci-cold-archive): remove message when load new page

kqesar:
ref: DTCORE-3168

## Description

Ticket Reference: DTCORE-3168

### DIFF
--- a/packages/manager/apps/pci-cold-archive/src/pages/containers/new/New.page.tsx
+++ b/packages/manager/apps/pci-cold-archive/src/pages/containers/new/New.page.tsx
@@ -45,7 +45,7 @@ export default function ContainerNewPage() {
     COLD_ARCHIVE_TRACKING.CONTAINERS.ADD_CONTAINER,
   );
 
-  const { addError, addSuccess } = useNotifications();
+  const { addError, addSuccess, clearNotifications } = useNotifications();
 
   const { ovhSubsidiary } = useContext(ShellContext).environment.getUser();
   const pricesLink =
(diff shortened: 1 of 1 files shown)


FabienHenon (approved):

packages/manager/apps/pci-cold-archive/src/pages/containers/new/New.page.tsx
```diff
@@ -55,9 +55,14 @@ export default function ContainerNewPage() {
 
   const goBack = () => navigate('..');
 
+  useEffect(() => {
+    clearNotifications();
+  }, [project]);
```
frenautvh, line 60: is project necessary as a dependency of useEffect here ?

packages/manager/apps/pci-cold-archive/src/pages/containers/new/New.page.tsx
```diff
@@ -55,9 +55,14 @@ export default function ContainerNewPage() {
 
   const goBack = () => navigate('..');
 
+  useEffect(() => {
+    clearNotifications();
```
fredericvilcot, line 59: I can't figure out the necessity of this code, it means that you want to clear notifications every time project changes? 

If so, why just not clearing notifications at unmount only?

kqesar: We need to clear notification when we arrive on this page not before to avoid notifications in double

fredericvilcot: I see the point, so can we achieve that with a simple `componentDidMount` lifecycle, without any dependency in the dependency array? 
```suggestion
    useEffect(() => {
     clearNotifications();
   }, []);
```

fredericvilcot (approved):

FabienHenon (approved):

seven-amid (approved):

PR #15872, opened 2025-03-07
feat(container): standardize contexts and context hooks

aottr:
## Description

The usage of useContext did not follow any standards and did not prevent the dev to use the context hooks outside the provider

Ticket Reference: #MANAGER-17283

### DIFF
--- a/packages/manager/apps/container/src/container/nav-reshuffle/index.tsx
+++ b/packages/manager/apps/container/src/container/nav-reshuffle/index.tsx
@@ -1,14 +1,8 @@
-import React, {
-  Suspense,
-  useContext,
-  useEffect,
-  useRef,
-  useState,
-} from 'react';
+import React, { Suspense, useEffect, useRef, useState } from 'react';
 
 import { IFrameMessageBus } from '@ovh-ux/shell';
 import { IFrameAppRouter } from '@/core/routing';
-import ApplicationContext from '@/context';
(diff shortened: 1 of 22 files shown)


packages/manager/apps/container/src/container/nav-reshuffle/index.tsx
```diff
@@ -26,7 +20,7 @@ function NavReshuffleContainer(): JSX.Element {
   const iframeRef = useRef(null);
   const [iframe, setIframe] = useState(null);
   const [showOverlay, setShowOverlay] = useState(false);
-  const { shell } = useContext(ApplicationContext);
+  const shell = useShell();
```
JacquesLarique, line 23:
```suggestion
  const { shell } = useApplication();
```
Shouldn't we use the `useApplication` hook (since it basically does the same thing) ?

aottr: Yes, I very much agree and thought about that. This would require to rewrite some of the tests though, because they explicitly mock the `useShell` hook.

JacquesLarique: As seen together, let's tackle this in your other ticket

JacquesLarique (approved):

oalkabouss (approved):

PR #15883, opened 2025-03-10
Missing traduction

kb-ovh:
## Description

Ticket Reference: #Manager 17353


ghyenne (approved):

ghyenne (changes requested):
Can you add the ref of ticket in your commit 
ref: #MANAGER-17353

PR #15908, opened 2025-03-10
fix(dedicated): prevent breaking words

sefiani-ovh:
## Description

ref: #MANAGER-17147

### DIFF
--- a/packages/manager/apps/dedicated/client/app/components/dedicated-cloud/dashboard/tiles/options/options.html
+++ b/packages/manager/apps/dedicated/client/app/components/dedicated-cloud/dashboard/tiles/options/options.html
@@ -13,6 +13,7 @@
         >
             <oui-tile-description>
                 <div
+                    class="text-break"
                     data-translate="ovhManagerPccDashboardOptions_security_drp"
                 ></div>
 


packages/manager/apps/dedicated/client/app/components/dedicated-cloud/dashboard/dedicatedCloud-dashboard.scss
```diff
@@ -1,3 +1,9 @@
+* {
```
rjamet-ovh, line 1: you should avoid `*` css rules here

packages/manager/apps/dedicated/client/app/components/dedicated-cloud/dashboard/tiles/options/options.html
```diff
@@ -13,6 +13,7 @@
         >
             <oui-tile-description>
                 <div
+                    class="no-break-word"
```
tibs245, line 16: Possibility use default bootstrap classes ? 

`text-break` or `test-nowrap` ?

https://getbootstrap.com/docs/4.4/utilities/text/#text-wrapping-and-overflow

tibs245 (approved):

PR #15917, opened 2025-03-11
refactor(pci-gateway): Combine catalog and product availability call to get easily addons and regions

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-2990

Combine catalog and product availability call and :
- create hook to get easily all gateway addons with regions available
- create a hook to get all sizes available for the selected region

### DIFF
--- a/packages/manager/apps/pci-gateway/src/api/hooks/useGateways/useGateways.ts
+++ b/packages/manager/apps/pci-gateway/src/api/hooks/useGateways/useGateways.ts
@@ -1,0 +1,66 @@
+import { useMemo } from 'react';
+import {
+  getCatalogQuery,
+  getProductAvailabilityQuery,
+} from '@ovh-ux/manager-pci-common';
+import { useQueries } from '@tanstack/react-query';
+import { GATEWAY_ADDON_FAMILY, SIZE_ORDER } from './useGateways.constant';
+import { TAddonRegions, TProductAddonDetail } from '@/types/addon.type';
+
+export const useGatewayHourlyAddons = (
+  ovhSubsidiary: string,
+  projectId: string,
+) =>
+  useQueries({
+    queries: [
+      getCatalogQuery(ovhSubsidiary),
+      getProductAvailabilityQuery(projectId, ovhSubsidiary, {
+        addonFamily: GATEWAY_ADDON_FAMILY,
+      }),
+    ],
+    combine: ([
+      { data: catalog, isFetching: isCatalogFetching },
+      { data: plans, isFetching: isPlansFetching },
+    ]): { addons: TAddonRegions; isFetching: boolean } => {
+      const addons = (plans?.plans ?? [])
+        .flatMap(({ code, regions }) => {
+          const addon = catalog?.addons.find(
+            ({ planCode }) => code === planCode,
+          );
+          return addon ? [{ ...addon, regions }] : [];
+        })
+        .filter((addon) =>
+          addon.pricings?.find(({ intervalUnit }) =>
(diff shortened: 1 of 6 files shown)


packages/manager/apps/pci-gateway/src/api/hooks/useGateways/useGateways.ts
```diff
@@ -0,0 +1,66 @@
...
+        .flatMap(({ code, regions }) => {
+          const addon = catalog?.addons.find(
+            ({ planCode }) => code === planCode,
+          );
+          return addon ? [{ ...addon, regions }] : [];
+        })
+        .filter((addon) =>
+          addon.pricings?.find(({ intervalUnit }) =>
+            ['none', 'hour'].includes(intervalUnit),
+          ),
```
SimonChaumet, line 35:
```suggestion
        .map(({ code, regions }) => {
          return catalog?.addons.find(
            ({ planCode }) => code === planCode,
          );
        })
        .filter((addon) =>
          addon && addon.pricings?.find(({ intervalUnit }) =>
            ['none', 'hour'].includes(intervalUnit),
          ),
```

packages/manager/apps/pci-gateway/src/types/region.ts
```diff
@@ -3,3 +3,11 @@ export enum RegionType {
   '3AZ' = 'region-3-az',
   LZ = 'localzone',
 }
+
+export type TProductAvailabilityRegion = {
```
SimonChaumet, line 7: you can import the type ProductAvailbility and get the region inside for now

packages/manager/apps/pci-gateway/src/api/hooks/useGateways/useGateways.ts
fredericvilcot: IMO is more elegant to use `flatMap` and to not map uncesseary `undefined` values that need to be filtered afterwards.

packages/manager/apps/pci-gateway/src/api/hooks/useGateways/useGateways.ts
```diff
@@ -0,0 +1,66 @@
...
+        );
+
+      return {
+        addons,
+        isFetching: isCatalogFetching || isPlansFetching,
+      };
+    },
+  });
+
+export const useRegionGatewayAddons = (
```
fredericvilcot, line 45: This hooks has no extra logic value as it could be a plain `select` function given to `useGatewayHourlyAddons` hook as parameter

packages/manager/apps/pci-gateway/src/api/hooks/useGateways/useGateways.ts
Tsiorifamonjena: I agree @fredericvilcot. If you have no objection @SimonChaumet I will let the `flatMap`. Otherwise we can discuss

packages/manager/apps/pci-gateway/src/types/region.ts
Tsiorifamonjena: Yes, I change it. (I will re fix it when psi-common export the good type)

packages/manager/apps/pci-gateway/src/api/hooks/useGateways/useGateways.ts
Tsiorifamonjena: As we discussed, I will let the hook for the moment since the `select` shouldn't get the `region` as params

SimonChaumet (approved):

fredericvilcot (approved):

PR #15936, opened 2025-03-12
fix(ovh-at-internet): force trackpage to use page params

mhelhali-soufien:
ref: #MANAGER-17275

## Description

Ticket Reference: #MANAGER-17275

### DIFF
--- a/packages/components/ovh-at-internet/src/ovh-at-internet.ts
+++ b/packages/components/ovh-at-internet/src/ovh-at-internet.ts
@@ -308,10 +308,14 @@ export default class OvhAtInternet extends OvhAtInternetConfig {
   }
 
   trackPage(data: LegacyTrackingData): void {
+    const trackingData = data;
+    if (trackingData?.page?.name) {
+      trackingData.name = trackingData.page.name;
+    }
     if (this.canTrack()) {
       const tracking = {
-        ...this.getGenericTrackingData(data),
-        ...getPageTrackingData(data?.page || data),
+        ...this.getGenericTrackingData(trackingData),
+        ...getPageTrackingData(trackingData),
       };
       if (tracking.page) {
         this.sendEvent('page.display', filterTrackingData(tracking));


packages/components/ovh-at-internet/src/ovh-at-internet.ts
```diff
@@ -308,10 +308,14 @@ export default class OvhAtInternet extends OvhAtInternetConfig {
   }
 
   trackPage(data: LegacyTrackingData): void {
+    const trackingData = data;
+    if (trackingData.page && !trackingData.page.name) {
+      trackingData.page = { name: trackingData.page.name };
+    }
```
anooparveti, line 314: **question:** Does not this override the `page` property with just name? Does not it cause loss of other information?

packages/components/ovh-at-internet/src/ovh-at-internet.ts
```diff
@@ -308,10 +308,14 @@ export default class OvhAtInternet extends OvhAtInternetConfig {
   }
 
   trackPage(data: LegacyTrackingData): void {
+    const trackingData = data;
+    if (trackingData.page && !trackingData.page.name) {
+      trackingData.page = { name: trackingData.page.name };
```
anooparveti, line 313: Also, it does not make sense to check `!trackingData.page.name` and then resetting the value to undefined/null/false as the code block is only executed when name does not exist on page.

packages/components/ovh-at-internet/src/ovh-at-internet.ts
```diff
@@ -308,10 +308,14 @@ export default class OvhAtInternet extends OvhAtInternetConfig {
   }
 
   trackPage(data: LegacyTrackingData): void {
+    const trackingData = data;
+    if (trackingData.page && trackingData.page.name) {
+      trackingData.name = trackingData.page.name;
+    }
```
anooparveti, line 314:
```suggestion
    if (trackingData?.page?.name) {
      trackingData.name = trackingData.page.name;
    }
```

anooparveti (approved):

qpavy (approved):

packages/components/ovh-at-internet/src/ovh-at-internet.ts
```diff
@@ -308,10 +308,14 @@ export default class OvhAtInternet extends OvhAtInternetConfig {
...
+    const trackingData = data;
+    if (trackingData?.page?.name) {
+      trackingData.name = trackingData.page.name;
+    }
     if (this.canTrack()) {
       const tracking = {
-        ...this.getGenericTrackingData(data),
-        ...getPageTrackingData(data?.page || data),
+        ...this.getGenericTrackingData(trackingData),
+        ...getPageTrackingData(trackingData),
```
darsene, line 318: All those tracking functions use `getPageTrackingData` : 
- trackClick
- trackEvent
- trackMixCommanderS3
- trackPage

So why not do it directly into the function `getPageTrackingData` ?
I find all the calls with different parameters confusing and not really maintanable

PR #15952, opened 2025-03-12
feat(pci-common): add deployment tiles

SimonChaumet:
ref: TAPC-2442

## Description

Ticket Reference: #...


packages/manager/modules/manager-pci-common/src/api/data/regions.ts
```diff
@@ -1,10 +1,16 @@
...
+
+export type TRegionBase = {
   name: string;
-  type: 'region' | 'localzone' | 'region-3-az' | string;
+  type: TRegionType;
+  availabilityZones: string[];
+  continentCode: 'ASIA' | 'EU' | 'NA' | 'US' | string;
+};
+
+export type TRegion = TRegionBase & {
```
fredericvilcot, line 12: A bit strange to split these two types, as a `Region` is made of all these properties.
Maybe we could consider using `Omit` to create a `TProductAvailabilityRegion`

packages/manager/modules/manager-pci-common/src/api/hook/useFeaturedDeploymentModes.ts
```diff
@@ -0,0 +1,41 @@
+import { useMemo } from 'react';
+import { TRegion } from '@/api/data';
+import { Deployment } from '@/components/deployment-tiles-input';
```
fredericvilcot: We should import a type from an UI component in the business logic folder

packages/manager/modules/manager-pci-common/src/api/hook/usePCIFeatureAvailability.ts
```diff
@@ -0,0 +1,30 @@
+import { useQuery } from '@tanstack/react-query';
+import { apiClient } from '@ovh-ux/manager-core-api';
+import { useApplicationName } from '@/hooks/useApplicationName/useApplicationName';
+
+type FeatureAvailabilityQueryParams = Partial<{
```
fredericvilcot, line 5: If there is only one property the `Partial` looks useless

packages/manager/modules/manager-pci-common/src/api/hook/usePCIFeatureAvailability.ts
```diff
@@ -0,0 +1,30 @@
...
+  app: string;
+}>;
+
+const getFeatureAvailability = (
```
fredericvilcot, line 9: Could be moved into the data folder

packages/manager/modules/manager-pci-common/src/api/hook/usePCIFeatureAvailability.ts
```diff
@@ -0,0 +1,30 @@
...
+  features: string[],
+  params?: FeatureAvailabilityQueryParams,
+) =>
+  apiClient.aapi.get(`/feature/${features.join(',')}/availability`, {
+    params,
+  });
+
+const useFeatureAvailability = <T extends string[]>(
```
fredericvilcot, line 17: We could narrow the type as const :
```suggestion
const useFeatureAvailability = <T extends readonly string[]>(
```

packages/manager/modules/manager-pci-common/src/api/hook/usePCIFeatureAvailability.ts
```diff
@@ -0,0 +1,30 @@
...
+  features: T,
+  params?: FeatureAvailabilityQueryParams,
+) =>
+  useQuery({
+    queryKey: ['feature-availability', ...features, params],
+    queryFn: () => getFeatureAvailability(features, params),
+    select: ({ data }) => new Map(features.map((f) => [f, data[f] ?? false])),
```
fredericvilcot, line 24: We could prevent `data[f]` to be `any`:

```suggestion
    select: ({ data }: { data: Record<T[number], boolean> })) => new Map(features.map((f) => [f, data[f] ?? false])),
```

packages/manager/modules/manager-pci-common/src/components/deployment-mode-card/DeploymentModeCard.tsx
```diff
@@ -0,0 +1,110 @@
...
+  beta?: boolean;
+  comingSoon?: boolean;
+  leastPrice?: ReactNode;
+};
+
+export type DeploymentModeCardProps = Readonly<
+  {
+    labelId?: string;
+    ariaDetailsId?: string;
+  } & TDeployment
```
fredericvilcot, line 26: `Readonly` marker has no effect on `TDeployment` object as it's not recursive

packages/manager/modules/manager-pci-common/src/components/deployment-mode-card/DeploymentModeCard.tsx
```diff
@@ -0,0 +1,101 @@
...
+      <div className="deployment-mode-card__content">
+        <div className="deployment-mode-card__header">
+          <OsdsText
+            color={ODS_THEME_COLOR_INTENT.text}
+            level={ODS_THEME_TYPOGRAPHY_LEVEL.body}
+            size={ODS_THEME_TYPOGRAPHY_SIZE._400}
+            className={clsx('leading-8')}
+            id={labelId}
+          >
+            {t(`deployment_mode_title_${name}`)}
```
fredericvilcot, line 50: Should be `deployment_mode_card_${name}` to keep consistency

packages/manager/modules/manager-pci-common/src/components/deployment-tiles-input/DeploymentTilesInputForApp.tsx
```diff
@@ -0,0 +1,32 @@
...
+  DeploymentTilesInputProps,
+} from '@/components/deployment-tiles-input/DeploymentTilesInput';
+import { useFeaturedDeploymentModes } from '@/api/hook/useFeaturedDeploymentModes';
+
+export type DeploymentTilesInputForAppProps = Omit<
+  DeploymentTilesInputProps,
+  'deployments'
+>;
+
+export const DeploymentTilesInputForApp = ({
```
fredericvilcot, line 15: This one is an abstraction for µapps, but I feel a bit confused with the name.

BTW, we could consider to move it out of components folder or to find a way to export these kinds of components.

packages/manager/modules/manager-pci-common/src/hooks/useApplicationName/useApplicationName.ts
```diff
@@ -0,0 +1,20 @@
...
+    meta &&
+    'applicationName' in meta &&
+    typeof meta.applicationName === 'string'
+  ) {
+    return meta.applicationName;
+  }
+  if (shellContext) {
+    return shellContext.environment.applicationName;
+  }
+  return 'public-cloud';
```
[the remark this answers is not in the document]
SimonChaumet: By default app is ignored in FF, so the idea was to fallback to whole public cloud if app wasn't provided, but we could thing of something else or remove this

It is an edge case so either is good

packages/manager/modules/manager-pci-common/src/api/data/regions.ts
SimonChaumet: The best would be to rename TRegion but would cause a breaking change

Omit isn't a good idea since some of these properties are generated by backend and not related to the object itself, Pick could do the trick also

packages/manager/modules/manager-pci-common/src/components/deployment-mode-card/DeploymentModeCard.tsx
```diff
@@ -0,0 +1,101 @@
...
+} from '@ovhcloud/ods-common-theming';
+import { ODS_CHIP_SIZE } from '@ovhcloud/ods-components';
+import { clsx } from 'clsx';
+import './style.scss';
+import '@/translations/deployment-mode';
+import { ReactNode } from 'react';
+import { RegionChipByType } from '@/components/region-selector/RegionChipByType';
+
+export type TDeployment = {
+  name: string;
```
Tsiorifamonjena, line 16: Maybe name is `TRegionType` here ?

packages/manager/modules/manager-pci-common/src/components/deployment-tiles-input/DeploymentTilesInputForApp.tsx
Tsiorifamonjena, line 15: I don't sure to understand the naming

packages/manager/modules/manager-pci-common/src/api/hook/useFeaturedDeploymentModes.ts
SimonChaumet: The typing is only related to UI, this call is only a helper on it since it is not a real domain object, or I could put it inside a DTO folder ?

packages/manager/modules/manager-pci-common/src/api/hook/usePCIFeatureAvailability.ts
SimonChaumet: It is to mark all properties partial since they are query params, today there is only but there could more tomorrow and other devs might not think of marking it as not required

Tsiorifamonjena:
great work 🫡

packages/manager/modules/manager-pci-common/src/api/hook/usePCIFeatureAvailability.ts
SimonChaumet: Why only this property ?

packages/manager/modules/manager-pci-common/src/api/hook/usePCIFeatureAvailability.ts
SimonChaumet: data isn't any, the type comes from getFeatureAvailability

packages/manager/modules/manager-pci-common/src/components/deployment-tiles-input/DeploymentTilesInputForApp.tsx
SimonChaumet: Yes, I don't know what is the best way for this, maybe with the `Featued` name as in the hook ?

packages/manager/modules/manager-pci-common/src/components/deployment-mode-card/DeploymentModeCard.tsx
fredericvilcot: AFAIS, `leastPrice ` is a React Node that should be recursively marked as `readonly`

packages/manager/modules/manager-pci-common/src/api/hook/usePCIFeatureAvailability.ts
fredericvilcot: Doing this will infer T as the union instead of simple strings

packages/manager/modules/manager-pci-common/src/api/hook/useFeaturedDeploymentModes.ts
fredericvilcot: Yes please, to follow the separation of concerns principle

packages/manager/modules/manager-pci-common/src/api/hook/usePCIFeatureAvailability.ts
fredericvilcot: Here, `getFeatureAvailability` returns a `Promise<AxiosResponse<any, any>>

packages/manager/modules/manager-pci-common/src/components/deployment-tiles-input/FeaturedDeploymentTilesInput.tsx
```diff
@@ -0,0 +1,32 @@
+import React from 'react';
+import { ODS_SPINNER_SIZE } from '@ovhcloud/ods-components';
+import { OsdsSpinner } from '@ovhcloud/ods-components/react';
+import {
+  DeploymentTilesInput,
+  DeploymentTilesInputProps,
+} from '@/components/deployment-tiles-input/DeploymentTilesInput';
+import { useFeaturedDeploymentModes } from '@/api/hook/useFeaturedDeploymentModes';
+
+export type DeploymentTilesInputForAppProps = Omit<
```
Tsiorifamonjena, line 10: Why does not change the type name too just to be related with the component ?

SimonChaumet: I missed this part, will fix

packages/manager/modules/manager-pci-common/src/hooks/useHas3AZ/useHas3AZ.ts
```diff
@@ -1,10 +1,19 @@
 import { useContext } from 'react';
-import { PCICommonContext } from '../../contexts/PCICommonContext/PCICommonContext';
+import {
+  hasMetaProperty,
+  InternalMeta,
+  PCICommonContext,
+  PCICommonMetaType,
+} from '@/contexts/PCICommonContext/PCICommonContext';
+
+function hasMetaHas3AZ(
```
frenautvh, line 9: nitpick: the double "has" in the function name is weird, i would suggest renaming it to something more straightforward such as "has3AZ"

SimonChaumet: I can rename it to `hasMetaPropertyHas3AZ` if it is better but renaming it has3AZ wouldn't be the same function

packages/manager/modules/manager-pci-common/src/dto/deployment.ts
```diff
@@ -0,0 +1,9 @@
+import { ReactNode } from 'react';
+import { TRegionType } from '@/api/data';
+
+export type TDeployment = {
+  name: TRegionType;
+  beta?: boolean;
+  comingSoon?: boolean;
+  price?: ReactNode;
```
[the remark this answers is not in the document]
SimonChaumet: Price isn't something fixed, sometimes it is a least price, sometimes it could be the price itself, will check if we can have a common usage on it

packages/manager/modules/manager-pci-common/src/api/hook/usePCIFeatureAvailability.ts
```diff
@@ -0,0 +1,24 @@
...
+  FeatureAvailabilityQueryParams,
+  getFeatureAvailability,
+} from '@/api/data/feature-availability';
+
+const useFeatureAvailability = <T extends string = string>(
+  features: T[],
+  params?: FeatureAvailabilityQueryParams,
+) =>
+  useQuery({
+    queryKey: ['feature-availability', ...features, params],
```
frenautvh, line 13: i think the spread of `...features` is wrong in terms of query key and potential query cache invalidation since a join is performed to build the query on the api side

SimonChaumet: It's a copy from pci-common, I don't know how this would cause an issue

packages/manager/modules/manager-pci-common/src/components/input-adapter/radio-adapter/RadioField.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { ODS_TEXT_LEVEL, ODS_TEXT_SIZE } from '@ovhcloud/ods-components';
+import { ODS_THEME_COLOR_INTENT } from '@ovhcloud/ods-common-theming';
+import { OsdsText } from '@ovhcloud/ods-components/react';
+
+export type RadioFieldProps = PropsWithChildren<{
+  label: ReactNode;
+  subtitle?: ReactNode;
+}>;
+
+export const RadioField: FC<RadioFieldProps> = ({
```
frenautvh, line 12: it doesn't look specific to pci to me, shouldn't this be in manager react components ?

packages/manager/modules/manager-pci-common/src/components/deployment-mode-card/DeploymentModeCard.tsx
```diff
@@ -0,0 +1,94 @@
...
+            level={ODS_THEME_TYPOGRAPHY_LEVEL.body}
+            size={ODS_THEME_TYPOGRAPHY_SIZE._400}
+            className={clsx('leading-8')}
+            id={labelId}
+          >
+            {t(`deployment_mode_card_title_${name}`)}
+          </OsdsText>
+        </div>
+
+        <div id={ariaDetailsId} className="deployment-mode-card__content">
```
seven-amid, line 47: You can use a flex gap on the parent div instead of using the same mb-3 on every child

packages/manager/modules/manager-pci-common/src/api/data/feature-availability.ts
```diff
@@ -0,0 +1,25 @@
...
+export type FeatureAvailabilityResponse<T extends string = string> = {
+  [Key in T]?: boolean;
+} & {
+  /* @deprecated */
+  feature: string;
+  /* @deprecated */
+  result: boolean;
+};
+
+export const getFeatureAvailability = <T extends string = string>(
```
frenautvh, line 16: i don't think this is specific to pci common this should belong in MRC

packages/manager/modules/manager-pci-common/src/components/deployment-mode-card/style.scss
```diff
@@ -0,0 +1,34 @@
+.deployment-mode-card {
+  @apply border-[--ods-color-blue-100];
+  @apply w-full px-[24px] py-[16px];
+  min-height: 200px;
```
frenautvh, line 4: it is preferable to use rem instead of pixels

packages/manager/modules/manager-pci-common/src/components/deployment-mode-card/style.scss
```diff
@@ -0,0 +1,34 @@
...
+
+  .deployment-mode-card__content {
+    @apply flex flex-col w-full items-center;
+  }
+
+  .deployment-mode-card__header {
+    @apply mb-4;
```
frenautvh, line 11: shouldn't you use tailwind classes directly, if not you are kind of losing the advantage of using tailwind

packages/manager/modules/manager-pci-common/src/components/input-adapter/radio-adapter/RadioField.tsx
SimonChaumet: Yes, most of it should be in it but now we can't add more features to MRC in ODS17, only fix, as far as I know

packages/manager/modules/manager-pci-common/src/api/data/feature-availability.ts
SimonChaumet: We can't add more features to MRC in ODS17

packages/manager/modules/manager-pci-common/src/components/deployment-mode-card/style.scss
SimonChaumet: Tailwind can't handle css selectors so we need to have css, so either I duplicate everything that tailwind is doing in CSS, or I have a mix of tailwind inside react components and other styles inside css

packages/manager/modules/manager-pci-common/src/components/input-adapter/radio-adapter/RadioField.tsx
frenautvh: so you are pushing new features to the maintenance branch of pci-common, shouldn't the branch be used for maintenance only ? do you plan to backport thoses changes to the develop branch as well ?

packages/manager/modules/manager-pci-common/src/components/deployment-tiles-input/DeploymentTilesInput.tsx
```diff
@@ -0,0 +1,79 @@
...
+        disabled: e.comingSoon,
+      })}
+      label={t('deployment_mode_title')}
+      subtitle={
+        <Trans
+          t={t}
+          i18nKey="deployment_mode_description"
+          components={{
+            Link: (
+              <OsdsLink
```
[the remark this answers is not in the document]
SimonChaumet: Could you provide design for this ?

Tsiorifamonjena (dismissed):

Tsiorifamonjena (dismissed):

fredericvilcot (approved):

frenautvh (dismissed):

Eric-ciccotti (approved):

Tsiorifamonjena (approved):

fredericvilcot (approved):

fredericvilcot (approved):

PR #15967, opened 2025-03-12
feat(pci-instances): remove disable property for actions

fredericvilcot:
## Description

Ticket Reference: #...


fredericvilcot:
Duplicate of #15960

PR #15977, opened 2025-03-13
feat(hub): display notifications when email unreachable

oalkabouss:
ref: #MANAGER-17027

## Description

Hub: display notifications when email unreachable

Ticket Reference: #MANAGER-17027


packages/manager/apps/hub-react/src/pages/layout/layout.tsx
```diff
@@ -138,8 +139,9 @@ export default function Layout() {
                   <Welcome />
                   <Banner />
                   {/* I doubt this height as we will have dots when we have more than 1 notification */}
-                  <div className="h-[56px]">
+                  <div>
```
JacquesLarique, line 142: If we remove this class, we should remove the comment as well.

Since this class was setup in order to improve the WCV, let's check that removing this class does not degrade them

packages/manager/apps/hub-react/src/pages/layout/NotificationsEmailUnreachable.component.tsx
MaximeBajeux, line 1: Are check failures false positives?

MaximeBajeux (dismissed):

JacquesLarique (approved):

JacquesLarique (dismissed):

aottr (dismissed):

MaximeBajeux (dismissed):

JacquesLarique (approved):

MaximeBajeux (approved):

packages/manager/modules/account/src/user/components/newAccountForm/field/new-account-form-field-component.html
```diff
@@ -1,4 +1,5 @@
 <div
+    data-ng-attr-id="{{ 'ovh_field_group_' + $ctrl.rule.fieldName }}"
```
JacquesLarique, line 2: This does not make much sense since the field already has an id and would add to the (already) long list of watchers.
I would recommend to focus the section `signup_section_contact`, since `email` is the first field of the section it would give the same result without deteriorating the performances

JacquesLarique (changes requested):

packages/manager/modules/account/src/user/components/newAccountForm/field/new-account-form-field-component.html
oalkabouss: i had focused on the email field because it is possible that the email field's position may change. I have fixed it, now i focus on the section with an already existing id

JacquesLarique (approved):

aottr (dismissed):

JacquesLarique (approved):

MaximeBajeux (approved):

PR #15985, opened 2025-03-13
refactor(pci-gateway): Remove useless hook call

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-2990

This PR contain clean code

- remove useless hook and effect call since there is new hook to get catalog and available addons on a selected region
- clean props or type which is no longer used


packages/manager/apps/pci-gateway/src/pages/add/SizeStep.tsx
```diff
@@ -165,6 +151,7 @@ export const SizeStep = ({
         )}
         onInput={(item) => {
           store.updateForm.size(item.size);
+          setSize(addons.find((addon) => addon.size === item.size));
```
fredericvilcot, line 154: Could be moved to a `handleInput` function.

fredericvilcot (approved):

PR #16010, opened 2025-03-14
feat(pci-volume-snapshot): initialize react app

MarcMonchablon:
ref: DTCORE-3208

## Description

Ticket Reference: DTCORE-3208

### DIFF
--- a/packages/manager/apps/pci-volume-snapshot/src/core/HidePreloader.tsx
+++ b/packages/manager/apps/pci-volume-snapshot/src/core/HidePreloader.tsx
@@ -1,0 +1,12 @@
+import { ShellContext } from '@ovh-ux/manager-react-shell-client';
+import { useContext, useEffect } from 'react';
+
+export default function HidePreloader() {
+  const { ux } = useContext(ShellContext).shell;
+
+  useEffect(() => {
+    ux.hidePreloader();
+  }, []);
+
+  return null;
+}
--- a/packages/manager/apps/pci-volume-snapshot/src/helpers/index.ts
+++ b/packages/manager/apps/pci-volume-snapshot/src/helpers/index.ts
@@ -1,0 +1,39 @@
+import { PaginationState } from '@ovh-ux/manager-react-components';
+import { ColumnSort } from '@tanstack/react-table';
+
+export const paginateResults = <T>(
+  items: T[],
+  pagination: PaginationState,
+) => ({
+  rows: items.slice(
+    pagination.pageIndex * pagination.pageSize,
+    (pagination.pageIndex + 1) * pagination.pageSize,
+  ),
+  pageCount: Math.ceil(items.length / pagination.pageSize),
+  totalRows: items.length,
+});
+
+export const compareFunction = <T>(key: keyof T) => (a: T, b: T) => {
+  const aValue = a[key] || '';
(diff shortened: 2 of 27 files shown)


kqesar (approved):

FabienHenon (approved):

packages/manager/apps/pci-volume-snapshot/src/core/HidePreloader.tsx
```diff
@@ -0,0 +1,12 @@
+import { ShellContext } from '@ovh-ux/manager-react-shell-client';
```
fredericvilcot, line 1: Maybe we can imagine rework these two components (HidePreloader & ShellRoutingSync) that are not real components and turn them into hooks.

We have made this transformation into future `pci-instances` µapp, I can push them into `pci-common` if you want :)

packages/manager/apps/pci-volume-snapshot/tsconfig.json
```diff
@@ -0,0 +1,27 @@
+{
+  "compilerOptions": {
```
fredericvilcot, line 2: I will suggest to take the opportunity to turn on the `strict` mode to fine tune ts engine :)

packages/manager/apps/pci-volume-snapshot/src/pages/Layout.tsx
```diff
@@ -0,0 +1,55 @@
...
+            <Outlet />
+          </>
+        )}
+      </Suspense>
+    </div>
+  );
+}
+
+export const ErrorBoundary = () => {
+  const error = useRouteError() as ApiError;
```
fredericvilcot, line 32: This cast assertion could be misleading as router can intercept any kind of Error and not only Api ones that don't share the same structure.

An idea to enhance that could be a type guard that check the error itself before consuming it into the `ErrorBoundary`  component.

I implemented it as well in `pci-instances`, tell me if it has an interest to push it into `pci-common` :)

packages/manager/apps/pci-volume-snapshot/src/helpers/index.ts
```diff
@@ -0,0 +1,39 @@
+import { PaginationState } from '@ovh-ux/manager-react-components';
```
fredericvilcot, line 1: We could consider adding a test suite for these helper functions.

packages/manager/apps/pci-volume-snapshot/src/pages/Layout.tsx
FabienHenon: Sure, thank you :)

fredericvilcot: @FabienHenon @MarcMonchablon 
Additions in `pci-common` can be reviewed here: https://github.com/ovh/manager/pull/16021

vovh (approved):

packages/manager/apps/pci-volume-snapshot/src/pages/Layout.tsx
MarcMonchablon: If that's ok with you, I propose that we merge this PR as-is, and create a new ticket to use ErrorBoundary from pci-common once it's merged.

PR #16013, opened 2025-03-14
fix(release.sh): revert lerna pre-installed changes

hbelkhir:
ref: MANAGER-17160

## Description

revert lerna pre-installed changes

Ticket Reference: #...


anooparveti:
Bypassing checks as it is just a CI change to revert the recent changes.

anooparveti (approved):

PR #16057, opened 2025-03-18
feat(pci.kubernetes): add deployement selector

Eric-ciccotti:
## Description

Ticket Reference: #...


packages/manager/apps/pci-kubernetes/src/components/region-selector/KubeRegionSelector.css
```diff
@@ -0,0 +1,4 @@
+hr {
```
SimonChaumet, line 1: You might want something more restrictive to select the element, or put this into the project css file

PR #16093, opened 2025-03-19
feat(pci.public-ip): display the smallest price of a floating ip

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-3076

image

## Additional Information

this PR depend on https://github.com/ovh/manager/pull/16090


packages/manager/apps/pci-public-ip/src/pages/order/steps/IpTypeStep.tsx
```diff
@@ -1,37 +1,128 @@
...
+    addonFamily: 'floatingip',
+    select: sortProductByPrice,
+  });
+
+  const { getFormattedHourlyCatalogPrice } = useCatalogPrice(4);
+
+  const selectPublicIp = (value: TPublicIp) => {
+    setPublicIp(value);
+    setForm({ ...form, ipType: value.type });
+  };
```
SimonChaumet, line 73: I don't understand the goal to have 2 states if they have the same values each time, is there another usage ?

SimonChaumet (approved):

packages/manager/apps/pci-public-ip/src/pages/order/steps/IpTypeStep.tsx
Tsiorifamonjena: the form is the zustand store that I really want but I had to use a local state for the MRC `TilesInput` since it only takes as value the same type as its data items

fredericvilcot (approved):

PR #16101, opened 2025-03-19
feat(hub): filter data for us

MaximeBajeux:
## Description

Some data on the new changelog page don't exist for US, so we hide them.

Ticket Reference: #MANAGER-17091

### DIFF
--- a/packages/manager/apps/container/src/container/nav-reshuffle/sidebar/Assistance/index.tsx
+++ b/packages/manager/apps/container/src/container/nav-reshuffle/sidebar/Assistance/index.tsx
@@ -116,11 +116,6 @@ const AssistanceSidebar: React.FC<ComponentProps<AssistanceProps>> = ({
 
   return (
     <ul className="mt-auto pb-3 flex-none" id="useful-links" role="menu" data-testid="assistance-sidebar">
-      <li className="assistance_header px-3 mb-3">
-        <h2 className="flex justify-between">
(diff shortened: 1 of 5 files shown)


packages/manager/apps/container/src/container/nav-reshuffle/sidebar/Assistance/index.tsx
```diff
@@ -116,11 +116,6 @@ const AssistanceSidebar: React.FC<ComponentProps<AssistanceProps>> = ({
 
   return (
     <ul className="mt-auto pb-3 flex-none" id="useful-links" role="menu" data-testid="assistance-sidebar">
-      <li className="assistance_header px-3 mb-3">
-        <h2 className="flex justify-between">
-          <span>{t('sidebar_assistance_title')}</span>
-        </h2>
-      </li>
```
JacquesLarique, line 123: I don't see the request to remove this for all region, is this correct ?

JacquesLarique (commented):
Tests on container are failing

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/Assistance/index.tsx
MaximeBajeux: Yes confirmed with the PO

JacquesLarique (approved):

PR #16102, opened 2025-03-19
fix(billing): restore sshKeySwitch directive

JacquesLarique:
## Description

Added sshKeySwitch directive that was forgotten during the splicing of billing

Ticket Reference: #MANAGER-17509

### DIFF
--- a/packages/manager/modules/new-billing/src/autoRenew/ssh/sshkeySwitch/sshkeySwitch.js
+++ b/packages/manager/modules/new-billing/src/autoRenew/ssh/sshkeySwitch/sshkeySwitch.js
@@ -1,0 +1,53 @@
+import './sshkeySwitch.css';
+
+const SSH_REGEX = [
+  {
+    name: 'RSA',
+    regex: /\b(ssh-rsa)\s+(A{4}[0-9A-Za-z +/]+[=]{0,3})(\s+.*)?$/,
+  },
+  {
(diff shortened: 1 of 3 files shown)


packages/manager/modules/new-billing/src/autoRenew/ssh/sshkeySwitch/sshkeySwitch.js
```diff
@@ -0,0 +1,53 @@
+import './sshkeySwitch.css';
+
+const SSH_REGEX = [
+  {
+    name: 'RSA',
+    regex: /\b(ssh-rsa)\s+(A{4}[0-9A-Za-z +/]+[=]{0,3})(\s+.*)?$/,
```
aottr, line 6: Are we sure that we want to test for the comment, since we don't use it anyway ?

aottr (approved):

packages/manager/modules/new-billing/src/autoRenew/ssh/sshkeySwitch/sshkeySwitch.js
JacquesLarique: No idea, I just retrieved the already existing code for this directive.
But let's check this with this ticket: MANAGER-17511

oalkabouss (approved):

PR #16104, opened 2025-03-20
feat(billing): handle new 2api renew period correctly

oalkabouss:
ref: #MANAGER-16329

## Description

Billign: handle new 2api renew period correctly
Ensures accurate frequency preselect, display, and tracking

Ticket Reference: #MANAGER-16329

### DIFF
--- a/packages/manager/modules/models/BillingService/BillingService.class.js
+++ b/packages/manager/modules/models/BillingService/BillingService.class.js
@@ -36,6 +36,25 @@ export default class BillingService {
     return this.renew.period;
   }
 
+  /**
+   * Note:
+   * Converts the renewal period from ISO 8601 format to the total number of months.
+   * Only years Y and months M are considered.
+   */
+  get cleanRenewPeriod() {
+    if (this.renew.period && !Number.isInteger(this.renew.period)) {
+      const matches = this.renew.period.match(/P(?:(\d+)Y)?(?:(\d+)M)?/);
+      if (!matches) return this.renew.period;
+
+      const years = Number(matches[1]) || 0;
+      const months = Number(matches[2]) || 0;
+
+      const totalMonths = years * 12 + months;
+      return Number(totalMonths);
+    }
+    return this.renew.period;
+  }
+
   isBillingSuspended() {
     return this.status === 'BILLING_SUSPENDED';
   }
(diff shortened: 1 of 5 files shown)


packages/manager/modules/new-billing/src/autoRenew/autorenew.html
```diff
@@ -217,7 +217,7 @@
             data-type="options"
         >
             <span
-                data-ng-bind-html="$row.hasAutomaticRenew() && !$row.isResiliated() && $row.renewPeriod ? 'billing_autorenew_service_every_' + $row.renewPeriod + '_month': 'billing_autorenew_service_none' | translate"
+                data-ng-bind-html="$row.hasAutomaticRenew() && !$row.isResiliated() && $row.renewPeriod ? 'billing_autorenew_service_every_' + $row.cleanRenewPeriod + '_month': 'billing_autorenew_service_none' | translate"
```
JacquesLarique, line 220: This will display an incorrect value for service with a renew period equal to "P1Y" for example (as it would display "Every month" instead of "Every year")

I think we have a utils to convert any period in this format to a number of months

JacquesLarique (changes requested):

packages/manager/modules/new-billing/src/autoRenew/autorenew.html
oalkabouss: i fixed it, but i could not find a function that meets the specific need of returning the number of months from the ISO 8601 format, and i avoid using the moment.js lib

packages/manager/modules/models/BillingService/BillingService.class.js
```diff
@@ -36,6 +36,25 @@ export default class BillingService {
...
+  get cleanRenewPeriod() {
+    if (this.renew.period && !Number.isInteger(this.renew.period)) {
+      const matches = this.renew.period.match(/P(?:(\d+)Y)?(?:(\d+)M)?/);
+      if (!matches) return this.renew.period;
+
+      const years = Number(matches[1]) || 0;
+      const months = Number(matches[2]) || 0;
+
+      const totalMonths = years * 12 + months;
+      return Number(totalMonths);
```
JacquesLarique, line 53: I don't think this call to Number constructor is necessary (as both `years` and `months` are already Numbers)

JacquesLarique (approved):

MaximeBajeux (approved):

PR #16117, opened 2025-03-20
fix(pci-instances): handle actions refresh & enhance cache set

fredericvilcot:
## Description

Ticket Reference: #...


packages/manager/apps/pci-instances/src/data/hooks/instance/useInstances.ts
```diff
@@ -29,7 +30,10 @@ export type TUseInstancesQueryParams = DeepReadonly<{
 
 export type TUpdateInstanceFromCache = (
   queryClient: QueryClient,
-  payload: { projectId: string; instance: Partial<TInstanceDto> },
+  payload: {
+    projectId: string;
+    instance: { id: string } & Partial<TInstanceDto>;
```
SimonChaumet, line 35:
```suggestion
    instance: Pick<TInstanceDto, 'id'> & Partial<TInstanceDto>;
```

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstances.ts
```diff
@@ -42,26 +46,32 @@ export const updateInstanceFromCache: TUpdateInstanceFromCache = (
...
+
+    const updatedPages = queryData.pages.map((page) =>
+      page.map((prevInstance) =>
+        prevInstance.id === instance.id
+          ? { ...prevInstance, ...instance }
+          : prevInstance,
+      ),
+    );
+
+    if (isEqual(queryData.pages, updatedPages)) return;
```
SimonChaumet, line 65: You could have a findIndex, if found you copy the array and replace the value, if not you return

This would avoid a lot of computations (double map and isEqual which does deep compare)

You could also have a boolean which handles if the condition was true at a time

SimonChaumet (approved):

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstances.ts
fredericvilcot: Yes, I added a deep compare at instance level, that is much more efficient, thx 👍🏻

PR #16129, opened 2025-03-21
feat(pci-volume-snapshot): implement onboarding page UI and functionality and test suite 

seven-amid:
## Description

Ticket Reference: DTCORE-3210

PR #16133, opened 2025-03-21
Feat/dtcore 3209

FabienHenon:
## Description

Listing page for volume snapshot

Ticket Reference: DTCORE-3209

### DIFF
--- a/packages/manager/apps/pci-volume-snapshot/.eslintrc.cjs
+++ b/packages/manager/apps/pci-volume-snapshot/.eslintrc.cjs
@@ -1,0 +1,27 @@
+module.exports = {
+  extends: [
+    '../../../../.eslintrc.js',
+    'plugin:react/recommended',
+    'plugin:react/jsx-runtime',
+    'plugin:react-hooks/recommended-legacy',
+  ],
+  settings: {
+    react: {
+      version: 'detect',
+    },
+  },
+  globals: {
+    __VERSION__: true,
+  },
+  rules: {
+    'import/extensions': 'off',
+    '@typescript-eslint/no-non-null-assertion': 'off',
+    'arrow-body-style': ['error', 'as-needed'],
+    'prefer-const': 'error',
+    'react/jsx-curly-brace-presence': [
+      'error',
+      { props: 'never', children: 'never' },
+    ],
+    'react/jsx-boolean-value': ['error', 'never'],
+  },
+};
--- a/packages/manager/apps/pci-volume-snapshot/package.json
+++ b/packages/manager/apps/pci-volume-snapshot/package.json
@@ -17,7 +17,7 @@
   "dependencies": {
     "@ovh-ux/manager-config": "^8.0.2",
     "@ovh-ux/manager-core-api": "^0.10.1",
-    "@ovh-ux/manager-core-utils": "*",
+    "@ovh-ux/manager-core-utils": "^0.3.0",
     "@ovh-ux/manager-pci-common": "^2.3.1",
     "@ovh-ux/manager-react-components": "^2.15.1",
     "@ovh-ux/manager-react-core-application": "^0.11.13",
@@ -55,6 +55,7 @@
     "@vitest/coverage-v8": "^1.6.0",
     "autoprefixer": "^10.4.19",
     "eslint": "^8.57.0",
+    "eslint-plugin-react-hooks": "^5.2.0",
     "postcss": "^8.4.38",
     "rollup": "^4.18.0",
     "tailwindcss": "^3.4.4",
--- a/packages/manager/apps/pci-volume-snapshot/package.json
+++ b/packages/manager/apps/pci-volume-snapshot/package.json
@@ -17,6 +17,7 @@
   "dependencies": {
     "@ovh-ux/manager-config": "^8.0.2",
     "@ovh-ux/manager-core-api": "^0.10.1",
+    "@ovh-ux/manager-core-utils": "*",
     "@ovh-ux/manager-pci-common": "^2.3.1",
     "@ovh-ux/manager-react-components": "^2.15.1",
     "@ovh-ux/manager-react-core-application": "^0.11.13",
@@ -27,6 +28,7 @@
     "@ovhcloud/ods-themes": "18.5.0",
     "@tanstack/react-query": "^5.51.21",
     "@tanstack/react-table": "^8.20.1",
+    "date-fns": "^3.6.0",
     "element-internals-polyfill": "^1.3.12",
     "i18next": "^23.8.2",
     "i18next-http-backend": "^2.5.2",
--- a/packages/manager/apps/pci-volume-snapshot/src/api/api.types.ts
+++ b/packages/manager/apps/pci-volume-snapshot/src/api/api.types.ts
@@ -1,0 +1,66 @@
+export type TVolumeType =
+  | 'classic'
+  | 'classic-BETA'
+  | 'high-speed'
+  | 'high-speed-BETA'
+  | 'high-speed-gen2';
+
+export type TSnapshotStatus =
+  | 'available'
+  | 'creating'
+  | 'deleting'
+  | 'error'
+  | 'error_deleting';
+
+export type TVolumeStatus =
+  | 'attaching'
+  | 'available'
+  | 'awaiting-transfer'
+  | 'backing-up'
+  | 'creating'
+  | 'deleting'
+  | 'detaching'
+  | 'downloading'
+  | 'error'
+  | 'error_backing-up'
+  | 'error_deleting'
+  | 'error_extending'
+  | 'error_restoring'
+  | 'extending'
+  | 'in-use'
+  | 'maintenance'
+  | 'reserved'
+  | 'restoring-backup'
+  | 'retyping'
+  | 'uploading';
+
+export type TSnapshot = {
+  id: string;
+  creationDate: string;
+  name: string;
+  description: string;
+  size: number;
+  volumeId: string;
+  region: string;
+  status: TSnapshotStatus;
+  planCode: string | null;
+};
+
+export type TVolume = {
+  id: string;
+  attachedTo: string[];
+  creationDate: string;
+  name: string;
+  description: string;
+  size: number;
+  status: TVolumeStatus;
+  region: string;
+  bootable: boolean;
+  planCode: string | null;
+  availabilityZone: string | null;
+  type: TVolumeType;
+};
+
+export type TVolumeSnapshot = TSnapshot & {
+  volume?: TVolume | null;
+};
--- a/packages/manager/apps/pci-volume-snapshot/src/api/data/snapshots.ts
+++ b/packages/manager/apps/pci-volume-snapshot/src/api/data/snapshots.ts
@@ -1,56 +1,5 @@
 import { v6 } from '@ovh-ux/manager-core-api';
-
-export interface TSnapshot {
-  id: string;
-  creationDate: string;
-  name: string;
-  description: string;
-  size: number;
-  volumeId: string;
-  region: string;
-  status: 'available' | 'creating' | 'deleting' | 'error' | 'error_deleting';
-  planCode: string | null;
-}
-
-export interface TVolume {
-  id: string;
-  attachedTo: string[];
-  creationDate: string;
-  name: string;
-  description: string;
-  size: number;
-  status:
-    | 'attaching'
-    | 'available'
-    | 'awaiting-transfer'
-    | 'backing-up'
-    | 'creating'
-    | 'deleting'
-    | 'detaching'
-    | 'downloading'
-    | 'error'
-    | 'error_backing-up'
-    | 'error_deleting'
-    | 'error_extending'
-    | 'error_restoring'
-    | 'extending'
-    | 'in-use'
-    | 'maintenance'
-    | 'reserved'
-    | 'restoring-backup'
-    | 'retyping'
-    | 'uploading';
-  region: string;
-  bootable: boolean;
-  planCode: string | null;
-  availabilityZone: string | null;
-  type:
-    | 'classic'
-    | 'classic-BETA'
-    | 'high-speed'
-    | 'high-speed-BETA'
-    | 'high-speed-gen2';
-}
+import { TSnapshot, TVolume } from '@/api/api.types';
 
 export const getSnapshots = async (projectId: string): Promise<TSnapshot[]> => {
   const { data } = await v6.get(`/cloud/project/${projectId}/volume/snapshot`);
--- a/packages/manager/apps/pci-volume-snapshot/src/api/data/snapshots.ts
+++ b/packages/manager/apps/pci-volume-snapshot/src/api/data/snapshots.ts
@@ -1,0 +1,68 @@
+import { v6 } from '@ovh-ux/manager-core-api';
+
+export interface TSnapshot {
+  id: string;
+  creationDate: string;
+  name: string;
+  description: string;
+  size: number;
+  volumeId: string;
+  region: string;
+  status: 'available' | 'creating' | 'deleting' | 'error' | 'error_deleting';
+  planCode: string | null;
+}
+
+export interface TVolume {
+  id: string;
+  attachedTo: string[];
+  creationDate: string;
+  name: string;
+  description: string;
+  size: number;
+  status:
+    | 'attaching'
+    | 'available'
+    | 'awaiting-transfer'
+    | 'backing-up'
+    | 'creating'
+    | 'deleting'
+    | 'detaching'
+    | 'downloading'
+    | 'error'
+    | 'error_backing-up'
+    | 'error_deleting'
+    | 'error_extending'
+    | 'error_restoring'
+    | 'extending'
+    | 'in-use'
+    | 'maintenance'
+    | 'reserved'
+    | 'restoring-backup'
+    | 'retyping'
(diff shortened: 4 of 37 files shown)


packages/manager/apps/pci-volume-snapshot/src/pages/listing/Listing.page.tsx
```diff
@@ -0,0 +1,231 @@
...
+            <div className="sm:flex">
+              <OdsButton
+                size="sm"
+                color="primary"
+                variant="outline"
+                className="xs:mb-0.5 sm:mb-0"
+                icon="plus"
+                label={t('pci_projects_project_storages_snapshots_add_label')}
+                onClick={() => {
+                  window.location.href = `${hrefProject}/storages/volume-backup/create`;
```
frenautvh, line 113: for external links (links between apps) you can use the shell navigation plugin, exemple of usage here : https://github.com/ovh/manager/blob/master/packages/manager/apps/veeam-backup/src/components/Breadcrumb/Breadcrumb.tsx#L46

since the app is displayed in the container, the correct way to naviage to an external url is to use window.top.location (you don't have to care about this if you are using the shell plugin)

packages/manager/apps/pci-volume-snapshot/src/api/data/snapshots.ts
```diff
@@ -0,0 +1,68 @@
+import { v6 } from '@ovh-ux/manager-core-api';
+
+export interface TSnapshot {
```
fredericvilcot, line 3: We can consider moving types into a dedicated api.types.folder :)

packages/manager/apps/pci-volume-snapshot/src/pages/listing/Status.component.tsx
```diff
@@ -0,0 +1,48 @@
...
+export default function StatusComponent({
+  status,
+}: Readonly<StatusComponentProps>) {
+  const { t } = useTranslation('volumes');
+
+  const statusGroup = ['error', 'error_deleting'].includes(status)
+    ? 'ERROR'
+    : status.toUpperCase();
+
+  const [badgeColor, setBadgeColor] = useState<OdsBadgeColor>('information');
```
frenautvh, line 20: i think you can replace the useState + useEffect with an useMemo, since having a state for badgeColor is not necessary here

packages/manager/apps/pci-volume-snapshot/src/api/hooks/useSnapshots.ts
```diff
@@ -0,0 +1,68 @@
...
+      data: (snapshots || []).map((snapshot) => ({
+        ...snapshot,
+        volume: volumes.find(({ data }) => data?.id === snapshot.volumeId)
+          ?.data,
+      })),
+    }),
+  });
+};
+
+export const usePaginatedVolumeSnapshot = (
```
fredericvilcot, line 43: This should could be a plain `select` function given as parameter to the `useVolumeSnapshots` hook.
This way we could extend the initial hook with agnostic computing behavior without rewriting some hooks that have no direct value :)

packages/manager/apps/pci-volume-snapshot/src/hooks/useFormattedDate.tsx
```diff
@@ -0,0 +1,25 @@
+import { getDateFnsLocale } from '@ovh-ux/manager-core-utils';
+import { format } from 'date-fns';
+import * as dateFnsLocales from 'date-fns/locale';
+import { useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
+
+export const useFormattedDate = (date: string | null, formatStr = 'PPpp') => {
```
frenautvh, line 7: could this be replaced with useFormattedDate from manager-react-component ?

packages/manager/apps/pci-volume-snapshot/src/pages/listing/Listing.page.tsx
```diff
@@ -0,0 +1,231 @@
...
+            </div>
+          </div>
+          <div>
+            <FilterList filters={filters} onRemoveFilter={removeFilter} />
+          </div>
+
+          <div className="overflow-x-auto max-w-full">
+            <Datagrid
+              columns={columns}
+              items={isLoading ? [] : paginatedSnapshots.rows}
```
fredericvilcot, line 218: Why this?
items could be an empty array no ?

packages/manager/apps/pci-volume-snapshot/src/hooks/useFormattedDate.tsx
seven-amid: In this case we should rework the implementation of the hook in mrc, because they didn't use date-fns the manage dates, and this didn't handle the display wanted in volume-snapshot neither.

kqesar: the hook from MRC don't permissive https://github.com/ovh/manager/blob/master/packages/manager-react-components/src/hooks/date/useFormattedDate.ts

Formats allowed are not from date-fns but custom

packages/manager/apps/pci-volume-snapshot/src/pages/listing/useDatagridColumn.tsx
```diff
@@ -0,0 +1,88 @@
...
+  useTranslatedMicroRegions,
+} from '@ovh-ux/manager-react-components';
+import { useTranslation } from 'react-i18next';
+import { useBytes } from '@ovh-ux/manager-pci-common';
+import { TVolumeSnapshot } from '@/api/hooks/useSnapshots';
+import { useFormattedDate } from '@/hooks/useFormattedDate';
+import StatusComponent from './Status.component';
+import ActionsComponent from './Actions.component';
+
+export const useDatagridColumn = () => {
```
fredericvilcot, line 13: As I mentioned in previous µapps, this should not be a hook as, indeed, it's only a plain function that returns static columns consumed by one single component.

If the reason to create this hook is to only handle other hooks, we can consider given a formatter object as parameter:)

packages/manager/apps/pci-volume-snapshot/src/pages/listing/Status.component.tsx
fredericvilcot, line 20: State and effects are not useful, we could consider using a simple useMemo function :)

packages/manager/apps/pci-volume-snapshot/src/api/data/snapshots.ts
```diff
@@ -0,0 +1,68 @@
...
+  name: string;
+  description: string;
+  size: number;
+  volumeId: string;
+  region: string;
+  status: 'available' | 'creating' | 'deleting' | 'error' | 'error_deleting';
+  planCode: string | null;
+}
+
+export interface TVolume {
```
kqesar, line 15: Use type instead of Interface

seven-amid:
The majority of the mock used in come tests files are already exist globally within the setupTests.ts file.
IMO: some comments used within test files are useless, but still my opinion for sure ;)

packages/manager/apps/pci-volume-snapshot/src/api/hooks/useSnapshots.ts
```diff
@@ -0,0 +1,68 @@
...
+    queryKey: ['snapshots', projectId],
+    enabled: !!projectId,
+    queryFn: () => getSnapshots(projectId),
+  });
+
+export const useVolumeSnapshots = (projectId: string) => {
+  const { data: snapshots, isPending, isLoading } = useAllSnapshots(projectId);
+
+  const volumeIds = snapshots?.map((volume) => volume.volumeId);
+  const uniqueVolumeIds = [...new Set(volumeIds)];
```
fredericvilcot, line 23:
```suggestion
  const uniqueVolumeIds = [...new Set(snapshots?.map((volume) => volume.volumeId))];
```

packages/manager/apps/pci-volume-snapshot/src/api/hooks/useSnapshots.ts
```diff
@@ -0,0 +1,68 @@
...
+import { paginateResults, sortResults } from '@/helpers';
+
+export type TVolumeSnapshot = TSnapshot & {
+  volume?: TVolume | null;
+};
+
+export const useAllSnapshots = (projectId: string) =>
+  useQuery({
+    queryKey: ['snapshots', projectId],
+    enabled: !!projectId,
```
fredericvilcot, line 15: If you make this predicate, I guess `projectId` parameter's type in missing `| undefined` :)

packages/manager/apps/pci-volume-snapshot/src/hooks/useFormattedDate.tsx
```diff
@@ -0,0 +1,25 @@
...
+  const { i18n } = useTranslation();
+
+  const locales = useMemo(() => ({ ...dateFnsLocales }), []);
+
+  return useMemo(() => {
+    if (!date) {
+      return '';
```
fredericvilcot, line 14: Could me more explicit by returning `null` :)

packages/manager/apps/pci-volume-snapshot/src/pages/listing/Listing.page.tsx
```diff
@@ -0,0 +1,231 @@
...
+      key: 'name',
+      value: searchField.trim(),
+      comparator: FilterComparator.Includes,
+      label: '',
+    });
+    setSearchField('');
+  };
+
+  return (
+    <RedirectionGuard condition={false} isLoading={false} route={''}>
```
fredericvilcot, line 73: Route could more explicit?

seven-amid: should be handled in another PR, check the commit bellow https://github.com/ovh/manager/pull/16129/commits/22434b4001bb69ef42a4e5d3598caa942ca45a34

packages/manager/apps/pci-volume-snapshot/package.json
```diff
@@ -17,6 +17,7 @@
   "dependencies": {
     "@ovh-ux/manager-config": "^8.0.2",
     "@ovh-ux/manager-core-api": "^0.10.1",
+    "@ovh-ux/manager-core-utils": "*",
```
ppprevost, line 20: no version for core-utils ?

packages/manager/apps/pci-volume-snapshot/src/hooks/useFormattedDate.spec.tsx
```diff
@@ -0,0 +1,212 @@
+import { describe, it, expect, vi, beforeEach } from 'vitest';
+import * as i18nApi from 'react-i18next';
+import { renderHook } from '@testing-library/react';
+import * as dateFns from 'date-fns';
+import * as MRCApi from '@ovh-ux/manager-core-utils';
+import { KeyPrefix } from 'i18next';
+import { useFormattedDate } from './useFormattedDate';
+
+// Mock dependencies
+vi.mock('react-i18next', () => ({
```
ppprevost, line 10: use a setupTest.ts file in order to mock i18n

packages/manager/apps/pci-volume-snapshot/src/pages/listing/Status.component.spec.tsx
```diff
@@ -0,0 +1,117 @@
...
+    'data-testid': string;
+  }) => (
+    <div data-testid={testId} data-color={color}>
+      {label}
+    </div>
+  ),
+}));
+
+// Mock react-i18next
+vi.mock('react-i18next', () => ({
```
ppprevost, line 24: same, go to setupTest

packages/manager/apps/pci-volume-snapshot/src/api/hooks/useSnapshots.ts
FabienHenon: I like the idea but it must be a hook so that when filters change, the list is updated (without any server roundtrip).

If I pass a plain select function to `useVolumeSnapshots` I loose this capability.
Data won't update when filters update because `useVolumeSnapshots` has no idea about `pagination`, `sorting`, `filters`.

Or that means, in the view, we have to create a `useMemo` that calls `useVolumeSnapshots` as soon as `pagination`, `sorting` or `filters` changes.
But that would only move a part of the code to the view.

What do you think?
Did you have any other solution in mind?

packages/manager/apps/pci-volume-snapshot/src/hooks/useFormattedDate.tsx
```diff
@@ -0,0 +1,25 @@
+import { getDateFnsLocale } from '@ovh-ux/manager-core-utils';
+import { format } from 'date-fns';
+import * as dateFnsLocales from 'date-fns/locale';
+import { useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
+
+export const useFormattedDate = (date: string | null, formatStr = 'PPpp') => {
+  const { i18n } = useTranslation();
+
+  const locales = useMemo(() => ({ ...dateFnsLocales }), []);
```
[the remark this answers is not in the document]
FabienHenon: I honestly don't know

packages/manager/apps/pci-volume-snapshot/src/pages/listing/useDatagridColumn.tsx
FabienHenon: It's complicated to transform this hook into a function because it uses the hook `useFormattedDate` for each item.
Because of that, this hook must be called inside the component otherwise React will complain about hooks called a different number of times.

Or, we must also convert `useFormattedDate` to a simple function but it uses `useMemo`.

packages/manager/apps/pci-volume-snapshot/src/api/hooks/useSnapshots.ts
FabienHenon: It can't be undefined but it can be empty (which is falsy too)

packages/manager/apps/pci-volume-snapshot/src/pages/listing/Listing.page.tsx
```diff
@@ -0,0 +1,231 @@
...
+                name="searchField"
+                className="min-w-[15rem]"
+                value={searchField}
+                onKeyDown={(event) => {
+                  if (event.key === 'Enter') {
+                    onHandleSearch();
+                  }
+                }}
+                onOdsChange={({ detail }) =>
+                  setSearchField(detail.value as string)
```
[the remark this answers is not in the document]
FabienHenon: I had to force it because it be be nulled according to its type

packages/manager/apps/pci-volume-snapshot/package.json
FabienHenon: That's what I saw for most apps but I can just it

packages/manager/apps/pci-volume-snapshot/src/pages/listing/Listing.page.tsx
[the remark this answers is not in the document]
fredericvilcot: My bad, I misled myself with Select Event type :)

So, I think we should handle nullity to prevent from runtime errors: 

```suggestion
onOdsChange={({ detail }) => {
  if (detail.value != null) {
    setSearchField(String(detail.value));
   }
}}
```

packages/manager/apps/pci-volume-snapshot/src/api/hooks/useSnapshots.ts
fredericvilcot: Yes, it's pretty hard to implement it this way because of the current architecture of both the app (react-query) and the API.

We could achieve this using functional composition (currification) but it would complicate a lot the existing code, so IMO the best solution would be to redesign API and handle sort, filter & pagination on backend-side, so data will be filtered in a reactive way with no extra logic on the view :)

But agree that, for now, we could go on with the hook, thanks for your analysis :)

packages/manager/apps/pci-volume-snapshot/.eslintrc.cjs
```diff
@@ -0,0 +1,26 @@
+module.exports = {
+  extends: [
+    '../../../../.eslintrc.js',
+    'plugin:react/recommended',
+    'plugin:react/jsx-runtime',
```
fredericvilcot, line 5: We can imagine adding react-hooks eslint plugin to enhance rules :)

packages/manager/apps/pci-volume-snapshot/src/api/api.types.ts
```diff
@@ -0,0 +1,55 @@
+export type TSnapshot = {
+  id: string;
+  creationDate: string;
+  name: string;
+  description: string;
+  size: number;
+  volumeId: string;
+  region: string;
+  status: 'available' | 'creating' | 'deleting' | 'error' | 'error_deleting';
```
fredericvilcot, line 9: We could be more exhaustive and create dedicated types for `enums` or `Records`, so you can easily import and manipulate them without knowing the whole structure.

packages/manager/apps/pci-volume-snapshot/src/hooks/useFormattedDate.tsx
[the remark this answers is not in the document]
fredericvilcot: I think there is a subject around this, to think about alter because it affects other µapps 👍🏻

frenautvh (approved):

kqesar (approved):

PR #16165, opened 2025-03-24
feat(container): add skip to the main content button

MaximeBajeux:
## Description

To improve the accessibility we added a skip to the main content button into the container in place of in the micro applications

Ticket Reference: #MANAGER-15487


packages/manager/apps/container/src/core/dev/index.ts
JacquesLarique, line 1: This should not be commited

packages/manager/apps/container/src/container/nav-reshuffle/header/skip-to-main-content/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  const { iframeRef } = props;
+  const { isFirstTabDone, setIsFirstTabDone, firstFocusableElement } = useProductNavReshuffle();
+  const [isButtonDisplayed, setIsButtonDisplayed] = useState(false);
+  const buttonRef = useRef<HTMLOsdsButtonElement>(null);
+  const shell = useShell();
+  const { t } = useTranslation('sidebar');
+
+  const handleGlobalKeyDown = (event: KeyboardEvent) => {
+    if (isFirstTabDone === true) return;
+    if (event.key === "Tab" && isFirstTabDone === false) {
```
JacquesLarique, line 24: The test `isFirstTabDone === false` seems to be unnecessary as `isFirstTabDone` is never something else than `true` or `false` (and you can't access this test if it is `true`)

packages/manager/apps/container/src/container/nav-reshuffle/header/skip-to-main-content/index.tsx
```diff
@@ -0,0 +1,78 @@
...
+  }, [setIsFirstTabDone, isFirstTabDone]);
+
+  useEffect(() => {
+    if (isButtonDisplayed) buttonRef?.current?.focus();
+  }, [isButtonDisplayed])
+
+  const handleKeyDown = (e: any) => {
+    if (e.key === "Enter" && iframeRef?.current) {
+      iframeRef.current.focus();
+      shell.emitEvent('skipToTheMainContent');
```
JacquesLarique, line 46: What is the purpose of this event ? It does not seem to be listened to

JacquesLarique (changes requested):

packages/manager/apps/container/src/container/nav-reshuffle/header/skip-to-main-content/index.tsx
MaximeBajeux: When we focus the iframe, we focus the first focusable element into the iframe. That's not always the "main content". So I emit an event to allow applications to redirect the focus on the real main content.

JacquesLarique: I'm wondering if we should use the ux plugin for this

JacquesLarique (dismissed):

packages/manager/apps/container/src/core/product-nav-reshuffle/ProductNavReshuffleProvider.tsx
```diff
@@ -15,13 +15,15 @@ export const ProductNavReshuffleProvider = ({
...
 
+  const [isFirstTabDone, setIsFirstTabDone] = useState<boolean>(false);
   const location = useLocation();
   const [currentNavigationNode, setCurrentNavigationNode] = useState<Node>(null);
   const [navigationTree, setNavigationTree] = useState({});
   const shell = useShell();
   const [isMobile, setIsMobile] = useState(useMediaQuery({
     query: `(max-width: ${MOBILE_WIDTH_RESOLUTION}px)`,
   }));
+  const firstFocusableElement = useRef(null)
```
JacquesLarique, line 26: You're missing a `;` here

packages/manager/apps/hub/src/pages/dashboard/dashboard.tsx
```diff
@@ -69,20 +69,29 @@ export default function Dashboard() {
...
   useRouteSynchro();
 
   useEffect(() => {
     trackCurrentPage();
   }, [location]);
 
+  const onMainContentSelected = () => {
+    window.document.getElementById('skipTarget').focus();
+  };
+
```
JacquesLarique, line 82: Isn't this irrelevant after dropping the idea of using the UX plugin ?

packages/manager/apps/hub/src/pages/dashboard/dashboard.tsx
```diff
@@ -69,20 +69,29 @@ export default function Dashboard() {
...
       const newValueIsAccountSidebarVisible = (await shell.ux.isAccountSidebarVisible()) as boolean;
       setIsAccountSidebarVisible(() => newValueIsAccountSidebarVisible);
     };
+
+    const getOnSkipToTheMainContent = async () => {
+      await shell.ux.onSkipToTheMainContent(onMainContentSelected);
+    };
+
     getIsAccountSidebarVisible();
+    getOnSkipToTheMainContent();
```
JacquesLarique, line 94: Isn't this irrelevant after dropping the idea of using the UX plugin ?

packages/manager/apps/container/src/container/nav-reshuffle/header/skip-to-main-content/index.tsx
JacquesLarique, line 1: Reading this topic: https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Global_attributes/tabindex, it seems like a bad idea to force this element to be focused by the first press of "Tab", could we rather move it in the dom and adjust its position using css ?

JacquesLarique: After checking with design team, the need to have the button focused on the first tab press is correct, however they agree with the recommandation (linked in my previous comment) to avoid overriding the keypress behavior.
To overcome this issue, contrary to my previous comment, I would suggest to use a Portal to keep the button in the current component but have it being appear first in the DOM (well at least before the first focusable element)
_This is an interesting article regarding this aspect of accessibility: https://www.a11y-collective.com/blog/skip-to-main-content/_

packages/manager/apps/container/src/container/nav-reshuffle/header/skip-to-main-content/index.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+  const skipToMainTheContent = () => {
+    const mainContent = iframeRef.current.contentWindow.document.getElementById(
+      'maincontent',
+    );
+    const mainView = iframeRef.current.contentWindow.document.getElementById(
+      'mainview',
+    );
+
+    iframeRef.current.focus();
```
JacquesLarique: Can't `iframeRef.current` be null ? if so please had some `?` to avoid crashes

JacquesLarique (approved):

aottr (approved):

PR #16238, opened 2025-03-27
fix(web): a null value in rows break the display

stif59100:
ref: #MANAGER-17615

## Description

I fixed locale compare for database

Ticket Reference: #MANAGER-17615


tristanwagner (approved):
`.filter(Boolean)` also works :)

PR #16273, opened 2025-03-28
feat(*): migrate from reket to axios

aottr:
## Description

Goal of this feature is to migrate from reket to axios (via manager-core-api) when possible.

Ticket Reference: #MANAGER-15627


MaximeBajeux (approved):

JacquesLarique (approved):

MaximeBajeux (approved):

aottr:
Hey and thanks for the detailed review.

> Remove duplicate exports – useOvhPaymentMethod is declared twice; please keep only the final Axios-based version to avoid confusion.

There are two exports (direct with `export const` and the `export default`) but both refer to the same implementation. I can happily remove the default export since the result will be the same, but this is something you'll see very often in MANAGER code (old and new) for "readability". I did not change many aspects of the package, since the scope (and sizing) of the ticket is to remove reket and it was already extended twice. What we can do though is plan a proper refactoring of the package (including proper TypeScript and a breaking change with new major version)

If you refer to the other "useOvhPaymentMethod" in the container, this is a custom hook for the react app.

> Clean up unused parameters – in payment-method.service.js, userLocale and paymentMethodPageUrl are still injected, while coreConfig/region were removed. Verify which dependencies are still needed.

`userLocale` is still used for the `getUserLocale()` service-function. `paymentMethodPageUrl` is still used in the controller when accessing the service for translations.

> Standardize error handling – switch all error checks to error.response.status under Axios rather than mixing error.status.

Could you please highlight occaisons of `error.status` ? This property does not exist anymore when using axios so it would need to be changed. I didn't find any occaisons.

> Reinforce SSO hooks – ensure the SSO re-authentication logic from Reket isn’t lost; if it’s handled elsewhere, please document where.

The migration has been done via the `@ovh-ux/manager-core-api` package, which exposes an axios instance with all necessary headers and the `handleAuthenticationError` interceptors which takes care of necessary auth redirects.

> Add unit tests – cover the new useOvhPaymentMethod service for success (200) and not-found (404) scenarios.

Like explained above, it's sadly very much out of scope and this will be included in a package revamp.

packages/manager/apps/container/src/container/legacy/server-sidebar/universe/AccountSidebar.tsx
```diff
@@ -1,4 +1,5 @@
 import { useEffect, useState } from 'react';
+import { v6 } from '@ovh-ux/manager-core-api';
```
JacquesLarique, line 2: Why do we import v6 here, it does not seem to be used

aottr: It got redundant after the last rebase :/

tristanwagner (approved):

Issue #16164, opened 2025-03-24
[manager-components]: Datagrid load all button


aboungnaseng-ovhcloud:
### Have you already checked if a similar item is present on manager-components?

- [ ] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

![Image](https://github.com/user-attachments/assets/df0a9e92-d2db-4e6a-bd6a-36fd29526c42)

### Where do you expect to use this?

now

### Do you have mock-up?

yes

### Additional Information

- Need some specs

tristanwagner:
How I see the implementation in `packages/manager-react-components/src/components/datagrid/datagrid.component.tsx:459`:

```ts
      {hasNextPage && (
        <div className="grid justify-items-center gap-4 my-5">
          <OdsButton
            data-testid="load-more-btn"
            variant={ODS_BUTTON_VARIANT.outline}
            label={t('common_pagination_load_more')}
            onClick={onFetchNextPage}
            isLoading={isLoading}
          />
          {onFetchAllPages && (
            <OdsButton
              data-testid="load-all-btn"
              variant={ODS_BUTTON_VARIANT.outline}
              label={t('common_pagination_load_all')}
              onClick={onFetchAllPages}
              isLoading={isLoading}
            />
         )}
        </div>
      )}
```

This way, if you define the `onFetchAllPages` prop on `<Datagrid />` the button will show, you have to manage this state outside of the `Datagrid` and pass it to your hook that will fetch the data, it allows you to have a good query key management.

When all pages will be loaded, `hasNextPage` will be false, thus it will hide both "load more" and "load all" buttons

aboungnaseng-ovhcloud:
<img width="917" alt="Image" src="https://github.com/user-attachments/assets/555a2ec0-3c6e-4162-9559-f27c4e9cb7a5" />
<img width="924" alt="Image" src="https://github.com/user-attachments/assets/0a406eea-e27e-4277-a90a-57deae8fb0fe" />
<img width="230" alt="Image" src="https://github.com/user-attachments/assets/c1080a37-79fa-4f0e-ac39-a19b0e00bb45" />
<img width="694" alt="Image" src="https://github.com/user-attachments/assets/14d0039d-d976-4e9f-9b46-ce7d648f736d" />

anooparveti:
@tristanwagner @ghyenne @stif59100, we have updated the specification on the ticket description.
Can you please check and come back to us if you've any further queries.

[...]

PR #16347, opened 2025-04-02
feat(mrc): add load all button to datagrid

tristanwagner:
ref: #16164

## Description

image

Ticket Reference: #16164    

## Additional Information

This PR handle the datagrid responsibilities from the specs

### DIFF
--- a/packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
+++ b/packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
@@ -108,10 +108,12 @@ export interface DatagridProps<T> {
   className?: string;
   /** option to adjust content on the left */
   contentAlignLeft?: boolean;
-  /** boolean to display load more button */
+  /** boolean to display load more button, and load all button if onFetchAllPages is defined */
   hasNextPage?: boolean;
-  /** callback when click on load more button */
+  /** callback on load more button click */
   onFetchNextPage?: any;
+  /** callback on load all button click, the button will be displayed only if this function is defined and hasNextPage is true */
+  onFetchAllPages?: React.MouseEventHandler<HTMLOdsButtonElement>;
   /** Enables manual sorting for the table */
   manualSorting?: boolean;
   /** Enables manual pagination */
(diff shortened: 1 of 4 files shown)


ghyenne (approved):

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
```diff
@@ -493,17 +497,26 @@ export const Datagrid = <T,>({
       ) : (
         <></>
       )}
-      {hasNextPage && (
-        <div className="grid justify-items-center my-5">
+      {hasNextPage ? (
```
helabenkhalfallah, line 500: the hasNextPage condition is like this: 

```
hasNextPage={data?.length && data.length < maxPages * pageSize}
```

I'm wondering if in real case we'll always have the information `maxPages` and  `pageSize`.

aboungnaseng-ovhcloud, line 500:
```suggestion
      {hasNextPage && (
```

aboungnaseng-ovhcloud: we cannot not let this ? or there is eslint error ?

tristanwagner: I did it like that because since it's a boolean, and you check it like `hasNextPage={data.length}` with a falsy number or something like that, you will have a "0" showing instead of nothing showing, it's to avoid that :)

tristanwagner: this condition is just to simulate it in the tests, in real case we are using the `hasNextPage` boolean from `useQuery` etc :)

helabenkhalfallah: oh yes, cool thank you then  :) :)

helabenkhalfallah (approved):

aboungnaseng-ovhcloud (approved):

Issue #16266, opened 2025-03-28
[manager-components]: Datagrid search bar use OdsInput type search


tristanwagner:
### Have you already checked if a similar item is present on manager-components?

- [x] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

the current implementation of datagrid's search bar in mrc looks like this

![Image](https://github.com/user-attachments/assets/ba2dc657-6548-4819-bf8a-971b597964a8)

but in ODS we now have an input type search

![Image](https://github.com/user-attachments/assets/cf082dae-38ab-489c-a04a-75138713c3e8)

We should make the change as it seems better

also another point is that currently the input is resizing when the "clear" button appears/disappears it's the behavior of ODS input

![Image](https://github.com/user-attachments/assets/018f76d2-0249-42e8-85d9-1dcdb31f5f64)

![Image](https://github.com/user-attachments/assets/87cb2ec5-3d23-4866-a572-57e7e136e1af)

personally I added a min-width to the input in my project to prevent that but I think it should be either fixed on ODS side (adding the "clear" button doesn't change input width) or put a default min-width in mrc

### Where do you expect to use this?

*

tristanwagner:
Seems like right now you can't pass a placeholder to the search bar, which would be nice to have

anooparveti:
@feoche can you give us the specification regarding the width of the search box? Can it variable width depending on the content or should it be of fixed width?

tristanwagner:
handled in #16354 , there's still this question of the width of the search bar that is resizing when clear button appear or disappear when there's no min-width set to be answered

PR #16354, opened 2025-04-02
feat(mrc): set datagrid searchbar as input search

tristanwagner:
## Description

Screenshot 2025-04-02 at 15 35 17

Ticket Reference: #16266

### DIFF
--- a/packages/manager/apps/zimbra/src/index.scss
+++ b/packages/manager/apps/zimbra/src/index.scss
@@ -44,7 +44,7 @@ ods-tag.org-tag::part(tag) {
 // put min-width otherwise the input is
 // changing size when the clear button appear/disappear
 ods-input#datagrid-searchbar::part(input) {
-  min-width: 220px;
+  min-width: 240px;
 }
 
 .dns-field {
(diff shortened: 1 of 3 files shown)


packages/manager/apps/zimbra/src/index.scss
```diff
@@ -44,7 +44,7 @@ ods-tag.org-tag::part(tag) {
 // put min-width otherwise the input is
 // changing size when the clear button appear/disappear
 ods-input#datagrid-searchbar::part(input) {
-  min-width: 220px;
+  min-width: 240px;
```
helabenkhalfallah, line 47: is this true for all device (responsive)?

tristanwagner: this is a workaround for now, to avoid the default behavior of OdsInput resizing when the clear button appearing/disappearing, so the best workaround that I found for now is to set this min-width, but a subject has been raised on ODS side and we are expecting a fix :)

helabenkhalfallah: all right! thank you for the clarification:) :)

helabenkhalfallah (approved):

aboungnaseng-ovhcloud (approved):

ghyenne (approved):

PR #16360, opened 2025-04-02
fix(pci-public-ip): land user to step 3 to assign Floating IP from instance

Tsiorifamonjena:
## Description

When user want assign a floating ip to an instance from the compute app, land him to the step 3

Ticket Reference: #TAPC-3628

### DIFF
--- a/packages/manager/apps/pci-public-ip/src/hooks/order/useStore.ts
+++ b/packages/manager/apps/pci-public-ip/src/hooks/order/useStore.ts
@@ -83,9 +83,6 @@ export const useOrderStore = create<Store>()((set) => {
         ...state,
         floatingIpCreation: true,
       })),
-    reset: () =>
-      set({
-        ...initialState,
-      }),
+    reset: () => set(initialState),
   };
 });
--- a/packages/manager/apps/pci-public-ip/src/hooks/order/useStore.ts
+++ b/packages/manager/apps/pci-public-ip/src/hooks/order/useStore.ts
@@ -11,20 +11,25 @@ type Store = {
   closeAllSteps: () => void;
   floatingIpCreation: boolean;
   setFloatingIpCreation: () => void;
+  reset: () => void;
+};
+
+const initialState = {
+  form: {
+    ipType: null,
+    failoverCountry: null,
+    floatingRegion: null,
+    instance: null,
+    ipAddress: null,
+    floatingGatewaySize: null,
+    isSubmitting: false,
+  },
+  steps: new Map(),
 };
 
 export const useOrderStore = create<Store>()((set) => {
   return {
-    form: {
-      ipType: null,
-      failoverCountry: null,
-      floatingRegion: null,
-      instance: null,
-      ipAddress: null,
-      floatingGatewaySize: null,
-      isSubmitting: false,
-    },
-    steps: new Map(),
+    ...initialState,
     setSteps: (newSteps: Map<StepIdsEnum, TStepState>) => {
       set(() => {
         return {
@@ -78,5 +83,9 @@ export const useOrderStore = create<Store>()((set) => {
         ...state,
         floatingIpCreation: true,
       })),
+    reset: () =>
+      set({
+        ...initialState,
+      }),
   };
 });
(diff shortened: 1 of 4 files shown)


packages/manager/apps/pci-public-ip/src/pages/order/steps/FloatingSteps.tsx
```diff
@@ -34,12 +32,17 @@ const isRegionWith3AZ = (regions: TRegion[]) =>
...
   regionName,
+  instanceId,
 }: {
   projectId: string;
   regionName: string;
+  instanceId?: string;
 }): JSX.Element => {
   const { t: tOrder } = useTranslation('order');
-  const { state: DataState, getInstanceById } = useData(projectId, regionName);
+  const { state: DataState, getInstanceById, isInstanceFetching } = useData(
```
SimonChaumet, line 42: can you rename it to use camelCase instead ?

SimonChaumet (dismissed):

packages/manager/apps/pci-public-ip/src/hooks/order/useStore.ts
```diff
@@ -78,5 +83,9 @@ export const useOrderStore = create<Store>()((set) => {
         ...state,
         floatingIpCreation: true,
       })),
+    reset: () =>
+      set({
+        ...initialState,
+      }),
```
fredericvilcot, line 89: You don't need to spread initial state
```suggestion
 reset: () => set(initialState),
```

packages/manager/apps/pci-public-ip/src/pages/order/Order.page.tsx
```diff
@@ -24,18 +24,26 @@ import { useOrderParams } from '@/hooks/order/useParams';
...
 
-  const { form, setSteps, setForm } = useOrderStore();
+  const {
+    form,
+    setSteps,
+    setForm,
+    reset,
+    closeStep,
+    openStep,
+  } = useOrderStore();
```
fredericvilcot, line 46: Could be great to use shallowing to prevent from re-render each time a property changes.
(https://zustand.docs.pmnd.rs/guides/prevent-rerenders-with-use-shallow)

packages/manager/apps/pci-public-ip/src/pages/order/steps/FloatingSteps.tsx
```diff
@@ -180,92 +177,106 @@ export const FloatingSteps = ({
...
-                  value={form.ipAddress?.ip}
+                  value={form.instance?.id}
                   onOdsValueChange={(event) => {
                     setForm({
                       ...form,
-                      ipAddress: form.instance.ipAddresses.find(
-                        (ipAddress) =>
-                          ipAddress.ip === (event.detail.value as string),
-                      ),
+                      instance: getInstanceById(event.detail.value as string),
```
fredericvilcot, line 210: Better to stringify `value` instead of this type cast:
```suggestion
                     instance: getInstanceById(String(event.detail.value)),
```

packages/manager/apps/pci-public-ip/src/pages/order/steps/FloatingSteps.tsx
```diff
@@ -180,92 +177,106 @@ export const FloatingSteps = ({
...
+                    </div>
+                    <OsdsSelect
+                      required
+                      value={form.ipAddress?.ip}
+                      onOdsValueChange={(event) => {
+                        setForm({
+                          ...form,
+                          ipAddress: form.instance.ipAddresses.find(
+                            (ipAddress) =>
+                              ipAddress.ip === (event.detail.value as string),
```
fredericvilcot, line 240: Same here

packages/manager/apps/pci-public-ip/src/pages/order/Order.page.tsx
Tsiorifamonjena: I'm not sure to understand the use case here. I have to extract all these methods from the `stores`.

fredericvilcot: There is no use case here :)
Using `shallow` acts as memoization for your selectors, it means that you will not re-render the whole page is a single property of the state changes.
So you have to retrieve property from the state using `useShallow` :)

fredericvilcot (approved):

Issue #16372, opened 2025-04-03
[manager-components]: Export formatted date component


tibs245:
### Have you already checked if a similar item is present on manager-components?

- [x] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

`packages/manager-react-components/src/components/formatted-date/FormattedDate.tsx` is not exported

### Where do you expect to use this?

sap

tibs245:
About `getDateFnsLocale` : 

The impact on tree-shaking and bundle size is significant:
String-based approach (current implementation):
❌ Negative impact on tree-shaking: When using string parameters, the bundler cannot determine which locales will be used at runtime
❌ Larger bundle size: All locale files might be included in the bundle since the bundler can't determine which ones are actually needed
❌ Runtime overhead: Requires additional string manipulation and mapping
Direct import approach:
✅ Better tree-shaking: The bundler can statically analyze which locales are imported and only include those in the bundle
✅ Smaller bundle size: Only the explicitly imported locales are included
✅ Type safety: Better TypeScript support and compile-time checking
✅ No runtime overhead: No need for string manipulation or mapping
Looking at the codebase, I can see that many components are already using the direct import approach, which is the recommended way. For example:
Apply to VrackNetwork...
;
To optimize the bundle size and improve tree-shaking, I recommend:
Replace the string-based getDateFnsLocale function with a direct import approach
Create a centralized locale mapping that uses direct imports
Update all components to use the new approach

I purpose import this 2 hooks in core/utils : 

```
import { useCallback } from 'react';
import { format as formatDateFns, isValid, Locale } from 'date-fns';
import { useTranslation } from 'react-i18next';
import { enGB, fr, de, es, it, pl, pt } from 'date-fns/locale';

export const DEFAULT_UNKNOWN_DATE_LABEL = 'N/A';

const localeMap: Record<string, Locale> = {
  en_GB: enGB,
  fr_FR: fr,
  fr_CA: fr,
  de_DE: de,
  es_ES: es,
  it_IT: it,
  pl_PL: pl,
  pt_PT: pt,
};

export const useDateFnsLocale = () => {
  const { i18n } = useTranslation();
  const language = i18n?.language || 'en_GB';
  return localeMap[language] || enGB;
};

/**
 * Format options for date strings
 *
 * @param P - Long localized date (e.g. 04/29/1453)
 * @param PP - Long localized date with abbreviated month (e.g. Apr 29, 1453)
[...]

PR #16375, opened 2025-04-03
feat(manager-react-component): export formatted date component

tibs245:
## Description

Ticket Reference: #16372

## Additional Information

Screenshot 2025-04-14 at 16 22 29

### DIFF
--- a/packages/manager-react-components/src/components/index.ts
+++ b/packages/manager-react-components/src/components/index.ts
@@ -24,6 +24,8 @@ export * from './notifications/useNotifications';
 
 export * from './filters';
 
+export * from './formatted-date/FormattedDate';
+
 export * from './ManagerButton/ManagerButton';
 export * from './ManagerText/ManagerText';
 
--- a/packages/manager-react-components/src/components/formatted-date/FormattedDate.spec.tsx
+++ b/packages/manager-react-components/src/components/formatted-date/FormattedDate.spec.tsx
@@ -1,0 +1,21 @@
+import { vi } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import { FormattedDate } from './FormattedDate';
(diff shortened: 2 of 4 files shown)


pauldkn (approved):

aboungnaseng-ovhcloud (approved):

anooparveti (changes requested):
I dont see story or unit tests for the component. So the component is not ready to be exported and moreover i dont see the need for the component as it just calls the hook and displays it.

tibs245:
> I dont see story or unit tests for the component. So the component is not ready to be exported and moreover i dont see the need for the component as it just calls the hook and displays it.

Yes I can add unit test but the easy use it's not very necessary for me. But I can do test of simple render.

The need is use the call easier. Because If I have one object with 5 dates. I need add 5 lines to call this hook to show it with good format.

So we add one step, the component is bigger, and when we see the JSX we need go in the top of file to see the format of date.

It's why this simple component is intrusting

packages/manager-react-components/src/components/index.ts
darsene, line 1: Does this component manage datetime display also ?

tibs245:
> I dont see story or unit tests for the component. So the component is not ready to be exported and moreover i dont see the need for the component as it just calls the hook and displays it.

Following us discussion. I render `useFormattedDate` deprecated. And I create unit test for component

I think it's may be not necessary. But with that we can do the change softly. Maybe fix a date for remove it ?

Not hesitate take decision and change during my holiday

anooparveti (approved):

anooparveti (changes requested):
Story is still missing for the component.

anooparveti (approved):

pauldkn (approved):

tibs245:
Closed to use better approache : https://github.com/ovh/manager/pull/16576

PR #16376, opened 2025-04-03
feat(manager-react-component): add use-auto-refetch hook

### LINKED ISSUE #16373: [manager-components]: Import useAutoRefetch on MRC

Component

### Description

We use a hook name `useAutoRefetch` on project to do pooling of data conditionned.

As we use it on multiple project It's intrusting to implement it on Manager React Component

See : `packages/manager/apps/hpc-vmware-managed-vcd/src/data/hooks/useAutoRefetch.ts`

HPC-Vmware, SAP

### Do you have mock-up?

_No response_

_No response_

### Additional Information

_No response_

tibs245:
Resolve #16373

## Description

Ticket Reference: #16373

### DIFF
--- a/packages/manager-react-components/src/hooks/auto-refetch/useAutoRefetch.ts
+++ b/packages/manager-react-components/src/hooks/auto-refetch/useAutoRefetch.ts
@@ -1,0 +1,28 @@
+import { useQueryClient } from '@tanstack/react-query';
+import { useEffect } from 'react';
+
+type AutoRefetchProps = {
+  queryKey: string[];
+  enabled: boolean;
+  interval?: number;
+};
+
+export const useAutoRefetch = ({
+  queryKey,
+  enabled,
+  interval = 10_000,
+}: AutoRefetchProps) => {
+  const queryClient = useQueryClient();
+
+  useEffect(() => {
+    if (!enabled) return undefined;
+
+    const refetchQueries = () => {
+      queryClient.invalidateQueries({ queryKey });
+    };
+
+    const refetchInterval = setInterval(() => refetchQueries(), interval);
+
+    return () => clearInterval(refetchInterval);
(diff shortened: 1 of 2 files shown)


pauldkn (approved):

tristanwagner (approved):
interesting, the use case of using this hook instead of `refetchInterval` query option is that you can use it in any part of the code just by passing the queryKey I guess ?

**nitpick**: it would be nice that it could support all [filters and options](https://tanstack.com/query/latest/docs/reference/QueryClient#queryclientinvalidatequeries) of `invalidateQueries`, mainly the `predicate` filter

anooparveti (changes requested):
**question:** What's missing from using `refetchInterval` from [useQuery](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery)?

packages/manager-react-components/src/hooks/auto-refetch/useAutoRefetch.ts
darsene, line 1: What about adding also a timeout like inhttps://github.com/ovh/manager/blob/develop/packages/manager/apps/key-management-service/src/data/hooks/useAutoRefetch.ts ?

tibs245: Why not

tibs245:
> `refetchInterval`



> **question:** What's missing from using `refetchInterval` from [useQuery](https://tanstack.com/query/latest/docs/framework/react/reference/useQuery)?

Yes it's true in my case it's a bad use I should use refetch interval. 

But I think it is intrusting to have because in specific case this can simplify the pooling. I will write documentation to explain differents specific cases and if this PR is not approved. The documentations can be in minimum help next developper to implement pooling

PR #16379, opened 2025-04-03
feat(generator): update prompt and package json

aboungnaseng-ovhcloud:
ref: #MANAGER-16960

## Description

In the generator update : 

- package.json
- tailwind.config.json
- vitest.config.json 

Ticket Reference: #MANAGER-16960

### DIFF
--- a/packages/manager/core/generator/app/index.js
+++ b/packages/manager/core/generator/app/index.js
@@ -8,7 +8,12 @@ import {
   createTranslations,
   createApiQueryFilesActions,
 } from '../utils/create-structure-helpers.js';
-import { UNIVERSES, SUB_UNIVERSES, LEVEL2 } from './universes.constant.js';
+import {
+  UNIVERSES,
+  SUB_UNIVERSES,
+  LEVEL2,
+  REGIONS,
+} from './universes.constant.js';
 
 const appDirectory = dirname(fileURLToPath(import.meta.url));
 
@@ -84,6 +89,27 @@ export default (plop) => {
         name: 'description',
         message: 'How would you describe the new app?',
         validate: (description) => description.length > 1,
+      },
+      {
+        type: 'checkbox',
+        name: 'regions',
+        message: 'what are the regions of the new app ?',
+        choices: REGIONS.map((element) => ({
+          name: element,
+          value: element,
+        })),
+        validate: (regions) => regions.length > 0,
+        default: REGIONS,
+      },
+      {
+        type: 'checkbox',
+        name: 'universes',
+        message: 'what are the universes of the new app ?',
+        choices: UNIVERSES.map((element) => ({
+          name: element,
+          value: element,
+        })),
+        validate: (UNIVERSES) => UNIVERSES.length > 0,
       },
       {
         type: 'checkbox',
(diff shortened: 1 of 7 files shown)


packages/manager/core/generator/app/templates/package.json.hbs
```diff
@@ -32,6 +36,7 @@
     "@ovh-ux/manager-react-components": "^2.2.0",
     "axios": "^1.1.2",
     "clsx": "^1.2.1",
+    "element-internals-polyfill": "^3.0.0",
```
helabenkhalfallah, line 39: why we need this package?

packages/manager/core/generator/app/index.js
```diff
@@ -8,7 +8,12 @@ import {
   createTranslations,
   createApiQueryFilesActions,
 } from '../utils/create-structure-helpers.js';
-import { UNIVERSES, SUB_UNIVERSES, LEVEL2 } from './universes.constant.js';
+import {
+  UNIVERSES,
+  SUB_UNIVERSES,
+  LEVEL2,
```
helabenkhalfallah, line 14: generally is not recommended to use number or type in name, is not possible to use more functional terms?

packages/manager/core/generator/app/templates/src/setupTests.ts
anooparveti, line 1: **question:** we also included setting up more wrappers like mocked shell-context no?

packages/manager/core/generator/app/templates/package.json.hbs
aboungnaseng-ovhcloud: mainly for the unit test with vitest, we have to pollify some ods component

packages/manager/core/generator/app/templates/package.json.hbs
```diff
@@ -41,19 +46,24 @@
...
+    "@testing-library/react": "^16.0.0",
     "@tanstack/react-query-devtools": "^5.51.21",
     "@ovh-ux/manager-vite-config": "*",
-    "@vitest/coverage-v8": "^2.1.2",
+    "@vitest/coverage-v8": "^2.1.9",
     "typescript": "^5.1.6",
-    "vite": "^5.2.13",
-    "vitest": "^2.1.2"
+    "vite": "^6.0.7",
+    "vitest": "^2.1.9"
```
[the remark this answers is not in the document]
aboungnaseng-ovhcloud: we put the '^' cause we don't want to have different version in the manager it can increase the size of the manager

packages/manager/core/generator/app/index.js
aboungnaseng-ovhcloud: Its more use as constant variable in our case

packages/manager/core/generator/app/templates/src/setupTests.ts
aboungnaseng-ovhcloud: I can also add it, I take a look

packages/manager/core/generator/app/index.js
aboungnaseng-ovhcloud: you mean type enum is more appropriate in this use case ? we need to refactor all the files

helabenkhalfallah: ok, we can keep like this time, but in the future if it's possible to avoid using number or types in name, we should have more functional terms, thank you so much @aboungnaseng-ovhcloud ! :)

packages/manager/core/generator/app/templates/package.json.hbs
helabenkhalfallah: ok.

helabenkhalfallah (approved):

PR #16400, opened 2025-04-04
fix(manager-react-component): params not used on iceberg v6 hook

### LINKED ISSUE #16399: [manager-components]: useResourcesIcebergV6 has disableCache but not used

packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx

### Description

`useResourcesIcebergV6` accept `disableCache` on params but it's not used.

### Steps to reproduce

Try use this hook with `disableCache` and see `Pragma: no-cache` is not present on header request.

### Browser

_No response_

### Operating System

- [x] macOS
- [x] Windows
- [x] Linux

### Additional Information

No impact on actual feature (Or not reported)

tibs245:
## Description

Ticket Reference: Resolve #16399

### DIFF
--- a/packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
+++ b/packages/manager-react-components/src/hooks/datagrid/useIcebergV6.tsx
@@ -29,6 +29,7 @@ export function useResourcesIcebergV6<T = unknown>({
   defaultSorting = undefined,
   shouldFetchAll = false,
   columns,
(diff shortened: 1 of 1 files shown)


anooparveti (changes requested):
I think we can more generalise to accept all params of `useInfiniteQuery` and pass them.

tristanwagner (approved):

PR #16410, opened 2025-04-04
feat(kms): add buttons for downloading public CA certificates

mathieumousnier:
## Description

Add buttons on kms dashboard for downloading public CA certificates

Ticket Reference: #MANAGER-17309

## Notes 

I made small refactorings before implementing the feature : 
- Some fetch functions and hooks were a bit mixed, so I moved some code to group functions by subjects
- Most files and functions use "Kms" and not "KMS" in there names, so i renamed a few ones I had to use


packages/manager/apps/key-management-service/src/components/dashboard/downloadKmsPublicCaLink/DownloadKmsPublicCaLink.tsx
```diff
@@ -0,0 +1,84 @@
...
+}: DownloadCaButtonProps) => {
+  const { t } = useTranslation('key-management-service/dashboard');
+  const [loading, setLoading] = useState(false);
+  const { addError } = useNotifications();
+
+  const resources = useMemo(
+    () => ({
+      publicCa: {
+        label: t('key_management_service_dashboard_button_label_download_ca'),
+        filename: PUBLIC_CA_FILENAME,
```
rjamet-ovh, line 32: filename should include a kms part (e.g. `${okmsId}-${PUBLIC_CA_FILENAME}` )

mathieumousnier: I talked to Gabriel, turns out that CA are relative to the region and not the kms, he proposed : 
- okms_eu-west-sbg_public_ca.pem
- okms_eu-west-sbg_public_rsa_ca.pem

packages/manager/apps/key-management-service/src/components/dashboard/downloadKmsPublicCaLink/DownloadKmsPublicCaLink.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+  return (
+    <div className="flex items-center gap-2">
+      <OdsLink
+        href="#"
+        color={ODS_BUTTON_COLOR.primary}
+        onClick={handleDownloadCa}
+        label={resources[type].label}
+        data-testid={`download_kms_public_ca`}
+      />
```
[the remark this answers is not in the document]
mathieumousnier: Right, i thought i couldn't disable a link

packages/manager/apps/key-management-service/src/data/api/iam.ts
```diff
@@ -0,0 +1,90 @@
...
+export type GetiamPolicyListParams = {
+  /** Pagination cursor */
+  'X-Pagination-Cursor': string;
+  /** Add extra information about resources in output */
+  details: boolean;
+  /** Filter on the readOnly attribute */
+  readOnly: boolean;
+};
+
+export const getiamPolicyListQueryKey = ['get/iam/policy'];
```
rjamet-ovh, line 19: theses iam api functions were not used anymore, so you can delete this file

packages/manager/apps/key-management-service/src/components/dashboard/downloadKmsPublicCaLink/DownloadKmsPublicCaLink.tsx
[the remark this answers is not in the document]
rjamet-ovh: if we use a lib to download the file, we should use a button instead of a link.

packages/manager/apps/key-management-service/src/components/dashboard/downloadKmsPublicCaLink/DownloadKmsPublicCaLink.tsx
```diff
@@ -0,0 +1,84 @@
...
+    } catch {
+      setLoading(false);
+      addError(t('key_management_service_dashboard_error_download_ca'));
+    }
+  };
+
+  return (
+    <div className="flex items-center gap-2">
+      <OdsLink
+        href="#"
```
helabenkhalfallah, line 75: Non-Semantic Link (href="#")
https://www.a11yproject.com/posts/creating-valid-and-accessible-links/

Perhaps you need more button than link

packages/manager/apps/key-management-service/src/components/dashboard/downloadKmsPublicCaLink/DownloadKmsPublicCaLink.tsx
```diff
@@ -0,0 +1,84 @@
...
+        label: t(
+          'key_management_service_dashboard_button_label_download_rsa_ca',
+        ),
+        filename: PUBLIC_RSA_CA_FILENAME,
+      },
+    }),
+    [t],
+  );
+
+  const handleDownloadCa = async (
```
helabenkhalfallah, line 44: I'm wondering if this function handleDownloadCa is not recreated on every render? If yes, we need a `useCallback`

packages/manager/apps/key-management-service/src/components/dashboard/downloadKmsPublicCaLink/DownloadKmsPublicCaLink.tsx
```diff
@@ -0,0 +1,84 @@
...
+        publicCa: certificates.publicCA,
+        publicRsaCa: certificates.publicRsaCA,
+      };
+
+      initiateTextFileDownload({
+        text: content[type],
+        filename: resources[type].filename,
+      });
+    } catch {
+      setLoading(false);
```
helabenkhalfallah, line 67: You're duplicating setLoading(false) in both the try and catch blocks.

packages/manager/apps/key-management-service/src/components/dashboard/downloadKmsPublicCaLink/DownloadKmsPublicCaLink.tsx
helabenkhalfallah: More general, I'm wondering if this code will not cause memory leaks when the component unmounts, but the awaited promise is still pending:
- The setLoading(false) (or any setState) will try to update state on an unmounted component (React can warn against in development).
- While React no longer throws this as an error, it's still considered a memory leak and should be avoided.

Perhaps we need something like: 
```js
function useAsyncSafeState<T>(initialValue: T): [T, (val: T) => void] {
  const isMounted = useRef(true);
  const [state, setState] = useState(initialValue);

  useEffect(() => () => {
    isMounted.current = false;
  }, []);

  const safeSetState = useCallback(
    (val: T) => {
      if (isMounted.current) setState(val);
    },
    [],
  );

  return [state, safeSetState];
}

const [loading, setLoading] = useAsyncSafeState(false);
```

This avoid setting state after async operations unless we check if the component is still mounted.

packages/manager/apps/key-management-service/src/components/layout-helpers/Create/RegionSelector.tsx
```diff
@@ -23,7 +23,7 @@ import {
 } from '@ovh-ux/manager-react-shell-client';
 import { useNavigate } from 'react-router-dom';
 import { ErrorBanner, Region } from '@ovh-ux/manager-react-components';
-import { useOrderCatalogOKMS } from '@/data/hooks/useOrderCatalogOKMS';
+import { useOrderCatalogOkms } from '@/data/hooks/useOrderCatalogOkms';
```
helabenkhalfallah, line 26: Generally we should avoid mixing PR purposes, we should differentiate between a PR that add functionalities from a PR that make refactor, to not loose everything when we rollback, and to avoid a significant software entropy (disorder).

packages/manager/apps/key-management-service/src/utils/dom/download.ts
```diff
@@ -0,0 +1,20 @@
...
+  const blob = new Blob([text], { type: 'text/plain' });
+  const link = document.createElement('a');
+  link.href = URL.createObjectURL(blob);
+  link.download = filename;
+
+  document.body.appendChild(link);
+  link.click();
+
+  document.body.removeChild(link);
+}
```
helabenkhalfallah, line 20: To well clean things:

```js
export function initiateTextFileDownload({
  filename,
  text,
  mimeType = 'text/plain',
}: {
  filename: string;
  text: string;
  mimeType?: string;
}) {
  const blob = new Blob([text], { type: mimeType });
  const link = document.createElement('a');
  const url = URL.createObjectURL(blob);
  link.href = url;
  link.download = filename;

  document.body.appendChild(link);
  try {
    link.click();
  } finally {
    document.body.removeChild(link);
    URL.revokeObjectURL(url);
  }
}
```

packages/manager/apps/key-management-service/src/components/dashboard/downloadKmsPublicCaLink/DownloadKmsPublicCaLink.tsx
```diff
@@ -0,0 +1,84 @@
...
+        filename: PUBLIC_CA_FILENAME,
+      },
+      publicRsaCa: {
+        label: t(
+          'key_management_service_dashboard_button_label_download_rsa_ca',
+        ),
+        filename: PUBLIC_RSA_CA_FILENAME,
+      },
+    }),
+    [t],
```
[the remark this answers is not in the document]
mathieumousnier: Yep you're right

packages/manager/apps/key-management-service/src/components/dashboard/downloadKmsPublicCaLink/DownloadKmsPublicCaLink.tsx
mathieumousnier: It is definitely a button, but the OdsButton component does not offer the style I have to implement here, so I didn't see any better solution

packages/manager/apps/key-management-service/src/components/dashboard/downloadKmsPublicCaLink/DownloadKmsPublicCaLink.tsx
mathieumousnier: Great, some real react discussions :)

### Adding useCallback on handleDownloadCa
I gently disagree :)

Of course the function will be recreated on every render, but I really think this is common react code and should not be an issue. The impact performance of this is negligible and I think we should not add useCallbacks on every functions declared inside components.

For me useCallback should be used to stabilize functions when they are put as props on memoized components, where we specificaly want that every props are stable.

I usually refer to this article on that particular subject:
https://kentcdodds.com/blog/usememo-and-usecallback

--
### State update on unmounted components

According to this PR on react, is seems that setting state on an unmounted component is not an issue.
The warning has been removed (i remember it really well !), and our use case is described here as a false positive.
https://github.com/facebook/react/pull/22114

--

packages/manager/apps/key-management-service/src/components/layout-helpers/Dashboard/GeneralInformationsTiles/KmipTile.tsx
```diff
@@ -16,7 +17,10 @@ const KmipTile = ({ okmsData }: KmipTileProps) => {
       id: 'kmip',
       label: KMIP_ENPOINT_LABEL,
       value: (
-        <Clipboard className="block w-full" value={okmsData?.kmipEndpoint} />
+        <div className="flex flex-col gap-2">
```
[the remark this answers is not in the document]
mathieumousnier: I don't get this one, could you please elaborate

packages/manager/apps/key-management-service/src/components/dashboard/downloadKmsPublicCaLink/DownloadKmsPublicCaLink.tsx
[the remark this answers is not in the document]
mathieumousnier: I don't think I can make this style with OdsButton ?
The `ghost` version has a padding that aims to show a background on hover.

image

packages/manager/apps/key-management-service/src/utils/dom/download.ts
mathieumousnier: This code will be erased with the use of the `file-saver` lib.

packages/manager/apps/key-management-service/src/data/api/iam.ts
```diff
@@ -0,0 +1,90 @@
...
+    const List = await fetchIcebergV6({
+      route: '/iam/policy',
+      pageSize,
+      page,
+    }).then(({ data, status, totalCount }) => ({ data, status, totalCount }));
+    return List;
+  } catch (error) {
+    return null;
+  }
+};
```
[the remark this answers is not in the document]
mathieumousnier: Did not realized but iam code was dead code, I removed it.

darsene (dismissed):

rjamet-ovh (dismissed):

rjamet-ovh (dismissed):

darsene (approved):

pauldkn (dismissed):

darsene (dismissed):

tibs245 (approved):

rjamet-ovh (approved):

PR #16446, opened 2025-04-08
fix(mrc): fix the issue on count visibility columns

ghyenne:
## Description

Ticket Reference: #MANAGER-17732

### DIFF
--- a/packages/manager-react-components/src/components/datagrid/datagrid.constants.ts
+++ b/packages/manager-react-components/src/components/datagrid/datagrid.constants.ts
@@ -9,3 +9,16 @@ export const DEFAULT_PAGINATION: PaginationState = {
 };
 
 export const defaultNumberOfLoadingRows = 5;
+
+export type ColumnsVisibility = {
+  id: string;
+  isDisabled: boolean;
+  label: string;
+  enableHiding: boolean;
+  isVisible: () => boolean;
+  onChange: () => void;
+};
+
+export type ColumnsVisibilityProps = {
+  columnsVisibility: ColumnsVisibility[];
+};
--- a/packages/manager-react-components/src/components/datagrid/datagrid.constants.ts
+++ b/packages/manager-react-components/src/components/datagrid/datagrid.constants.ts
@@ -10,15 +10,4 @@ export const DEFAULT_PAGINATION: PaginationState = {
 
 export const defaultNumberOfLoadingRows = 5;
 
-export type ColumnsVisibility = {
-  id: string;
-  isDisabled: boolean;
-  label: string;
-  enableHiding: boolean;
-  isVisible: () => boolean;
-  onChange: () => void;
-};
-
-export type ColumnsVisibilityProps = {
-  columnsVisibility: ColumnsVisibility[];
-};
+export const INTERNAL_COLUMNS = ['expander', 'actions'];
--- a/packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
+++ b/packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
@@ -13,19 +13,31 @@ import {
 } from '@ovhcloud/ods-components/react';
 import { useRef } from 'react';
 import { useTranslation } from 'react-i18next';
-import { ColumnsVisibilityProps } from '../datagrid.constants';
+import { INTERNAL_COLUMNS } from '../datagrid.constants';
+
+export type ColumnsVisibility = {
+  id: string;
+  isDisabled: boolean;
+  label: string;
+  enableHiding: boolean;
+  isVisible: () => boolean;
+  onChange: () => void;
+};
+
+export type ColumnsVisibilityProps = {
+  columnsVisibility: ColumnsVisibility[];
+};
 
 export function VisibilityManagement({
   columnsVisibility,
 }: Readonly<ColumnsVisibilityProps>) {
   const { t } = useTranslation('datagrid');
   const visibilityPopoverRef = useRef(null);
-  const columnsIncludedInCounter = columnsVisibility.filter(
-    (column) =>
-      !['expander', 'actions'].includes(column.id) && column.label !== '',
+  const eligibleColumns = columnsVisibility.filter(
+    (column) => !INTERNAL_COLUMNS.includes(column.id) && column.label !== '',
   );
 
-  const visibleColumnsCount = columnsIncludedInCounter.filter((column) =>
+  const visibleColumnsCount = eligibleColumns.filter((column) =>
     column.isVisible(),
   ).length;
 
@@ -40,7 +52,7 @@ export function VisibilityManagement({
         icon={ODS_ICON_NAME.columns}
         aria-label={t('common_topbar_columns')}
         label={`${t('common_topbar_columns')}${
-          visibleColumnsCount < columnsIncludedInCounter.length
+          visibleColumnsCount < eligibleColumns.length
             ? ` (${visibleColumnsCount})`
             : ''
         }`}
@@ -51,7 +63,7 @@ export function VisibilityManagement({
         with-arrow
       >
         <div className="flex flex-col">
-          {columnsIncludedInCounter.map((column) => (
+          {eligibleColumns.map((column) => (
             <OdsFormField key={column.id}>
               <div className="flex flex-row items-center gap-x-2">
                 <OdsCheckbox
--- a/packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
+++ b/packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
@@ -32,12 +32,24 @@ export function VisibilityManagement({
 }: Readonly<ColumnsVisibilityProps>) {
   const { t } = useTranslation('datagrid');
   const visibilityPopoverRef = useRef(null);
-  const numberVisibleColumns = columnsVisibility?.filter((col) =>
-    col.isVisible(),
-  ).length;
-  const disabledVisibleColumns = columnsVisibility?.filter(
-    (col) => col.enableHiding === false,
-  ).length;
+  const columnsIncludedInCounter = columnsVisibility.filter(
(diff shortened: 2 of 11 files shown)


aboungnaseng-ovhcloud (dismissed):

aboungnaseng-ovhcloud (approved):

anooparveti (changes requested):
**nitpick(blocking):** As discussed, relying on label is not a good solution as label may not be empty for columns displaying checkbox.

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
```diff
@@ -50,8 +50,8 @@ export function VisibilityManagement({
         icon={ODS_ICON_NAME.columns}
         aria-label={t('common_topbar_columns')}
         label={`${t('common_topbar_columns')} ${
```
helabenkhalfallah, line 52: This expression is complexe to read and understand, is it possible to create intermediate constants before return perhaps? 

```
label={`${t('common_topbar_columns')} ${
          numberVisibleColumns < columnsVisibility.length - columnsWithoutLabel
            ? `(${numberVisibleColumns})`
            : ''
        }`}
```

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
```diff
@@ -32,11 +32,11 @@ export function VisibilityManagement({
 }: Readonly<ColumnsVisibilityProps>) {
   const { t } = useTranslation('datagrid');
   const visibilityPopoverRef = useRef(null);
-  const numberVisibleColumns = columnsVisibility?.filter((col) =>
-    col.isVisible(),
+  const numberVisibleColumns = columnsVisibility?.filter(
```
helabenkhalfallah, line 35: I think we can use reduce instead of filter + length : 
```js
  const numberVisibleColumns = columnsVisibility?.filter(
    (col) => col.isVisible() && col.label !== '',
  ).length;
  const columnsWithoutLabel = columnsVisibility?.filter(
    ({ label }) => label === '',
  ).length;
```

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
```diff
@@ -17,7 +17,7 @@ import { useTranslation } from 'react-i18next';
 export type ColumnsVisibility = {
   id: string;
   isDisabled: boolean;
-  label: string;
+  label: string | JSX.Element;
```
ghyenne, line 20: I anticipate that in the future we might have labels other than strings (a checkbox for example).

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
ghyenne: I changed the way. I includes in the reduce

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.spec.tsx
```diff
@@ -54,7 +54,7 @@ describe('visibility management button part', () => {
 
     expect(visibilityElement).toHaveAttribute(
       'label',
-      'common_topbar_columns (0)',
+      'common_topbar_columns (1)',
```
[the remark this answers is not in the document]
ghyenne: In fact, the counter was wrong. We need to take those that are visible as well as those that are disabled and exclude only those with an empty label or a label other than a string.

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
tristanwagner: I would recommend using `React.ReactNode` instead of `string | JSX.Element` then

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
```diff
@@ -62,7 +79,7 @@ export function VisibilityManagement({
       >
         <div className="flex flex-col">
           {columnsVisibility
-            .filter(({ label }) => label !== '')
+            .filter(({ label }) => typeof label === 'string' && label !== '')
```
tristanwagner, line 82: So it means only labels typed as strings can be selected/unselected ? Is this limitation necessary ? Like if I put an icon as a column header why wouldn't I be able to show or hide this column

tristanwagner (approved):
nitpick on the label typing and also the limitation that only label of type string can be show/hidden but beside that it seems ok

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
```diff
@@ -32,12 +32,29 @@ export function VisibilityManagement({
...
+        acc.enabled < columnsVisibility.length - acc.withoutLabel;
+      return acc;
+    },
+    {
+      enabled: 0,
+      withoutLabel: 0,
+      total: columnsVisibility.length,
+      haveToDisplay: false,
+    },
+  );
```
anooparveti, line 57: **nitpick:** The logic seems complex to understand (maybe because the variable names are over-used) and also `columnsVisibilityParams` is not a parameter, so variable is misleading.

tristanwagner (dismissed):

tristanwagner (dismissed):

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
```diff
@@ -61,27 +73,25 @@ export function VisibilityManagement({
...
+                  ariaLabel={column.label}
+                />
+                <label slot="label" htmlFor={column.id}>
+                  <OdsText preset={ODS_TEXT_PRESET.paragraph}>
+                    {column.label}
+                  </OdsText>
+                </label>
+              </div>
+            </OdsFormField>
+          ))}
```
anooparveti, line 94: **question:** We should also filter the columns like expander and actions from being displayed in the form no?

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
```diff
@@ -32,12 +32,24 @@ export function VisibilityManagement({
...
   const { t } = useTranslation('datagrid');
   const visibilityPopoverRef = useRef(null);
-  const numberVisibleColumns = columnsVisibility?.filter((col) =>
-    col.isVisible(),
-  ).length;
-  const disabledVisibleColumns = columnsVisibility?.filter(
-    (col) => col.enableHiding === false,
-  ).length;
+  const columnsIncludedInCounter = columnsVisibility.filter(
+    (column) => !['expander'].includes(column.id) && column.label !== '',
```
anooparveti, line 36: **nitpick:** We can add the the array as constant in `packages/manager-react-components/src/components/datagrid/datagrid.contants.ts` to avoid duplication to find out the columns added by Datagrid internally.

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
```diff
@@ -32,12 +32,24 @@ export function VisibilityManagement({
...
+  );
+
+  const columnsVisibilityState = columnsIncludedInCounter.reduce(
+    (acc, current) => {
+      if (current.isVisible()) {
+        acc.enabled += 1;
```
anooparveti, line 42:
```suggestion
        acc.visibleCoulmns += 1;
```

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
```diff
@@ -32,12 +32,24 @@ export function VisibilityManagement({
...
+      }
+      acc.haveToDisplay = acc.enabled < acc.total;
+      return acc;
+    },
+    {
+      enabled: 0,
+      total: columnsIncludedInCounter.length,
+      haveToDisplay: false,
```
anooparveti, line 50: **nitpick:** Unnecessary variable, can easily be included as an expression to decide whether or not to display label.

tristanwagner (approved):

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
ghyenne: It's already filtered L23 - L26

packages/manager-react-components/src/components/datagrid/datagrid.constants.ts
```diff
@@ -9,3 +9,5 @@ export const DEFAULT_PAGINATION: PaginationState = {
 };
 
 export const defaultNumberOfLoadingRows = 5;
+
+export const EXCLUDED_COLUMNS = ['expander', 'actions'];
```
anooparveti, line 13:
```suggestion
export const INTERNAL_COLUMNS = ['expander', 'actions'];
```

ghyenne: It's fixed

packages/manager-react-components/src/components/datagrid/visibility/visibility-management.component.tsx
```diff
@@ -32,11 +33,12 @@ export function VisibilityManagement({
 }: Readonly<ColumnsVisibilityProps>) {
   const { t } = useTranslation('datagrid');
   const visibilityPopoverRef = useRef(null);
-  const numberVisibleColumns = columnsVisibility?.filter((col) =>
-    col.isVisible(),
-  ).length;
-  const disabledVisibleColumns = columnsVisibility?.filter(
-    (col) => col.enableHiding === false,
+  const columnsIncludedInCounter = columnsVisibility.filter(
```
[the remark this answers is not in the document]
ghyenne: It's fixed

tristanwagner (dismissed):

tristanwagner (approved):

anooparveti (approved):

PR #16468, opened 2025-04-09
feat(sap-features-hub): add menu entry dashboard

pauldkn:
ref: #MANAGER-16148

## Description

Ticket Reference: #MANAGER-16148

### DIFF
--- a/packages/manager/apps/sap-features-hub/public/translations/dashboard/Messages_fr_FR.json
+++ b/packages/manager/apps/sap-features-hub/public/translations/dashboard/Messages_fr_FR.json
@@ -1,7 +1,4 @@
 {
-  "title": "Dashboard page",
   "error_service": "No services info",
-  "general_informations": "Informations générales",
-  "tab2": "Tab 2",
-  "back_link": "Retour à la liste"
+  "description": "Découvrez notre gamme de fonctionnalités SAP optionnelles pour faciliter vos déploiements SAP."
 }
--- a/packages/manager/apps/sap-features-hub/src/pages/dashboard/general-informations/index.tsx
+++ b/packages/manager/apps/sap-features-hub/src/pages/dashboard/general-informations/index.tsx
@@ -1,7 +1,8 @@
 import React from 'react';
 
 function GeneralInfos() {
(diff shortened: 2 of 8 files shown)


packages/manager/apps/sap-features-hub/src/pages/dashboard/general-informations/index.tsx
```diff
@@ -1,7 +1,7 @@
 import React from 'react';
 
 function GeneralInfos() {
-  return <div>Information Générales Tab</div>;
+  return <div>Building blocks</div>;
```
vovh, line 4: I would add a const in a constants file to explicit that this is a value we don't want to trad instead of a static value we forgot to translate. Like we do in other app.

What do you think ? @darsene  your opinion on this take ? since you're gonna review this

vovh (approved):

packages/manager/apps/sap-features-hub/public/translations/dashboard/Messages_fr_FR.json
```diff
@@ -1,7 +1,5 @@
 {
-  "title": "Dashboard page",
   "error_service": "No services info",
-  "general_informations": "Informations générales",
-  "tab2": "Tab 2",
-  "back_link": "Retour à la liste"
+  "back_link": "Retour à la liste",
```
darsene, line 3: back link translation must exist in the common-translation in my opinion

packages/manager/apps/sap-features-hub/src/pages/dashboard/general-informations/index.tsx
pauldkn: Actually, it's not really a value here, it's just to indicate that this is the place to implement building blocks (for the next US), this is not a label supposed to stay

packages/manager/apps/sap-features-hub/public/translations/dashboard/Messages_fr_FR.json
pauldkn: I missed deleting this line, since we use `Retour au SAP Features Hub` rather than this one

darsene (approved):

PR #16479, opened 2025-04-09
fix(mrc): click on searchbar icon

tristanwagner:
ref: #NONE

## Description

Ticket Reference: #...


packages/manager-react-components/src/components/datagrid/datagrid-topbar.component.tsx
```diff
@@ -80,28 +80,35 @@ export const DatagridTopbar = <T,>({
...
+                    e.preventDefault();
+                    search?.onSearch(search?.searchInput);
                   }}
-                  value={search?.searchInput}
-                />
+                >
+                  <OdsInput
+                    isClearable
+                    onOdsClear={() => {
+                      search?.onSearch(null);
```
fredericvilcot, line 93: We could take the opportunity to fix the `onSearch` signature, because this is not correct: 
  
```ts
onSearch: (search: string | null) => void;
```

packages/manager-react-components/src/components/datagrid/datagrid-topbar.component.tsx
```diff
@@ -80,28 +80,35 @@ export const DatagridTopbar = <T,>({
...
+                      search?.setSearchInput('');
```
fredericvilcot, line 94: BTW, I would have explicitly add nullity, over an empty string.
```suggestion
                      search?.setSearchInput(null);
```

packages/manager-react-components/src/components/datagrid/datagrid-topbar.component.tsx
```diff
@@ -80,28 +80,35 @@ export const DatagridTopbar = <T,>({
...
+                    }}
+                    type={ODS_INPUT_TYPE.search}
+                    id="datagrid-searchbar"
+                    name="datagrid-searchbar"
+                    placeholder={search?.placeholder}
+                    defaultValue={search?.searchInput}
+                    data-testid="datagrid-searchbar"
+                    onOdsChange={(event) =>
+                      search?.setSearchInput(event?.detail?.value?.toString())
```
fredericvilcot, line 103: Doing this could invoke `setSearchInput` method with possibly `undefined`, that violates function signature. (`setSearchInput: React.Dispatch<React.SetStateAction<string>>;`)

packages/manager-react-components/src/components/datagrid/datagrid-topbar.component.tsx
```diff
@@ -80,28 +80,35 @@ export const DatagridTopbar = <T,>({
...
+                    }
+                    onKeyDown={(event) => {
+                      if (event.key === 'Enter')
+                        search?.onSearch(search?.searchInput);
```
fredericvilcot, line 107: Idem, could invoke `onSearch` with `undefined`

packages/manager-react-components/src/components/datagrid/datagrid-topbar.component.tsx
```diff
@@ -80,28 +80,35 @@ export const DatagridTopbar = <T,>({
...
-                  data-testid="datagrid-searchbar"
-                  onOdsChange={(event) =>
-                    search?.setSearchInput(event?.detail?.value?.toString())
-                  }
-                  onKeyDown={(event) => {
-                    if (event.key === 'Enter')
-                      search?.onSearch(search?.searchInput);
+                  onSubmit={(e) => {
+                    e.preventDefault();
+                    search?.onSearch(search?.searchInput);
```
fredericvilcot, line 87: `search?.searchInput` could be `undefined`

fredericvilcot (changes requested):

packages/manager-react-components/src/components/datagrid/datagrid-topbar.component.tsx
tristanwagner: Please check again, now it won't be ever called again with null, if it's empty then it will be an empty string instead, onSearch will never be called with undefined or null value, `search.searchInput` can't be undefined unless `search` it self is not defined according to the type of `SearchProps`

packages/manager-react-components/src/components/datagrid/datagrid-topbar.component.tsx
tristanwagner: it can't according to `SearchProps`, unless `search` is itself undefined, which would mean that `onSearch` won't be called,

fredericvilcot (approved):

anooparveti (approved):

PR #16485, opened 2025-04-09
fix(pci-gateway): get the value of the size input from global store

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-3626

### DIFF
--- a/packages/manager/apps/pci-gateway/src/pages/add/SizeStep.tsx
+++ b/packages/manager/apps/pci-gateway/src/pages/add/SizeStep.tsx
@@ -1,5 +1,5 @@
 import { useTranslation } from 'react-i18next';
-import { useContext, useState } from 'react';
+import { useContext } from 'react';
 import { OsdsText } from '@ovhcloud/ods-components/react';
 import {
   StepComponent,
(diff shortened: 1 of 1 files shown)


packages/manager/apps/pci-gateway/src/pages/add/SizeStep.tsx
```diff
@@ -38,6 +38,11 @@ export const SizeStep = (): JSX.Element => {
     store.form.regionName,
   );
 
+  useEffect(() => {
+    // reset on every region change
+    setSize(null);
```
fredericvilcot, line 43: Is there a way to achieve that in the store instead of here?
I mean, we could consider each time we set the regionName in the form, we set the size to null ?

SimonChaumet (approved):

packages/manager/apps/pci-gateway/src/pages/add/SizeStep.tsx
fredericvilcot: I can't figure out the purpose of the local state?
You store the size object, but you can extract the size itself to give it to the `TilesInputComponent` value no?

Tsiorifamonjena: Yes, you are right. The local state is no more needed

SimonChaumet (approved):

fredericvilcot (approved):

PR #16514, opened 2025-04-11
feat(manager-react-components): datagrid component add filterVariant and filterLabel props

kb-ovh:
## Description

Ticket Reference: #MANAGER-17858


anooparveti (changes requested):
**question:** For all the datagrids, Filter button must be same (variant and label). What's the need to change it?

helabenkhalfallah (approved):

anooparveti (approved):

PR #16524, opened 2025-04-11
refactor(web-office): apply the good folder structure

ghyenne:
## Description

Ticket Reference: #MANAGER-17108


tristanwagner (dismissed):

tristanwagner:
I feel like overall it's kinda verbose, like why put "Office" everywhere, like "useOfficeLicenseDetails" "useOfficeLicenses" etc, to me it's obvious that it's related to Office, so we could omit that.

Same for modals, I don't really think we should do `users/modalChangePasswordUsers/ModalChangePasswordUsers.page.tsx` can't we just go simple like `users/changePassword/ChangePassword.modal.tsx`

It really makes a lot of unnecessary noise

tristanwagner (dismissed):

SimonChaumet (approved):

tristanwagner (approved):

tristanwagner (approved):

PR #16539, opened 2025-04-14
fix(kms): verticaly align buttons in credentials

mathieumousnier:
ref: #MANAGER-16970

### DIFF
--- a/packages/manager/apps/key-management-service/src/pages/credential/generalInformations/generalInformations.page.tsx
+++ b/packages/manager/apps/key-management-service/src/pages/credential/generalInformations/generalInformations.page.tsx
@@ -99,7 +99,7 @@ const CredentialGeneralInformations = () => {
       id: 'actions',
       label: t('key_management_service_credential_dashboard_actions'),
       value: (
-        <div className="flex items-center gap-4">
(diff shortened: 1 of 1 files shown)


packages/manager/apps/key-management-service/src/pages/credential/generalInformations/generalInformations.page.tsx
```diff
@@ -99,7 +99,7 @@ const CredentialGeneralInformations = () => {
       id: 'actions',
       label: t('key_management_service_credential_dashboard_actions'),
       value: (
-        <div className="flex items-center gap-4">
+        <div className="flex gap-2 flex-col">
```
helabenkhalfallah, line 102: is div the correct semantic tag to use? it should not be article?

darsene (approved):

packages/manager/apps/key-management-service/src/pages/credential/generalInformations/generalInformations.page.tsx
mathieumousnier: This div contains a link and a button.
Doesn't feel like what I would put in an article tag, but if you have some details I would be opened to consider it.

PR #16553, opened 2025-04-14
Build/upgrade turbo lerna

kqesar:
## Description

Ticket Reference: #...


scripts/start-application.mjs
```diff
@@ -105,16 +105,16 @@ inquirer
         const appId = await getApplicationId(packageName);
         await concurrently(
           [
-            `VITE_CONTAINER_APP=${appId} yarn workspace ${containerPackageName} run start:dev`,
-            `CONTAINER=1 yarn workspace ${packageName} run start:dev`,
+            `VITE_CONTAINER_APP=${appId} turbo run start --filter=${containerPackageName}`,
+            `CONTAINER=1 turbo run start --filter=${packageName}`,
```
SimonChaumet, line 109: On my computer I have a lot of issues when using turbo due to a lot of files access, I think we need to do some performances checks before migrating

turbo.json
```diff
@@ -1,13 +1,22 @@
 {
-  "$schema": "https://turborepo.org/schema.json",
-  "pipeline": {
+  "$schema": "https://turbo.build/schema.json",
+  "daemon": true,
```
SimonChaumet, line 3: This could be removed since true is the default value

scripts/start-application.mjs
frenautvh: What kind of issues & on which platform ?

If it's too many disk IO, @kqesar i would suggest disabling turbo daemon as i find the performance gain to be negligible and i did encouter some issues on OSX with it.

Regarding benchmarks, we did some test on our dev environement, starting an application the fist time (without cache) is taking the same time as before (~30s-1min). However starting an application a second time once the turbo cache is available, takes only a few seconds (<5s)

SimonChaumet: on Linux my CPU was at 100% for 10 minutes only due to the daemon

SimonChaumet: I will do a test this afternoon on this branch to check, it might be due to an old turbo versions that might have been fixed

SimonChaumet: After some tests this issues doesn't exist anymore, everything is fine except some errors on turbo daemon, but it is not specific to this project

package.json
```diff
@@ -125,10 +124,11 @@
     "stylelint-config-standard": "^20.0.0",
     "stylelint-prettier": "^1.1.2",
     "stylelint-scss": "^3.18.0",
-    "turbo": "^1.4.3",
+    "turbo": "^2.5.0",
     "vite": "^5.4.12",
     "whatwg-fetch": "^3.6.2"
   },
+  "packageManager": "yarn@1.22.22",
```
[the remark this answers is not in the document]
kqesar: We can't use carret with `packageManager` entry

This version is the last of yarn 1 so if necessary you can update the version of local yarn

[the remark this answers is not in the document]
kqesar: Mandatory to upgrade turbo to 2.x https://turbo.build/docs/crafting-your-repository/upgrading#add-a-packagemanager-field-to-root-packagejson

SimonChaumet (dismissed):

tristanwagner (approved):

SimonChaumet (dismissed):

tristanwagner (approved):

SimonChaumet (approved):

tristanwagner (approved):

frenautvh (approved):

helabenkhalfallah (approved):

aboungnaseng-ovhcloud (approved):

ghyenne (approved):

jperchoc (approved):

kb-ovh (approved):

PR #16563, opened 2025-04-14
feat(manager-react-components): add position props to actionmenu comp…

kb-ovh:
## Description
Add `position` props to ActionMenu component.

Ticket Reference: #MANAGER-17893

### DIFF
--- a/packages/manager-react-components/src/components/navigation/menus/action/action.component.tsx
+++ b/packages/manager-react-components/src/components/navigation/menus/action/action.component.tsx
@@ -4,6 +4,7 @@ import {
   ODS_BUTTON_SIZE,
   ODS_ICON_NAME,
   ODS_BUTTON_COLOR,
+  ODS_POPOVER_POSITION,
 } from '@ovhcloud/ods-components';
 import { OdsButton, OdsPopover } from '@ovhcloud/ods-components/react';
 import { useTranslation } from 'react-i18next';
@@ -37,6 +38,7 @@ export interface ActionMenuProps {
   id: string;
   isDisabled?: boolean;
   isLoading?: boolean;
+  popoverPosition?: ODS_POPOVER_POSITION;
 }
 
 const MenuItem = ({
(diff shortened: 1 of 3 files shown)


packages/manager-react-components/src/components/navigation/menus/action/action.component.tsx
```diff
@@ -79,6 +81,7 @@ export const ActionMenu: React.FC<ActionMenuProps> = ({
   isDisabled = false,
   isLoading = false,
   id,
+  popoverPosition,
```
helabenkhalfallah, line 84: we did not need a default value here?

packages/manager-react-components/src/components/navigation/menus/action/action.spec.tsx
```diff
@@ -6,11 +6,13 @@ import { ActionMenu, ActionMenuProps } from './action.component';
 import { render } from '../../../../utils/test.provider';
 import { useAuthorizationIam } from '../../../../hooks/iam';
 import { IamAuthorizationResponse } from '../../../../hooks/iam/iam.interface';
+import { ODS_OVERLAY_POSITION } from '@ovhcloud/ods-components/dist/types/utils/overlay';
 
 vitest.mock('../../../../hooks/iam');
 
 const actionItems: ActionMenuProps = {
   id: 'action-menu-test-id',
+  popoverPosition: ODS_OVERLAY_POSITION.bottomEnd,
```
helabenkhalfallah, line 15: if the default position is bottomEnd then we should not change the test and the test should continue to work :)

packages/manager-react-components/src/components/navigation/menus/action/action.component.tsx
kb-ovh: field can be null

packages/manager-react-components/src/components/navigation/menus/action/action.spec.tsx
kb-ovh: ok i remove this line

packages/manager-react-components/src/components/navigation/menus/action/action.component.tsx
helabenkhalfallah: Oki :)

helabenkhalfallah (approved):
Thank you :)

anooparveti (approved):

PR #16574, opened 2025-04-15
fix(manager-react-components): target property

louisbensiovh:
ref: #MANAGER-17877

## Description
In the action menu, the target is missing in the link, so you can't do a target="_blank"

Ticket Reference: #MANAGER-17877

### DIFF
--- a/packages/manager-react-components/src/components/navigation/menus/action/action.component.tsx
+++ b/packages/manager-react-components/src/components/navigation/menus/action/action.component.tsx
@@ -58,7 +58,7 @@ const MenuItem = ({
 
   if (item.href) {
     return (
-      <a href={item.href} download={item.download}>
+      <a href={item.href} download={item.download} target={item.target}>
         <OdsButton {...buttonProps} />
       </a>
     );


tristanwagner (approved):

anooparveti (approved):

kb-ovh (approved):

packages/manager-react-components/src/components/navigation/menus/action/action.component.tsx
```diff
@@ -58,7 +58,7 @@ const MenuItem = ({
 
   if (item.href) {
     return (
-      <a href={item.href} download={item.download}>
+      <a href={item.href} download={item.download} target={item.target}>
```
aboungnaseng-ovhcloud, line 61: 'target' can be optional no as 'download' ?

louisbensiovh: Yes target is optionnal in interface ActionMenuItem !

aboungnaseng-ovhcloud (approved):

Issue #16454, opened 2025-04-08
[manager-components]: ServiceCancelModal


aottr:
### Have you already checked if a similar item is present on manager-components?

- [x] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

Requested is a generic cancellation modal for services, which does not require the input-field anymore and shows 2 buttons instead.

The headline should display the name of the service that is about to be canceled.

### Texts:
**Title:** `Résilier mon [type of service]`
**Body:**
```
La résiliation de votre service effacera définitivement toutes les données correspondantes.
Souhaitez-vous procéder à la résiliation de votre service ?
```

**2 CTAs:**
- `Non, refuser`
- `Oui, résilier`

![Image](https://github.com/user-attachments/assets/b89e10a4-5c26-4565-8f79-39aa5a781135)

### Where do you expect to use this?

kms

### Do you have mock-up?

yes

### When do you expect this to be delivered?

25-04-2025

### Additional Information

Idea is to:
- re-use the current Modal component in MRCv2 and backport it to MRCv1
- create template components for the ServiceCancelModal in MRCv1 + MRCv2
- decommission the DeleteModal that is currently present in MRCv1 + MRCv2 without introducing a breaking change on uApps that are using it.

aboungnaseng-ovhcloud:
Hello Dustin, you can modify the Delete Modal in MRC V1 and V2, to delete the input text

PR #16585, opened 2025-04-15
feat(mrc): update delete-modal to FTC requests (mrc2) #16454

aottr:
## Description

Changes the behavior of the Delete-Modal to be compliant with MANAGER-17707 while ensuring backwards-compatibility with current implementations.

Ticket Reference: #MANAGER-17707


packages/manager-react-components/src/components/templates/delete-modal/delete-modal.component.tsx
```diff
@@ -66,51 +82,37 @@ export const DeleteModal: React.FC<DeleteModalProps> = ({
...
-            value={deleteInput}
-            onOdsChange={(e: OdsInputCustomEvent<OdsInputChangeEventDetail>) =>
-              setDeleteInput(e.detail.value as string)
-            }
-          />
-        </OdsFormField>
+        <OdsText preset={ODS_TEXT_PRESET.paragraph}>
+          {t('deleteModalDescription')}
+        </OdsText>
+        {children}
```
anooparveti, line 98: **question:** If the modal is always the same, why do we need `children`?

aottr: Some teams might need to add additional information about the service that needs to be canceled, aside of the service name. @JacquesLarique

JacquesLarique: I would even say it is almost guaranteed that feature teams will want to add specific text (ex: vrack services or KMS have requirement before the resiliation is possible which they will want to indicate to user in this modal)

anooparveti (dismissed):

packages/manager-react-components/src/components/templates/delete-modal/__snapshots__/delete-modal.spec.tsx.snap
JacquesLarique, line 1: Is this file required ?

JacquesLarique (dismissed):

anooparveti (dismissed):

JacquesLarique (approved):

MaximeBajeux (approved):

aboungnaseng-ovhcloud (approved):

PR #16589, opened 2025-04-16
feat(pnr): remove switchback from pnr eu ca

MaximeBajeux:
## Description

As US is not ready to fully adopt pnr, we keep the switch back button only for us

Ticket Reference: #MANAGER-16732


oalkabouss (approved):

packages/manager/apps/container/src/container/legacy/navbar/Navbar.tsx
```diff
@@ -47,6 +47,7 @@ function Navbar({ environment }: Props): JSX.Element {
   const [isDropdownOpen, setIsDropdownOpen] = useState<boolean>(false);
   const { setIsNotificationsSidebarVisible } = useHeader();
   const { t } = useTranslation(TRANSLATE_NAMESPACE);
+  const isUS = shell?.getPlugin('environment')?.getEnvironment()?.getRegion() === 'US';
```
SimonChaumet, line 50: Why not use feature flipping instead ?

ghyenne (dismissed):

oalkabouss (approved):

PR #16593, opened 2025-04-16
refactor(pci-gateway): use common hook to get catalog

Tsiorifamonjena:
## Description

The goal of this PR is to use a common hook to get catalog and product availability. Some µapps have the same logic and then this hook will be moved to a pci-common type library.

Ticket Reference: #...


packages/manager/apps/pci-gateway/src/types/product.type.ts
```diff
@@ -0,0 +1,9 @@
+import { TProductAvailabilityRegion } from '@ovh-ux/manager-pci-common';
+
+export type TProductAddonDetail = {
+  size: string;
+  price: number;
+  technicalName: string;
+  bandwidth: number;
+  regions: TProductAvailabilityRegion[];
+};
```
SimonChaumet, line 9: isn't this type already defined with TAddon ?

SimonChaumet (dismissed):

SimonChaumet (approved):

PR #16602, opened 2025-04-17
feat(dedicated): add guid link for ip listing

mhelhali-soufien:
ref: #MANAGER-17943

## Description

Ticket Reference: #MANAGER-17943


qpavy (approved):

sachinrameshn (changes requested):
You need to update guide links for other subs as well as of now it as only for FR

sachinrameshn (dismissed):

qpavy (approved):

qpavy (dismissed):

packages/manager/apps/dedicated/client/app/app.less
```diff
@@ -2,6 +2,11 @@
 
 @import 'css/source.less';
 
+.ip-guide-block {
+  flex-direction: column;
+  align-items: flex-end;
```
SimonChaumet, line 7: aren't these available with bootstrap ?

packages/manager/apps/dedicated/client/app/ip/ip.controller.js
```diff
@@ -58,6 +58,12 @@ export default /* @ngInject */ function IpMainCtrl(
     );
   }
 
+  const { ovhSubsidiary } = $scope.currentUser;
+  $scope.guides = Object.keys(GUIDE_LINKS).map((key) => ({
+    name: $translate.instant(`ip_guide_link_${key}`),
```
SimonChaumet, line 63: why not transate it inside the DOM like others ?

sachinrameshn (dismissed):

packages/manager/apps/dedicated/client/app/ip/ip.controller.js
mhelhali-soufien: I prefer function call directly more than use directive in the DOM to translate

packages/manager/apps/dedicated/client/app/app.less
mhelhali-soufien: I think yes but anyway this code will be removed soon when bump to react be done

qpavy (approved):

tibs245 (approved):

packages/manager/apps/dedicated/client/app/app.less
aderghamov: It looks like this class is specific to the ip (should-be)component.
I think it should be defined into the `packages/manager/apps/dedicated/client/app/` directory into a dedicated file.

PR #16613, opened 2025-04-17
feat(pci-instances): add edit route & action link

fredericvilcot:
ref: #TAPC-2163

## Description

Ticket Reference: #...


Eric-ciccotti (approved):

packages/manager/apps/pci-instances/src/routes/routes.tsx
```diff
@@ -96,6 +97,12 @@ const routes: RouteObject[] = [
           import('@/pages/instances/instance/Instance.page'),
         ),
       },
+      {
+        path: SECTIONS.edit,
+        ...lazyRouteConfig(() =>
+          import('@/pages/instances/instance/edit/Edit.page'),
+        ),
+      },
```
SimonChaumet, line 105: If we plan to use the migration I don't think this is useful, the link is sufficient I think

Tsiorifamonjena (approved):

packages/manager/apps/pci-instances/src/routes/routes.tsx
fredericvilcot: I'll keep it to prevent from QA false negative, and to initiate the work on next pages

SimonChaumet (approved):

PR #16647, opened 2025-04-18
feat(web-hosting): add virtual row for load all

stif59100:
ref: #MANAGER-17890

## Description

I added virtual row for load all

Ticket Reference: #MANAGER-17890


tristanwagner (approved):

anooparveti (changes requested):
**nitpick:** This must be added to the Datagrid component itself to avoid implementing this in every page.

anooparveti (approved):

PR #16655, opened 2025-04-21
feat(ips.onboarding): format onboarding content based on feedback

sachinrameshn:
ref: #MANAGER-15777

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/ips/src/pages/listing/index.tsx
+++ b/packages/manager/apps/ips/src/pages/listing/index.tsx
@@ -4,7 +4,7 @@ import {
   useResolvedPath,
   useLocation,
   Outlet,
-  NavLink,
+  useNavigate,
 } from 'react-router-dom';
 import { OdsTabs, OdsTab } from '@ovhcloud/ods-components/react';
 import { ShellContext } from '@ovh-ux/manager-react-shell-client';
@@ -39,6 +39,7 @@ export default function Listing() {
       label: t('ips_dashboard_guide'),
     },
   ];
+  const navigate = useNavigate();
 
   const tabsList = [
     {
@@ -69,10 +70,9 @@ export default function Listing() {
               isSelected={location.pathname === tab.to}
               className="flex items-center justify-center"
               title={tab.title}
+              onClick={() => navigate(tab.to)}
             >
-              <NavLink to={tab.to} className="flex no-underline">
-                {tab.title}
-              </NavLink>
+              {tab.title}
             </OdsTab>
           ))}
         </OdsTabs>
--- a/packages/manager/apps/ips/src/pages/onboarding/index.tsx
+++ b/packages/manager/apps/ips/src/pages/onboarding/index.tsx
@@ -1,6 +1,6 @@
 import React from 'react';
 import { useTranslation } from 'react-i18next';
-import { ShellContext } from '@ovh-ux/manager-react-shell-client';
+import { useNavigate } from 'react-router-dom';
 import {
   Card,
   OnboardingLayout,
@@ -8,19 +8,16 @@ import {
 } from '@ovh-ux/manager-react-components';
 import { OdsText, OdsTable } from '@ovhcloud/ods-components/react';
 import { ODS_TEXT_PRESET } from '@ovhcloud/ods-components';
-import { useNavigate } from 'react-router-dom';
 
 import useGuideUtils from '@/pages/onboarding/useGuideUtils';
 import onboardingImgSrc from './onboarding-img.png';
 import { urls } from '@/routes/routes.constant';
-import { useCatalogLowestPrice } from '@/data/hooks/catalog';
-import { PriceDescription } from '@/components/PriceDescription/PriceDescription';
+import { OnboardingIpOptions } from './onboardingIpOptionsAdvantages';
+import { IpOptionRow } from './onboardingIpOptionRow';
 
 export default function Onboarding() {
   const { t } = useTranslation('onboarding');
-  const { t: tCommon } = useTranslation();
   const link = useGuideUtils();
-  const { ipv4LowestPrice, ipv6LowestPrice } = useCatalogLowestPrice();
   const navigate = useNavigate();
 
   const tileList = [
@@ -53,61 +50,6 @@ export default function Onboarding() {
     },
   ];
 
-  const IP_OPTIONS_ROWS = [
-    {
-      feature: t('optionsAddressTypes'),
-      ipv4: t('optionsAddressTypesIpv4'),
-      ipv6: t('optionsAddressTypesIpv6'),
-    },
-    {
-      feature: t('optionsModes'),
-      ipv4: t('optionsModesIpv4'),
-      ipv6: t('optionsModesIpv6'),
-    },
-    {
-      feature: t('optionsGeolocation'),
-      ipv4: t('optionsGeolocationIpv4'),
-      ipv6: t('optionsGeolocationIpv6'),
-    },
-    {
-      feature: t('optionsCost'),
-      ipv4: (
-        <PriceDescription
-          shouldOverrideStyle={true}
-          isStartingPrice
-          price={ipv4LowestPrice}
-          suffix={tCommon('per_ip')}
-        />
-      ),
-      ipv6: (
-        <PriceDescription price={ipv6LowestPrice} shouldOverrideStyle={true} />
-      ),
-    },
-    {
-      feature: t('optionsCompatibleServices'),
-      ipv4: t('optionsCompatibleServicesIpv4'),
-      ipv6: t('optionsCompatibleServicesIpv6'),
-    },
-    {
-      feature: t('optionsAvailability'),
-      ipv4: t('optionsAvailabilityIpv4'),
-      ipv6: (
-        <>
-          {t('optionsAvailabilityCheck')}{' '}
-          <Links
-            href={link?.presentationLink}
-            label={t('optionsAvailabilityHere')}
-          />
-        </>
-      ),
-    },
-    {
-      feature: t('optionsByoip'),
-      ipv4: t('optionsByoipIpv4'),
-      ipv6: t('optionsByoipIpv6'),
-    },
-  ];
-
   const imgSrc = {
     src: onboardingImgSrc,
   };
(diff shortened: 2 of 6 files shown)


aderghamov (approved):

chipp972 (approved):

packages/manager/apps/ips/src/pages/onboarding/index.tsx
```diff
@@ -184,22 +201,44 @@ export default function Onboarding() {
                           className="text-left pl-4"
                           style={{ whiteSpace: 'pre-line' }}
                         >
-                          {row.ipv4}
+                          {typeof row?.ipv4 === 'string' && row.applyBulletStyle
+                            ? row.ipv4
+                                ?.split('\n')
+                                .map((line: string, i: number) => (
+                                  <span key={i}>
```
helabenkhalfallah, line 208: it's not a good pattern to have index as key ;) 
https://robinpokorny.com/blog/index-as-a-key-is-an-anti-pattern/
https://react.dev/learn/rendering-lists#keeping-list-items-in-order-with-key

packages/manager/apps/ips/src/pages/onboarding/index.tsx
```diff
@@ -184,22 +201,44 @@ export default function Onboarding() {
...
+                                  </span>
+                                ))
+                            : row.ipv4}
                         </td>
                         <td
                           className="text-left pl-4"
                           style={{ whiteSpace: 'pre-line' }}
                         >
-                          {row.ipv6}
+                          {typeof row?.ipv6 === 'string' && row.applyBulletStyle
```
helabenkhalfallah, line 219: this code is duplicated as for `typeof row?.ipv4 === 'string'` you should create a common logic and component I think. And the same issue for key: https://github.com/ovh/manager/pull/16655/files#r2053800464

packages/manager/apps/ips/src/pages/listing/index.tsx
```diff
@@ -54,10 +55,9 @@ export default function Listing() {
               isSelected={location.pathname === tab.to}
               className="flex items-center justify-center"
               title={tab.title}
+              onClick={() => navigate(tab.to)}
```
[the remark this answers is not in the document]
sachinrameshn: You are right. Sure I will use useCallback to create a memoized handleOnTabClick function that only changes when navigate changes.

packages/manager/apps/ips/src/pages/onboarding/index.tsx
```diff
@@ -127,32 +129,47 @@ export default function Onboarding() {
                 {t('advantagesTitle')}
               </OdsText>
               <div className="text-left mb-8">
-                <ol className="list-decimal pl-6 space-y-4">
+                <ul className="list-disc pl-6 space-y-4">
```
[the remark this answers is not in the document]
sachinrameshn: Okay I will try to use map here.

packages/manager/apps/ips/src/pages/onboarding/index.tsx
sachinrameshn: Created a component for creating the bullet point and also modified the usage of index as key.

packages/manager/apps/ips/src/pages/onboarding/index.tsx
sachinrameshn: done

packages/manager/apps/ips/src/pages/listing/index.tsx
```diff
@@ -39,6 +39,14 @@ export default function Listing() {
...
     },
   ];
+  const navigate = useNavigate();
+
+  const onTabClick = useCallback(
+    (path: string) => {
+      navigate(path);
+    },
+    [navigate],
+  );
```
chipp972, line 49: That is equivalent to using navigate directly since you recreate a function every time the function changes.

packages/manager/apps/ips/src/pages/onboarding/index.tsx
```diff
@@ -127,32 +170,24 @@ export default function Onboarding() {
...
-                  <li>
-                    <strong>{t('advantage3Title')}:</strong>{' '}
-                    {t('advantage3Description')}
-                  </li>
-                  <li>
-                    <strong>{t('advantage4Title')}:</strong>{' '}
-                    {t('advantage4Description')}
-                  </li>
+                <ul className="list-disc pl-6 space-y-4">
+                  {IP_OPTIONS_ADVANTAGES}
```
chipp972, line 174: Prefer creating a separate component and use jsx instead
```
<IpOptionsAdvantages />
```

packages/manager/apps/ips/src/pages/onboarding/index.tsx
```diff
@@ -174,32 +209,54 @@ export default function Onboarding() {
...
+                          {typeof row?.ipv6 === 'string' &&
+                          row.applyBulletStyle ? (
+                            <BulletPointList
+                              content={row.ipv6}
+                              feature={row.feature}
+                              type="ipv6"
+                            />
+                          ) : (
+                            row.ipv6
+                          )}
```
chipp972, line 246: You can simplify this by returning a ReactNode from ipv4 and ipv6 properties :
- remove applyBulletStyle boolean
- put the jsx for ipv4 and ipv6 directly in the array
- {row.ipv4} or {row.ipv6} here

packages/manager/apps/ips/src/pages/onboarding/index.tsx
```diff
@@ -174,32 +209,54 @@ export default function Onboarding() {
                     </tr>
                   </thead>
                   <tbody>
-                    {IP_OPTIONS_ROWS.map((row, index) => (
+                    {IP_OPTIONS_ROWS.map((row) => (
                       <tr
```
chipp972, line 213: if would be easier to read if you create a separate component for each line of the table:

```
<IpOptionRow key={row.feature} {...row} />
```

packages/manager/apps/ips/src/pages/onboarding/index.tsx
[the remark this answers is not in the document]
chipp972: There is no unwanted re-rendering if the data is static though
No need to create dynamic data sources in this case, unless we want to factor the code and create subcomponents

chipp972:
Also we can add a simple snapshot test to make sure everything renders well and track any breaking changes in case we update our dependencies

packages/manager/apps/ips/src/pages/onboarding/index.tsx
```diff
@@ -16,9 +15,24 @@ import { urls } from '@/routes/routes.constant';
...
+  content: string;
+  feature: string;
+  type: 'ipv4' | 'ipv6';
+}
+
+const BulletPointList = ({ content, feature, type }: BulletPointListProps) => {
+  return content?.split('\n').map((line: string) => (
+    <span key={`${feature}-${type}-${line}`}>
+      {'• '}
+      {`${line}\n`}
```
chipp972, line 28: if you use a div you can just skip the \n or maybe use `{line}<br />` instead

packages/manager/apps/ips/src/pages/onboarding/index.tsx
sachinrameshn: done

packages/manager/apps/ips/src/pages/onboarding/index.tsx
sachinrameshn: done

packages/manager/apps/ips/src/pages/onboarding/onboardingIpOptionRow.tsx
```diff
@@ -0,0 +1,132 @@
...
+import { useTranslation } from 'react-i18next';
+import { Links } from '@ovh-ux/manager-react-components';
+import { PriceDescription } from '@/components/PriceDescription/PriceDescription';
+import { useCatalogLowestPrice } from '@/data/hooks/catalog';
+import useGuideUtils from '@/pages/onboarding/useGuideUtils';
+
+interface IpOptionRowProps {
+  feature: string;
+  ipv4: string | React.ReactNode;
+  ipv6: string | React.ReactNode;
```
chipp972, line 11:
```suggestion
  ipv4: React.ReactNode;
  ipv6: React.ReactNode;
```

string is included inside ReactNode

packages/manager/apps/ips/src/pages/onboarding/onboardingIpOptionRow.tsx
```diff
@@ -0,0 +1,132 @@
...
+interface BulletPointListProps {
+  content: string;
+  feature: string;
+}
+
+const BulletPointList = ({ content, feature }: BulletPointListProps) => {
+  return content?.split('\n').map((line: string) => (
+    <div key={`${feature}-${line}`}>
+      {'• '}
+      {`${line}`}
```
chipp972, line 22:
```suggestion
      {line}
```

chipp972, line 22:
```suggestion
      {'• ${line}'}
```

packages/manager/apps/ips/src/pages/onboarding/onboardingIpOptionRow.tsx
```diff
@@ -0,0 +1,132 @@
...
+      ipv4: (
+        <PriceDescription
+          shouldOverrideStyle={true}
+          isStartingPrice
+          price={ipv4LowestPrice}
+          suffix={tOrder('per_ip')}
+        />
+      ),
+      ipv6: (
+        <PriceDescription price={ipv6LowestPrice} shouldOverrideStyle={true} />
```
chipp972, line 79:
```suggestion
        <PriceDescription price={ipv6LowestPrice} shouldOverrideStyle />
```

packages/manager/apps/ips/src/pages/onboarding/onboardingIpOptionRow.tsx
```diff
@@ -0,0 +1,132 @@
...
+      ipv6: t('optionsByoipIpv6'),
+    },
+  ];
+};
+
+export function IpOptionRow({ feature, ipv4, ipv6 }: IpOptionRowProps) {
+  return (
+    <tr className="border-solid border-[1px] border-[--ods-color-blue-200]">
+      <td className="text-left pl-4">{feature}</td>
+      <td className="text-left pl-4" style={{ whiteSpace: 'pre-line' }}>
```
chipp972, line 122:
```suggestion
      <td className="text-left pl-4 whitespace-pre-line">
```

It doesn't work when using tailwind ?

packages/manager/apps/ips/src/pages/listing/index.tsx
```diff
@@ -69,10 +77,9 @@ export default function Listing() {
               isSelected={location.pathname === tab.to}
               className="flex items-center justify-center"
               title={tab.title}
+              onClick={() => onTabClick(tab.to)}
```
chipp972, line 80:
```suggestion
              onClick={() => navigate(tab.to)}
```

onTabClick does not prevent re-render here so no need to create this function

sachinrameshn: ok

packages/manager/apps/ips/src/pages/onboarding/onboardingIpOptionRow.tsx
```diff
@@ -0,0 +1,123 @@
...
+  feature: string;
+}
+
+const BulletPointList = ({ content, feature }: BulletPointListProps) => {
+  return content
+    ?.split('\n')
+    .map((line: string) => <div key={`${feature}-${line}`}>{`• ${line}`}</div>);
+};
+
+export const getIpOptionsRows = () => {
```
chipp972, line 24:
```suggestion
export const useIpOptionsRows = () => {
```

Actually it's a hook since you use hooks inside it

And I think you can just do a component `IpOptionsRows` by returning the rendered rows directly since the hook will not be reused so no need to create a hook and export it.

packages/manager/apps/ips/src/pages/onboarding/onboardingIpOptionRow.tsx
```diff
@@ -0,0 +1,127 @@
...
+      feature: t('optionsByoip'),
+      ipv4: t('optionsByoipIpv4'),
+      ipv6: t('optionsByoipIpv6'),
+    },
+  ];
+  return (
+    <>
+      {optionList.map((option: IpOptionRowProps) => (
+        <tr
+          key={`${option.feature}-${option.ipv4}-${option.ipv6}`}
```
chipp972, line 115:
```suggestion
          key={option.feature}
```

Since ipv4 and ipv6 are ReactNode we should avoid putting them in key
feature parameter is enough since it differ for each item

sachinrameshn: Okay

packages/manager/apps/ips/src/pages/listing/index.tsx
```diff
@@ -1,10 +1,10 @@
-import React, { useContext } from 'react';
+import React, { useContext, useCallback } from 'react';
```
chipp972, line 1:
```suggestion
import React, { useContext } from 'react';
```

PR #16674, opened 2025-04-22
fix(dedicated): number of disks in os reinstallation

JayBeeDe:
## Description

Number of disks parameter is not taken into account in the API payload to reinstall the OS on the dedicated servers.
It is quasi always set to 0, meaning all disks.
This is this issue that this PR will fix.

Ticket Reference: MANAGER-18029

## Additional Information

No dependencies, no translation, ready to be deployed once acked

### DIFF
--- a/packages/manager/modules/bm-server-components/src/general-information/installation/ovh/server-installation-ovh.controller.js
+++ b/packages/manager/modules/bm-server-components/src/general-information/installation/ovh/server-installation-ovh.controller.js
@@ -2207,8 +2207,8 @@ export default class ServerInstallationOvhCtrl {
       hardwareRaid: [],
       partitioning: {
         disks:
-          this.$scope.informations.nbDisk > 2 &&
-          this.$scope.installation.nbDiskUse > 1
+          this.$scope.informations.nbDisk > 1 &&
+          this.$scope.installation.nbDiskUse
             ? this.$scope.installation.nbDiskUse
             : 0,
       },


sachinrameshn (approved):

qpavy (approved):

packages/manager/modules/bm-server-components/src/general-information/installation/ovh/server-installation-ovh.controller.js
```diff
@@ -2207,8 +2207,8 @@ export default class ServerInstallationOvhCtrl {
       hardwareRaid: [],
       partitioning: {
         disks:
-          this.$scope.informations.nbDisk > 2 &&
-          this.$scope.installation.nbDiskUse > 1
+          this.$scope.informations.nbDisk > 1 &&
+          this.$scope.installation.nbDiskUse
```
tibs245, line 2211: Why delete `> 1` on the second line ?

tibs245 (approved):

packages/manager/modules/bm-server-components/src/general-information/installation/ovh/server-installation-ovh.controller.js
tibs245: `>= 1` may be more readable ? As you want

JayBeeDe: because if server has only 1 single disk, then the dropdown to choose the number of disks is not displayed since user can only choose 1 disk. Therefore we don't rely on nbDiskUse and systematically set the value to 0 meaning all disks in the API

PR #16686, opened 2025-04-23
feat(hpc): add vRack column

sefiani-ovh:
ref: #MANAGER-17492


packages/manager/apps/hpc-vmware-managed-vcd/src/pages/listing/datacentres/Datacentres.constants.ts
```diff
@@ -0,0 +1 @@
+export const VRAK_LINK = 'https://www.ovh.com/manager/#/dedicated/vrack/';
```
vovh, line 1: Shouldn't it be VRACK_LINK ?

Also having the base domain as const will it not be an issue with US / EU subsidiary ?

vovh (changes requested):

tibs245 (approved):

pauldkn (approved):

pauldkn (changes requested):
Tests are failing

Issue #16714, opened 2025-04-23
[manager-components]: Add missing guides for volume backup


FabienHenon:
### Have you already checked if a similar item is present on manager-components?

- [x] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

Volume Backup is using different guides. These guides are missing

### Where do you expect to use this?

volume backup

### When do you expect this to be delivered?

24-04-2025

anooparveti:
**question:** Can we avoid putting product related information in the component library?

PR #16715, opened 2025-04-23
feat(mrc): add volume backup guides

FabienHenon:
ref: #16714

## Description

Add guides for volume backup

Ticket Reference: #16714

### DIFF
--- a/packages/manager-react-components/src/components/guides-header/pci/pci-guides-header.constants.ts
+++ b/packages/manager-react-components/src/components/guides-header/pci/pci-guides-header.constants.ts
@@ -12,8 +12,7 @@ export const PUBLIC_CLOUD_GUIDES: GuideLinks = {
   PT: 'https://docs.ovh.com/pt/public-cloud/',
   IE: 'https://docs.ovh.com/ie/en/public-cloud/',
   DEFAULT: 'https://docs.ovh.com/gb/en/public-cloud/',
-  US:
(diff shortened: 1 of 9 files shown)


sachinrameshn:
I will let control team review this before as I see linked manager issue is not yet reviewed from them.

anooparveti (changes requested):
**nitpick:** I dont like the idea of adding guides for PCI in MRC. Is it something that can be added on PCI COmmon?

helabenkhalfallah (approved):

FabienHenon:
> **nitpick:** I dont like the idea of adding guides for PCI in MRC. Is it something that can be added on PCI COmmon?

@anooparveti there are already a lot of PCI guides in this component (all of them I think). The entire component, maybe, should be moved to pci common. But it will break a lot of apps as soon as they update their MRC deps. Should I proceed?

anooparveti (changes requested):
We are not supposed to updated locale files apart from `fr_FR`.
Can you please get the changes from the translation team?

aboungnaseng-ovhcloud (approved):

FabienHenon:
> We are not supposed to updated locale files apart from `fr_FR`. Can you please get the changes from the translation team?

@anooparveti these were existing translations from the manager. I just moved them inside MRC and changed the key to be alike the other keys.
Translations themselves have not changed and existed before

anooparveti (approved):

PR #16720, opened 2025-04-23
feat(pci-block-storage): add multi detach

SimonChaumet:
ref: #TAPC-3761

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-block-storage/src/api/hooks/useInstance.ts
+++ b/packages/manager/apps/pci-block-storage/src/api/hooks/useInstance.ts
@@ -2,7 +2,10 @@ import { useQuery } from '@tanstack/react-query';
 import { useMemo } from 'react';
 import { getInstancesByRegion } from '@/api/data/instance';
 import { useVolume } from '@/api/hooks/useVolume';
-import { selectAttachableInstances } from '@/api/select/instances';
+import {
+  selectAttachableInstances,
+  selectAttachedInstances,
+} from '@/api/select/instances';
 
 const getInstancesQueryKey = (projectId: string, region?: string) => [
   'instances',
@@ -16,6 +19,22 @@ export const useAttachableInstances = (projectId: string, volumeId: string) => {
   const select = useMemo(
     () =>
       volume ? selectAttachableInstances(volume.availabilityZone) : undefined,
+    [volume],
+  );
+
+  return useQuery({
+    queryKey: getInstancesQueryKey(projectId, volume?.region),
+    queryFn: () => getInstancesByRegion(projectId, volume.region),
+    enabled: !!volume,
+    select,
+  });
+};
+
+export const useAttachedInstances = (projectId: string, volumeId: string) => {
+  const { data: volume } = useVolume(projectId, volumeId);
+
+  const select = useMemo(
+    () => (volume ? selectAttachedInstances(volume.attachedTo) : undefined),
     [volume],
   );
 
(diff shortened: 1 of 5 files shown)


Tsiorifamonjena (approved):

Tsiorifamonjena (approved):

packages/manager/apps/pci-block-storage/src/pages/detach/DetachStorage.page.tsx
```diff
@@ -10,30 +12,41 @@ import {
...
+    isPending: isInstancesPending,
+  } = useAttachedInstances(projectId, volumeId);
+
+  const [selectedInstanceId, setSelectedInstanceId] = useState<
+    TAttachedInstance['id'] | null
+  >(null);
+
+  const selectedInstance = useMemo(() => {
+    if (!!selectedInstanceId && !!instances)
+      return instances.find(({ id }) => id === selectedInstanceId)!;
```
ppprevost, line 42: why putting ! at the end for the type ?

packages/manager/apps/pci-block-storage/src/pages/detach/DetachStorage.page.tsx
```diff
@@ -85,27 +100,49 @@ export default function DetachStorage() {
...
-              },
-            )}
-          </OsdsText>
+        )}
+        {instances?.length > 1 && (
+          <div>
+            <OsdsSelect
+              value={selectedInstanceId}
+              inline
+              className="mt-5 w-[100%]"
```
ppprevost, line 116: w-full

packages/manager/apps/pci-block-storage/src/pages/detach/DetachStorage.page.tsx
```diff
@@ -69,13 +82,15 @@ export default function DetachStorage() {
...
-  const isPending = isInstancePending || isDetachPending;
-  const canDetach = !isPending;
+  const isPending = isInstancesPending || isDetachPending;
+  const canDetach = !isPending && !!selectedInstanceId;
 
   // redirect to listing if volume is not attached
   useEffect(() => {
     if (!isPending && volume?.attachedTo?.length === 0) {
       navigate(`/pci/projects/${projectId}/storages/blocks`);
+    } else if (volume?.attachedTo.length === 1) {
```
ppprevost, line 92: means that you want to test the length even if isPending is true ?

packages/manager/apps/pci-block-storage/src/api/hooks/useInstance.ts
```diff
@@ -26,3 +29,19 @@ export const useAttachableInstances = (projectId: string, volumeId: string) => {
     select,
   });
 };
+
+export const useAttachedInstances = (projectId: string, volumeId: string) => {
```
ppprevost, line 33: you don' need a use memo here since useVolume should be properly memoize within the hooks usevolume

`export const useAttachedInstances = (projectId: string, volumeId: string) => {

  return useQuery({
    queryKey: getInstancesQueryKey(projectId, volume?.region),
    queryFn: () => getInstancesByRegion(projectId, volume!.region),
    enabled: !!volume,
    select: volume ? selectAttachedInstances(volume.attachedTo) : undefined,
  });
};`

packages/manager/apps/pci-block-storage/src/pages/detach/DetachStorage.page.tsx
SimonChaumet: To make typescript happy, or else it would say that it is possibly undefined while at this point it cannot

packages/manager/apps/pci-block-storage/src/api/hooks/useInstance.ts
SimonChaumet: Without the useMemo the function is recreated on every render and the function is called

Tsiorifamonjena (approved):

Issue #16721, opened 2025-04-23
[manager-components]: improving useResourcesV6


FabienHenon:
### Have you already checked if a similar item is present on manager-components?

- [x] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

Function `useResourcesV6` has several flaws when using it:

- It uses another type for columns which makes it incompatible with `datagrid`
- It does not support endpoints that does not support iceberg headers and/or that returns another kind of payload
- It does not support default sorting configuration
- It does not support polling
- It sorts in the wrong order (asc = desc)

### Where do you expect to use this?

volume backup

### When do you expect this to be delivered?

24-04-2025

PR #16722, opened 2025-04-23
feat(mrc): improve useResourcesV6

FabienHenon:
## Description

Function useResourcesV6 has several flaws when using it:

- It uses another type for columns which makes it incompatible with datagrid
- It does not support endpoints that does not support iceberg headers and/or that returns another kind of payload
- It does not support default sorting configuration
- It does not support polling
- It sorts in the wrong order (asc = desc)

Ticket Reference: #16721

### DIFF
--- a/packages/manager-react-components/src/components/datagrid/documentation.mdx
+++ b/packages/manager-react-components/src/components/datagrid/documentation.mdx
@@ -38,11 +38,11 @@ Why Use TanStack Table?
 
 #### How to use It
 
-1 - In your columns definition, fill `type` or `comparator` attributes
+1 - In your columns definition, fill `type` and/or `comparator` attributes
 
-- `type`comparator is a `FilterTypeCategories`
+- `type` is a `FilterTypeCategories` and will allow you to have enhanced filters (datepicker for date type, etc...)
 
-- `comparator`is a `string` of `FilterCategories`, if you fill this attribute, it override the `type` attributes
+- `comparator` is a `string` of `FilterCategories`. Use this option to define the set of comparators that can be applied on the data. By default, Datagrid determines the set of comparators that can be applied depending on the `type` field. **Only use this option if default comparators does not match your need.**
 
 <Source
   code={`
@@ -62,6 +62,15 @@ Why Use TanStack Table?
             },
             label: 'Price',
             comparator: FilterCategories.String,
+        },
+        {
+            id: 'creationDate',
+            cell: (item: Item) => {
+                return <DataGridTextCell>{useFormattedDate(item.creationDate)}</DataGridTextCell>;
+            },
+            label: 'Creation date',
+            type: FilterTypeCategories.Date
+            comparator: FilterCategories.Date,
         },
     ];
     `}
--- a/packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
+++ b/packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
@@ -1,26 +1,27 @@
 import React, { useState, useEffect, useMemo, useCallback } from 'react';
 import isDate from 'lodash.isdate';
 import {
-  IcebergFetchParamsV6,
-  fetchIcebergV6,
   applyFilters,
   FilterTypeCategories,
   FilterComparator,
+  Filter,
+  v6,
 } from '@ovh-ux/manager-core-api';
-import { useQuery } from '@tanstack/react-query';
-import { useColumnFilters, ColumnSort } from '../../components';
+import { Query, useQuery } from '@tanstack/react-query';
+import { DatagridColumn, useColumnFilters, ColumnSort } from '../../components';
 
-export interface ColumnDatagrid {
-  cell: (props: any) => React.JSX.Element;
-  id: string;
-  label: string;
-  type: string;
-  isSearchable?: boolean;
-}
+export type FetchResultV6<T> = {
+  data: T[];
+};
 
-export interface ResourcesV6Hook {
+export interface ResourcesV6Params<T> {
   queryKey: string[];
-  columns: ColumnDatagrid[];
+  queryFn?: (route: string) => Promise<FetchResultV6<T>>;
+  refetchInterval?: (query: Query<FetchResultV6<T>>) => number | false;
+  columns: DatagridColumn<T>[];
+  defaultSorting?: ColumnSort;
+  route: string;
+  pageSize?: number;
 }
 
 export function dataType(a) {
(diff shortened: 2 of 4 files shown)


anooparveti:
> It does not support endpoints that does not support iceberg headers and/or that returns another kind of payload

@FabienHenon can you describe this part please?

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
```diff
@@ -6,21 +6,21 @@ import {
...
+import { DatagridColumn, useColumnFilters, ColumnSort } from '../../components';
 
-export interface ColumnDatagrid {
-  cell: (props: any) => React.JSX.Element;
-  id: string;
-  label: string;
-  type: string;
-  isSearchable?: boolean;
-}
+export type ApiData<T> = {
```
tristanwagner, line 14: You could use the `IcebergFetchResultV6<T>` instead

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
```diff
@@ -31,91 +31,86 @@ export function dataType(a) {
...
+export function useResourcesV6<T extends Record<string, unknown>>({
   route,
   queryKey,
+  queryFn,
+  refetchInterval,
   pageSize = 10,
   columns = [],
-}: IcebergFetchParamsV6 & ResourcesV6Hook) {
+  defaultSorting,
+}: IcebergFetchParamsV6 & ResourcesV6Hook<T>) {
```
tristanwagner, line 76: actually we could updated the type to `Pick<IcebergFetchParamsV6, 'route'> & ResourcesV6Hook<T>` since we only allow to pass the `route` param from `IcebergFetchParamsV6`

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
```diff
@@ -31,91 +31,86 @@ export function dataType(a) {
...
   const [searchInput, setSearchInput] = useState('');
```
aboungnaseng-ovhcloud, line 77: You don't use searchInput state, so the search don't work, I have tried on my side after generated a project

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
```diff
@@ -31,91 +31,86 @@ export function dataType(a) {
...
+        : a
             .trim()
             ?.toString()
             ?.toLowerCase()
-            ?.localeCompare?.(a?.trim()?.toLowerCase());
+            ?.localeCompare?.(b?.trim()?.toLowerCase());
     default:
       return -1;
   }
 }
```
tibs245, line 66: Maybe another ticket for refactor but in reading that I see a beautiful refactor with maybe fix for empty value :

(Example generated with Copilot)

```ts
function sortColumnAsc(
  type: FilterTypeCategories,
  a: string,
  b: string
): number {
  if (!a || !b) return -1;
  switch (type) {
    case FilterTypeCategories.Numeric:
      return parseFloat(a) - parseFloat(b);
    case FilterTypeCategories.Date:
      return new Date(a).getTime() - new Date(b).getTime();
    case FilterTypeCategories.Boolean:
      return Number(a) - Number(b);
    case FilterTypeCategories.String:
      return a
        ?.trim()
        .toLowerCase()
        ?.localeCompare(b?.trim()?.toLowerCase());
    default:
      return -1;
  }
}
```

```ts
function sortColumn(
  type: FilterTypeCategories,
  a: string,
  b: string,
  desc: boolean
): number {
  const result = sortColumnAsc(type, a, b);
  return desc ? result * -1 : result;
}function sortColumn(
  type: FilterTypeCategories,
  a: string,
  b: string,
  desc: boolean
): number {
  const result = sortColumnAsc(type, a, b);
  return desc ? result * -1 : result;
}
```

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
FabienHenon: `searchInput` was not use before I updated the code. `onSearch` function receives the search value.

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
FabienHenon: Acutally, pageSize is also used, but I get your point.
I think we should remove the dependency to `IcebergFetchParamsV6` and update `ResourcesV6Hook` by adding the missing params and renaming it to `ResourcesV6Params`

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
FabienHenon: Done

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
FabienHenon: I pushed a new version

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
aboungnaseng-ovhcloud: Ok I understand, I think, we have a conflicts during a rebase cause I added this commit in March

https://github.com/ovh/manager/commit/0ba78df728d46fa0a85f0b8122a7ed31bf70e59f#diff-13089f1b3964d084432b095c0390ddb3b92ea72146cd9ef260cc1010df963ac3

Can we check together ?

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
```diff
@@ -31,91 +30,88 @@ export function dataType(a) {
...
+}: ResourcesV6Params<T>) {
   const [searchInput, setSearchInput] = useState('');
-  const [searchFilter, setSearchFilter] = useState<any>(null);
+  const [searchFilter, setSearchFilter] = useState<Filter[]>([]);
   const { data, isError, isLoading, error, status } = useQuery({
     queryKey: [queryKey],
-    queryFn: () => fetchIcebergV6<T>({ route }),
+    queryFn: queryFn
+      ? () => queryFn(route)
+      : () => fetchIcebergV6<T>({ route }),
```
anooparveti, line 82: **nitpick:** The goal of this hook is to support v6 calls without Iceberg. So, imo, instead of `fetchIcebergV6` we should use `v6.get`.

anooparveti (changes requested):

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
FabienHenon: I fixed it

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
tristanwagner: Actually sorry I didn't understood that you wanted to support non iceberg calls, so I was suggesting to use the type directly provided by core api but it's not relevant

anooparveti (changes requested):
This is a breaking change - https://github.com/ovh/manager/pull/16722.
Can you please mention that in the commit?

aboungnaseng-ovhcloud (dismissed):

packages/manager-react-components/src/components/datagrid/documentation.mdx
```diff
@@ -38,11 +38,11 @@ Why Use TanStack Table?
 
 #### How to use It
 
-1 - In your columns definition, fill `type` or `comparator` attributes
+1 - In your columns definition, fill `type` and/or `comparator` attributes (ideally both)
```
anooparveti, line 41: **nitpick:** `Ideally both` is not requiredimo. Because comparators should be identified depending on the type and for some reason if some of the comparators cannot be applied (maybe like API does not support that comparator or logically it does not make sense to apply the filter depending on data), then comparators can be used to hide such unwanted comparatrors.
```suggestion
1 - In your columns definition, fill `type` and/or `comparator` attributes
```

packages/manager-react-components/src/components/datagrid/documentation.mdx
```diff
@@ -38,11 +38,11 @@ Why Use TanStack Table?
...
 
-- `type`comparator is a `FilterTypeCategories`
+- `type` is a `FilterTypeCategories` and will allow you to have enhanced filters (datepicker for date type, etc...)
 
-- `comparator`is a `string` of `FilterCategories`, if you fill this attribute, it override the `type` attributes
+- `comparator` is a `string` of `FilterCategories`
```
anooparveti, line 45:
```suggestion
- `comparator` is a `string` of `FilterCategories`. Use this option to define the set of comparators that can be applied on the data. By default, Datagrid determines the set of comparators that can be applied depending on the `type` field. **Only use this option if default comparators does not match your need.**
```

packages/manager-react-components/src/components/datagrid/documentation.mdx
```diff
@@ -266,3 +275,17 @@ Exemple of column definition
...
+
+`useResourcesV6` takes an object with the following definition as parameter:
+
+- `queryKey` (`string[]`, mandatory): The unique key identifying the query (used for caching purpose)
+- `queryFn` (`(route: string) => Promise<FetchResultV6<T>>`, optional): Define this function to override the query function used to retrieve the data. By default, `V6` api is used.
+- `refetchInterval` (`(query: Query<FetchResultV6<T>>) => number | false`, optional): If you need polling, you can define this function. It takes the query and must return the interval in ms before the next query, or `false` to stop polling.
+- `columns` (`DatagridColumn<T>[]`, mandatory): Columns used in the datagrid (used for searching, sorting and filtering).
+- `defaultSorting` (`ColumnSort`, optional): By default, no sorting is done. The object take the name (`id` property) of the column and a `desc` boolean property.
+- `route` (`string`, mandatory): The endpoint to get the data from.
+- `pageSize` (`number`, optional): The number of items to return per page (defaults to 10).
```
anooparveti, line 291: **nitpick:** Not the right place to document about the `useResourcesV6` hook. It can be documented separately under the `hooks` section.

anooparveti (changes requested):

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
```diff
@@ -31,91 +32,105 @@ export function dataType(a) {
...
-}: IcebergFetchParamsV6 & ResourcesV6Hook) {
+  defaultSorting,
+}: ResourcesV6Params<T>) {
   const [searchInput, setSearchInput] = useState('');
-  const [searchFilter, setSearchFilter] = useState<any>(null);
+  const [searchFilters, setSearchFilters] = useState<Filter[]>([]);
   const { data, isError, isLoading, error, status } = useQuery({
     queryKey: [queryKey],
-    queryFn: () => fetchIcebergV6<T>({ route }),
+    queryFn: queryFn ? () => queryFn(route) : () => v6.get(route),
```
anooparveti, line 98: **question:** When was it decided to add custom `queryFn` instead of route? I dont see the need.

anooparveti (changes requested):

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
FabienHenon: APIs can have different kind of structures in their get API.

We encountered the "classical" type : a data property with a list of items.
But we also encountered another one : a `resources` property with a list of items.

Thus, We thought it was a good idea (and a necessity) to be able to override the default getter

anooparveti: > But we also encountered another one : a resources property with a list of items.

`resources` is in V2. Do you have an example of it in v6?

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
```diff
@@ -1,26 +1,27 @@
...
 
-export interface ResourcesV6Hook {
+export interface ResourcesV6Params<T> {
   queryKey: string[];
-  columns: ColumnDatagrid[];
+  queryFn?: (route: string) => Promise<FetchResultV6<T>>;
+  refetchInterval?: (query: Query<FetchResultV6<T>>) => number | false;
+  columns: DatagridColumn<T>[];
+  defaultSorting?: ColumnSort;
+  route: string;
```
anooparveti, line 23: **nitpick:** If we introduce `queryFn` then `route` can be made an optional as well.

anooparveti (approved):

PR #16725, opened 2025-04-23
fix(pci-instances): remove auto-refetch for instances

fredericvilcot:
## Description

Ticket Reference: #...


fredericvilcot:
replaced by #16747 16747

PR #16780, opened 2025-04-25
fix(mrc): update useResourcesV6 issues search

aboungnaseng-ovhcloud:
ref: #MANAGER-16703

## Description

- useResourcesV6 update search feature

Ticket Reference: #MANAGER-16703


FabienHenon (dismissed):

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
```diff
@@ -60,13 +59,20 @@ function sortColumn(type: string, a: any, b: any, desc: boolean) {
...
+  return items?.filter(
+    (item) =>
+      filters?.some((columnId) => {
+        const value = item[columnId];
+        if (value === null || value === undefined) return false;
+        return value
+          ?.toString()
+          ?.toLowerCase()
+          ?.includes(searchInput?.toLowerCase());
+      }),
```
SimonChaumet, line 75:
```suggestion
  if (!searchInput?.length === 0 || !filters?.length)
    return items;
  
  const searchInputLowerCase = searchInput.toLowerCase();

  return items?.filter(
    (item) =>
      filters.some((columnId) => {
        const value = item[columnId];
        if (value === null || value === undefined) return false;
        return value
          .toString()
          .toLowerCase()
          .includes(searchInputLowerCase);
      }),
```
It avoids unused checks and avoids call to lower case for each value

packages/manager-react-components/src/hooks/datagrid/useResourcesV6.tsx
```diff
@@ -94,17 +100,28 @@ export function useResourcesV6<T = unknown>({
   const [flattenData, setFlattenData] = useState<T[]>([]);
   const { filters, addFilter, removeFilter } = useColumnFilters();
 
+  const searchableColumns = useMemo(() => {
+    return columns
+      .filter((col: ColumnDatagrid) => col.isSearchable)
+      .map((filter) => filter?.id);
```
SimonChaumet, line 106:
```suggestion
      .map((filter) => filter.id);
```
The object cannot be nullish here or it would have crash on line above

PR #16783, opened 2025-04-25
feat(ips): invalidate query key after a update

sachinrameshn:
ref: #MANAGER-17443

## Description

Ticket Reference: #...


chipp972 (approved):

packages/manager/apps/ips/src/pages/listing/manageOrganisations/OpenOrganisations/OpenOrganisationsModal.page.tsx
```diff
@@ -94,6 +95,14 @@ export const OpenOrganisationsModal: React.FC<{ isOpen: boolean }> = ({
         </OdsText>,
         true,
       );
+      queryClient.invalidateQueries({
+        queryKey: [`get/me/ipOrganisation`],
```
tristanwagner, line 99: I would recommend you to handle the query keys as arrays like that `['get', 'me', 'ipOrganisation', id]` with that you can invalidate in cascade so for exemple if you invalidate `['get', 'me', 'ipOrganisation']` it will also invalidate all `['get', 'me', 'ipOrgansation', id]`

PR #16804, opened 2025-04-28
fix(ips): resolve sidebar issue for legacy app

sachinrameshn:
ref: #MANAGER-15804

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/container/src/container/legacy/server-sidebar/index.tsx
+++ b/packages/manager/apps/container/src/container/legacy/server-sidebar/index.tsx
@@ -102,7 +102,7 @@ export default function ServerSidebarIndex() {
         }`}
       >
         {universe === 'public-cloud' && <PublicCloudSidebar />}
-        {universe === 'server' || universe === 'dedicated' && <DedicatedSidebar />}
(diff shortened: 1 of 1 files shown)


packages/manager/apps/container/src/container/legacy/server-sidebar/index.tsx
```diff
@@ -102,7 +102,8 @@ export default function ServerSidebarIndex() {
         }`}
       >
         {universe === 'public-cloud' && <PublicCloudSidebar />}
-        {universe === 'server' || universe === 'dedicated' && <DedicatedSidebar />}
+        {universe === 'server' && <DedicatedSidebar />}
+        {universe === 'dedicated' && <DedicatedSidebar />}
```
qpavy, line 106:
```suggestion
        {(universe === 'server' || universe === 'dedicated') && <DedicatedSidebar />}
```

Should work

sachinrameshn: yes it should but I kept it like that just for better readabaility. I will make the modification

qpavy (approved):

PR #16836, opened 2025-04-29
fix(dedicated): using iceberg utilities and fix iceberg call of stati…

sefiani-ovh:
fixing an issue in iceberg utilities! :(


packages/manager/apps/dedicated/client/app/components/dedicated-cloud/dedicatedCloud.service.js
```diff
@@ -339,14 +339,14 @@ class DedicatedCloudService {
   }
 
   getDatacenterInfoVm(serviceName, datacenterId, params = {}) {
-    return this.icebergQuery(
+    return this.icerbergUtils.icebergQuery(
```
tibs245, line 342: The error is not here but on the initial name `icerbergUtils`

Sorry It's me I think the origin of this problem

PR #16858, opened 2025-04-30
feat(zimbra): account POST requests in sequence

tristanwagner:
ref: #MANAGER-18180

## Description

Ticket Reference: #MANAGER-18180  

## Additional Information

Currently there's an issue on backend side that when you send multiple account POST request in parallel, it can happen that some accounts will not show in the list of accounts and it requires manual action on backend side to fix it.

Sending the requests sequentially seems to avoid this issue in most cases, so it will be needed until a solution on backend side is found

### DIFF
--- a/packages/manager/apps/zimbra/src/utils/promise.ts
+++ b/packages/manager/apps/zimbra/src/utils/promise.ts
@@ -1,0 +1,10 @@
+export const allSettledSequential = (funcs: Array<() => Promise<unknown>>) =>
+  funcs.reduce(
+    (promise, func) =>
+      promise.then((result) =>
+        func()
+          .then((value) => result.concat([{ status: 'fulfilled', value }]))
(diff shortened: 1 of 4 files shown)


packages/manager/apps/zimbra/src/utils/promise.ts
```diff
@@ -0,0 +1,10 @@
+export const allSequential = (funcs: Array<() => Promise<unknown>>) =>
+  funcs.reduce(
+    (promise, func) =>
+      promise.then((result) =>
+        // use finally so all promises are ran once
+        // even if one fails
+        func().finally(Array.prototype.concat.bind(result)),
+      ),
+    Promise.resolve([]),
+  );
```
tibs245, line 10: No unit test ?

tibs245 (approved):

tibs245 (approved):

tibs245 (approved):

ghyenne (approved):

PR #16868, opened 2025-04-30
fix(web-hosting): fix issue on csv export menu

ghyenne:
## Description

Ticket Reference: #MANAGER-18126

### DIFF
--- a/packages/manager/apps/web-hosting/src/index.scss
+++ b/packages/manager/apps/web-hosting/src/index.scss
@@ -13,3 +13,12 @@ html {
 ods-table td {
   line-height: unset !important;
 }
+
+.menu-item-button {
+  &::part(button) {
+    width: 100%;
+    justify-content: left;
+    height: 32px;
+    border-radius: 0;
+  }
+}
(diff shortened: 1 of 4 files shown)


tristanwagner (approved):

packages/manager/apps/web-hosting/src/index.scss
```diff
@@ -13,3 +13,12 @@ html {
 ods-table td {
   line-height: unset !important;
 }
+
+.menu-item-button {
+  &::part(button) {
+    width: 100%;
+    justify-content: left;
+    height: 32px;
```
tibs245, line 21: Maybe useful to use em size unit

packages/manager/apps/web-hosting/src/pages/websites/Websites.page.tsx
```diff
@@ -399,6 +401,10 @@ export default function Websites() {
             <OdsPopover
```
tibs245, line 401: I don't understand why it's so complex to use.

I think the implemention is bad ( May be since migration ?) : https://ovh.github.io/design-system/latest/?path=/docs/ods-components-popover--migration-from-17-x

May be fix the implementation can fix your problem ?

PR #16874, opened 2025-05-02
feat(*): update datagrid vitualization

aboungnaseng-ovhcloud:
ref: #MANAGER-X

## Description

- add virtualization dans datagrid

Ticket Reference: #...


packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
```diff
@@ -160,6 +164,7 @@ export const Datagrid = <T,>({
   search,
   topbar,
   totalItems,
+  isFetchAll,
```
tristanwagner, line 167: name this `isVirtualized` instead, it doesn't have to be tied to the fact that all the data is all loaded or not

Issue #16515, opened 2025-04-11
[manager-components]: Datagrid - Add possibility to filter on tags


qpavy:
### Have you already checked if a similar item is present on manager-components?

- [x] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

For Tagging resource project, we need to improve the datagrid filter to accept key + value instead of only value when the filter type is tag see example below.

Also, mockup shows that key and values should be combobox. But value combobox is populated by key combobox. How can we manage to do this ?

### Where do you expect to use this?

All product listing page in the futur and dedicated listing page at first

### Do you have mock-up?

yes https://www.figma.com/design/EDzxlSKDncRee21PEfGgmB/-IAM--Tagging-system?node-id=764-56920&p=f&t=usge2re05RPyyLUP-0

### When do you expect this to be delivered?

Before end of q3

### Additional Information

![Image](https://github.com/user-attachments/assets/eb42d6f6-e5ba-48ed-9776-1458029bb2e8)

To filter the resources with tags, We now have the possibility to use query string parameters on each resource get (for exemple 
https://eu.api.ovh.com/console/?section=%2Fdedicated%2Fserver&branch=v1#get-/dedicated/server) with at least 1 key and 0 ore mutliple values. For our need only 0 or 1 values should be enough.

To get the resources tags, there is 2 possibilities: 
1- Call the root Get for the resource with iceberg (GET /dedicated/servers) and we are able to see the tags on the iam object.
2- If the listing does not use iceberg (ip for example) we can get the tags doing a call on APIv2 GET /iam/resource filtered with resourceType (ip for exemple)

## Implementation Details:

1. Add a new Type named “Tags” in the column definition
2. Accept resourceType and primaryResourceUrn as a prop to the datagrid
3. In the filter-add component, if theselectedCoulmn type is Tag, display 2 OdsInput, one for key and the other for value.
4. Validation must be applied on key field as it is required field
5. Update Filter list component to display the key:value structure in the OdsTag if the column type is Tag.

### Handling Filtering for different API engines:

**Common Util: (In Core api package)**

1. Create an util that accepts list of objects, each object must contain column Name, key, value and operator. Util returns the queryString as an output.
	Input:
	[
	   {
              name: ‘iamTags’,
             operator: EQ and so on,
             value: {environment: production}
	   },
	  {
              name: ‘iamTags’,
             operator: EQ and so on,
             value: {environment: production2}
	   },
          {
[...]

PR #16911, opened 2025-05-05
feat(mrc): add iam tags filter on apiv6 iceberg call

qpavy:
## Description

Add new iam tags filter feature:
- add new column type "Tags" to the datagrid
- add new filter type "Tags" which fetch all the tags from the resource type and display 2 ODS combobox to filter by tag=key:value
- add util to format tags from API response iam tags object
- Improve FetchIcebergV6 to use tag filters as query params instead of iceberg header

Ticket Reference: #MANAGER-17692, [#16515](https://github.com/ovh/manager/issues/16515)

### DIFF
--- a/packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
+++ b/packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
@@ -149,6 +149,8 @@ export interface DatagridProps<T> {
   resetExpandedRowsOnItemsChange?: boolean;
   /** When true, will fix the columns size by column definition size */
   tableLayoutFixed?: boolean;
+  /** To use if tag column is present and filter is enabled. This allows to fetch all tags from iam only for this resource type */
+  resourceType?: string;
 }
 
 export const Datagrid = <T,>({
@@ -179,6 +181,7 @@ export const Datagrid = <T,>({
   resetExpandedRowsOnItemsChange,
   hideHeader,
   tableLayoutFixed,
+  resourceType,
 }: DatagridProps<T>) => {
   const { t } = useTranslation('datagrid');
   const pageCount = pagination
@@ -326,6 +329,7 @@ export const Datagrid = <T,>({
         filters={filters}
         search={search}
         topbar={topbar}
+        resourceType={resourceType}
       />
       <div className={`contents px-[1px] ${className || ''}`}>
         <OdsTable className="overflow-x-visible">
--- a/packages/manager-react-components/src/components/datagrid/documentation.mdx
+++ b/packages/manager-react-components/src/components/datagrid/documentation.mdx
@@ -105,6 +105,52 @@ language="javascript"
 
 ---
 
+### Filters with Tags
+
+#### How to use It
+
+1 - In your columns definition, use `type: Tags` attribute to use iam Tag filters for this column
+
+- `type` is a `FilterTypeCategories`
+
+<Source
+  code={`
+    export const columns = [
+        {
+            id: 'tags',
+            cell: (item: Item) => {
+                return <DataGridTextCell>{JSON.stringify(item.iam?.tags)}</DataGridTextCell>;
+            },
+            label: 'Tags',
+            type: FilterTypeCategories.Tags
+        }
+    ];
+  `}
+  language="javascript"
+/>
+
+2 - Get only tags for corresponding resource
+
+By default, the tag filter will fetch all the user existing tags but most of the time you would like to have only the tags for the resource displayed in the datagrid.
+To be able to get only the tags associate to the resource, you can use `resourceType` property on Datagrid definition.
+
+- `resourceType` can be found on APIv2 /iam/reference/resource/type.
+
+<Source
+  code={`
+<Datagrid
+    {...}
+    resourceType="dedicatedServer"
+/>
+    `}
+
+language="javascript"
+/>
+
+**Note:** Filters on Tags type of column is currently supported only on the primary resources.
+
+---
+
 ### Search
 
 #### How to use It
--- a/packages/manager-react-components/src/components/filters/filter-add.component.tsx
+++ b/packages/manager-react-components/src/components/filters/filter-add.component.tsx
@@ -15,6 +15,7 @@ import {
 } from '@ovhcloud/ods-components/react';
 import { useTranslation } from 'react-i18next';
 import './translations';
+import { TagsFilterForm } from './tags-filter-form.component';
 
 export type Option = {
   label: string;
@@ -31,10 +32,15 @@ export type ColumnFilter = {
 
 export type FilterAddProps = {
   columns: ColumnFilter[];
+  resourceType?: string;
   onAddFilter: (filter: Filter, column: ColumnFilter) => void;
 };
 
-export function FilterAdd({ columns, onAddFilter }: Readonly<FilterAddProps>) {
+export function FilterAdd({
+  columns,
+  onAddFilter,
+  resourceType,
+}: Readonly<FilterAddProps>) {
   const { t } = useTranslation('filters');
 
   const [selectedId, setSelectedId] = useState(columns?.[0]?.id || '');
@@ -43,6 +49,7 @@ export function FilterAdd({ columns, onAddFilter }: Readonly<FilterAddProps>) {
   );
   const [value, setValue] = useState('');
   const [dateValue, setDateValue] = useState<Date | null>(null);
+  const [tagKey, setTagKey] = useState('');
 
   const selectedColumn = useMemo(
     () => columns.find(({ id }) => selectedId === id),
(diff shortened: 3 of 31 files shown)


packages/manager-react-components/package.json
```diff
@@ -57,8 +57,8 @@
     "@ovh-ux/manager-react-shell-client": "^0.8.7",
     "@ovh-ux/manager-tailwind-config": "^0.3.0",
     "@ovh-ux/manager-vite-config": "^0.9.3",
-    "@ovhcloud/ods-components": "^18.5.1",
-    "@ovhcloud/ods-themes": "^18.5.1",
+    "@ovhcloud/ods-components": "^18.6.2",
+    "@ovhcloud/ods-themes": "^18.6.2",
```
anooparveti, line 61: **nitpick:** Mostly out of the context of this PR. I say this because, we dont know when this PR will be merged and by the time this PR is merged, the version may be outdated or someother PR might already update it. So, imo, better to isolate this bump.

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
anooparveti, line 1: > Accept resourceType and primaryResourceUrn as a prop to the datagrid

We also need to add `primaryResourceUrn` as a prop to support the feature for secondary datagrids.

qpavy: We will have to discuss about primaryResource Urn because after few talks with enabler team it seems that we should find another way to manage "sub resources" as 1 day we can have more than 1 level of sub resource

packages/manager-react-components/src/components/filters/filter-add.component.tsx
```diff
@@ -197,16 +228,73 @@ export function FilterAdd({ columns, onAddFilter }: Readonly<FilterAddProps>) {
...
-            </span>
-          </div>
-          {inputComponent}
-        </OdsFormField>
-      </div>
+      {selectedColumn?.type !== FilterTypeCategories.Tags && (
+        <div>
+          <OdsFormField className="mt-2 w-full">
+            <div slot="label">
+              <span className="text-[--ods-color-heading] leading-[22px]">
```
anooparveti, line 235: **question:** Why 22px? Is it mentioned in the specs?

packages/manager-react-components/src/components/filters/filter-add.component.tsx
```diff
@@ -197,16 +228,73 @@ export function FilterAdd({ columns, onAddFilter }: Readonly<FilterAddProps>) {
...
+                  ?.find((tag) => tag.key === tagKey)
+                  ?.values?.map((tagValue) => (
+                    <OdsComboboxItem key={tagValue} value={tagValue}>
+                      {tagValue}
+                    </OdsComboboxItem>
+                  ))}
+              </OdsCombobox>
+            )}
+          </OdsFormField>
+        </div>
```
anooparveti, line 296: **nitpick:** I think this can be split into a different component and just display the component in the main component.

packages/manager-react-components/src/components/filters/filter-list.component.tsx
```diff
@@ -5,6 +5,7 @@ import { useTranslation } from 'react-i18next';
 import { FilterTypeCategories } from '@ovh-ux/manager-core-api';
 import { FilterWithLabel } from './interface';
 import './translations';
+import { formatFilter } from '../../utils/format-filter';
```
anooparveti, line 8: **nitpick:** Use absolute path.
```suggestion
import { formatFilter } from '@/utils/format-filter.ts';
```

packages/manager-react-components/src/hooks/iam/useOvhIam.tsx
```diff
@@ -1,6 +1,7 @@
-import { apiClient } from '@ovh-ux/manager-core-api';
+import { apiClient, fetchIcebergV2 } from '@ovh-ux/manager-core-api';
 import { useQuery, keepPreviousData } from '@tanstack/react-query';
-import { IamCheckResponse, IamInterface } from './iam.interface';
+import { IamCheckResponse, IamInterface, IamObject } from './iam.interface';
+import { formatIamTagsFromResources } from '../../utils/format-tags';
```
anooparveti, line 4:
```suggestion
import { formatIamTagsFromResources } from '@/utils/format-tags';
```

packages/manager-react-components/src/utils/format-tags.ts
```diff
@@ -0,0 +1,34 @@
...
+          ];
+        } else {
+          allTags.push({
+            key,
+            values: [resource.tags[key]],
+          });
+        }
+      });
+    });
+  return allTags;
```
anooparveti, line 33: **nitpick:** Since the end result is the formation of new array with grouped tags, usage of `Array.prototype.reduce` can be better suited and avoids two loops (one for filtering the resources without tag and one for forming the tags array).

packages/manager/core/api/src/filters.ts
```diff
@@ -110,3 +122,34 @@ export function applyFilters<T>(items: T[] = [], filters: Filter[] = []) {
...
+
+export function transformTagsFiltersToQuery(filters: Filter[] = []): string {
+  const queryObject: Record<
+    string,
+    Array<{ operator: string; value?: string }>
+  > = {};
+
+  const filteredFilters = filters.filter(
+    ({ type }) => type === FilterTypeCategories.Tags,
+  );
```
anooparveti, line 134:
```suggestion
  const tagFilters = filters.filter(
    ({ type }) => type === FilterTypeCategories.Tags,
  );
```

anooparveti (changes requested):

packages/manager/core/api/src/iceberg.ts
```diff
@@ -99,22 +104,38 @@ export async function fetchIcebergV6<T>({
...
+  if (requestHeaders['x-pagination-filter'] === '') {
+    delete requestHeaders['x-pagination-filter'];
   }
+
   if (disableCache) {
     requestHeaders.Pragma = 'no-cache';
   }
-  const { data, headers, status } = await apiClient.v6.get(route, {
+
+  const routeWithParmas = params.size ? `${route}?${params.toString()}` : route;
```
anooparveti, line 136: **nitpick:** We also need to handle the case if route already contains some queryParams.
For example, route: `/a/b?c=d`

anooparveti (changes requested):
As discussed, add more unit tests and a dedicated documentation page to explain how Datagrid handles the Tags.

packages/manager/core/api/src/iceberg.ts
qpavy: ok I didn't thought about that use case, I check

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
qpavy: After discussion with API team, this feature is not ready yet, I remove it from the props then

anooparveti (changes requested):
Screenshot 2025-05-12 at 1 16 16 PM

I am not able to select the key to apply filters.

aboungnaseng-ovhcloud (approved):

aboungnaseng-ovhcloud (dismissed):

packages/manager-react-components/src/components/datagrid/documentation.mdx
```diff
@@ -105,6 +105,50 @@ language="javascript"
...
+    {...}
+    resourceType="dedicatedServer"
+/>
+    `}
+
+language="javascript"
+/>
+
+---
+
```
anooparveti, line 151:
```suggestion
2 - Get only tags for corresponding resource

By default, the tag filter will fetch all the user existing tags but most of the time you would like to have only the tags for the resource displayed in the datagrid.
To be able to get only the tags associate to the resource, you can use `resourceType` property on Datagrid definition.

- `resourceType` can be found on APIv2 /iam/reference/resource/type.

<Source
  code={`
<Datagrid
    {...}
    resourceType="dedicatedServer"
/>
    `}

language="javascript"
/>

---

**Note:** Filters on Tags type of column is currently supported only on the primary resources.
```

aboungnaseng-ovhcloud (dismissed):

packages/manager-react-components/src/utils/format-tags.ts
```diff
@@ -0,0 +1,35 @@
...
+        ];
+      } else {
+        updatedTags.push({
+          key,
+          values: [resource.tags[key]],
+        });
+      }
+    });
+    return updatedTags;
+  }, [] as Tag[]);
```
anooparveti, line 34: **perf:**  time-complexity can be improved by having O(n) instead of O(n^2). (Lets discuss over call)

aboungnaseng-ovhcloud (dismissed):

anooparveti (dismissed):

anooparveti (approved):

packages/manager-react-components/src/utils/format-tags.ts
```diff
@@ -0,0 +1,29 @@
...
+  resources.forEach((resource) => {
+    if (!resource.tags) return;
+
+    Object.entries(resource.tags).forEach(([key, value]) => {
+      if (key.startsWith('ovh:')) return;
+
+      if (!tagsMap.has(key)) {
+        tagsMap.set(key, new Set([value]));
+      } else {
+        tagsMap.get(key)?.add(value);
```
SimonChaumet, line 20:
```suggestion
        tagsMap.get(key)!.add(value);
```

SimonChaumet (approved):
snapshots are good

packages/manager/core/api/src/iceberg.ts
```diff
@@ -99,22 +104,44 @@ export async function fetchIcebergV6<T>({
...
   }
-  const { data, headers, status } = await apiClient.v6.get(route, {
+
+  let routeWithParams = route;
+  if (params.size) {
+    routeWithParams =
+      route.indexOf('?') > -1
+        ? `${route}&${params.toString()}`
+        : `${route}?${params.toString()}`;
+  }
```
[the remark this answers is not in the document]
qpavy: we cannot do this 
```javascript
let routeWithParams = new URL(route);
```

Because route is not a valid url it is route and does not include protocol nether domain

SimonChaumet (approved):

anooparveti (approved):

PR #16914, opened 2025-05-05
fix(instance): use flavorGroup prices

lionel95200x:
ref: #INC0128395

## Description

- Set the correct price for pci-instances

Ticket Reference: #INC0128395


Tsiorifamonjena (dismissed):

ppprevost (approved):

Eric-ciccotti (dismissed):

fredericvilcot (changes requested):
We should have two commits with two references towards OneGate to describe both incidents.

PR #16938, opened 2025-05-06
fix(pci-volume-backup): filter available action on backup status

MarcMonchablon:
ref: #DTCORE-3314

## Description

Ticket Reference: #DTCORE-3314

### DIFF
--- a/packages/manager/apps/pci-volume-backup/src/pages/listing/Actions.component.tsx
+++ b/packages/manager/apps/pci-volume-backup/src/pages/listing/Actions.component.tsx
@@ -1,7 +1,6 @@
 import { useTranslation } from 'react-i18next';
 import { useHref } from 'react-router-dom';
 import { ActionMenu } from '@ovh-ux/manager-react-components';
-import { useMemo } from 'react';
 import { TVolumeBackup } from '@/data/api/api.types';
 import { VOLUME_BACKUP_STATUS } from '@/constants';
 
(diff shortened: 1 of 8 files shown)


packages/manager/apps/pci-volume-backup/src/pages/listing/Actions.component.tsx
```diff
@@ -12,38 +11,34 @@ type ActionsProps = {
...
+  const restoreBackupItem = {
+    id: 1,
+    href: useHref(`./restore-volume/${backup.volumeId}`),
+    label: t(
+      'pci_projects_project_storages_volume_backup_list_datagrid_menu_action_restore',
+    ),
+  };
+  const createVolumeItem = {
+    id: 2,
+    href: useHref(`./${backup.id}/create-volume`),
```
ppprevost, line 23: why is this the only inversed uri ? 
./restore-volume/${backup.volumeId} opposite to ./${backup.id}/create-volume`

packages/manager/apps/pci-volume-backup/src/pages/listing/Actions.component.tsx
```diff
@@ -12,38 +11,34 @@ type ActionsProps = {
...
-    },
-    {
-      id: 3,
-      href: useHref(`./delete/${backup.id}`),
-      label: t(
-        'pci_projects_project_storages_volume_backup_list_datagrid_menu_action_delete',
-      ),
-    },
-  ];
+  let items: { id: number; href: string; label: string }[] = [];
```
FabienHenon, line 36: There are other statuses for this item (restoring, deleting, creating, ok, error:

```
export type TBackupStatus =
  | 'creating'
  | 'deleting'
  | 'error'
  | 'ok'
  | 'restoring'
  | string;
```

Maybe you should provide a fallback for the other cases

FabienHenon (changes requested):

packages/manager/apps/pci-volume-backup/src/pages/listing/Actions.component.tsx
MarcMonchablon: We have to be ISO with prod, and currently the other url aren't, so we're working on a fix for that, however the new urls will either have the required ids as queryparams, or in the case of create-volume be left as-is, since it's what's in prod right now.

FabienHenon (approved):

ppprevost (approved):

PR #16955, opened 2025-05-06
feat(vrack-services): update termination modal

chipp972:
ref: #MANAGER-18197

## Description

Ticket Reference: #...


mhelhali-soufien (approved):

qpavy (approved):

sachinrameshn (approved):

packages/manager/apps/vrack-services/src/test-utils/test-i18n.ts
```diff
@@ -12,6 +12,7 @@ import endpoints from '../../public/translations/vrack-services/endpoints/Messag
 import subnets from '../../public/translations/vrack-services/subnets/Messages_fr_FR.json';
 import actions from '../../../../modules/common-translations/public/translations/actions/Messages_fr_FR.json';
 import regionSelector from '../../public/translations/region-selector/Messages_fr_FR.json';
+import deleteModal from '../../../../../manager-react-components/src/components/templates/delete-modal/translations/Messages_fr_FR.json';
```
anooparveti, line 15: **nitpick:** It is not correct to load the files like this especially when the package is versioned. If your app is not using the latest version of the lib, then there is always a chance that this breaks.

anooparveti (changes requested):

anooparveti (approved):

PR #16964, opened 2025-05-07
fix(vrack-services): fix and skipe some tests

mhelhali-soufien:
ref: #MANAGER-16777

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/vrack-services/src/test-utils/setupTests.tsx
+++ b/packages/manager/apps/vrack-services/src/test-utils/setupTests.tsx
@@ -1,32 +1,22 @@
 import { vi } from 'vitest';
 import React from 'react';
 import '@testing-library/jest-dom';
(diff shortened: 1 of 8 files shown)


packages/manager/apps/vrack-services/src/test-utils/setupTests.tsx
anooparveti, line 1: **question:** What is the need to mock the components? I dont think it's good to do that.

anooparveti, line 1: **question:** What is the need to mock the components? I dont think it's good to do that.

anooparveti (changes requested):

chipp972 (changes requested):
This PR will fix vrack-services build and tests https://github.com/ovh/manager/pull/16955

PR #16965, opened 2025-05-07
Feat/manager 17913 ftc veeam enterprise

tibs245:
## Description

Ticket Reference: #MANAGER-17913


packages/manager/modules/veeam-enterprise/src/details/veeam-enterprise.module.js
```diff
@@ -1,5 +1,6 @@
 import angular from 'angular';
 
+import ovhManagerBillingComponents from '@ovh-ux/manager-billing-components';
```
JacquesLarique, line 3: I'm not sure this is the correct module to be able to use the `billingAutorenewTerminateAgoraService`, I think you should use the module '@ovh-ux/manager-new-billing' (even if it would make more sense if the component was stored in the '@ovh-ux/manager-billing-components' module)

tibs245: Good catch

JacquesLarique (approved):

packages/manager/modules/veeam-enterprise/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "peerDependencies": {
     "@ovh-ux/manager-core": "^12.0.0 || ^13.0.0",
+    "@ovh-ux/manager-new-billing": "*",
```
darsene, line 22: dedicated app uses `@ovh-ux/manager-billing` so I recommend to use the same to avoid having conflicts

packages/manager/modules/veeam-enterprise/package.json
```diff
@@ -19,6 +19,7 @@
   },
   "peerDependencies": {
     "@ovh-ux/manager-core": "^12.0.0 || ^13.0.0",
+    "@ovh-ux/manager-new-billing": "^0.37.0",
```
darsene, line 22:
```suggestion
```

sefiani-ovh (approved):

PR #17007, opened 2025-05-09
fix(pci-object-storage): apply the deployment mode tags with correct…

selm3n:
… contetnt and style

ref: #TAPC-3559, #TAPC-3124

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-object-storage/src/pages/objects/container/new/step/DeploymentModeStepTile.component.tsx
+++ b/packages/manager/apps/pci-object-storage/src/pages/objects/container/new/step/DeploymentModeStepTile.component.tsx
@@ -56,7 +56,7 @@ export function DeploymentModeStepTile({ item: mode, isItemSelected }) {
         preset="paragraph"
         className={clsx('leading-8', isItemSelected && 'font-bold')}
       >
-        <span className="mr-4">
+        <span className={clsx(isItemSelected && 'font-bold', 'mr-4')}>
           {t(
             isLocalZoneAvailable && is3azAvailable
               ? `pci_projects_project_storages_containers_add_deployment_mode_${mode}`
@@ -66,9 +66,7 @@ export function DeploymentModeStepTile({ item: mode, isItemSelected }) {
         {OBJECT_CONTAINER_DEPLOYMENT_MODES_LABELS[mode]?.isMultiZone && (
           <OdsBadge
             size="sm"
-            className={`${
-              isItemSelected ? 'font-bold' : ''
-            } bg-[--ods-color-blue-200]`}
+            className="chip-3AZ"
             label={t(
               `pci_projects_project_storages_containers_add_deployment_mode_${mode}_label`,
             )}
@@ -79,9 +77,7 @@ export function DeploymentModeStepTile({ item: mode, isItemSelected }) {
           is3azAvailable && (
             <OdsBadge
               size="sm"
-              className={`${
-                isItemSelected ? 'font-bold' : ''
-              } bg-[--ods-color-blue-200]`}
+              className="chip-1AZ"
               label={t(
                 `pci_projects_project_storages_containers_add_deployment_mode_${mode}_label`,
               )}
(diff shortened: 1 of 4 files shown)


Tsiorifamonjena (dismissed):

packages/manager/apps/pci-object-storage/src/pages/objects/container/useDatagridColumn.tsx
```diff
@@ -56,25 +67,43 @@ export const useDatagridColumn = () => {
...
+                )}
               />
             </DataGridTextCell>
           );
         }
         if (props.deploymentMode === OBJECT_CONTAINER_MODE_MONO_ZONE) {
           return (
             <DataGridTextCell>
-              <OdsBadge color="information" size="sm" label={props.mode} />
+              <OdsBadge
```
ppprevost, line 82: why not using ChipBadge from pci-common ?

packages/manager/apps/pci-object-storage/src/pages/objects/container/new/step/DeploymentModeStepTile.component.tsx
```diff
@@ -56,7 +56,7 @@ export function DeploymentModeStepTile({ item: mode, isItemSelected }) {
         preset="paragraph"
         className={clsx('leading-8', isItemSelected && 'font-bold')}
       >
-        <span className="mr-4">
+        <span className={`${isItemSelected ? 'font-bold' : ''} mr-4`}>
```
SimonChaumet, line 59: Can you use clsx instead ?

SimonChaumet (commented):
Why not use components from pci-common ?

packages/manager/apps/pci-object-storage/src/pages/objects/container/new/step/DeploymentModeStepTile.component.tsx
selm3n: Done

packages/manager/apps/pci-object-storage/src/pages/objects/container/useDatagridColumn.tsx
selm3n: I'm using OdsBadge  Because  it's the dedecated component of ods for tags and it gives me exactly what i need and it's used all over the manager

SimonChaumet: Ods badge doesn't give you the right style and you need to duplicate some styling so there is an issue here

selm3n: I do use the badges in two differents places of object storage so i have to add the style for every file. In ods 18 we can customise the ods badges with custom style

SimonChaumet: You can use the same style in both files

selm3n: Chip - migrate from v17 to v18
Chip has been removed from ODS components. 
Removed

It has been split into two different components.

If you need a presentational component, with no behaviour, use 
[ods-badge](https://ovh.github.io/design-system/v18.4.1/iframe.html?viewMode=docs&id=ods-components-chip--docs#)
.

If you need a clickable element, use 
[ods-tag](https://ovh.github.io/design-system/v18.4.1/iframe.html?viewMode=docs&id=ods-components-chip--docs#)
.

https://ovh.github.io/design-system/v18.4.1/?path=/docs/ods-components-chip--docs

selm3n: We can use the same style for both files but if we add some custom style need for the first one it may affect the second page so it is better to keep them in two different files

SimonChaumet: Files are only for organization purpose, both files are included and will apply

fredericvilcot (dismissed):

packages/manager/apps/pci-object-storage/src/pages/objects/container/useDatagridColumn.tsx
selm3n: Yes but the css included in the container creation file will not be applyed for the objects page

Eric-ciccotti (approved):

fredericvilcot (approved):

Eric-ciccotti (approved):

PR #17009, opened 2025-05-09
feat(*): bump prettier version

aottr:
## Description
bumping version of prettier for support of TypeScript 5.2

Ticket Reference: #MANAGER-17808


anooparveti (changes requested):
> Lint Code Base / Lint Code Base (push) Failing after 9m

There is a lint failure on the PR. And moreover, bumping prettier would require to address all the breaking changes on the entire repo. We at CT, have a different ticket to enhance the linting on the repo. So, imo, you can close this PR.

PR #17050, opened 2025-05-12
feat(common-api): add services delete hook

sefiani-ovh:
ref: #MANAGER-18300
## Description

Update the hook deleteService to be able to call diffrents routes depending of the ovhSubsidiary

**common-api** needs now  the dependency **ovh-ux/manager-react-shell-client**

### DIFF
--- a/packages/manager/modules/common-api/package.json
+++ b/packages/manager/modules/common-api/package.json
@@ -26,6 +26,7 @@
   "devDependencies": {
     "@ovh-ux/manager-core-api": "^0.11.1",
     "@ovh-ux/manager-core-test-utils": "^0.4.4",
+    "@ovh-ux/manager-react-shell-client": "^0.8.7",
     "@testing-library/jest-dom": "^6.6.3",
     "@testing-library/react": "^16.0.1",
     "@vitejs/plugin-react": "^4.3.3",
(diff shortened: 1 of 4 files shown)


packages/manager/modules/common-api/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@ovh-ux/manager-module-common-api",
-  "version": "0.3.8",
+  "version": "0.3.9",
```
anooparveti, line 3: Version will be automatically bumped by CI, you should not modify it.

anooparveti (changes requested):

packages/manager/modules/common-api/package.json
sefiani-ovh: updated

aboungnaseng-ovhcloud (dismissed):

vovh (approved):

packages/manager/modules/common-api/src/services/hooks/useDeleteService.ts
```diff
@@ -27,6 +30,8 @@ export const useDeleteService = ({
   ...options
 }: UseDeleteServiceParams) => {
   const queryClient = useQueryClient();
+  const { ovhSubsidiary } = useContext(ShellContext).environment.getUser();
```
anooparveti, line 33:
```suggestion
  const { ovhSubsidiary } = useContext(ShellContext)?.environment?.getUser();
```

anooparveti (approved):

tibs245 (approved):

PR #16633, opened 2025-04-17
fix(mrc): add select all on visibility columns component

ghyenne:
## Description

Ticket Reference: #MANAGER-17733  #16175

## Additional Information

Screenshot 2025-04-17 at 18-07-45 Components _ Datagrid Cursor - Visibility ⋅ Storybook
Screenshot 2025-04-17 at 18-07-35 Components _ Datagrid Cursor - Visibility ⋅ Storybook

PR #17060, opened 2025-05-13
fix(mrc): add select all on visibility columns component (#16633)

ghyenne:
ref: #MANAGER-17733

## Description

Ticket Reference: #...


anooparveti (changes requested):
**question:** Why do you need to add this manually on your branch instead of rebasing your branch with develop? Adding this commit manually on your branch will duplicate this on your branch.

PR #17101, opened 2025-05-14
refactor(pci-load-balancer): remove octavia-load-balancer part 1

FabienHenon:
ref: #DTCORE-3341

## Description

Ticket Reference: #DTCORE-3341

### DIFF
--- a/.github/labeler.yml
+++ b/.github/labeler.yml
@@ -19,13 +19,7 @@ translation required:
 universe-public-cloud:
   - packages/manager/*/pci/**
   - packages/manager/apps/public-cloud/**
-  - packages/manager/apps/octavia-load-balancer/**
-  - packages/manager/apps/pci-rancher/**
-  - packages/manager/apps/pci-ssh-keys/**
-  - packages/manager/apps/pci-users/**
-  - packages/manager/apps/pci-vouchers/**
-  - packages/manager/apps/pci-gateway/**
-  - packages/manager/apps/pci-private-network/**
+  - packages/manager/apps/pci-*/**
 universe-bare-metal-cloud:
   - packages/manager/apps/dedicated/**/!(dedicatedCloud)/**
   - packages/manager/*/billing/**
--- a/packages/manager/apps/container/src/core/routing/iframe-app-router.tsx
+++ b/packages/manager/apps/container/src/core/routing/iframe-app-router.tsx
@@ -119,8 +119,8 @@ export function IFrameAppRouter({
         .map((config) => makeRoute({ ...config, iframeRef })),
     [mockedConfiguration],
   );
-  const redirections = useMemo(() => Redirections(configuration), [
-    configuration,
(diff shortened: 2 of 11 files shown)


packages/manager/apps/container/src/core/routing/iframe-app-router.tsx
```diff
@@ -119,8 +119,8 @@ export function IFrameAppRouter({
         .map((config) => makeRoute({ ...config, iframeRef })),
     [mockedConfiguration],
   );
-  const redirections = useMemo(() => Redirections(configuration), [
-    configuration,
+  const redirections = useMemo(() => Redirections(mockedConfiguration), [
```
SimonChaumet, line 122: If all the FF are migrated shouldn't the mockedConfiguration be removed ?

SimonChaumet (dismissed):

packages/manager/apps/container/src/core/routing/iframe-app-router.tsx
FabienHenon: Actually not yet.
I tried without it and it wasn't working.
As far as I understood it, `pci-load-balancer` and `octavia-load-balancer` match the same url.
Thus, while both are present in the configuration, `octavia-load-balancer` will be chosen (because first in the list I guess), which makes it not working properly.
That's why this PR is the first step.
I first have to change the feature name from `octavia-load-balancer` to `pci-load-balancer`.
Then, we can remove `octavia-load-balancer` from feature-availability and bff.
And finally, I can remove the `mockedConfiguration`

SimonChaumet (dismissed):

Tsiorifamonjena (dismissed):

anooparveti (approved):

anooparveti (approved):

JacquesLarique (approved):

Tsiorifamonjena (approved):

.github/labeler.yml
```diff
@@ -19,13 +19,7 @@ translation required:
...
   - packages/manager/*/pci/**
   - packages/manager/apps/public-cloud/**
-  - packages/manager/apps/octavia-load-balancer/**
-  - packages/manager/apps/pci-rancher/**
-  - packages/manager/apps/pci-ssh-keys/**
-  - packages/manager/apps/pci-users/**
-  - packages/manager/apps/pci-vouchers/**
-  - packages/manager/apps/pci-gateway/**
-  - packages/manager/apps/pci-private-network/**
+  - packages/manager/apps/pci-*/**
```
SimonChaumet, line 22: Not sure about this change since there is pci-ai and database

fredericvilcot: @SimonChaumet Yes it's OK as it gathers all pci applications under the folder, it has been validated with @anooparveti 👍🏻

SimonChaumet (approved):

PR #17105, opened 2025-05-14
fix(translation): use 'type' placeholder instead of Plesk...

AzodFR:
… when no upgrade is available on licenses

## Description

When no upgrades are availables on any type of licenses, the alert message says "Plesk"


AzodFR:
Hi, the contributing docs says: "Update only Messages_fr_FR.json locales for any content changes."
But in this case, every lang.json need to be changed, tell me if I need to remove others files

PR #17120, opened 2025-05-15
fix(pci-volume-backup): hide action menu when no items inside exists

seven-amid:
ref: #DTCORE-3338

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-volume-backup/src/pages/listing/Actions.component.tsx
+++ b/packages/manager/apps/pci-volume-backup/src/pages/listing/Actions.component.tsx
@@ -10,35 +10,43 @@ type ActionsProps = {
 
 export default function Actions({ backup }: Readonly<ActionsProps>) {
   const { t } = useTranslation('listing');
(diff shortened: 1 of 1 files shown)


FabienHenon (approved):

StephaneRavet (commented):
You can improve clarity (and even a bit of performance) by moving the guard `if` up front to return `null` when the status is neither `OK` nor `ERROR`. That way you don’t declare the constants or the `items` array if you’re not going to render the menu anyway.

MarcMonchablon (approved):

FabienHenon:
> You can improve clarity (and even a bit of performance) by moving the guard `if` up front to return `null` when the status is neither `OK` nor `ERROR`. That way you don’t declare the constants or the `items` array if you’re not going to render the menu anyway.

Be careful though because you are using `useHref` hook and putting this hook after the condition is not allowed be react because it makes the usage of the hook conditional

SimonChaumet (approved):

Issue #17155, opened 2025-05-16
[manager-components]: <Changelog> & <PciGuidesHeader>: add 'onClick' props for alternative tracking


MarcMonchablon:
### Have you already checked if a similar item is present on manager-components?

- [x] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Others

### Description

While migrating the "Volume Backup" micro-app, we would like to implement the new tracking nomenclature, which also covers the tracking sent by the Changelog button & the pci-guide-header button.

As such, we need to be able to overwrite the current tracking behavior to customize it in a specific micro-application, by specifying an optional 'onClick' props, which would bypass the current behavior.

### Where do you expect to use this?

pci-volume-backup

MarcMonchablon:
Here's the related PR => https://github.com/ovh/manager/pull/17157 😃

PR #17157, opened 2025-05-16
feat(mrc): add onClick props to Changelog & PciGuideHeader

MarcMonchablon:
ref: #DTCORE-3308

## Description
Related to [this issue](https://github.com/ovh/manager/issues/17155).

Ticket Reference: #DTCORE-3308

### DIFF
--- a/packages/manager-react-components/src/components/guides-header/pci/pci-guides-header.component.tsx
+++ b/packages/manager-react-components/src/components/guides-header/pci/pci-guides-header.component.tsx
@@ -11,9 +11,10 @@ import './translations';
 
 interface PciGuidesHeaderProps {
   category: string;
+  onClick?: (key: string) => void;
 }
 
-export function PciGuidesHeader({ category }: PciGuidesHeaderProps) {
+export function PciGuidesHeader({ category, onClick }: PciGuidesHeaderProps) {
   const { ovhSubsidiary } = useEnvironment().getUser();
   const { trackClick } = useTracking();
   const [t] = useTranslation('pci-guides-header');
@@ -26,10 +27,14 @@ export function PciGuidesHeader({ category }: PciGuidesHeaderProps) {
         t(`pci_project_guides_header_${guide.key}`)
       }
       onGuideClick={(guide: Guide) => {
-        trackClick({
-          name: `public-cloud_credit_and_vouchers${guide.tracking}`,
-          type: 'action',
-        });
+        if (onClick) {
+          onClick(guide.key);
+        } else {
+          trackClick({
+            name: `public-cloud_credit_and_vouchers${guide.tracking}`,
+            type: 'action',
+          });
+        }
       }}
     />
   );
--- a/packages/manager-react-components/src/components/navigation/menus/changelog/changelog.component.tsx
+++ b/packages/manager-react-components/src/components/navigation/menus/changelog/changelog.component.tsx
@@ -19,6 +19,7 @@ export interface ChangelogLinks {
 export interface ChangelogButtonProps {
   links: ChangelogLinks;
   chapters?: string[];
+  onClick?: (key: string) => void;
 }
 
 export const CHANGELOG_PREFIXES = ['tile-changelog-roadmap', 'external-link'];
@@ -27,6 +28,7 @@ const GO_TO = (link: string) => `go-to-${link}`;
 export const ChangelogButton: React.FC<ChangelogButtonProps> = ({
   links,
   chapters = [],
+  onClick,
 }) => {
   const { t } = useTranslation('buttons');
   const { trackClick } = useOvhTracking();
@@ -55,12 +57,16 @@ export const ChangelogButton: React.FC<ChangelogButtonProps> = ({
               type={LinkType.external}
               rel={LinkType.external}
               label={t(`mrc_changelog_${key}`)}
-              onClickReturn={() =>
-                trackClick({
-                  actionType: 'navigation',
-                  actions: [...chapters, ...CHANGELOG_PREFIXES, GO_TO(key)],
-                })
-              }
+              onClickReturn={() => {
+                if (onClick) {
+                  onClick(key);
+                } else {
+                  trackClick({
+                    actionType: 'navigation',
+                    actions: [...chapters, ...CHANGELOG_PREFIXES, GO_TO(key)],
+                  });
+                }
+              }}
             />
           </div>
         ))}


packages/manager-react-components/src/components/navigation/menus/changelog/changelog.component.tsx
```diff
@@ -55,12 +57,16 @@ export const ChangelogButton: React.FC<ChangelogButtonProps> = ({
...
+              onClickReturn={() => {
+                if (onClick) {
+                  onClick(key);
+                } else {
+                  trackClick({
+                    actionType: 'navigation',
+                    actions: [...chapters, ...CHANGELOG_PREFIXES, GO_TO(key)],
+                  });
+                }
+              }}
```
tibs245, line 69: May be good to dissociate onClick specify props and "isTrackingDisabled" props ?

And maybe create "handleClick" function

MarcMonchablon: For the 'handleClick' function, those two components are quite small & aren't expected to grow anymore, so I'm not sure it's worth the (however small) indirection.

packages/manager-react-components/src/components/guides-header/pci/pci-guides-header.component.tsx
```diff
@@ -26,10 +27,14 @@ export function PciGuidesHeader({ category }: PciGuidesHeaderProps) {
...
-          type: 'action',
-        });
+        if (onClick) {
+          onClick(guide.key);
+        } else {
+          trackClick({
+            name: `public-cloud_credit_and_vouchers${guide.tracking}`,
+            type: 'action',
+          });
+        }
```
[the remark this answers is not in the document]
MarcMonchablon: Mmhhh, it's true that the alternative 'onClick' is only to handle tracking issue, and in fact won't do any type of preventDefault() or stopPropagation(). Maybe instead of a 'onClick' props we pass a 'handleTracking' props, expecting the same function, but making it more explicit the fact that it cancels the default tracking ?

packages/manager-react-components/src/components/navigation/menus/changelog/changelog.component.tsx
```diff
@@ -21,7 +21,7 @@ export interface ChangelogButtonProps {
   chapters?: string[];
 }
 
-export const CHANGELOG_PREFIXES = ['tile-changelog-roadmap', 'external-link'];
+export const CHANGELOG_PREFIXES = ['page', 'title-tutorial'];
```
anooparveti: **question:** If i understand your case correctly, you only wanted to update the tracking only in some apps which is on old nomenclature. Updating it here will update both Old and new tracking nomenclature that cause regression. Can you confirm the required behavior please?

MarcMonchablon: Mmhhh, you're right. It believe that the `tile-changelog-roadmap` & `external-link` is outdated, but I'm not sure and I'm not getting an answer right now, so to avoid any potential regression, I'll instead switch to an optional prop to override those default prefixes, if that's all-right with you.

packages/manager-react-components/src/components/guides-header/pci/pci-guides-header.component.tsx
```diff
@@ -26,10 +30,14 @@ export function PciGuidesHeader({ category }: PciGuidesHeaderProps) {
         t(`pci_project_guides_header_${guide.key}`)
       }
       onGuideClick={(guide: Guide) => {
-        trackClick({
-          name: `public-cloud_credit_and_vouchers${guide.tracking}`,
-          type: 'action',
-        });
+        if (onGuideClick) {
+          onGuideClick(guide.key);
+        } else {
```
helabenkhalfallah, line 35: the trackClick should be clicked only if we dont have onGuideClick?

packages/manager-react-components/src/components/navigation/menus/changelog/changelog.component.tsx
```diff
@@ -58,7 +60,11 @@ export const ChangelogButton: React.FC<ChangelogButtonProps> = ({
               onClickReturn={() =>
                 trackClick({
                   actionType: 'navigation',
-                  actions: [...chapters, ...CHANGELOG_PREFIXES, GO_TO(key)],
+                  actions: [
+                    ...chapters,
+                    ...(prefixes || CHANGELOG_PREFIXES),
```
helabenkhalfallah, line 65: prefixes does not have a default value, I have a doubt that this can cause a crash if prefixes is undefined or null.

packages/manager-react-components/src/components/guides-header/pci/pci-guides-header.component.tsx
```diff
@@ -26,10 +30,14 @@ export function PciGuidesHeader({ category }: PciGuidesHeaderProps) {
...
+          trackClick({
+            name: `public-cloud_credit_and_vouchers${guide.tracking}`,
```
aboungnaseng-ovhcloud, line 37: are we sure we want to harcoded this name with credit and voucher ?

packages/manager-react-components/src/components/navigation/menus/changelog/changelog.component.tsx
MarcMonchablon: But then `CHANGELOG_PREFIXES` will be used as a fallback value when `prefixes` are falsy, no ?

packages/manager-react-components/src/components/guides-header/pci/pci-guides-header.component.tsx
MarcMonchablon: We don't, but it's the current behavior, (it was already present in the original code, now behind an if/else) hence the optional prop to bypass the default tracking without impacting the other micro-app :)

packages/manager-react-components/src/components/guides-header/pci/pci-guides-header.component.tsx
MarcMonchablon: Yes. The issue is:
 - We need to update the tracking (to an updated tracking nomenclature) for this component
 - We don't want to potentially alter the tracking of already-deployed micro-app
 - But we also need to bypass the default tracking behavior for new one.

Hence the fact that trackClick is only called when the optional prop isn't set, as a way to update the component without introducing unexpected tracking change the other micro-apps.

helabenkhalfallah: Ok, clear, thank you for the detailed explanation:)

packages/manager-react-components/src/components/navigation/menus/changelog/changelog.component.tsx
helabenkhalfallah: Yes! since the parenthesis surrounding both, good!

PR #17178, opened 2025-05-19
Feat/manager 16280

mathieumousnier:
ref: #MANAGER-16280

Adds the service URN to the general information tab

* I hard coded the label "URN" to avoid going through the translation process for this.
Other URN labels are in translations, but in different namespaces.
There is several solutions : 
1. hard code URN like I did
2. create a file with some constant labels, most of the time technical ones, because this is never translated
3. use the common translation file because it is used everywhere
4. put URN label in translation file for the servicekey namespace

I think we should do 1 or 2 with and plan a refactoring for existing labels


packages/manager/apps/key-management-service/src/pages/serviceKey/ServiceKey.page.tsx
```diff
@@ -186,6 +186,11 @@ export default function Key() {
                 ),
                 value: <Clipboard className="w-full" value={kmsKey.id} />,
               },
+              {
+                id: 'urn',
+                label: 'URN',
```
mathieumousnier, line 191: I hard coded the label "URN" to avoid going through the translation process for this.
Other URN labels are in translations, but in different namespaces.
There is several solutions : 
1. hard code URN like I did
2. create a file with some constant labels, most of the time technical ones, because this is never translated
3. use the common translation file because it is used everywhere
4. put URN label in translation file for the servicekey namespace

I think we should do 1 or 2 with and plan a refactoring for existing labels.

@rjamet-ovh what's your opinion ?

rjamet-ovh: we already have a `SERVICE_KEYS_LABEL` global constant for the same use-case, as it'll never be translated.

pauldkn (approved):

PR #17183, opened 2025-05-19
Revert/feat tapc 3783 tapc 3786

Eric-ciccotti:
## Description

Ticket Reference: #...


SimonChaumet (approved):

fredericvilcot (changes requested):
Commits should be reverted using proper `git revert` command

PR #17193, opened 2025-05-20
feat(container): improved pnr design with ods feedbacks

oalkabouss:
## Description

Ticket Reference: #MANAGER-14863


packages/manager/apps/container/src/container/nav-reshuffle/header/language-menu/LanguageMenu.component.tsx
```diff
@@ -0,0 +1,115 @@
...
+    userLocale?: string;
+};
+
+export const LanguageMenu: FunctionComponent<Props> = ({
+    onChange,
+    setUserLocale,
+    userLocale = '',
+}) => {
+    const { i18n } = useTranslation('language');
+    const ref = useRef<any>();
```
aottr, line 22: Could you try to type this please ?

packages/manager/apps/container/src/container/nav-reshuffle/header/notifications-button.tsx/NotificationsButton.component.tsx
aottr, line 1: the component path contains `.tsx`

packages/manager/apps/container/src/container/nav-reshuffle/header/style.module.scss
aottr, line 1: please don't use ui-kit in PNR

packages/manager/apps/container/src/container/nav-reshuffle/header/notifications-button/NotificationsButton.component.tsx
JacquesLarique, line 1: Why duplicating this component and not update the old one ?

packages/manager/apps/container/src/container/nav-reshuffle/header/language-menu/LanguageMenu.component.tsx
JacquesLarique, line 1: Why duplicating this component and not update the old one ?

packages/manager/apps/container/src/container/nav-reshuffle/header/language-menu/LanguageMenu.component.tsx
```diff
@@ -0,0 +1,115 @@
...
+    const trackingPlugin = shell.getPlugin('tracking');
+    const isSmallDevice = useMediaQuery({
+        query: `(max-width: ${SMALL_DEVICE_MAX_SIZE})`,
+    });
+    const { t } = useTranslation('language');
+    const [currentLanguage, setCurrentLanguage] = useState(null);
+    const [availableLanguages, setAvailableLanguages] = useState([]);
+
+    useEffect(() => {
+        const shadowRoot = ref.current?.shadowRoot;
```
JacquesLarique, line 33: Why do we need to do this ?
I don't see it being done in other places where we use OsdsMenu

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/ProjectSelector/ProjectSelector.constants.tsx
JacquesLarique, line 1: What is the use of these ? I don't see them being imported anywhere

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/SubTree/PublicCloudPanel.tsx
```diff
@@ -163,15 +165,20 @@ export const PublicCloudPanel: React.FC<ComponentProps<
           seeAllLabel={t('sidebar_pci_all')}
         />
         {pciError && (
-          <button
-            className={style.sidebar_pci_refresh}
-            onClick={() => refetchPciProjects()}
-            role="button"
-            title={t('sidebar_pci_load_error')}
+          <div
```
JacquesLarique, line 168: Shouldn't we use an OsdsButton here ?

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/SubTree/SubTreeSection.tsx
```diff
@@ -88,6 +95,8 @@ const SubTreeSection: React.FC<ComponentProps<SubTreeSectionProps>> = ({
                 {childNode.separator && <hr role="separator" />}
               </li>
             ))}
+
+          <OsdsDivider separator color={ODS_THEME_COLOR_INTENT.primary} className='px-3' size={ODS_DIVIDER_SIZE.four} />
```
JacquesLarique, line 99: Shouldn't this replace the `<hr role="separator" />` from line 95 ?

packages/manager/apps/container/src/container/nav-reshuffle/header/notifications-button/NotificationsButton.component.tsx
oalkabouss: Because it is used by the legacy version, whereas the one i implemented is for pnr. The request only concerns the pnr version

packages/manager/apps/container/src/container/nav-reshuffle/header/language-menu/LanguageMenu.component.tsx
oalkabouss: https://github.com/ovh/manager/pull/17193#discussion_r2101780442

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/SubTree/PublicCloudPanel.tsx
oalkabouss: I think osds was not used here because we did not want the hover effect with a bg. I'll confirm with the PO.

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/SubTree/SubTreeSection.tsx
oalkabouss: nop, that replaces the subtree_section class on line 57.

packages/manager/apps/container/src/container/nav-reshuffle/header/language-menu/LanguageMenu.component.tsx
oalkabouss: It is to trigger the onChange, OsdsMenu doesn't expose onChange. The onChange allows us to call these function: setIsNotificationsSidebarVisible(false); //Navabar.tsx

packages/manager/apps/container/src/container/nav-reshuffle/header/notifications-button/NotificationsButton.component.tsx
JacquesLarique: Seen with PO, there is no need to have two differents versions of this component so please make these changes directly on the previous version of the component

packages/manager/apps/container/src/container/nav-reshuffle/header/language-menu/LanguageMenu.component.tsx
JacquesLarique: Seen with PO, there is no need to have two differents versions of this component so please make these changes directly on the previous version of the component

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/SubTree/PublicCloudPanel.tsx
oalkabouss: Seen with the PO, he prefers to keep it like this.

anooparveti (dismissed):
Approved for files owned by CT.

packages/manager/apps/container/src/container/nav-reshuffle/header/language-menu/LanguageMenu.component.tsx
JacquesLarique: Yes but the `OsdsButton` that you display in the `OsdsMenuItem` has an onClick attribute where you can call `setIsNotificationsSidebarVisible(false);`

packages/manager/apps/container/src/container/common/language/index.tsx
```diff
@@ -1,91 +1,118 @@
...
-      <LanguageButton show={show} onClick={(nextShow) => setShow(nextShow)}>
-        {getLanguageLabel()}
-      </LanguageButton>
-      <LanguageList
-        languages={availableLanguages}
-        onSelect={onLocaleChange}
-      ></LanguageList>
-    </div>
-  );
+    console.log(currentLanguage, "currentLanguage", getLanguageLabel())
```
JacquesLarique: Cleanup

packages/manager/apps/container/src/container/nav-reshuffle/header/language-menu/LanguageMenu.component.tsx
oalkabouss: I thought about it, but I’m already passing a variable isOpened in the onChange(isOpened) callback. It's also used earlier with setIsDropdownOpen(show).
I think it’s better to check whether the menu is open or closed with:
const isOpened = surfaceElement.hasAttribute('opened');
onChange(isOpened);

JacquesLarique: Ok after checking the code I see the issue is to make the call when opening the language menu, and indeed it seems only possible using `MutationObserver` which I don't like to do here (too complex).

So I propose another solution: use `OsdsPopover` instead of `OsdsMenu`
It offer the `onClick` props (letting us avoid using `MutationObserver` and also `OsdsMenu` is dropped in version 18 of ODS (https://ovh.github.io/design-system/latest/?path=/docs/ods-components-menu--docs)
You can find an implementation example here: https://github.com/ovh/manager/blob/master/packages/manager/apps/hub/src/billing/components/services-actions/ServicesActions.component.tsx

oalkabouss: i was initialy going with a onClick and useClickAway to handle pop-up open and close. But with OSDS, even when the pop-up is closed, the content still stay in the DOM, and it cause weird behavior like value inversion in onChange. Without a strong solution, I keep the MutationObserver way for now to handle onChange with good value.

aottr (approved):

PR #17210, opened 2025-05-20
fix(vcd): fix breadcrumb translation and rerender

tibs245:
## Description

Ticket Reference:  #MANAGER-18185

### DIFF
--- a/packages/manager/apps/hpc-vmware-managed-vcd/src/components/breadcrumb/Breadcrumb.component.tsx
+++ b/packages/manager/apps/hpc-vmware-managed-vcd/src/components/breadcrumb/Breadcrumb.component.tsx
@@ -20,10 +20,15 @@ function Breadcrumb({ customRootLabel, items }: BreadcrumbProps): JSX.Element {
 
   const breadcrumbItems = useBreadcrumb({ rootLabel, items });
 
+  // The key is used to force re-render breadcrumb items when the breadcrumb length changes, to update items states (`collapsed` / `last`)
   return (
     <OdsBreadcrumb>
-      {breadcrumbItems.map(({ label, ...props }) => (
-        <OdsBreadcrumbItem key={label} label={label} {...props} />
+      {breadcrumbItems.map(({ label, ...props }, index) => (
+        <OdsBreadcrumbItem
+          key={`${label}-${index}-${breadcrumbItems.length}`}
+          label={label}
+          {...props}
+        />
       ))}
     </OdsBreadcrumb>
   );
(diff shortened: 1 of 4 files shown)


packages/manager/apps/hpc-vmware-managed-vcd/src/components/breadcrumb/Breadcrumb.component.tsx
```diff
@@ -20,10 +20,15 @@ function Breadcrumb({ customRootLabel, items }: BreadcrumbProps): JSX.Element {
...
   const breadcrumbItems = useBreadcrumb({ rootLabel, items });
 
+  // The key is used to force re-render the last item component when his state change
   return (
     <OdsBreadcrumb>
-      {breadcrumbItems.map(({ label, ...props }) => (
-        <OdsBreadcrumbItem key={label} label={label} {...props} />
+      {breadcrumbItems.map(({ label, ...props }, index) => (
+        <OdsBreadcrumbItem
+          key={`${label}-${index + 1 === breadcrumbItems.length}`}
```
pauldkn, line 28: Looks like `index + 1 === breadcrumbItems.length` returns a boolean, so I guess that could result in a duplicated key

tibs245: No can't be send duplicated key because already uniq with label

pauldkn: I guess it could, even if having the same label would be a bit unusual
Was that the result you intended, a boolean right after the label?

packages/manager/apps/hpc-vmware-managed-vcd/src/components/breadcrumb/Breadcrumb.component.tsx
```diff
@@ -20,10 +20,15 @@ function Breadcrumb({ customRootLabel, items }: BreadcrumbProps): JSX.Element {
 
   const breadcrumbItems = useBreadcrumb({ rootLabel, items });
 
+  // The key is used to force re-render the last item component when his state change
```
pauldkn, line 23: **nitpick**
```suggestion
  // The key is used to force re-render breadcrumb items when the breadcrumb length changes, to update items states (`collapsed` / `last`)
```

pauldkn (approved):

vovh (approved):

PR #17212, opened 2025-05-20
fix(cloud-volume-backup): preserve original volume type when creating from backup #TAPC-4081

StephaneRavet:
ref: #TAPC-4081


packages/manager/apps/pci-volume-backup/src/data/api/volume.ts
StephaneRavet, line 1: This type is unused.

PR #17241, opened 2025-05-21
style(pci-volume-backup): make styles compliant

FabienHenon:
ref: #DTCORE-3086

## Description

Ticket Reference: #DTCORE-3086


SimonChaumet (approved):
Please use correct commit type, style is only for code styling, therefore it should not modify any logic

MarcMonchablon (approved):

seven-amid (approved):

Issue #13901, opened 2024-11-04
[manager-components]: Dont bundle the translations in the manager-react-components


anooparveti:
### On which entity, you would like to report a bug?

Whole library

### Description

Currently, ~50% of the size is from the translations files being included in the bundle.
![Screenshot 2024-11-04 at 12 16 07 PM](https://github.com/user-attachments/assets/13ee4853-e95a-4c60-8416-a514bab9e085)

Dont include static content in the bundle and load it on demand.

### Steps to reproduce

N/A

### Operating System

- [ ] macOS
- [ ] Windows
- [ ] Linux

PR #17246, opened 2025-05-21
[FEATURE]: Dont bundle the translations in the manager-react-components

helabenkhalfallah:
ref: MANAGER-18402

## Description

Original issue: https://github.com/ovh/manager/issues/13901

## Additional Information

Before: 
BEFORE_1

After:
AFTER_1

### DIFF
--- a/packages/manager-react-components/src/components/ManagerText/translations/Messages_fr_FR.json
+++ b/packages/manager-react-components/src/components/ManagerText/translations/Messages_fr_FR.json
@@ -1,5 +1,5 @@
 {
-  "iam_hidden_text": "Hidden",
+  "iam_hidden_text": "Caché",
   "common_iam_get_message": "Vous ne disposez pas des permissions nécessaires pour afficher les données de cette ressource. Veuillez contacter votre administrateur.",
   "common_iam_actions_message": "Vous ne disposez pas des permissions nécessaires pour effectuer cette action, veuillez contacter votre administrateur"
 }
--- a/packages/manager-react-components/src/components/ManagerText/translations/index.ts
+++ b/packages/manager-react-components/src/components/ManagerText/translations/index.ts
@@ -1,9 +1,11 @@
 import i18next from 'i18next';
 
 import fr_FR from './Messages_fr_FR.json';
+import en_GB from './Messages_en_GB.json';
 
 function addTranslations() {
   i18next.addResources('fr_FR', 'iam', fr_FR);
+  i18next.addResources('en_GB', 'iam', en_GB);
 }
 
 if (i18next.isInitialized) {
--- a/packages/manager-react-components/src/hooks/region/translations/index.ts
+++ b/packages/manager-react-components/src/hooks/region/translations/index.ts
@@ -1,27 +1,92 @@
 import i18next from 'i18next';
 
-import de_DE from './Messages_de_DE.json';
-import en_GB from './Messages_en_GB.json';
-import es_ES from './Messages_es_ES.json';
-import fr_CA from './Messages_fr_CA.json';
-import fr_FR from './Messages_fr_FR.json';
-import it_IT from './Messages_it_IT.json';
-import pl_PL from './Messages_pl_PL.json';
-import pt_PT from './Messages_pt_PT.json';
+/**
+ * @typedef {Object} TranslationLoaders
+ * @property {() => Promise<{default: object}>} de_DE - German (Germany) translations loader
+ * @property {() => Promise<{default: object}>} en_GB - English (UK) translations loader
+ * @property {() => Promise<{default: object}>} es_ES - Spanish (Spain) translations loader
+ * @property {() => Promise<{default: object}>} fr_CA - French (Canada) translations loader
+ * @property {() => Promise<{default: object}>} fr_FR - French (France) translations loader
+ * @property {() => Promise<{default: object}>} it_IT - Italian (Italy) translations loader
+ * @property {() => Promise<{default: object}>} pl_PL - Polish (Poland) translations loader
+ * @property {() => Promise<{default: object}>} pt_PT - Portuguese (Portugal) translations loader
+ */
 
-function addTranslations() {
-  i18next.addResources('de_DE', 'region', de_DE);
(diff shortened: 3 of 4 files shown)


StephaneRavet (dismissed):

packages/manager-react-components/src/components/ManagerText/translations/Messages_de_DE.json
anooparveti, line 1: Are you sure to merge this change in `develop`. I assume this change is for testing purpose.

packages/manager-react-components/src/components/ManagerText/translations/Messages_en_GB.json
anooparveti, line 1: Are you sure to merge this change in `develop`. I assume this change is for testing purpose.

packages/manager-react-components/src/components/ManagerText/translations/Messages_es_ES.json
anooparveti, line 1: Are you sure to merge this change in `develop`. I assume this change is for testing purpose.

packages/manager-react-components/src/components/ManagerText/translations/Messages_fr_FR.json
anooparveti, line 1: Are you sure to merge this change in `develop`. I assume this change is for testing purpose.

packages/manager-react-components/src/components/ManagerText/translations/Messages_it_IT.json
anooparveti, line 1: Are you sure to merge this change in `develop`. I assume this change is for testing purpose.

packages/manager-react-components/src/components/ManagerText/translations/Messages_pl_PL.json
anooparveti, line 1: Are you sure to merge this change in `develop`. I assume this change is for testing purpose.

packages/manager-react-components/src/components/ManagerText/translations/Messages_pt_PT.json
anooparveti, line 1: Are you sure to merge this change in `develop`. I assume this change is for testing purpose.

anooparveti (changes requested):

packages/manager-react-components/src/utils/TranslationUtils.ts
```diff
@@ -0,0 +1,85 @@
...
+      handleLanguageChange(i18next.language);
+    });
+  }
+
+  i18next.on('languageChanged', handleLanguageChange);
+
+  window.addEventListener('languagechange', () => {
+    const normalizedLang = normalizeLanguageCode(navigator.language);
+    i18next.changeLanguage(normalizedLang);
+  });
```
anooparveti, line 84: **nitpick(performance):** Since this function needs to be called for every component that uses translations, we have an added eventListener. So imo, it's not good to have 1 eventListener per component and moreover I dont see that it's being removed when the component is removed from the DOM.

anooparveti (changes requested):

packages/manager-react-components/src/components/ManagerText/translations/Messages_pl_PL.json
helabenkhalfallah: Yes I'll clean all :)

anooparveti (changes requested):
Screenshot 2025-05-29 at 12 54 22 PM

Irrespective of whether a component is present in the page or not, existing translations of all the components are loaded by default. **IMO, it's bad and not scalable when we have lots of components with translations, difference can be significant to load all the unwanted translations.**

I think there will be an issue with fallback language as well. Current behavior, if translation key does not exist on the current locale file, content from fallback language (FR) will be displayed. **Since the current implementation does not load/configure the fallback language, there might be an issue with the current expected behaviour.**

helabenkhalfallah:
> Screenshot 2025-05-29 at 12 54 22 PM
> Irrespective of whether a component is present in the page or not, existing translations of all the components are loaded by default. **IMO, it's bad and not scalable when we have lots of components with translations, difference can be significant to load all the unwanted translations.**
> 
> I think there will be an issue with fallback language as well. Current behavior, if translation key does not exist on the current locale file, content from fallback language (FR) will be displayed. **Since the current implementation does not load/configure the fallback language, there might be an issue with the current expected behaviour.**

Did you see that I did not change anything comparing to what's loaded before? I just changed  the type of import The same code is preserved and I moved it to a new utils file. So what behaviour did I change? 

> Since the current implementation does not load/configure the fallback language, there might be an issue with the current expected behaviour.

I did not change on any application file, only in MRC:

Capture d’écran 2025-05-30 à 08 42 49
Capture d’écran 2025-05-30 à 08 43 14

"Since the current implementation does not load/configure the fallback language, there might be an issue with the current expected behaviour."

helabenkhalfallah:
> Screenshot 2025-05-29 at 12 54 22 PM
> Irrespective of whether a component is present in the page or not, existing translations of all the components are loaded by default. **IMO, it's bad and not scalable when we have lots of components with translations, difference can be significant to load all the unwanted translations.**
> 
> I think there will be an issue with fallback language as well. Current behavior, if translation key does not exist on the current locale file, content from fallback language (FR) will be displayed. **Since the current implementation does not load/configure the fallback language, there might be an issue with the current expected behaviour.**

As we see together:
-> I'll take a look at the fallback language.
-> But for the chuncks that are all loaded, this is another concern related to MRC. MRC should be modular and tree-shaked so language chuncks can be loaded as need. This is a concern of another ticket.

Thanks :)

helabenkhalfallah:
## Handling default fallback

**The `test-key` is not declared for `pl_PL`, `it_IT` and `pt_PT`..**

### Before

We can see that the default French language is not loaded:
image
image
image
image
image

## After

We can see below that the French language is used as fallback: 
image
image
image
image

anooparveti (dismissed):
For the namespace, I think it's better to add prefix like `@ovh-ux/manager-react-components` to avoid collision if the µ-app also add a same namespace.
For example, in `iam` µ-app, we can have a namespace with name `iam` and we also have a namespace `iam` in MRC.

PR #17286, opened 2025-05-23
[FIX]: Routes migration from json to components version - add Suspense to wrap lazy imports

helabenkhalfallah:
## Description

An issue encountered by Tristan when using lazy components version due to absence of Suspense around the RouterProvider.

Ticket Reference: #...    

## Additional Information

```
Vous • So as we see together, the best practices is to lazy load by default, and since also everything was lazy before, I cant make the decision in the CLI to choose by myself the approach. But what can I do: :
-> add a suspense in the App that wrap the routerprovider.
-> enhance the documentation to tell developers that since it's a component version now, it'll be a real lazy evaluated so we have a loader displayed before loading the routes/chuncks. 
-> it's the responsibility of the project to choose lazy or not based on the route size (bundle size) and performances.

Tristan Wagner (EXT) • Yes and for the release I made 2 separated PRs that address the 2 issues found,

1st one  https://github.com/ovh/manager/pull/17275 about this lazyloading issue -> decision is to not lazyload Layout components and to avoid having a suspense at the root of the app, as it makes the user experience better (no blinking whitescreens with Loading icon when something is lazy loaded)

2nd one https://github.com/ovh/manager/pull/17278 about the bug in the email account form

I ran the E2E test in local with both and it's working as expected but to double check ofc
```


packages/manager-cli/routes-migrate/steps/update-routers-init-cli.mjs
```diff
@@ -98,7 +104,7 @@ const updateRouterInitialization = (code) => {
   // Replace JSX usage of router
   transformed = transformed.replace(
     /<RouterProvider\s+router=\{router\}\s*\/>/g,
-    `<RouterProvider router={routes} />`,
+    `<Suspense fallback={<span>Loading routes ...</span>}><RouterProvider router={routes} /></Suspense>`,
```
tristanwagner, line 107: as I said adding this automatically will prevent crash in case of issue, but developers really need to check and understand these behavior and modify accordingly if needed, so maybe we could add a comment to make sure it's not overlooked, but actually I hope the fallback will be enough for raising a flag 😁

tristanwagner (approved):

packages/manager-cli/routes-migrate/steps/update-routers-init-cli.mjs
helabenkhalfallah: Yes, I updated the doc for that : https://stash.ovh.net/projects/UXMANAGERS/repos/documentation/pull-requests/188/diff#docs%2Ftechnical-documentation%2Fmanager%2Fdevelopment-guidelines%2Fupdate-react-routes.md

And I kept intentionally a bad loader to be easy founded :) :)

tristanwagner (approved):

PR #17360, opened 2025-05-28
feat(okms): add new secret-management-service entry and create first routes

mathieumousnier:
## Description

- add new secret-management-service menu entry
- declare first secrets routes

Ticket Reference: #MANAGER-17869

## Additionnal Informations

The "secret-management-service" features will be added in a folder called modules/sms.
Files related to "key-management-service" won't be moved for now, but the final goal is to move them to modules/kms to get a clear separation between the two modules, with a DDD strategy in mind.

For urls, this is a first step where we declare a secret-management-service/ path to host all secrets related pages. At some point we will move kms pages to a key-management-service/ path.

## Demo

_With corresponding feature-flipping for the menu entry_

https://github.com/user-attachments/assets/b948d89a-3cd0-4a47-978a-792bb86e1c3a

### DIFF
--- a/packages/manager/apps/key-management-service/src/modules/sms/routes/routes.constants.ts
+++ b/packages/manager/apps/key-management-service/src/modules/sms/routes/routes.constants.ts
@@ -1,0 +1,23 @@
+const URIS = {
+  root: 'secret-management-service',
+  versions: 'versions',
+  secrets: 'secrets',
+  create: 'create',
+};
+
+const URLS = {
+  smsRoot: `/${URIS.root}`,
+  secretCreate: `/${URIS.root}/${URIS.create}`,
+  secretListing: (domainId: string) =>
+    `/${URIS.root}/${domainId}/${URIS.secrets}`,
+  secretDashboard: (domainId: string, secretId: string) =>
+    `/${URIS.root}/${domainId}/${URIS.secrets}/${secretId}`,
+};
+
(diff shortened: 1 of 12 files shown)


packages/manager/apps/key-management-service/src/modules/sms/routes/routes.constants.ts
```diff
@@ -0,0 +1,23 @@
+export const URIS = {
+  root: 'secret-management-service',
+  versions: 'versions',
+  secrets: 'secrets',
+  create: 'create',
+};
+
+export const URLS = {
```
rjamet-ovh, line 8:
```suggestion
const URLS = {
```

packages/manager/apps/key-management-service/src/modules/sms/routes/routes.constants.ts
```diff
@@ -0,0 +1,23 @@
...
+    `/${URIS.root}/${domainId}/${URIS.secrets}/${secretId}`,
+};
+
+export const SMS_URL_PARAMS = {
+  domainId: ':domainId',
+  secretId: ':secretId',
+};
+
+export const SMS_ROUTES_URIS = URIS;
+export const SMS_ROUTES_URLS = URLS;
```
rjamet-ovh, line 23: why renaming in some cases and not the others ?

PR #17391, opened 2025-06-02
test(*): fix test-cases during pnpm migration

anooparveti:
## Description

Test-case fix for common-api, network-common and dedicated packages.

Ticket Reference: #...


playwright-helpers/network.ts
```diff
@@ -1,38 +1,38 @@
...
-            typeof response === 'function' ? await response(request) : response,
-          body: responseText,
-        });
-      }
-      return route.fallback();
-    },
-    { times: once ? 1 : undefined },
-  );
-};
+// export const toPlaywrightMockHandler = (context: BrowserContext) => ({
```
helabenkhalfallah, line 5: commented -> not used -> should not be committed nor pushed. We can find history on git.

packages/manager/modules/common-api/package.json
```diff
@@ -32,7 +32,7 @@
     "@vitejs/plugin-react": "^4.3.3",
     "@vitest/coverage-v8": "^2.1.9",
     "msw": "2.1.7",
-    "react-dom": "^18.2.19",
+    "react-dom": "^18.3.1",
```
helabenkhalfallah, line 35: why?

packages/manager/modules/network-common/package.json
```diff
@@ -37,6 +37,7 @@
     "@vitejs/plugin-react": "^4.3.3",
     "@vitest/coverage-v8": "^2.1.9",
     "msw": "2.1.7",
+    "react-dom": "^18.3.1",
```
helabenkhalfallah, line 40: why?

pnpm-workspace.yaml
```diff
@@ -2,7 +2,18 @@ packages:
   - 'docs'
   - 'packages/components/*'
   - 'packages/manager/apps/*'
-  - 'packages/manager/core/*'
+  - 'packages/manager/core/api'
```
helabenkhalfallah, line 5: why?

ovh-cds:
CDS Report build-v5#49.0 ✘  
* 
  * Build ✘

packages/manager/modules/network-common/package.json
anooparveti: Ref: https://github.com/testing-library/react-testing-library/issues/1078

pnpm-workspace.yaml
anooparveti: It was the same issue that we encountered while publishing MRCv2, multiple instances of react were created and that was causing the errors while accessing context.
So, we need to force shell-client to be consumed from NPM package and not local by removing it as pnpm workspace.

packages/manager/modules/common-api/package.json
anooparveti: Ref: https://github.com/testing-library/react-testing-library/issues/1078

playwright-helpers/network.ts
anooparveti: Agreed, we can remove it in the final PR if the removal solves the issue of running test-cases.

frenautvh (approved):

PR #17417, opened 2025-06-03
feat(pci-block-storage): add refresh button

SimonChaumet:
ref: #TAPC-4245

## Description

Ticket Reference: #...    

## Additional Information

I recommend to use the "hide whitespace" setting for the review

### DIFF
--- a/packages/manager/apps/pci-block-storage/src/pages/list/List.page.tsx
+++ b/packages/manager/apps/pci-block-storage/src/pages/list/List.page.tsx
@@ -1,5 +1,5 @@
-import { Suspense, useContext, useEffect, useRef, useState } from 'react';
-import { Outlet, useParams, useNavigate } from 'react-router-dom';
+import { Suspense, useRef, useState } from 'react';
+import { Outlet, useParams, useNavigate, Navigate } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 import {
   ChangelogButton,
@@ -9,12 +9,10 @@ import {
   Headers,
   Notifications,
   PciGuidesHeader,
-  PciMaintenanceBanner,
-  RedirectionGuard,
   useColumnFilters,
   useDataGrid,
   useNotifications,
-  useProductMaintenance,
+  useProjectUrl,
 } from '@ovh-ux/manager-react-components';
 import { ODS_THEME_COLOR_INTENT } from '@ovhcloud/ods-common-theming';
 import {
@@ -28,7 +26,6 @@ import {
   OsdsSpinner,
 } from '@ovhcloud/ods-components/react';
 import { FilterCategories, FilterComparator } from '@ovh-ux/manager-core-api';
-import { ShellContext } from '@ovh-ux/manager-react-shell-client';
 import {
   ODS_BUTTON_SIZE,
   ODS_BUTTON_VARIANT,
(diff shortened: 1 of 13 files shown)


packages/manager/apps/pci-instances/src/queryClient.ts
```diff
@@ -4,6 +4,7 @@ const queryClient = new QueryClient({
   defaultOptions: {
     queries: {
       staleTime: 300_000,
+      throwOnError: true,
```
fredericvilcot: This will cause the ErrorBoundary to be mounted each time an error is thrown, and we don't want this behavior as we use error banners to be displayed in specific use cases instead of the fallback UI

SimonChaumet: Since there is no page with ErrorBanner and the ErrorBanner takes the whole page I don't understand how you would handle the error ? If this setting is set to false there is no error thrown and the error banner isn't displayed

SimonChaumet: Wrong project, but I think it is still needed, otherwise there is no error handling

fredericvilcot: Throwing an error means that an error is caught by the router and calls the `ErrorBoundary` fallback UI.
In most of cases, we use the `error` object returned by the useQuery hook to handle manually errors, like in the listing page where errors are handled by a error banner (`Notifications`) above the data grid.

SimonChaumet: Can you provide an example for this ? I didn't saws any, in block there is no error handling like this

fredericvilcot: Here is the error handling in the instance listing page: https://github.com/ovh/manager/blob/6a183d7164a41a4e6f2fe56f41abe77ce1d84986/packages/manager/apps/pci-instances/src/components/datagrid/Datagrid.component.tsx#L311

Here we do not want to display the Error Boundary but rather the notification above the listing page to tell the user if an error occurred.

BTW, if you enable error throwing on the query client, keep in mind that your mutations will also throw an error and display a fallback Error Boundary. According to the way we handle the `onError` callbacks, I'm not sure that it's the way you want to handle errors.

Capture d’écran 2025-06-03 à 14 31 33

packages/manager/apps/pci-block-storage/src/pages/list/List.page.tsx
```diff
@@ -90,183 +66,184 @@ export default function ListingPage() {
...
+          <div className="flex items-center">
+            {isFetching ? (
+              <OsdsSpinner inline size={ODS_SPINNER_SIZE.md} />
+            ) : (
               <OsdsButton
-                slot="popover-trigger"
                 size={ODS_BUTTON_SIZE.sm}
                 color={ODS_THEME_COLOR_INTENT.primary}
                 variant={ODS_BUTTON_VARIANT.stroked}
+                onClick={() => refetch()}
```
fredericvilcot, line 133:
```suggestion
                onClick={refetch}
```

packages/manager/apps/pci-block-storage/src/pages/list/List.page.tsx
```diff
@@ -63,25 +56,8 @@ export default function ListingPage() {
...
-    isPending: isPendingAllVolumes,
-  } = useAllVolumes(projectId);
-  const {
-    data: volumes,
-    isLoading: isVolumesLoading,
-    isPending: isVolumesPending,
-    error,
-  } = useVolumes(
+  const { data: allVolumes } = useAllVolumes(projectId);
+  const { data: volumes, error, isFetching, isPending, refetch } = useVolumes(
```
fredericvilcot, line 60: I'm a bit confused with the naming as, at first reading, I cannot distinguish the difference between `volumes` and `allVolumes`

packages/manager/apps/pci-block-storage/src/pages/list/List.page.tsx
SimonChaumet: It doesn't work since onClick sends a MouseClickEvent and refetch accepts an object with options

packages/manager/apps/pci-block-storage/src/pages/list/List.page.tsx
SimonChaumet: It was the previous naming, useAllVolumes corresponds to the mapped query while useVolumes corresponds to the query for the datagrid

fredericvilcot: And can we enhance this?

SimonChaumet: When we will refactor the listing page yes, for now this feature only adds a refresh button so I think it is not the place to do a refactor on a hook that is used in multiple places

fredericvilcot (approved):

Rom123Soleil (approved):

Eric-ciccotti (approved):

PR #17419, opened 2025-06-03
chore(manager-wiki): build dependencies

anooparveti:
## Description

Build dependencies before launching manager-wiki in local

Ticket Reference: #MANAGER-17646

### DIFF
--- a/packages/manager-wiki/package.json
+++ b/packages/manager-wiki/package.json
@@ -6,9 +6,10 @@
   "license": "BSD-3-Clause",
   "author": "OVH SAS",
   "scripts": {
+    "build-dependencies": "turbo run build --filter=\"@ovh-ux/manager-wiki\"",
     "build:storybook": "storybook build",
     "dev": "tsc",
-    "start": "rm -rf dist/.cache/storybook && storybook dev -p 6006"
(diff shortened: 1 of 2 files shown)


packages/manager-wiki/package.json
```diff
@@ -6,9 +6,10 @@
   "license": "BSD-3-Clause",
   "author": "OVH SAS",
   "scripts": {
+    "build-dependencies": "turbo run build --filter=\"@ovh-ux/manager-wiki\"",
     "build:storybook": "storybook build",
     "dev": "tsc",
-    "start": "rm -rf dist/.cache/storybook && storybook dev -p 6006"
+    "start": "yarn run build-dependencies && rm -rf dist/.cache/storybook && storybook dev -p 6006"
```
helabenkhalfallah, line 12: we should not clean cache then build?

anooparveti: Since it is only required to build the dependent packages to start the storybook, I dont see the need to clean as rebuilding them can be quick with cache.

helabenkhalfallah (approved):

aboungnaseng-ovhcloud (dismissed):

PR #17420, opened 2025-06-03
feat(ovh-shell): implement location plugin

JacquesLarique:
## Description

Created a location plugin allowing the retrieval and sharing of location API response

Ticket Reference: #MANAGER-18341

### DIFF
--- a/packages/components/ovh-shell/src/plugin/location/index.ts
+++ b/packages/components/ovh-shell/src/plugin/location/index.ts
@@ -1,0 +1,55 @@
+import { v2 } from '@ovh-ux/manager-core-api';
+
+export type CardinalPoint =
+  | 'CENTRAL'
+  | 'EAST'
+  | 'NORTH'
+  | 'NORTHEAST'
+  | 'NORTHWEST'
+  | 'SOUTH'
+  | 'SOUTHEAST'
+  | 'SOUTHWEST'
+  | 'WEST';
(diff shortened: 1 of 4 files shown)


packages/components/ovh-shell/src/plugin/location/index.ts
```diff
@@ -0,0 +1,55 @@
...
+  string & keyof LocationPlugin,
+  CallableFunction
+>;
+
+export class LocationPlugin implements ILocationPlugin {
+  private locations: Location[] | null = null;
+
+  async getLocations(): Promise<Location[]> {
+    if (this.locations === null) {
+      const { data } = await v2.get('/location');
```
anooparveti, line 50: **nitpick:** Please implement error handling as well in case of failure from API. Since this is not critical/blocker data, we can have a retry or ignore the failure gracefully.

anooparveti (changes requested):

packages/components/ovh-shell/src/plugin/location/index.ts
JacquesLarique: As seen with you, we'll implement the retry logic using tanstack on MRC's side (#17465)

aboungnaseng-ovhcloud (dismissed):

anooparveti (dismissed):

anooparveti (approved):

PR #17421, opened 2025-06-03
fix(project-vmware-logs): fix datacenters url for legacy redirection

vovh:
ref: #MANAGER-17691

## Description

Ticket Reference: #MANAGER-17691


darsene (approved):

packages/manager/apps/hpc-vmware-vsphere/src/routes/routes.constant.ts
```diff
@@ -4,7 +4,7 @@ export const urls = {
   listing: '',
   dashboard: ':serviceName',
   logs: 'logs',
-  datacenters: 'datacenters',
+  datacenters: 'datacenter',
```
tibs245:
```suggestion
  datacenter: 'datacenter',
```

What do you think ?

tibs245 (approved):

PR #17444, opened 2025-06-04
[CI][pci-instances] ESlint config

fredericvilcot:
## Description

This PR adds `@typescript-eslint/no-unnecessary-condition` rule and refactors the linter issues


packages/manager/apps/pci-instances/src/pages/instances/action/InstanceAction.page.tsx
```diff
@@ -59,16 +59,12 @@ const InstanceAction: FC = () => {
...
-                  />,
-                ]
-              : []
-          }
+          components={[
+            <code
+              key="0"
+              className="px-1 py-0.5 text-[90%] text-[#c7254e] bg-[#f9f2f4] rounded"
+            />,
+          ]}
```
SimonChaumet, line 67: how is this modification related to eslint modification ? The 2 codes doesn't seem equivalent

fredericvilcot: `isRescue` ternary has been removed as predicate is already checked in line 53

packages/manager/apps/pci-instances/src/data/hooks/image/selector/image.selector.spec.tsx
```diff
@@ -10,12 +10,25 @@ const t: TFunction = ((key: string, options?: { name: string }) => {
...
+const fakeImage: TImageDto = {
+  name: 'Ubuntu',
+  osType: '',
+  baseImageRef: '',
+  createdAt: '',
+  id: 'foo',
+  region: '',
+  size: 0,
+  status: 'active',
+  type: 'official',
```
lolottetheclash, line 22: just a detail: previous ones are "standard" type, I don't know if it's important or not here

fredericvilcot: `standard` types didn't exist in the original code..

Here is the correct signature of the type :)
```ts
export type TImageType = 'custom' | 'official' | 'rescue' | 'snapshot';
```

SimonChaumet (approved):

lolottetheclash (approved):

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstances.ts
```diff
@@ -51,11 +51,11 @@ export const updateInstanceFromCache: TUpdateInstanceFromCache = (
     predicate: listQueryKeyPredicate(projectId),
   });
 
+  let isPageModified = false;
```
ppprevost: isPageModified, is called inside a map which is considered as a side effect and must be avoid.
You could directly test queryData.pages and updatedPages ? which is big computation too i admit

packages/manager/apps/pci-instances/src/pages/instances/onboarding/Onboarding.page.tsx
```diff
@@ -28,7 +28,7 @@ const Onboarding: FC = () => {
   const { ovhSubsidiary } = context.environment.getUser() as {
     ovhSubsidiary: OvhSubsidiary;
   };
-  const project = useRouteLoaderData('root') as TProject;
+  const project = useRouteLoaderData('root') as TProject | undefined;
```
ppprevost, line 31: why a casting ?

fredericvilcot: We need to cast with the loader return type as the `useRouteLoaderData` function returns `unknown` in its signature and accepts no type parameter.

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstances.ts
fredericvilcot: @ppprevost yes I totally agree that there could have a side effect mutating the variable inside the loop, but that was the original implementation and I'd prefer refactor this code in a dedicated PR with non regression test items

ppprevost (approved):

PR #17470, opened 2025-06-05
feat(mrc): add drawer component

mathieumousnier:
## Description

The Drawer component is an expandable panel that slides in from the right side of the screen, designed to display contextual content or forms without disrupting the main interface.

Ticket Reference: #MANAGER-17872

### Why there is no `slide-out` animation 

The OdsDrawer component comes with a `slide-in` and a `slide-out` animation.
To get the `slide-out` animation you need 2 things : 
- use the `open` and `close` callbacks of the OdsDrawer component and do not rely only on the `isOpen` prop
- if used with an associated route, wait for the animation to play out before changing the route

I could add this with a bit of complexity in the code inducing some more setTimeOuts on every drawer close event, solution I already don't like that much.

But it won't over, after this I would have to : 
- sync the slide-out animation of the drawer with the handle animation (the worst part I think)
- catch the escape key press to trigger the dismiss function with a setTimeout
- and I'm sure handle a few more edge cases I didn't think about

So, as I feel like the result without the slide-out animation is quiet ok, I prefered cutting it out and go for cleaner code.
Pauline from the design team who designed the drawer confirmed it is alright like this.

### Difficulties on the `handle` implementation

While I would have preferred to "absolute position" the handle relatively to its Drawer, and rely on Drawer animations to animate the handle, I couldn't get to display anything outside the frame of the `OdsDrawer`.
So I chose to position the handle as "fixed", and to duplicate slide-in slide-out animations on the drawer and the handle.

Also, the handle slides on the "collapse" and "uncollapse" animations, because I added those animations myself with simple tailwind classes. But, for the drawer opening animations I didn't want to replicate the drawer animation coded in ODS and I chose to fadein the handle, the result seems better.

## Demos

https://github.com/user-attachments/assets/7c46ae92-3b92-4193-a726-3f23392e4d59

https://github.com/user-attachments/assets/78bdea83-4175-45ad-a628-850cb8bcb5cf

https://github.com/user-attachments/assets/d0814b77-3bc8-4ae4-ae5e-3ee47782c717


packages/manager-react-components/src/components/drawer/Drawer.component.tsx
```diff
@@ -0,0 +1,158 @@
...
+                  data-testid="drawer-dismiss-button"
+                  label=""
+                  aria-label="Dismiss"
+                  icon="xmark"
+                  onClick={onDismiss}
+                  color={ODS_BUTTON_COLOR.primary}
+                  variant={ODS_BUTTON_VARIANT.ghost}
+                />
+              </div>
+            </header>
```
SimonChaumet: If I understand correctly you cannot close the drawer when it is loading ?

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
```diff
@@ -0,0 +1,158 @@
...
+  isPrimaryButtonLoading,
+  isPrimaryButtonDisabled,
+  onPrimaryButtonClick,
+  primaryButtonLabel,
+  isSecondaryButtonLoading,
+  isSecondaryButtonDisabled,
+  onSecondaryButtonClick,
+  secondaryButtonLabel,
+}: DrawerProps) => {
+  const [collapseState, setCollapsedState] =
```
SimonChaumet: I feel like the collapsible drawer and the classic are two different components : one is controlled by parent, one is controlled by himself

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
```diff
@@ -0,0 +1,158 @@
...
+  ODS_BUTTON_VARIANT,
+} from '@ovhcloud/ods-components';
+import clsx from 'clsx';
+import { DrawerVariant, DrawerCollapseState } from './Drawer.types';
+import DrawerBackdrop from './DrawerBackdrop.component';
+import DrawerHandle from './DrawerHandle.component';
+
+export type DrawerProps = PropsWithChildren & {
+  /** Drawer heading */
+  heading: string;
```
SimonChaumet: I think this should be a reactnode

aboungnaseng-ovhcloud:
Do we need the attribute 'onOpen' ? as we have defined in the ticket

aboungnaseng-ovhcloud (dismissed):

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
```diff
@@ -0,0 +1,158 @@
...
+  /** Open/close the drawer */
+  isOpen: boolean;
+  /** Called when the drawer is closed internally */
+  onDismiss: () => void;
```
SimonChaumet: What happens when we want a collapsible drawer every time ? This prop would be an empty function since we would not dismiss it

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
```diff
@@ -0,0 +1,158 @@
...
+
+        {!isLoading && (
+          <div className="h-screen overflow-hidden flex flex-col justify-between">
+            <header className="px-[24px] min-h-[100px] flex items-center">
+              <div className="flex justify-between items-center w-full py-[24px] space-x-[8px]">
+                <OdsText preset={ODS_TEXT_PRESET.heading2}>{heading}</OdsText>
+                <OdsButton
+                  data-testid="drawer-dismiss-button"
+                  label=""
+                  aria-label="Dismiss"
```
SimonChaumet: No translation ?

packages/manager-react-components/src/components/drawer/Drawer.spec.tsx
```diff
@@ -0,0 +1,130 @@
...
+  onDismiss: vi.fn(),
+};
+
+it('should display the drawer in its classic variant', () => {
+  render(<Drawer {...mockedProps} />);
+  expect(screen.getByTestId('drawer')).not.toBeNull();
+  expect(screen.queryByText('Drawer heading')).not.toBeNull();
+  expect(screen.queryByText('Drawer content')).not.toBeNull();
+
+  const dismissButton = screen.getByTestId('drawer-dismiss-button');
```
SimonChaumet: You could query by label directly, it would avoid unnecessary test id

packages/manager-react-components/src/components/drawer/Drawer.spec.tsx
SimonChaumet, line 1: Should the drawer be closed when pressing escape ? if so I don't see a test related

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
```diff
@@ -0,0 +1,81 @@
...
+        isDrawerSlideInFinished ? 'opacity-100' : 'opacity-0',
+        hasEscapeBeenPressed && 'hidden',
+        collapseState === 'hidden' && 'translate-x-[456px]',
+      )}
+    >
+      <div
+        className={clsx(
+          'right-0 w-[56px] h-[56px] bg-white',
+          'rounded-l-lg absolute justify-center items-center',
+          // Hide the handle on small screens to avoid design issues
```
SimonChaumet: So how do we open it on small screens ?

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
```diff
@@ -0,0 +1,81 @@
...
+          // Remove shadow on right side to avoid overlap with the drawer
+          clipPath: 'inset(-15px 0px -15px -15px)',
+        }}
+      >
+        <div>
+          <OdsButton
+            data-testid="drawer-handle"
+            label={undefined}
+            aria-label={
+              collapseState === 'visible' ? 'Hide drawer' : 'Show drawer'
```
SimonChaumet: No translation ?

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
mathieumousnier: Good question, I will show the xmark icon button while loading

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
mathieumousnier: This is more an additionnal feature, the collapsible drawer should be opened and closed by the parents just as the classic one. But you can also collapse it, and the collapse state is managed in the component

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
mathieumousnier: This opens uses we didn't specified, as replacing the title with buttons.
The design team already identified the need of more buttons on the right corner, but we cut the spec without it because this component generated enough discussions. This might be an amelioration in a second time.

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
mathieumousnier: The "close button" (the cross icon on the corner) is always displayed, either in the classic and in the collapsible version, so this `onDismiss` function is required since this is where the parent state is updated by this "close button".

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
mathieumousnier: @aboungnaseng-ovhcloud should we put translations for aria labels in MRC ?

packages/manager-react-components/src/components/drawer/Drawer.spec.tsx
mathieumousnier: you're right, but this button has no label

packages/manager-react-components/src/components/drawer/Drawer.spec.tsx
mathieumousnier: The drawer is closed when on "esc" press but this is managed inside OdsDrawer, and I had to mock the component.
Then this is covered by odsdrawer tests

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
mathieumousnier: I could have given more details on the comment, but this only hides the "collapse" handle on small screen when the handle is closed from getting out of the screen.
This avoids a style issue, where you see the handle "detaching" from the drawer.
The only downside is that you can't collapse the drawer, and it is on screens close to mobile sizes.

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
SimonChaumet: I feel like these 2 behaviors should be 2 components, when I have a collapsible drawer I want it to always be displayed and collapsed by default, while in the other case I want to trigger it with a button and it should disappear when closed

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
SimonChaumet: If no we suppose that accessibility is only available for english-speaking users ?

packages/manager-react-components/src/components/drawer/Drawer.spec.tsx
SimonChaumet: Since you test its label line 43 it should have one : https://github.com/ovh/manager/pull/17470/files/1579d59fe7cda9b023d3269fb4efbe7f132365b5#diff-d90596faee22579388b1d59a0ca5b40daa3afac80d26c11839098a92c177e741R43

packages/manager-react-components/src/components/drawer/Drawer.spec.tsx
SimonChaumet: So what does this code covers : https://github.com/ovh/manager/pull/17470/files#diff-981136c233e2c4c6894544a17c7b92fc474415f1705c818fbf59a48063c9e4f3R23-R34 ?

packages/manager-react-components/src/components/drawer/Drawer.spec.tsx
mathieumousnier: You pointed the line 39 with "drawer-dismiss-button" :)
I might change the tests of the other buttons by querying labels

packages/manager-react-components/src/components/drawer/Drawer.spec.tsx
mathieumousnier: As the comment says : 
```
// Handle Escape key press to hide the handle
```

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
mathieumousnier: "when I have a collapsible drawer I want it to always be displayed and collapsed by default".
I see what you mean, but this is not what this drawer has been designed for. To be honest this component is not totally "standard" and might be subject to many discussions, as it already has been :)

The use case that aims to be filled here is a drawer, but with the possibility to access the content on the page below. The main example would be a form related to some information on the page.

I feel that what you propose would be a very different use case.

If you want more information I can link you with Pauline who designed the Drawer.
And unfortunately I don't have a "real like" example, I only need the classic version for my project.

packages/manager-react-components/src/components/drawer/Drawer.spec.tsx
mathieumousnier: I'll look into adding a test for it

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
SimonChaumet: What I propose is what is already here : 2 use cases, so 2 components, to avoid having weird usages and a lot of code that isn't used like there is on current MRC components

packages/manager-react-components/src/components/drawer/Drawer.spec.tsx
SimonChaumet: Line 43 like I said

mathieumousnier:
> Do we need the attribute 'onOpen' ? as we have defined in the ticket

@aboungnaseng-ovhcloud 
I have a few arguments against it:

- I don't really see what this `onOpen` callback would be for.
- This is not on the modal component, that has a very close API and similar use cases
- I tried to wire direcly a `onOpen` callback on the `odsOnOpen` props of` OdsDrawer` but it does not work. It only works if we control the `OdsDrawer` with a ref and open it with `ref.current.open()`
- I would need to trigger this on a useEffect on `isOpen`, something I would prefered not to do for a feature that has no identified use case

So I would advise on not adding it.

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
mathieumousnier: See fixups

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
mathieumousnier: See fixups

packages/manager-react-components/src/components/drawer/Drawer.spec.tsx
mathieumousnier: See fixups

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
mathieumousnier: See fixups

SimonChaumet (dismissed):

packages/manager-react-components/src/components/drawer/Drawer.component.tsx
```diff
@@ -0,0 +1,14 @@
...
+import { DrawerBase, DrawerBaseProps } from './DrawerBase.component';
+import './translations';
+
+export type DrawerProps = Omit<DrawerBaseProps, 'className'>;
+
+export const Drawer = (props: DrawerProps) => {
+  return (
+    <div className="relative bg-transparent">
+      <DrawerBase {...props} />
+      {props.isOpen && <DrawerBackdrop onClick={props.onDismiss} />}
```
anooparveti, line 11:
```suggestion
      {props?.isOpen && <DrawerBackdrop onClick={props.onDismiss} />}
```

packages/manager-react-components/src/components/drawer/DrawerBase.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+} from '@ovhcloud/ods-components';
+import clsx from 'clsx';
+import { useTranslation } from 'react-i18next';
+import './translations';
+
+export type DrawerBaseProps = PropsWithChildren & {
+  /** Drawer heading */
+  heading: string;
+  /** Open/close the drawer */
+  isOpen: boolean;
```
anooparveti:
```suggestion
  isOpen?: boolean;
```

packages/manager-react-components/src/components/drawer/DrawerBase.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+  /** Called when the drawer is closed internally */
```
anooparveti:
```suggestion
  /** Callback function to be called on close of the Drawer */
```

packages/manager-react-components/src/components/drawer/DrawerBase.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+  const hasFooter = hasPrimaryButton || hasSecondaryButton;
+
+  return (
+    <OdsDrawer
+      data-testid="drawer"
+      isOpen={isOpen}
+      position={'right'}
+      onOdsClose={onDismiss}
+      className={clsx(
+        '[&::part(drawer)]:w-[456px] [&::part(drawer)]:p-0 [&::part(drawer)]:duration-300',
```
anooparveti: **question:** Why should it always be 456px. Should not it be responsive?

packages/manager-react-components/src/components/drawer/DrawerBase.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+        )}
+
+        {!isLoading && (
+          <>
+            <section className="px-[24px] flex-1 overflow-y-auto outline-none">
+              {children}
+            </section>
+
+            {hasFooter && (
+              <footer className="px-[24px] py-[24px] space-x-[8px]">
```
anooparveti: **suggestion:** Can be simplified.
```suggestion
              <footer className="p-[24px] space-x-[8px]">
```

packages/manager-react-components/src/components/drawer/DrawerBase.component.tsx
```diff
@@ -0,0 +1,145 @@
...
+  isSecondaryButtonLoading,
+  isSecondaryButtonDisabled,
+  onSecondaryButtonClick,
+  secondaryButtonLabel,
+  className,
+}: DrawerBaseProps) => {
+  const { t } = useTranslation('drawer');
+  const hasPrimaryButton = Boolean(primaryButtonLabel);
+  const hasSecondaryButton = Boolean(secondaryButtonLabel);
+  const hasFooter = hasPrimaryButton || hasSecondaryButton;
```
anooparveti: **nitpick:** Unnecessary variable can be directly put in TSX as it is only used in once.

packages/manager-react-components/src/components/drawer/DrawerBase.component.tsx
mathieumousnier: Could you give me some details on this ?
It feels a bit weird to me to make isOpen optional as you can't open the drawer in any other way.
To say it in an other way : Setting isOpen as optional gives the signal that you can use the drawer without it, but you can't

packages/manager-react-components/src/components/drawer/DrawerBase.component.tsx
mathieumousnier: Good question.
OdsDrawer doesn't do anything about responsiness, we could argue that this should be managed there.

I might set the witdh as `w-full` on `sm` screen,s what do you think ?

mathieumousnier: I managed to do this, I think it looks good. 

https://github.com/user-attachments/assets/fb391df0-26c4-42f0-9670-33aabf0264a0

packages/manager-react-components/src/components/drawer/DrawerBackdrop.component.tsx
```diff
@@ -0,0 +1,15 @@
+type DrawerBackdropProps = {
+  onClick: () => void;
+};
+
+const DrawerBackdrop = ({ onClick }: DrawerBackdropProps) => {
+  return (
+    <div
+      data-testid="drawer-backdrop"
+      className="fixed inset-0 bg-[#0050d7] z-40 drawer-backdrop-fade-in"
+      onClick={onClick}
```
helabenkhalfallah, line 16: You should use button instead of div, div is not a clickable element, this is not good semantically and accessibly.

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+    };
+
+    document.addEventListener('keydown', handleEscapeKey);
+
+    return () => document.removeEventListener('keydown', handleEscapeKey);
+  }, [collapseState]);
+
+  return (
+    <div
+      className={clsx(
```
helabenkhalfallah, line 33: why tooo many nested div? each div is an html object, this will affect rendering performance.

packages/manager-react-components/src/components/drawer/DrawerBackdrop.component.tsx
helabenkhalfallah: Logically backdrops are interactive. Without a role or aria-* attributes, screen readers may ignore this element, 

```
role="button" 
aria-label="Close drawer"
tabIndex={0} 
onKeyDown={(e) => { if (e.key === 'Enter' || e.key === ' ') onClick(); }}
```

packages/manager-react-components/src/components/drawer/DrawerCollapsible.component.tsx
```diff
@@ -0,0 +1,36 @@
...
+      prevState === 'visible' ? 'hidden' : 'visible',
+    );
+  };
+
+  return (
+    <div className="relative bg-transparent">
+      <DrawerBase
+        {...props}
+        className={
+          collapseState === 'hidden' && `[&::part(drawer)]:translate-x-[456px]`
```
helabenkhalfallah: I think this may result in false being passed to className when the state is not 'hidden', which can break class merging. To verify.

helabenkhalfallah: translate-x-[456px] is a magic number and we should avoid px as possible.

packages/manager-react-components/src/components/drawer/DrawerCollapsible.component.tsx
```diff
@@ -0,0 +1,36 @@
+import { useState } from 'react';
+import DrawerHandle from './DrawerHandle.component';
+import { DrawerBase, DrawerBaseProps } from './DrawerBase.component';
+import { DrawerCollapseState } from './Drawer.types';
+import './translations';
+
+export type DrawerCollapsibleProps = Omit<DrawerBaseProps, 'className'>;
+
+export const DrawerCollapsible = (props: DrawerCollapsibleProps) => {
+  const [collapseState, setCollapsedState] =
```
helabenkhalfallah: setCollapsedState → setCollapseState for consistency with the variable name collapseState.

packages/manager-react-components/src/components/drawer/DrawerCollapsible.component.tsx
```diff
@@ -0,0 +1,36 @@
...
+    useState<DrawerCollapseState>('visible');
+
+  const handleToggleCollapseState = () => {
```
helabenkhalfallah, line 14: O have a doubt about this, we should not use `useCallback`?

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+          'rounded-l-lg absolute justify-center items-center',
+          // Hide the handle on small screens when the drawer becomes w-full
+          collapseState === 'visible' ? 'hidden sm:flex' : 'flex',
+        )}
+        style={{
+          // Apply ODS Drawer shadow style
+          boxShadow: 'rgba(0, 14, 156, 0.2) 0px 2px 8px 0px',
+          // Remove shadow on right side to avoid overlap with the drawer
+          clipPath: 'inset(-15px 0px -15px -15px)',
+        }}
```
helabenkhalfallah: magic numbers: top-[100px], right-[456px], w-[56px], h-[56px]

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+            label={undefined}
+            aria-label={
+              collapseState === 'visible' ? t('collapse') : t('expand')
+            }
+            icon={
+              collapseState === 'visible'
+                ? 'chevron-double-right'
+                : 'chevron-double-left'
+            }
+            onClick={onClick}
```
helabenkhalfallah, line 64: If onClick is passed down from a parent and triggers unnecessary re-renders in OdsButton, we should wrap it in useCallback at the parent level.

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+      <div
+        className={clsx(
+          'right-0 w-[56px] h-[56px] bg-white',
+          'rounded-l-lg absolute justify-center items-center',
+          // Hide the handle on small screens when the drawer becomes w-full
+          collapseState === 'visible' ? 'hidden sm:flex' : 'flex',
+        )}
+        style={{
+          // Apply ODS Drawer shadow style
+          boxShadow: 'rgba(0, 14, 156, 0.2) 0px 2px 8px 0px',
```
helabenkhalfallah: inline style is not recommended. The inline boxShadow and clipPath could be extracted into a utility class or CSS module to improve readability and styling consistency.

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+
+const DrawerHandle = ({ onClick, collapseState }: DrawerHandleProps) => {
+  const { t } = useTranslation('drawer');
+  const [isDrawerSlideInFinished, setIsDrawerSlideInFinished] = useState(false);
+  const [hasEscapeBeenPressed, setHasEscapeBeenPressed] = useState(false);
+
+  // Wait for the drawer to slide in before showing the handle with a fade-in effect
+  // This is to ensure the handle is not visible during the initial slide-in animation
+  useEffect(() => {
+    const timer = setTimeout(() => setIsDrawerSlideInFinished(true), 300);
```
helabenkhalfallah: how do you ensure that 300 is enough waiting time?

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
```diff
@@ -0,0 +1,84 @@
...
+import './translations';
+
+type DrawerHandleProps = {
+  onClick: () => void;
+  collapseState: DrawerCollapseState;
+};
+
+const DrawerHandle = ({ onClick, collapseState }: DrawerHandleProps) => {
+  const { t } = useTranslation('drawer');
+  const [isDrawerSlideInFinished, setIsDrawerSlideInFinished] = useState(false);
```
helabenkhalfallah: this state cannot be derivate from collapseState?

mathieumousnier: I would have liked it but no, it doesn't serve the same purpose.
I had to add `isDrawerSlideInFinished` to avoid showing the handle during the drawer appearance.
Maybe I can do this with CSS I'll give it a try

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
mathieumousnier: I can't.
This is the limit of hacking features on a component coming from another library.
This is the case for several things here, where I had to mimic some OdsDrawer sizes and behaviors.
Here I have to wait for the slidein animation to finish, if I set it in CSS (as mentionned in another comment) in would still be a random value that would be obsolete if it changes on ODS side

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
mathieumousnier: I don't really agree.
On a normal use of this, if `onClick` changes it is because `DrawerHandle` has re-rendered.
If `DrawerHandle` re-renders then `OdsButton` re-renders, no `useCallback` will change this.

packages/manager-react-components/src/components/drawer/DrawerCollapsible.component.tsx
mathieumousnier: There is no need of useCallback here.
`useCallback` aims to stabilize functions passed to components memoized with the `memo` function.
If you add `useCallback` here it won't avoid any re-render.

The only thing it is going to do is to avoid the function to be re-defined on every `DrawerCollapsible` render. This a very small gain, it is so small it is smaller that what you lose by calling useCallback itself for caching the function.

https://kentcdodds.com/blog/usememo-and-usecallback

packages/manager-react-components/src/components/drawer/DrawerBackdrop.component.tsx
mathieumousnier: I am really not an expert in accessibility so I am eager to get any feedback, but I felt a bit a weird to use a `button` tag on a fullscreen overlay so I checked a few libs, and for the 3 bellow I can't see any button. 

https://ui.shadcn.com/docs/components/drawer
https://chakra-ui.com/docs/components/drawer
https://mui.com/material-ui/react-drawer/

Also, any IA I ask for this tells me not to use a button, because it is semantically inaccurate : 
> "If you use a `button`, it may confuse screen readers or other assistive technologies, as it implies that the overlay itself is a button that can be interacted with in a way that is not intended."

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
mathieumousnier: So, I changed this for a css only version.
It does not fix that the 300ms is a hard value that I replicate to mimic ods, but it is way better than this setTimeOut horror

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
mathieumousnier: I declared a css class

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
mathieumousnier: I used css variables for all this values

packages/manager-react-components/src/components/drawer/DrawerCollapsible.component.tsx
mathieumousnier: 1. I used clsx to resolve class merging properly
2. I moves the value in a css variable

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
mathieumousnier: On those 3 divs, the 1st (the parent) is for positioning, the 2nd is for styling, the 3rd is because without it the centering of the `OdsButton` does not work.
I agree we should not declare useless divs.
But here, I don't think a div will affect performance. There can be only 1 drawer per page, not 1000. And removing on of those divs is not that easy

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
mathieumousnier: I found a we better solution with only css

mathieumousnier:
## PR update
I squashed some commits and let only one fixup with all the changes relative to @anooparveti  and @helabenkhalfallah's feedbacks.

- [x] Add a readme file to explain the implementation and how to add features
- [x] Use css variables for element sizes
- [x] Replace all pixels values with tailwind tokens or em values
- [x] Use a `button` tag for the backdrop and + `aria-label`
- [x] Replace some state + tailwind with a css solution for the `DrawerHandle` appearance
- [x] Make a test file for `Drawer`, `DrawerCollapsible` and `DrawerBase` and factorize the mocks
- [x] Make isOpen optional and default to true
- [x] Make the drawer full-width on small screens
- [x] Use 1 div instead of 3 for the `DrawerHandle` : Couldn't do anything for this last month. @helabenkhalfallah if you want to give it a try you're welcome :) (edit: this PR follows the topic https://github.com/ovh/manager/issues/17885)

packages/manager-react-components/src/components/drawer/DrawerBase.component.tsx
```diff
@@ -0,0 +1,142 @@
...
+  isSecondaryButtonLoading,
+  isSecondaryButtonDisabled,
+  onSecondaryButtonClick,
+  secondaryButtonLabel,
+  className,
+}: DrawerBaseProps) => {
+  const { t } = useTranslation('drawer');
+
+  return (
+    <OdsDrawer
```
helabenkhalfallah, line 65: The component is some bit long, we cannot split it on: 
-> drawer header
-> drawer content
-> drawer footer

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  // Handle Escape key press to hide the handle
+  useEffect(() => {
+    const handleEscapeKey = (event: KeyboardEvent) => {
+      if (event.key === 'Escape' && collapseState === 'visible') {
+        setHasEscapeBeenPressed(true);
+      }
+    };
+
+    document.addEventListener('keydown', handleEscapeKey);
```
anooparveti: **nitpick** Including Escape in the function is not appropriate as we are listening to a generic keyDown. And in future if we want to handle some other key press along with Escape key, it complicates to name the function.
```suggestion
    document.addEventListener('keydown', onKeyDown);
```

packages/manager-react-components/src/components/drawer/DrawerBase.component.tsx
mathieumousnier: As we said in a meeting, we keep it like this for simplicity

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
mathieumousnier: I created this issue for future improvements: https://github.com/orgs/ovh/projects/22/views/1?pane=issue&itemId=117786309&issue=ovh%7Cmanager%7C17885

packages/manager-react-components/src/components/drawer/DrawerHandle.component.tsx
mathieumousnier: done

anooparveti (approved):

PR #17477, opened 2025-06-06
fix(mrc): fix onlick event on tag list

mhelhali-soufien:
ref: #MANAGER-18147

## Description

Ticket Reference: #...


packages/manager-react-components/src/components/tags-list/tags-list.component.tsx
```diff
@@ -86,7 +86,10 @@ export const TagsList: React.FC<TagsListProps> = ({
           <OdsLink
             href="#"
             className="text-xs"
-            onClick={onClick}
+            onClick={(e) => {
+              if (onClick) onClick();
+              e.preventDefault();
```
aboungnaseng-ovhcloud, line 91:
```suggestion
    else {
        e.preventDefault();
     }
```

mhelhali-soufien: no need of else here, should block the link every time

aboungnaseng-ovhcloud (approved):

helabenkhalfallah (approved):

PR #17480, opened 2025-06-06
fix(pci-instances): Sync UI design with ODS 18

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-4146

### DIFF
--- a/packages/manager/apps/pci-instances/src/components/modal/Modal.component.tsx
+++ b/packages/manager/apps/pci-instances/src/components/modal/Modal.component.tsx
@@ -18,6 +18,8 @@ import {
 } from '@ovhcloud/ods-components';
 import { ODS_THEME_COLOR_INTENT } from '@ovhcloud/ods-common-theming';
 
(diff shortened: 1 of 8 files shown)


SimonChaumet (approved):
PR is good but you have a typo in your commit

packages/manager/apps/pci-instances/src/components/modal/Modal.component.tsx
```diff
@@ -18,6 +18,8 @@ import {
 } from '@ovhcloud/ods-components';
 import { ODS_THEME_COLOR_INTENT } from '@ovhcloud/ods-common-theming';
 
+export type TVariant = 'primary' | 'warning' | 'critical';
```
fredericvilcot: As it's exported, it's better to enhance naming, because `TVariant` out of context means nothing.
```suggestion
export type TModalVariant = 'primary' | 'warning' | 'critical';
```

Tsiorifamonjena: You're right 🙏🏽

PR #17527, opened 2025-06-10
feat(manager-config): added user related types

JacquesLarique:
## Description

Added types related to user and account creation / edition

Ticket Reference: #MANAGER-17333


packages/manager/modules/config/src/environment/user/legalForm.ts
```diff
@@ -0,0 +1,7 @@
+export type LegalFrom =
```
fredericvilcot: Guess you have a typo here ;)

```suggestion
export type LegalForm =
```

packages/manager/modules/config/src/locale/country-code.enum.ts
```diff
@@ -22,4 +22,287 @@ export enum CountryCode {
...
+  | 'MA'
+  | 'SN'
+  | 'AU'
+  | 'ASIA'
+  | 'SG'
+  | 'IN'
+  | 'EU'
+  | 'WE';
+
+export type Country =
```
SimonChaumet, line 52: How is this different from CountryCode ?

packages/manager/modules/config/src/locale/country-code.enum.ts
```diff
@@ -22,4 +22,287 @@ export enum CountryCode {
   WS = 'WS',
 }
 
+export type Subsidiary =
```
SimonChaumet, line 25: Should this replace OvhSubsidiary from MRC ?

jperchoc (approved):

packages/manager/modules/config/src/locale/country-code.enum.ts
JacquesLarique: This is an exhaustive list of countries that user can configure their account with

packages/manager/modules/config/src/locale/country-code.enum.ts
JacquesLarique: No as this one is oriented to the user account configuration
The one on MRC is more oriented to links management (having a `DEFAULT` value that does not exist on API side)

MaximeBajeux (approved):

jperchoc (approved):

PR #17529, opened 2025-06-10
fix(storage): use enum to not duplicate storage

lionel95200x:
## Description

- Fix duplicate typing of storage

### DIFF
--- a/packages/manager/apps/pci-object-storage/src/constants.ts
+++ b/packages/manager/apps/pci-object-storage/src/constants.ts
@@ -468,11 +468,6 @@ export const CONTAINER_USER_ASSOCIATION_MODES = {
   LINKED: 'LINKED',
   CREATE: 'CREATE',
 };
-export const OBJECT_CONTAINER_STORAGE_CLASS = {
-  STANDARD: 'STANDARD',
-  HIGH_PERFORMANCE: 'HIGH_PERF',
-  STANDARD_INFREQUENT_ACCESS: 'STANDARD_IA',
-};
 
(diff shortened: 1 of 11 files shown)


packages/manager/apps/pci-object-storage/src/constants.ts
```diff
@@ -506,6 +501,8 @@ export const UNIVERSE = 'PublicCloud';
 export const SUB_UNIVERSE = 'storages';
 export const APP_NAME = 'objects_storage_container';
 
-export const STANDARD = 'STANDARD';
-export const HIGH_PERFORMANCE = 'HIGH_PERF';
-export const STANDARD_INFREQUENT_ACCESS = 'STANDARD_IA';
+export enum TReplicationStorageEnum {
```
fredericvilcot: Please remove the `T` prefix as it's not a type (and the `enum` keyword)

```suggestion
export enum ReplicationStorageClass {
```

lionel95200x: Yes of for the `T` but for the class why keep it as it's not a class also ?

fredericvilcot: As per contract, a `TObject` has a `storageClass` property: 

```ts
replication: {
    rules: {
      id: string;
      status: 'enabled' | 'disabled';
      priority: number;
       destination: {
         name: string;
         region: string;
         storageClass: ReplicationStorageEnum;
       };
...
```

So, it seems to be a `ReplicationStorageClass`, no?

PR #17552, opened 2025-06-11
fix(dedicated-servers): fix cluster tab and onboarding

mhelhali-soufien:
ref: #MANAGER-18609

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/dedicated-servers/src/hooks/useLinkUtils.tsx
+++ b/packages/manager/apps/dedicated-servers/src/hooks/useLinkUtils.tsx
@@ -16,7 +16,8 @@ function useLinkUtils<T extends string | UrlEntry | UrlLinks | LinkEntry>(
     const getSubSidiary = async () => {
       const env = await environment.getEnvironment();
       const { ovhSubsidiary } = env.getUser();
-      const localLinks = dictionaryLinks[ovhSubsidiary];
+      const localLinks =
+        dictionaryLinks[ovhSubsidiary] || dictionaryLinks.DEFAULT;
       setLinks(localLinks);
     };
     getSubSidiary();
--- a/packages/manager/apps/dedicated-servers/src/pages/listing/index.tsx
+++ b/packages/manager/apps/dedicated-servers/src/pages/listing/index.tsx
@@ -16,14 +16,17 @@ import {
   ChangelogButton,
   BaseLayout,
   useFeatureAvailability,
+  useResourcesIcebergV6,
 } from '@ovh-ux/manager-react-components';
 import { OdsTabs, OdsTab } from '@ovhcloud/ods-components/react';
 import { useTranslation } from 'react-i18next';
 import { CHANGELOG_LINKS } from '@/data/constants/changelogLinks';
 import { urls } from '@/routes/routes.constant';
+import { Cluster } from '@/data/types/cluster.type';
 
 export default function Layout() {
   const { t } = useTranslation('dedicated-servers');
+  const [hasCluster, setHasCluster] = useState(false);
   const location = useLocation();
   const { shell } = useContext(ShellContext);
   const matches = useMatches();
@@ -31,6 +34,12 @@ export default function Layout() {
   const { data: features, isSuccess } = useFeatureAvailability([
     'dedicated-server:cluster',
   ]);
+  const { flattenData, isSuccess: isSuccessCluster } = useResourcesIcebergV6<
+    Cluster[]
+  >({
+    route: `/dedicated/cluster`,
+    queryKey: ['dedicated-servers', `/dedicated/cluster`],
+  });
 
   const [activePanel, setActivePanel] = useState('');
   const navigate = useNavigate();
@@ -49,6 +58,15 @@ export default function Layout() {
   useEffect(() => {
     shell.ux.hidePreloader();
   }, []);
+
+  useEffect(() => {
+    setHasCluster(
+      isSuccess &&
+        features?.['dedicated-server:cluster'] &&
+        isSuccessCluster &&
+        flattenData?.length !== 0,
+    );
+  }, [isSuccessCluster, isSuccess]);
 
   useEffect(() => {
     if (!location.pathname) {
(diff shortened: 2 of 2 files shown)


qpavy (approved):

packages/manager/apps/dedicated-servers/src/pages/listing/index.tsx
```diff
@@ -16,21 +16,30 @@ import {
...
   const { trackCurrentPage } = useOvhTracking();
   const { data: features, isSuccess } = useFeatureAvailability([
     'dedicated-server:cluster',
   ]);
+  const { flattenData, isSuccess: isSuccessCluster } = useResourcesIcebergV6<
+    Cluster[]
+  >({
+    route: `/dedicated/cluster`,
+    queryKey: ['dedicated-servers', `/dedicated/cluster`],
+  });
```
sachinrameshn: Can you create a hook for this under hooks folder and also get Query key and Query function from api folder ?

packages/manager/apps/dedicated-servers/src/pages/listing/index.tsx
```diff
@@ -50,6 +59,15 @@ export default function Layout() {
...
 
+  useEffect(() => {
+    setHasCluster(
+      isSuccess &&
+        features?.['dedicated-server:cluster'] &&
+        isSuccessCluster &&
+        flattenData?.length !== 0,
+    );
+  }, [isSuccessCluster, isSuccess]);
+
```
sachinrameshn:
```suggestion
  useEffect(() => {
    const hasClusterFeature = features?.['dedicated-server:cluster'];
    const hasClusterData = isSuccess && isSuccessCluster && flattenData?.length > 0;
    setHasCluster(hasClusterFeature && hasClusterData);
  }, [isSuccessCluster, isSuccess, features, flattenData]);
```

packages/manager/apps/dedicated-servers/src/hooks/useCluster.tsx
```diff
@@ -0,0 +1,13 @@
...
+
+export const useCluster = () => {
+  return useQuery<string[], ApiError>({
+    queryKey: getClusterQueryKey,
+    queryFn: async () => {
+      const { data } = await getClusters();
+      return data;
+    },
+  });
+};
```
chipp972, line 13: why do you return only data ?
You may need to know if it's loading, there is an error etc.

packages/manager/apps/dedicated-servers/src/pages/listing/index.tsx
```diff
@@ -50,6 +54,12 @@ export default function Layout() {
     shell.ux.hidePreloader();
   }, []);
 
+  useEffect(() => {
+    const hasClusterFeature = features?.['dedicated-server:cluster'];
```
chipp972: You can pass this in the enabled parameter in the useCluster hook

packages/manager/apps/dedicated-servers/src/pages/listing/index.tsx
```diff
@@ -50,6 +54,12 @@ export default function Layout() {
...
+    const hasClusterData = isSuccess && isSuccessCluster && data?.length > 0;
+    setHasCluster(hasClusterFeature && hasClusterData);
+  }, [isSuccessCluster, isSuccess, features, data]);
```
chipp972: no need to create a new state `hasCluster`, just put the variables in the render

sachinrameshn: As features and data are objects this could cause unnecessary re-renders so relaying on its properties
```suggestion
  }, features?.['dedicated-server:cluster'], data?.length]);
```

mhelhali-soufien: this is why before I use just the status of call , success and you suggest to add the object

mhelhali-soufien: just tobe more clear

packages/manager/apps/dedicated-servers/src/pages/listing/index.tsx
mhelhali-soufien: I dont like to link the useCluster hook with feature hook

packages/manager/apps/dedicated-servers/src/pages/listing/index.tsx
sachinrameshn: yup can you remove it please.

packages/manager/apps/dedicated-servers/src/pages/listing/index.tsx
chipp972: Why not ?
It is what hooks are usefull for : abstract a bunch of logic and queries
And additionally it is actually linked : you shouldn't query clusters if the feature is not available

packages/manager/apps/dedicated-servers/src/pages/listing/index.tsx
chipp972: Duplicating state is not more clear
You should use `useMemo` instead

packages/manager/apps/dedicated-servers/src/hooks/useLinkUtils.tsx
```diff
@@ -16,7 +16,8 @@ function useLinkUtils<T extends string | UrlEntry | UrlLinks | LinkEntry>(
     const getSubSidiary = async () => {
       const env = await environment.getEnvironment();
       const { ovhSubsidiary } = env.getUser();
-      const localLinks = dictionaryLinks[ovhSubsidiary];
+      const localLinks =
+        dictionaryLinks[ovhSubsidiary] || dictionaryLinks.DEFAULT;
```
chipp972, line 20: We can use the synchronous version `environment.user.ovhSubsidiary` instead of doing all this in a `useEffect`

packages/manager/apps/dedicated-servers/src/pages/listing/index.tsx
```diff
@@ -93,22 +95,25 @@ export default function Layout() {
...
-                  className="no-underline"
-                >
-                  <OdsTab
-                    id={'cluster-tab'}
-                    role="tab"
-                    isSelected={activePanel === 'cluster'}
-                    isDisabled={false}
+              {isSuccess &&
+                features?.['dedicated-server:cluster'] &&
+                isSuccessCluster &&
```
chipp972:
```suggestion
              {isSuccessCluster &&
```

PR #17553, opened 2025-06-11
feat(dedicated.vps): hide error message on listing page when a vps is expired

carsso:
### VPS Listing page :
Hide error message related to location when a VPS is expired, as this error is not relevant.

Error message that will be hidden when service is expired : 

## Additional Information
Ticket Reference: MANAGER-18613


ghyenne (dismissed):

atefzed (dismissed):

carsso:
Rebase as requested because of 2API drop

ghyenne (approved):

PR #17569, opened 2025-06-11
feat(common-translations): add new translations

chipp972:
ref: #MANAGER-17549

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/modules/common-translations/public/translations/actions/Messages_fr_FR.json
+++ b/packages/manager/modules/common-translations/public/translations/actions/Messages_fr_FR.json
@@ -48,5 +48,6 @@
   "retry": "Réessayer",
   "start": "Commencer",
   "renew": "Renouveler",
-  "restore": "Restaurer"
+  "restore": "Restaurer",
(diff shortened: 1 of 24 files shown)


packages/manager/modules/common-translations/public/translations/actions/Messages_fr_FR.json
anooparveti, line 1: **question:** May i know the intention behind adding the placeholders?

sachinrameshn: May be its needed to handle few of the common translations like Ex: Supprimer Additional IP, Supprimer VCD

chipp972: Yes exactly
It is a solution but we can also create a new string each time `${t('delete')} ${t('subnet')}`

anooparveti (approved):

helabenkhalfallah (dismissed):

aboungnaseng-ovhcloud (approved):

Issue #17214, opened 2025-05-20
[manager-components]: Last item of breadcrumb not updated when collapsed


tibs245:
### Have you already checked if a similar item is present on manager-components?

- [x] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

When the breadcrumb is collapsed and modal is opened and closed. The last items is not opened.

We can add the last notion on key as this next breadcrumb on vcd. With this notion this force redender item when state is changed

```tsx
import React from 'react';
import {
  OdsBreadcrumb,
  OdsBreadcrumbItem,
} from '@ovhcloud/ods-components/react';
import {
  useBreadcrumb,
  BreadcrumbItem,
} from '@/hooks/breadcrumb/useBreadcrumb';
import appConfig from '@/hpc-vmware-managed-vcd.config';

export interface BreadcrumbProps {
  customRootLabel?: string;
  appName?: string;
  items?: BreadcrumbItem[];
}

function Breadcrumb({ customRootLabel, items }: BreadcrumbProps): JSX.Element {
  const rootLabel = customRootLabel || appConfig.rootLabel;

  const breadcrumbItems = useBreadcrumb({ rootLabel, items });

  // The key is used to force re-render the last item component when his state change
  return (
    <OdsBreadcrumb>
      {breadcrumbItems.map(({ label, ...props }, index) => (
        <OdsBreadcrumbItem
          key={`${label}-${index + 1 === breadcrumbItems.length ? 'last' : ''}`}
          label={label}
          {...props}
        />
      ))}
    </OdsBreadcrumb>
  );
}

export default Breadcrumb;
```

Also I think we can simplify the breadcrumb hook code with : 

```tsx
import { useLocation, useHref } from 'react-router-dom';
import { JSX } from '@ovhcloud/ods-components';
import { urls } from '@/routes/routes.constant';

export type BreadcrumbItem = {
  id: string;
  label: string | undefined;
  href?: string;
};

export interface BreadcrumbProps {
  rootLabel?: string;
  items?: BreadcrumbItem[];
}

export const useBreadcrumb = ({ rootLabel, items }: BreadcrumbProps) => {
  const location = useLocation();
  const rootHref = useHref(urls.root);

  const rootItem = {
    id: rootLabel,
    label: rootLabel,
    href: rootHref,
  };

  const pathnames = location.pathname.split('/').filter(Boolean);
  const pathsTab = pathnames.map((value, index) => {
[...]

PR #17573, opened 2025-06-11
feat(sap): update breadcrumb and routes navigation

pauldkn:
ref: #MANAGER-17751

## Description

Update use of params to have the correct navigation : 
- replace `params` by `searchParams` for `serviceName` and `taskId` (installation report & form)

Update breadcrumb labels and application routes : 
- `/` : dashboard
- `/installations` : listing
- `/installations/report` : installation report (`serviceName` & `taskId` as searchParams)
- `/installation` : installation wizard
- `/installation/initialization` : installation form (initialization)
- `/installation/:stepId` : installation form (steps)(`serviceName` as searchParams)

Ticket Reference: #MANAGER-17751


packages/manager/apps/sap-features-hub/src/hooks/breadcrumb/useBreadcrumb.tsx
```diff
@@ -37,12 +43,17 @@ export const useBreadcrumb = ({ rootLabel, appName }: BreadcrumbProps) => {
...
+      const item = items?.find(({ id }) => id === value);
+
+      return {
+        id: item?.id ?? value,
+        label: item?.label ?? value,
+        href: item?.href ?? `/#/${appName}/${value}`,
+      };
+    });
+    setPathsItems(pathsTab);
   }, [location]);
```
tibs245: That is exactly a useMemo because it's equal to : 
```ts
  const pathnames = useMemo(() => {
    return pathnames.map((value) => {
      const item = items?.find(({ id }) => id === value);

      return {
        id: item?.id ?? value,
        label: item?.label ?? value,
        href: item?.href ?? `/#/${appName}/${value}`,
      };
    });
  }, [location]);
  ```
  
  After as we have discuted. May be not useful to have useMemo. I leave you the choice. My I think we cannot use `useEffect` and `useState` because it's just a bad practice

pauldkn: Absolutely, I fixed this

FYI, this pattern (`useEffect` + `useState`) comes from our generator's default structure.
As we discussed, `useMemo` isn’t strictly necessary given the minimal computation cost, so I removed `useEffect` and `useState` without adding `useMemo`

(maybe we can plan to update the generator for future projects)

tibs245: On genetor we should use the MRC breadcrumb. And I discutted refactor of this component on this issue : https://github.com/ovh/manager/issues/17214

I should create new ticket to discuss specifically about the refactor. But I am on it 😉

PR #17586, opened 2025-06-12
feat(pci-private-registry): update IAM modal confirm handling

lolottetheclash:
ref: #TAPC-4339

This ticket improves the behavior of the confirm button to enable/disable IAM authentication.

## Description

Ticket Reference: [#TAPC-4339](https://jira.ovhcloud.tools/browse/TAPC-4339)

### DIFF
--- a/packages/manager/apps/pci-private-registry/src/pages/IAM/IAMModal/IAMModal.tsx
+++ b/packages/manager/apps/pci-private-registry/src/pages/IAM/IAMModal/IAMModal.tsx
@@ -91,9 +91,16 @@ export const IAMModal = ({
 
   const handleQuit = (action: TRegistryAction) => () => onQuit(action);
 
-  const { handleSubmit, control } = useForm<IAMchemaType>({
+  const {
+    handleSubmit,
+    control,
+    formState: { isValid },
+  } = useForm<IAMchemaType>({
     resolver: zodResolver(confirmIAMSchema()),
-    mode: 'onSubmit',
+    mode: 'onChange',
+    defaultValues: {
+      confirmIAM: '',
+    },
   });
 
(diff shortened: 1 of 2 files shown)


packages/manager/apps/pci-private-registry/src/pages/IAM/IAMModal/IAMModal.tsx
```diff
@@ -181,11 +189,18 @@ export const IAMModal = ({
...
+        render={({
+          field: { value, onChange },
+          fieldState: { error, isDirty },
+        }) => (
           <OsdsFormField
             className="mt-6"
             data-testid="confirm-formfield"
-            error={error ? t('common_field:common_field_error_required') : ''}
+            error={
+              error && isDirty
```
SimonChaumet: It seems weird to test if input is dirty, isn't there a mode more suitable to do this ? Something like onBlur ? Or maybe with revalidateMode ?

lolottetheclash: they want the same behavior as for Terminate modal (which doesn't use react hook form): the error message is visible as soon as the 1st character is written, and the error is lifted as soon as the word “CONFIRM” is written. If I use isTouched, it will wait for the blur ...

Tsiorifamonjena (approved):

packages/manager/apps/pci-private-registry/src/pages/IAM/IAMModal/IAMModal.tsx
lolottetheclash: checked together, thanks ;)

SimonChaumet (approved):

PR #17603, opened 2025-06-12
feat(pci-project): add listing page

FabienHenon:
ref: #MANAGER-18601

## Description

Ticket Reference: #...


packages/manager/apps/pci-project/src/components/ManagerBannerText.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+  },
+};
+
+const createWrapper = (contextValue: typeof mockShellContext) => {
+  return ({ children }: { children: React.ReactNode }) => (
+    <ShellContext.Provider value={contextValue as never}>
+      {children}
+    </ShellContext.Provider>
+  );
+};
```
seven-amid: Maybe use the one in wrapperRenders.tsx

packages/manager/apps/pci-project/src/components/ManagerBannerText.tsx
```diff
@@ -0,0 +1,29 @@
...
+
+  if (!message || !message.en) {
+    return null;
+  }
+
+  const userLanguage = shell.environment.getUserLanguage() as LangId;
+
+  const localizedMessage = Object.keys(message).includes(userLanguage)
+    ? message[userLanguage]
+    : message.en;
```
seven-amid: Suggestion: `const localizedMessage = message[userLanguage] || message.en;`

packages/manager/apps/pci-project/src/data/api/projects-with-services.test.ts
```diff
@@ -0,0 +1,310 @@
...
+      project_id: 'project-1',
+      description: 'Project Alpha',
+      status: 'ok',
+      planCode: 'standard',
+      unleash: true,
+      creationDate: '2023-01-01T00:00:00Z',
+      expiration: null,
+      orderId: 123,
+      access: 'full',
+    } as unknown) as TProject,
```
SimonChaumet: Please don't cast as unknown, if an issue is raised it means that your type isn't compatible with TProject, if you only have missing properties the "as TProject" is sufficient

packages/manager/apps/pci-project/src/data/api/projects-with-services.ts
```diff
@@ -0,0 +1,70 @@
...
+ * The aggregated status is set to 'unpaid' if the service's billing lifecycle state is 'unpaid', otherwise it uses the project's status.
+ * The resulting list is sorted by project description.
+ *
+ * @returns A promise that resolves to a `FetchResultV6` containing an array of `TProjectWithService` objects,
+ *          each representing a project with its associated service and aggregated status.
+ */
+export const getProjectsWithServices = async (): Promise<FetchResultV6<
+  TProjectWithService
+>> => {
+  return Promise.all([getProjects(), getServices(), getDefaultProject()]).then(
```
SimonChaumet, line 22: Why not use useQueries and use the combine method or useMemo ?

packages/manager/apps/pci-project/src/data/api/projects-with-services.ts
```diff
@@ -0,0 +1,70 @@
...
+            },
+            [] as TProjectWithService[],
+          )
+          // Sort by isDefault first, then by description
+          .sort((a: TProjectWithService, b: TProjectWithService) => {
+            if (a.isDefault === b.isDefault) {
+              return (a.description || '').localeCompare(b.description || '');
+            }
+            return a.isDefault ? -1 : 1;
+          }),
```
SimonChaumet, line 60: Can this sort be handled by the query instead of API ?

packages/manager/apps/pci-project/src/hooks/useRedirectAfterProjectSelection.ts
```diff
@@ -0,0 +1,132 @@
...
+        setRedirectParams(
+          (stateDetail.targetParamKeys || ['params']).reduce(
+            (params, targetKey) => ({
+              ...params,
+              ...redirectTarget[targetKey],
+            }),
+            {},
+          ),
+        );
+        setRedirectExternalUrl(stateDetail.url);
```
SimonChaumet: It seems that these calls shouldn't be state related. State should be kept as minimal as possible

packages/manager/apps/pci-project/src/hooks/useRedirectAfterProjectSelection.ts
```diff
@@ -0,0 +1,132 @@
...
+  const [isRedirectExternal, setIsRedirectExternal] = useState<boolean>(false);
+  const [redirectExternalUrl, setRedirectExternalUrl] = useState<string>('');
+  const [doCheckFF, setDoCheckFF] = useState<boolean>(false);
+  const [ffToCheck, setFFToCheck] = useState<string>('');
+
+  const redirectTarget = JSON.parse(searchParams.get('target') || '{}');
+  const { category, state } = redirectTarget;
+
+  const upperCategory = category?.toUpperCase() as keyof typeof PCI_FEATURES_STATES;
+  const upperState = state?.toUpperCase() as keyof typeof PCI_FEATURES_STATES[typeof upperCategory];
```
SimonChaumet: I see a lot of errors not handled here, could you handle these somehow ?

packages/manager/apps/pci-project/src/pages/listing/Listing.tsx
```diff
@@ -147,29 +113,30 @@ export default function Listing() {
...
-      <OdsButton
-        variant={ODS_BUTTON_VARIANT.ghost}
-        size={ODS_BUTTON_SIZE.sm}
-        label="Add service"
-      >
-        <OdsIcon name={ODS_ICON_NAME.plus} />
-        Add service
-      </OdsButton>
+      {!isTrustedZoneLoading && !isTrustedZone && (
+        <OdsButton
```
SimonChaumet, line 126: This seems like a link not a button

packages/manager/apps/pci-project/src/pages/listing/datagrid-columns.tsx
```diff
@@ -0,0 +1,64 @@
...
+  t: (key: string) => string,
+  goToProject: (projectId: string) => void,
+): DatagridColumn<TProjectWithService>[] => [
+  {
+    id: 'description',
+    label: t('pci_projects_description'),
+    cell: (props: TProjectWithService) => (
+      <DataGridTextCell>
+        <OdsButton
+          onClick={() => goToProject(props.project_id)}
```
SimonChaumet: Please use a link for accessibility

packages/manager/apps/pci-project/src/setupTests.tsx
```diff
@@ -0,0 +1,113 @@
...
+  };
+});
+
+vi.mock('@tanstack/react-query', async (importOriginal) => {
+  const actual: typeof import('@tanstack/react-query') = await importOriginal();
+  return {
+    ...actual,
+    useQuery: vi.fn(),
+  };
+});
```
SimonChaumet: This should belong in tests files not in setup, because after this you will not be able to test the hooks

packages/manager/apps/pci-project/src/setupTests.tsx
```diff
@@ -0,0 +1,113 @@
...
+
+vi.mock('@ovh-ux/manager-react-components', () => ({
```
SimonChaumet, line 35: Why mock all of this and not use the original ?

packages/manager/apps/pci-project/src/data/api/projects-with-services.ts
FabienHenon: MRC encourages us to use `useIcebergV6` and `useResourcesV6`. It's the only way we can do what we need using those hooks

packages/manager/apps/pci-project/src/data/api/projects-with-services.ts
FabienHenon: I guess it would be possible for `description` but we would still have to sort in frontend for `isDefault` which is not a field a project has (API speaking)

SimonChaumet: The issue is to have the sort inside the "data/api" folder, because someone could want to have the projects without it being sorted or sorted with other parameters.

packages/manager/apps/pci-project/src/pages/listing/Listing.tsx
FabienHenon: It has always been a button so far, and that's the way the generator generates it

packages/manager/apps/pci-project/src/hooks/useRedirectAfterProjectSelection.ts
FabienHenon: Which errors? If target does not exist, category or state not in target, or invalid, no crash would occur

packages/manager/apps/pci-project/src/pages/listing/datagrid-columns.tsx
FabienHenon: I agree, that would have been better, but it complicates the code and forces me to use promises to get the link URL which is not a perfect solution either

packages/manager/apps/pci-project/src/setupTests.tsx
FabienHenon: Why not? Just setting `vi.fn()` means you can use expect on function usages, and eventually mock the return value. But if you don't mock the return value the behavior remains the same

packages/manager/apps/pci-project/src/setupTests.tsx
FabienHenon: 1. Because it's not the purpose of this component to test MRC components
2. Mainly because MRC uses ODS which uses webcomponents, which creates some unexpected behaviors in tests sometimes because of their state not reset after each test execution. It's safer to mock ODS. But we can't mock ODS if it's used by MRC. In this case we have to mock MRC instead

packages/manager/apps/pci-project/src/data/api/projects-with-services.ts
SimonChaumet: Since there are only 2 columns and a search is it really useful to use useIcebergV6 or useResourcesV6 ?

packages/manager/apps/pci-project/src/setupTests.tsx
SimonChaumet: But why not mock MRC inside each test or use a global mock folder instead ? This would allow each test to specify what mock is necessary

packages/manager/apps/pci-project/src/setupTests.tsx
SimonChaumet: Because you disable testing useQuery in the whole project while this is something we might want to test

packages/manager/apps/pci-project/src/pages/listing/datagrid-columns.tsx
SimonChaumet: Why would you have a promise ? usePCIUrl and useProjectUrl are already doing it

packages/manager/apps/pci-project/src/hooks/useRedirectAfterProjectSelection.ts
SimonChaumet: JSON.parse can cause error, unwrapping category and state from user controllable input (search params) can cause errors, upperCategory and upperState are cast while they might not be of this type

packages/manager/apps/pci-project/src/data/api/projects-with-services.ts
FabienHenon: It allows us to leverage on the search/filter/sort algorithms. I think it's useful and allows us to write less code.
For me it's worth the tradeoff of not using useQueries

packages/manager/apps/pci-project/src/data/api/projects-with-services.ts
FabienHenon: Here it's not an issue, it's a default sorting to comply with the one in production at the moment.
But sorting here is done only during the first retrieval. Then, it's put in tanstack cache (useQuery is used under the hood in `useResourcesV6`). And it does not prevent the user to use another column to sort the data

packages/manager/apps/pci-project/src/hooks/useRedirectAfterProjectSelection.ts
FabienHenon: I agree for JSON.parse and will update the code.
About the other potential errors :
- unwrapping when category or state does not exist will set them to undefined, which is fine for the rest of the code
- upper... could be indeed, casted also to undefined. However, this is checked below so that I never use them if they are undefined

packages/manager/apps/pci-project/src/pages/listing/datagrid-columns.tsx
FabienHenon: Those hooks require a projectId in the URL which I don't have here.
And they use `getURL` under the hood, which returns a promise (the one I meant in my comment)

packages/manager/apps/pci-project/src/setupTests.tsx
FabienHenon: Well, if that happens we will be able to move it. But there is no case where we need to test it in this app for the moment

packages/manager/apps/pci-project/src/setupTests.tsx
FabienHenon: Because we can use these mocks for all our tests. And if at any moment we have another use case that requires to move a mock we will move it at that moment. No need to over organize things when it's not necessary in my opinion.

packages/manager/apps/pci-project/src/data/api/projects-with-services.test.ts
FabienHenon: Done

packages/manager/apps/pci-project/src/hooks/useRedirectAfterProjectSelection.ts
FabienHenon: Done

packages/manager/apps/pci-project/src/hooks/useRedirectAfterProjectSelection.ts
FabienHenon: I made changes on the typing and exception catching

packages/manager/apps/pci-project/src/data/api/projects-with-services.test.ts
SimonChaumet: There are still 19 occurrences of this

packages/manager/apps/pci-project/src/pages/listing/datagrid-columns.tsx
SimonChaumet: You could also create a component that takes the project_id as a prop and use the hook inside

packages/manager/apps/pci-project/src/data/api/projects-with-services.test.ts
FabienHenon: Unfortunately `useResourcesV6` does not explicitly define a return type and the one "guessed" by typescript is not exact. `data`, `flattenData`, `sorting`, `error` are fields that could be undefined|null if you look at the code, but typescript does not allow it, thinking they must be defined (which makes no sense here).
Thus, I cannot define an object that would match `useResourcesV6` return type and must go through an `unknown` first.

packages/manager/apps/pci-project/src/pages/listing/datagrid-columns.tsx
FabienHenon: Done

packages/manager/apps/pci-project/src/hooks/useRedirectAfterProjectSelection.ts
```diff
@@ -0,0 +1,165 @@
...
+  [key: string]: Record<string, string> | string | undefined;
+};
+
+export type RedirectStatus = {
+  isRedirectRequired: boolean;
+  redirectUrl: (projectId: string) => Promise<string>;
+  redirect: (projectId: string) => void;
+};
+
+const useRedirectAfterProjectSelection = (): RedirectStatus => {
```
lionel95200x: This hook seems very hard to read and maintain,

There is a mix of redirection, feature flag checking

Could you split it into smaller function ?

Pass in params the activeProjects and availability to avoid recaculcing the whole hook
Split  the redirection logic 

```
if (isRedirectRequired) {
const { ffToCheck, ... } = getRedirectInformation()
}
```

packages/manager/apps/pci-project/src/hooks/useRedirectAfterProjectSelection.ts
```diff
@@ -0,0 +1,165 @@
...
+  const {
+    shell: { navigation },
+  } = useContext(ShellContext);
+  const [searchParams] = useSearchParams();
+  let redirectPath = '';
+  let redirectParams = {};
+  let isRedirectExternal = false;
+  let redirectExternalUrl = '';
+  let doCheckFF = false;
```
lionel95200x:
```suggestion
  let shouldCheckFeatureFlipping  = false;
```

packages/manager/apps/pci-project/src/pages/listing/Status.component.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { OdsBadgeColor } from '@ovhcloud/ods-components';
+import { useMemo } from 'react';
+import { useTranslation } from 'react-i18next';
+import { TProjectWithService } from '@/data/types/project.type';
+
+type StatusComponentProps = {
+  project: TProjectWithService;
+};
+
+const isUnpaidAndNotSuspended = (project: TProjectWithService): boolean =>
```
lionel95200x: Could be move it an utils files with an unit test :)

packages/manager/apps/pci-project/src/pages/listing/Status.component.tsx
```diff
@@ -0,0 +1,49 @@
...
+  project.isUnpaid && project.status !== 'suspended';
+
+export default function StatusComponent({
+  project,
+}: Readonly<StatusComponentProps>) {
+  const { t } = useTranslation('listing');
+
```
lionel95200x, line 15: Could declare only one time the variable and reuse it 
```suggestion
const isUnpaid = isUnpaidAndNotSuspended(project)
```

seven-amid: `const isUnpaidAndNotSuspended =  project.isUnpaid && project.status !== 'suspended'; `
 IMO: there is no need of function and no need to refactor it in utils file, since it's something not generic

packages/manager/apps/pci-project/src/hooks/useRedirectAfterProjectSelection.ts
FabienHenon: Done

packages/manager/apps/pci-project/src/hooks/useRedirectAfterProjectSelection.ts
FabienHenon: Done

packages/manager/apps/pci-project/src/pages/listing/Status.component.tsx
FabienHenon: Done

seven-amid (approved):

PR #17611, opened 2025-06-12
feat(*): migrate from reket to axios

aottr:
## Description

Migrate all from reket to axios

Ticket Reference: #MANAGER-18742
originally: #MANAGER-15627

### DIFF
--- a/.prettierignore
+++ b/.prettierignore
@@ -1,5 +1,6 @@
 packages/manager/apps/container
 packages/manager/apps/web-office
+packages/manager/modules/config
 custom-elements-bundle
 **/dist/
 **/node_modules/
--- a/packages/manager/apps/container/src/container/nav-reshuffle/header/Header.spec.tsx
+++ b/packages/manager/apps/container/src/container/nav-reshuffle/header/Header.spec.tsx
@@ -10,6 +10,7 @@ import HeaderComponent from './index';
 import UserAccountMenu from './user-account-menu/Content';
 import { UserLink } from './user-account-menu/UserLink';
 import { fetchProcedureStatus } from '@/data/api/procedure/procedure';
+import { ProcedureStatus } from '@/types/procedure';
 
 /**
  * Mocked Data
@@ -187,7 +188,10 @@ describe('Header.component', () => {
         vi.mocked(fetchFeatureAvailabilityData).mockResolvedValue({
           [feature]: isEnabled,
         });
-        vi.mocked(fetchProcedureStatus).mockResolvedValue({ status: 'open', ticketId: 'XXXXXX' });
+        vi.mocked(fetchProcedureStatus).mockResolvedValue({
+          status: ProcedureStatus.Open,
+          ticketId: 'XXXXXX',
+        });
 
         await act(async () => {
           render(
--- a/packages/manager/apps/container/src/types/procedure.ts
+++ b/packages/manager/apps/container/src/types/procedure.ts
@@ -3,7 +3,13 @@ export enum Procedures {
   FRAUD,
 }
 
+export enum ProcedureStatus {
+  Ok = 'ok',
+  Open = 'open',
+  Required = 'required',
+}
+
 export type Procedure = {
-  status: 'ok' | 'open' | 'required';
+  status: ProcedureStatus;
   ticketId: string;
 };
--- a/packages/manager/modules/config/src/index.ts
+++ b/packages/manager/modules/config/src/index.ts
@@ -33,6 +33,92 @@ export type ConfigurationApiResponse = {
   class?: string;
 };
 
+export type ErrorResponseData =
+  | ConfigurationApiResponse
+  | {
+      message?: string;
+      code?: string;
+      details?: unknown;
+    };
+
+const updateEnvironment = (environment: Environment, data: Environment): Environment => {
+  environment.setRegion(data.region);
+  environment.setUser(data.user);
+  environment.setApplicationURLs(data.applicationURLs);
+  environment.setUniverse(data.universe || '');
+  environment.setMessage(data.message);
+  environment.setApplications(data.applications);
+  return environment;
+};
+
+const handleUnauthorizedError = (): void => {
+  if (!isTopLevelApplication()) {
+    window.parent.postMessage({
+      id: 'ovh-auth-redirect',
+      url: `/auth?action=disconnect&onsuccess=${encodeURIComponent(window.location.href)}`,
+    });
+  }
+};
+
+const handleForbiddenError = (data: ConfigurationApiResponse): void => {
+  const region = data?.region || RESTRICTED_DEFAULTS.region;
+  const publicURL = data?.applications?.restricted?.publicURL || RESTRICTED_DEFAULTS.publicURL;
+  const targetWindow = window.top || window;
+  targetWindow.location.href = `${publicURL}?region=${region}`;
+};
+
+const handleMaintenanceError = (
+  data: ConfigurationApiResponse,
+  environment: Environment,
+): never => {
+  const errorObj = new Error('Server::InternalServerError::ApiUnreachableMaintenance');
+  Object.assign(errorObj, { error: data, environment });
+  throw errorObj;
+};
+
+const handleServerError = (
+  data: ErrorResponseData,
+  err: ApiError,
+  environment: Environment,
+): never => {
+  const errorObj = new Error('Server Error');
+  Object.assign(errorObj, {
+    error: {
+      message: data || err?.response?.statusText,
+    },
+    environment,
+  });
+  throw errorObj;
+};
+
+const handleErrorResponse = (
+  err: ApiError & { response: { data: ConfigurationApiResponse } },
+  environment: Environment,
+): Environment => {
+  const { status, data } = err?.response || {};
+
+  switch (status) {
+    case 401:
+      handleUnauthorizedError();
+      break;
+    case 403:
+      handleForbiddenError(data);
+      break;
+    case 500:
+      if (data?.class === 'Server::InternalServerError::ApiUnreachableMaintenance') {
+        handleMaintenanceError(data, environment);
+      }
+      break;
+    default:
+      if (status && status >= 500) {
+        handleServerError(data, err, environment);
+      }
+  }
+
+  environment.setRegion(HOSTNAME_REGIONS[window.location.hostname]);
+  return environment;
+};
+
 export const fetchConfiguration = async (applicationName: string): Promise<Environment> => {
   const environment = new Environment();
   let configurationURL = '/configuration';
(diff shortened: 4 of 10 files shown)


packages/manager/apps/container/src/container/legacy/server-sidebar/universe/TelecomSidebar.tsx
```diff
@@ -42,13 +42,13 @@ export default function TelecomSidebar() {
...
-          .then((result: { value: unknown }) => {
-            resolve(result?.value || false);
+        v6
+          .get<Preference>(`/me/preferences/manager/${key}`)
+          .then(({ data }) => {
+            resolve(data.value || false);
           })
-          .catch((error: { status: number }) => {
-            if (error.status === 404) {
+          .catch((error: any) => {
```
JacquesLarique: Can't this `any` be replaced by `AxiosError` ?

packages/manager/modules/config/src/index.ts
```diff
@@ -27,47 +27,44 @@ export const fetchConfiguration = async (
...
         window.parent.postMessage({
           id: 'ovh-auth-redirect',
           url: `/auth?action=disconnect&onsuccess=${encodeURIComponent(
             window.location.href,
           )}`,
         });
       }
-      if (err?.status === 403) {
-        const region = err?.data?.region || RESTRICTED_DEFAULTS.region;
+      if (err?.response?.status === 403) {
```
JacquesLarique: Shouldn't this change (acessing `status` via `response`) be replicated on lines 74 and 84 ?

packages/components/ovh-payment-method/src/ovh-payment-method.js
```diff
@@ -48,23 +55,27 @@ export const useOvhPaymentMethod = ({ reketInstance }) => {
...
       })
-      .then((response) => {
-        if (params.orderId && response.paymentMethodId) {
-          return usedReketInstance.post(`/me/order/${params.orderId}/pay`, {
-            paymentMethod: {
-              id: response.paymentMethodId,
-            },
-          });
+      .then(({ data }) => data)
+      .then((paymentMethod) => {
```
JacquesLarique:
```suggestion
      .then(({ data: paymentMethod }) => {
```

packages/manager/apps/container/src/container/legacy/server-sidebar/universe/TelecomSidebar.tsx
aottr: ah yeah, must be a mistake by the rebase

aottr: Since we have axios as peer-dependency in the container I can import it

packages/manager/modules/config/src/index.ts
```diff
@@ -27,67 +27,64 @@ export const fetchConfiguration = async (
...
 
         window.top.location.href = `${publicURL}?region=${region}`;
       }
       environment.setRegion(HOSTNAME_REGIONS[window.location.hostname]);
       if (
-        err?.status === 500 &&
-        err?.data.class ===
-          'Server::InternalServerError::ApiUnreachableMaintenance'
+        status === 500 &&
+        data.class === 'Server::InternalServerError::ApiUnreachableMaintenance'
```
JacquesLarique: data can potentially be undefined here I think

MaximeBajeux (approved):

JacquesLarique (approved):

oalkabouss (approved):

MaximeBajeux (approved):

packages/manager/apps/container/src/container/common/notifications-sidebar/NotificationsSidebar.spec.tsx
```diff
@@ -64,8 +64,7 @@ describe('NotificationsSidebar', () => {
           updating: false,
           urlDetails: { href: '' },
           level: 'HIGH',
-        }]),
-        post: postNotificationsUpdate,
+        }],
```
JacquesLarique, line 67: I think this error is because you removed the mock for the post from here:
TypeError: Cannot read properties of undefined (reading 'catch')
 ❯ updateNotications src/core/notifications/useNotifications.ts:54:39

```suggestion
        }],
        vi.mocked(aapi.post).mockImplementationOnce(postNotificationsUpdate);
```
`vi.mocked(aapi.post).mockImplementationOnce(postNotificationsUpdate);` this would then be uncessary on line 140

packages/manager/apps/container/src/container/nav-reshuffle/header/Header.spec.tsx
```diff
@@ -53,6 +51,8 @@ vi.mock('@/core/notifications/useNotifications', () => ({
   }),
 }));
 
+vi.mock('@/api/procedure/procedure');
```
JacquesLarique: I think the failing test is due to a mispelling in this mock, it should be
```suggestion
vi.mock('@/data/api/procedure/procedure');
```

packages/manager/modules/config/src/index.ts
```diff
@@ -22,87 +21,76 @@ export * from './locale/locale.constants';
...
-      environment.setApplications(config.applications);
+  return aapi
+    .get(configurationURL, {
+      headers: {
+        'Content-Type': 'application/json;charset=utf-8',
+        Accept: 'application/json',
+        ...getHeaders('/engine/2api/configuration'),
+      },
+    })
+    .then(({ data }) => {
```
JacquesLarique: You could type `data` as an `Environment` to fix these linting errors, but this is not totally correct since `data` only have the `Environment` properties (that are not functions)
It could be nice to have a `Configuration` type that better match the result from the API

MaximeBajeux (approved):

MaximeBajeux (dismissed):

MaximeBajeux (approved):

packages/manager/modules/config/src/index.ts
```diff
@@ -22,87 +22,77 @@ export * from './locale/locale.constants';
...
+      environment.setApplicationURLs(data.applicationURLs);
+      environment.setUniverse(data.universe || '');
+      environment.setMessage(data.message);
+      environment.setApplications(data.applications);
       return environment;
     })
-    .catch((exception) => {
-      const apiError = exception as ApiError;
-      if (apiError && apiError.status === 401 && !isTopLevelApplication()) {
+    .catch((err: unknown) => {
```
JacquesLarique: You should be able to type `err` as an `AxiosError<ConfigurationApiResponse>` with `ConfigurationApiResponse` being:
```typescript
  AxiosError<{
    region?: string;
    applications?: {
      restricted?: { publicURL?: string };
    };
    class?: string;
  }>;
```
Then you wouldn't need the casting at line 53

aottr: AxiosError is not available, unless I import axios just for the Error type. That's why I made a union with `response` in `ApiError` (which is basically AxiosError)

packages/manager/modules/config/src/index.ts
```diff
@@ -22,89 +23,87 @@ export * from './locale/locale.constants';
...
+        ) {
+          const errorObj = new Error('Server::InternalServerError::ApiUnreachableMaintenance');
+          Object.assign(errorObj, { error: data, environment });
+          throw errorObj;
+        }
+        if (status && status >= 500) {
+          const errorObj = new Error('Server Error');
+          Object.assign(errorObj, {
+            error: {
+              message: data.message || err?.response?.statusText,
```
JacquesLarique: Shouldn't this be `message: data || err?.response?.statusText` as previously ?

packages/manager/modules/config/src/index.ts
```diff
@@ -22,89 +23,87 @@ export * from './locale/locale.constants';
 export * from './application';
 export * from './environment';
 
-export const isTopLevelApplication = () => window.top === window.self;
+export const isTopLevelApplication = () => window?.top === window.self;
```
JacquesLarique: Why are we using optional chaining on `window` in (what seems to be) random places ?

aottr: Fair point. If window is undefined, most parts of this is not working anyway. We'll have to see if this causes any issues in tests.

packages/manager/modules/config/src/index.ts
aottr: good point, this seemed to be inconsistent to me but that comes from the error response.

packages/manager/apps/container/src/types/procedure.ts
```diff
@@ -4,6 +4,6 @@ export enum Procedures {
 }
 
 export type Procedure = {
-  status: string;
+  status: 'ok' | 'open' | 'required';
```
achrafHafedh: I think we could use an enum here for better readability and type safety, for example : 
```
enum ProcedureStatus {
  Ok = 'ok',
  Open = 'open',
  Required = 'required',
}

```

packages/manager/modules/config/src/index.ts
```diff
@@ -24,87 +25,85 @@ export * from './environment';
...
+        ...getHeaders('/engine/2api/configuration'),
+      },
+    })
+    .then(({ data }: { data: Environment }) => {
+      environment.setRegion(data.region);
+      environment.setUser(data.user);
+      environment.setApplicationURLs(data.applicationURLs);
+      environment.setUniverse(data.universe || '');
+      environment.setMessage(data.message);
+      environment.setApplications(data.applications);
```
achrafHafedh: What do you think about moving these repeated setter calls into a small helper like updateEnvironment(env, data)?
```
function updateEnvironment(env, data)  {
  env.setRegion(data.region);
  env.setUser(data.user);
  env.setApplicationURLs(data.applicationURLs);
  env.setUniverse(data.universe ?? '');
  env.setMessage(data.message);
  env.setApplications(data.applications);
}
```

packages/manager/modules/config/src/index.ts
```diff
@@ -24,87 +25,85 @@ export * from './environment';
...
+        if (status && status >= 500) {
+          const errorObj = new Error('Server Error');
+          Object.assign(errorObj, {
+            error: {
+              message: data || err?.response?.statusText,
+            },
+            environment,
+          });
+          throw errorObj;
+        }
```
achrafHafedh: We could make this code a bit more DRY by extracting the redirection and error-building logic into two reusable helper functions, something like redirectToRestrictedRegion() and buildError() ?

packages/manager/modules/config/src/index.ts
helabenkhalfallah: Here it's more builder pattern: https://refactoring.guru/design-patterns/builder

packages/manager/modules/config/src/index.ts
```diff
@@ -24,87 +25,85 @@ export * from './environment';
...
-          url: `/auth?action=disconnect&onsuccess=${encodeURIComponent(window.location.href)}`,
-        });
-      }
-      if (apiError?.status === 403) {
-        const region = apiError?.data?.region || RESTRICTED_DEFAULTS.region;
-        const publicURL =
-          apiError?.data?.applications?.restricted?.publicURL || RESTRICTED_DEFAULTS.publicURL;
-        if (window?.top) {
-          window.top.location.href = `${publicURL}?region=${region}`;
+    .catch(
```
helabenkhalfallah: too long catch, perhaps we can compose by creating small SRP functions.

packages/manager/apps/container/src/types/procedure.ts
sachinrameshn: Its done you can tale a look now

packages/manager/modules/config/src/index.ts
sachinrameshn: Its done you can tale a look now

packages/manager/modules/config/src/index.ts
```diff
@@ -24,87 +25,108 @@ export * from './environment';
...
+      break;
+    case 500:
+      if (data?.class === 'Server::InternalServerError::ApiUnreachableMaintenance') {
+        handleMaintenanceError(data, environment);
+      }
+      break;
+    default:
+      if (status && status >= 500) {
+        handleServerError(data, err, environment);
+      }
```
JacquesLarique, line 115: Isn't there any impact of stop throwing the error in these cases ?

packages/components/ovh-shell/package.json
```diff
@@ -25,13 +25,12 @@
...
-    "@ovh-ux/ovh-at-internet": "^0.24.10",
-    "@ovh-ux/ovh-reket": "^2.1.19",
-    "@ovh-ux/request-tagger": "^0.5.7",
-    "@ovh-ux/url-builder": "^2.1.7",
+    "@ovh-ux/manager-config": "^8.6.7",
+    "@ovh-ux/manager-core-api": "^0.19.0",
+    "@ovh-ux/manager-core-sso": "^0.4.6",
+    "@ovh-ux/ovh-at-internet": "^0.24.9",
+    "@ovh-ux/request-tagger": "^0.5.6",
+    "@ovh-ux/url-builder": "^2.1.6",
```
JacquesLarique: Is there a reason for these downgrades ?

.prettierignore
```diff
@@ -1,5 +1,6 @@
 packages/manager/apps/container
 packages/manager/apps/web-office
+packages/manager/modules/config
```
anooparveti, line 3: **question:** What's the need to add it in ignore list?

sachinrameshn: The config module uses an older version of Prettier, so it needs to be excluded from coverage to prevent \issue with the modern linting rules introduced during the static kit integration.

packages/manager/modules/config/src/index.ts
sachinrameshn: I don't see any issue here.  Verified with 503 error I am able to see the error been thrown properly.

packages/components/ovh-shell/package.json
sachinrameshn: I guess its been downgraded by mistake during conflict resolution. I will fix this.

PR #17619, opened 2025-06-13
test(core.sso): add unit tests

anooparveti:
## Description

Set-up testing environment using common config
Add unit tests for core.sso package
Move `__tests__` folder of url-builder inside src folder

Ticket Reference: #MANAGER-18634


packages/manager/core/url-builder/src/__tests__/url-builder.spec.ts
```diff
@@ -1,5 +1,5 @@
 import { describe, expect, it } from 'vitest';
-import { buildURL, buildURLs } from '../src';
+import { buildURL, buildURLs } from '..';
```
helabenkhalfallah: you don't think that's more clear to keep `../src` ?

packages/manager/core/sso/src/__tests__/sso.spec.ts
```diff
@@ -0,0 +1,101 @@
+import { vi, describe, beforeEach, afterEach, it, expect } from 'vitest';
+import { redirectToLoginPage, redirectToLogoutPage } from '../index';
```
helabenkhalfallah: generally we dont put index in the path

helabenkhalfallah (approved):

aboungnaseng-ovhcloud (approved):

PR #17638, opened 2025-06-13
fix(pci-instances): fix translation encodage

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-4372


SimonChaumet (approved):
trans* or translation* not trad ;)

PR #17646, opened 2025-06-16
feat(ola-nics): adding aggregation/desaggregation buttons to nics lists

aderghamov:
ref: #MANAGER-17106

## Description

This PR introduces a new button to navigate to the ola-configuration screen or to trigger a disaggragation of the current aggregation.
In additon to that, this PR improves slightly the to-be-reworked ola-configuration screen

Ticket Reference: #MANAGER-17106


chipp972 (approved):

packages/manager/modules/bm-server-components/src/network-interfaces/controller.js
```diff
@@ -1,15 +1,20 @@
+import { OLA_MODES } from '../ola/ola.constants';
 import { LABELS } from './interfaces.constants';
 
 export default class BMNetworkInterfaceController {
   /* @ngInject */
-  constructor() {
+  constructor($state) {
+    this.$state = $state;
```
qpavy: Not sure you need to inject $state here

packages/manager/modules/bm-server-components/src/ola/ola-configuration/ola-configuration.controller.js
```diff
@@ -31,20 +28,19 @@ export default class {
...
     );
-    this.allowedInterfaces = this.interfaces.filter(
-      (i) => !this.notAllowedInterfaces.includes(i),
-    );
+    this.allowedInterfaces = this.interfaces
+      .filter((i) => !this.notAllowedInterfaces.includes(i))
+      .map((nic) => ({
+        ...nic,
+        displayedMacAdresses: nic.mac ? nic.mac.split(', ') : [],
+      }));
```
qpavy: You can replace this by an unique reduce that include filter and map

packages/manager/modules/bm-server-components/src/network-interfaces/controller.js
aderghamov: Right! I will fix that.

qpavy (approved):

PR #17654, opened 2025-06-16
feat(gateway): improve quota message error

lionel95200x:
ref: #TAPC-4100

## Description

- Handle quota error message if there is an api error MaxQuotaReached

Ticket Reference: #TAPC-4100


packages/manager/apps/pci-gateway/public/translations/add/Messages_fr_FR.json
```diff
@@ -36,6 +36,7 @@
   "pci_projects_project_public_gateways_add_modal_add_private_network_error": "Une erreur est survenue lors de la création du réseau privé : {{ message }}",
   "pci_projects_project_public_gateways_add_success": "Une nouvelle Gateway a été créée avec succès.",
   "pci_projects_project_public_gateways_add_error": "Une erreur est survenue lors de la création de la Gateway: {{ message }}",
+  "pci_projects_project_public_gateways_quota_error": "Vous n'avez plus de quota disponible sur cette localisation. Pour continuer, <Link>augmenter vos quotas</Link>.",
```
lolottetheclash: just a detail here: I think it should be either: 
=> Pour continuer, <Link>augmentez vos quotas</Link>." or 
=> Pour continuer, <Link>augmenter mes quotas</Link>."

packages/manager/apps/pci-gateway/src/pages/add/LocationStep.tsx
```diff
@@ -37,7 +37,7 @@ import { useDeployments } from '@/api/hooks/useDeployments/useDeployments';
 const isRegionWith3AZ = (regions: TProductAvailabilityRegion[]) =>
   regions.some((region) => region.type === RegionType['3AZ']);
 
-const GuideLink: FC<PropsWithChildren<{ href: string }>> = ({
+export const GuideLink: FC<PropsWithChildren<{ href: string }>> = ({
```
Tsiorifamonjena: Maybe it's time to move this in components directory ? If it's ok for you and if you have time of course

packages/manager/apps/pci-gateway/src/pages/add/NetworkStep.tsx
```diff
@@ -168,11 +174,23 @@ export const NetworkStep = (): JSX.Element => {
       });
     },
     onError: (error) => {
-      addError(
-        t('pci_projects_project_public_gateways_add_error', {
-          message: error,
-        }),
-      );
+      if (error.class?.includes(TGatewayErrorClass.MaxQuotaReached)) {
```
lolottetheclash: if `error` is a `TGatewayError`, class should be defined as string so I think you don't need the `?` here.
But, are you sure that you'll always have a `class` prop for the errors?

packages/manager/apps/pci-gateway/src/api/data/gateways.ts
```diff
@@ -22,7 +31,7 @@ export const createGateway = async (
     return data;
   } catch (e) {
     const error = e as AxiosError;
-    throw new Error((error.response.data as { message: string })?.message);
+    throw error.response.data;
```
lolottetheclash: I think that response could be `undefined` here as `AxiosError` `response` is optional

packages/manager/apps/pci-gateway/src/api/data/gateways.ts
```diff
@@ -7,6 +7,15 @@ export type TNewGateway = {
   model: string;
 };
 
+export type TGatewayError = {
```
SimonChaumet: Shouldn't this be a class extending Error ?

packages/manager/apps/pci-gateway/src/api/data/gateways.ts
SimonChaumet: Maybe you should check with axios.isAxiosError(e) and verify if you had a response ?

packages/manager/apps/pci-gateway/src/api/data/gateways.ts
lionel95200x: No it's a custom error from api 
```
{ class: string, message: string }
```

Does not contain id

packages/manager/apps/pci-gateway/src/api/data/gateways.ts
lionel95200x: Could you add an example ?

lolottetheclash (dismissed):

Tsiorifamonjena (dismissed):

packages/manager/apps/pci-gateway/src/pages/add/NetworkStep.tsx
```diff
@@ -168,11 +172,23 @@ export const NetworkStep = (): JSX.Element => {
       });
     },
     onError: (error) => {
-      addError(
-        t('pci_projects_project_public_gateways_add_error', {
-          message: error,
-        }),
-      );
+      if (error.class.includes(TGatewayErrorClass.MaxQuotaReached)) {
```
fredericvilcot: `error` is any here.

As I wrote in LBAAS pull request, you should handle error object with a type guard as `AxiosError` does not guarantee you to expose a `class` property in generic `data` object

packages/manager/apps/pci-gateway/src/api/data/gateways.ts
fredericvilcot: @lionel95200x 
How could you guarantee that `data` contains `class` property, if no type guard does the check?

SimonChaumet: https://github.com/axios/axios?tab=readme-ov-file#typescript

packages/manager/apps/pci-gateway/src/api/data/gateways.ts
```diff
@@ -1,12 +1,33 @@
...
+};
+
+export const isApiErrorResponse = (
+  error: unknown,
+): error is AxiosError<CustomApiError> => {
+  const err = error as AxiosError<CustomApiError>;
+  const data = err.response?.data;
+
+  return (
+    isAxiosError(err) &&
```
SimonChaumet: You should check first if it is an axios error before casting

fredericvilcot:
I convert this PR to draft because we have decided with @lionel95200x to :

- move type guards logic that will handle errors into dedicated manager-core-api library
- move UI logic (pci banner) to pci-common library

This will guarantee separation of concerns and will prevent from code duplication.

packages/manager/apps/pci-gateway/src/api/data/gateways.ts
```diff
@@ -1,5 +1,5 @@
 import { v6 } from '@ovh-ux/manager-core-api';
-import { AxiosError } from 'axios';
+import { AxiosError, isAxiosError } from 'axios';
```
SimonChaumet: import doesn't seem to be used anymore

SimonChaumet (approved):

lolottetheclash (approved):

packages/manager/apps/pci-gateway/src/pages/add/NetworkStep.tsx
```diff
@@ -168,11 +175,28 @@ export const NetworkStep = (): JSX.Element => {
...
+                <GuideLink
+                  href={`${hrefProject}/quota`}
+                  isTargetBlank={false}
+                />
+              ),
+            }}
+          />
+        ) : (
+          t('pci_projects_project_public_gateways_add_error', {
+            message: isApiCustomError(error) ? error.response.data.message : '',
```
lolottetheclash, line 194: if error is not an apiCustomError, the message will be:
"Une erreur est survenue lors de la création de la Gateway: ", I think it could be nice to end message without `:` in this case. It 's up to you :)

SimonChaumet (approved):

packages/manager/apps/pci-gateway/src/pages/add/NetworkStep.tsx
lionel95200x: I don't know exactly when the case without message can happen, i just handle it in a type safe way but for me it's when API came down without a message
I think we should catch it globally for all our api call in a dedicated hook with a common error message (not one by app)

PR #17677, opened 2025-06-17
feat(load-balancer): improve error quota message

lionel95200x:
ref: #TAPC-4099

## Description

- Improve error quota message

Ticket Reference: #TAPC-4099

### DIFF
--- a/packages/manager/apps/pci-load-balancer/src/api/data/load-balancer.ts
+++ b/packages/manager/apps/pci-load-balancer/src/api/data/load-balancer.ts
@@ -123,17 +123,6 @@ export type TCreateLoadBalancerParam = {
   name: string;
 };
 
-export type TCreateLoadBalancerError = {
-  response: {
-    data: { class: string; message: string };
-    headers: Record<string, unknown>;
-  };
-};
-
-export enum LoadBalancerErrorClass {
-  MAX_QUOTA_REACHED = 'MaxQuotaReached',
-}
-
 export const createLoadBalancer = async ({
   projectId,
   flavor,
--- a/packages/manager/apps/pci-load-balancer/src/pages/create/hooks/useCreateCallbacks.tsx
+++ b/packages/manager/apps/pci-load-balancer/src/pages/create/hooks/useCreateCallbacks.tsx
@@ -4,15 +4,15 @@ import {
   useProjectUrl,
 } from '@ovh-ux/manager-react-components';
 import { useNavigate, useParams } from 'react-router-dom';
+import {
+  isApiCustomError,
+  isMaxQuotaReachedError,
+} from '@ovh-ux/manager-core-api';
 import { LOAD_BALANCER_CREATION_TRACKING } from '@/constants';
 import queryClient from '@/queryClient';
 import { getAllLoadBalancersQueryKey } from '@/api/hook/useLoadBalancer';
 import { useTracking } from '@/pages/create/hooks/useTracking';
 import GuideLink from '@/components/GuideLink/GuideLink.component';
-import {
-  LoadBalancerErrorClass,
-  TCreateLoadBalancerError,
-} from '@/api/data/load-balancer';
 
 export const useCreateCallbacks = () => {
   const navigate = useNavigate();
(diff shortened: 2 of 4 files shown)


packages/manager/apps/pci-load-balancer/src/pages/create/hooks/useCreateCallbacks.tsx
```diff
@@ -29,30 +37,43 @@ export const useCreateCallbacks = () => {
         queryKey: getAllLoadBalancersQueryKey(projectId),
       });
     },
-    onError: (error: ApiError) => {
+    onError: (error: TCreateLoadBalancerError) => {
```
fredericvilcot: This does not work where `useCreateCallbacks` is invoked.
AFAIS, `onError` is used in the context of a store, with `create` function:
```ts
  create: (
    flavor: TFlavor,
    onSuccess: () => void,
    onError: (cause: ApiError) => void,
  ) => Promise<void>;
```

This generates a type error.

packages/manager/apps/pci-load-balancer/src/pages/create/hooks/useCreateCallbacks.tsx
```diff
@@ -29,30 +37,43 @@ export const useCreateCallbacks = () => {
...
-            ></span>
-          )}
-        </Translation>,
-        false,
-      );
+
+      if (
+        error.response.data.class.includes(
+          LoadBalancerErrorClass.MAX_QUOTA_REACHED,
+        )
```
fredericvilcot: Well, I'm not 100% confident with this assertion.

Natively, you are handling an `AxiosError` that can be returned by the `v6.post` method.

AFAIS, this object can contain:
- an optional `response` object (here you manipulate the response as a mandatory property)
- a `data` object of generic type (`data: T`) that must be checked to be sure that `class` property exists (that is not the case here)

So, if you want to be sure to handle this structure, you should add a type guard that guarantees the good shape of the error object before using it as is.

packages/manager/apps/pci-load-balancer/src/pages/create/hooks/useCreateCallbacks.tsx
```diff
@@ -29,30 +37,43 @@ export const useCreateCallbacks = () => {
...
+              Link: <GuideLink href={`${hrefProject}/quota`} />,
+            }}
+          />,
+        );
+      } else {
+        addError(
+          <Translation ns="load-balancer">
+            {(_t) => (
+              <span
+                dangerouslySetInnerHTML={{
```
SimonChaumet: Can we use this PR to remove this translation issue ?

packages/manager/apps/pci-load-balancer/src/utils/utils.ts
```diff
@@ -8,3 +9,14 @@ export const getAutoGeneratedName = (prefix = ''): string => {
...
+type CustomApiError = {
+  class: string;
+  message: string;
+};
+
+export const isApiErrorResponse = (
+  error: unknown,
+): error is AxiosError<CustomApiError> =>
+  isAxiosError(error) &&
+  (error as AxiosError<CustomApiError>).response?.data?.message !== undefined;
```
fredericvilcot: You should also check if `class` property exists in `error` :)

packages/manager/apps/pci-load-balancer/src/api/data/load-balancer.ts
```diff
@@ -123,6 +123,17 @@ export type TCreateLoadBalancerParam = {
   name: string;
 };
 
+export type TCreateLoadBalancerError = {
```
fredericvilcot: You should reuse the `CustomApiError` here :)

packages/manager/apps/pci-load-balancer/src/pages/create/hooks/useCreateCallbacks.tsx
```diff
@@ -29,30 +35,45 @@ export const useCreateCallbacks = () => {
...
-        false,
-      );
+            />,
+          );
+        } else {
+          addError(
+            <Translation ns="load-balancer">
+              {(_t) => (
+                <span
+                  dangerouslySetInnerHTML={{
```
SimonChaumet: Since you modified the translation can you remove this part ?

packages/manager/apps/pci-load-balancer/src/utils/utils.ts
```diff
@@ -8,3 +9,21 @@ export const getAutoGeneratedName = (prefix = ''): string => {
...
+};
+
+export const isApiErrorResponse = (
+  error: unknown,
+): error is AxiosError<CustomApiError> => {
+  const err = error as AxiosError<CustomApiError>;
+  const data = err.response?.data;
+
+  return (
+    isAxiosError(err) &&
```
SimonChaumet: Same as other PR : isAxiosError should be done before cast (also isAxiosError includes predicate and avoids casting)

fredericvilcot:
I convert this PR to draft because we have decided with @lionel95200x  to :

- move type guards logic that will handle errors into dedicated `manager-core-api` library
- move UI logic (pci banner) to `pci-common` library

This will guarantee separation of concerns and will prevent from code duplication.

packages/manager/apps/pci-load-balancer/package.json
```diff
@@ -32,6 +32,7 @@
     "@tanstack/react-query": "^5.51.21",
     "@tanstack/react-query-devtools": "^5.51.21",
     "@tanstack/react-table": "^8.20.1",
+    "axios": "^1.6.0",
```
fredericvilcot: Not needed anymore :)

fredericvilcot (approved):

fredericvilcot (approved):

packages/manager/apps/pci-load-balancer/src/pages/create/hooks/useCreateCallbacks.tsx
```diff
@@ -29,30 +37,42 @@ export const useCreateCallbacks = () => {
...
+          addError(
+            <Trans
+              ns="load-balancer"
+              i18nKey="octavia_load_balancer_global_error"
+              values={{
+                message: error.response.data.message,
+                requestId: error.response.headers['x-ovh-queryid'],
+              }}
+            />,
+            false,
```
lolottetheclash, line 72: just a question: for both errors, we don't want them to be dismissable?
(also, false `seems` to be the default value, if you don't want to bother adding it)

lionel95200x: I don't know exactly when the case without message can happen, i just handle it in a type safe way but for me it's when API came down without a message
I think we should catch it globally for all our api call in a dedicated hook with a common error message (not one by app)

lolottetheclash: @lionel95200x  I'm not sure to understand, is it an answer to my comment...?

SimonChaumet (approved):

aboungnaseng-ovhcloud (approved):

PR #17684, opened 2025-06-18
feat(pci-project): create Onboarding page

MarcMonchablon:
## Description

Ticket Reference: #MANAGER-18603


packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
```diff
@@ -0,0 +1,269 @@
...
+                  />
+                ))}
+              {(isCartLoading || isContractsLoading) && (
+                <OdsSkeleton className="w-full mt-4 mb-4" />
+              )}
+            </div>
+          </div>
+
+          {/* Italy-specific agreements */}
+          {isItalySubsidiary && (
```
lionel95200x, line 218: Could you split this italien specific part in another component 
Do we still need this custom agreement for a country ?

packages/manager/apps/pci-project/src/data/api/cart.ts
```diff
@@ -0,0 +1,93 @@
...
+  // Find the correct offer & pricing (matching planCode & capacities = 'renew')
+  const renewPrice = cloudProjectOffer.prices.find((price) =>
+    price.capacities.includes('renew'),
+  );
+  if (!renewPrice) {
+    throw new AxiosError(`could not find 'renew' price in product offer.`);
+  }
+  const { duration, pricingMode } = renewPrice;
+
+  // Now that we have the right order information, order the product
```
lionel95200x, line 63: Think you can remove this comment .. :)

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
MarcMonchablon: I'm on a task to do specifically that (moving Italian agreements to it's own component since it's used elsewhere).
I do think it's still needed.

packages/manager/apps/pci-project/src/data/api/cart.ts
MarcMonchablon: The one line 35 ? I could if you really want to, but while I don't do comment that explains the following code (since it's regarded as a code smell), I tend to do it when it's a function that do a whole lot of imperative "orchestration", with multiple sequential API calls whose purposes aren't that obvious.

The intent behind the 'orderCloudProject' and the 'createCart' project is to abstract away a dozen or so API calls that was before hard to untangle with indirection all over the place, doing  a whole front-end orchestration (that could honestly be done in the back-end potentially).

Potentially, as we work on the other aspect of project creation & payment method, we might have to split up those method into more granular one, but that will be a loss since it will means that the component will do the orchestration instead. In the meantime, I believe that the kind of function that does a lot of orchestration is more readable with a kind of literate programming approach, instead of a sequence of calls without comments that you then have to untangle.

It's a long Github comment, because I know that it goes against both rules of "if your comment explain your code, your code can be simpler" and "if you have a huge function, break it into smaller chunks", that I generally agree with, except in cases like this where many steps are actually accidental complexity of something that could be set in backend in 1 or 2 API calls ideally.

fredericvilcot: @MarcMonchablon 
Yes, but what are your thoughts about SRP principle ?

Here, we have a single function with multiple responsibilities with a lot of software interruptions (`throw`), so you have to comment it :)

I think we may consider splitting the responsibilities with function composition to make the code:

- easier to maintain, as each function would have a single purpose
- more understandable, because the functions would be lighter, clearer, and more readable
- more predictable, since each function would have only one reason to change its behavior

And about throwing, I'm not a big fan of throwing everywhere as it makes the code very unpredictable (will be catch by `ErrorBoundary` but you never know when..) , so we may consider in the future handling errors as values , as we could do in a more functional orientation (e.g. monads) 😎

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
MarcMonchablon: Done :)

packages/manager/apps/pci-project/src/data/api/cart.ts
MarcMonchablon: In the `orderCloudProject` function, throwing serves more as an assert, catching case that should never happen in production, hence the fact that it's handled by ErrorBoundary instead of upstream with values.

I do agree the Single Responsibility Principle, but I'm more wary of Clean Code's notion that a function should always be very short, and should have as few parameters as possible. (In my experience it's often a good idea, but sometime leaves a codebase that replace a relatively straightforward 30-line function with a few dozens smaller methods that aren't used in another context, create early abstraction, and adds unneeded layers of indirection. I personally calls this kind of code "Code where YOU are the hero", because instead of having a linear reading of the code you find yourself having to jump everywhere to follow the plot.)

 I believe that in this case the SRP is respected, since the only reason why the method would change is if data-flow needed to order a cloud product needs to change.  If I were to split up this function into many smaller functions, I believe here that I would only have added a layer of indirection to handle a sequence of API calls that aren't really atomic (since for now we haven't seen case were they've been used separately), and are I believe an accidental complexity of having the front-end orchestrate something that could be done in way fewer API call in the backend.

When we will work more in the (regular) project creation flow, the HDS flow or the payment flow, maybe we will need to have a more granular control of the API calls, but for now it doesn't seems to be the case.

packages/manager/apps/pci-project/src/data/api/cart.ts
```diff
@@ -0,0 +1,93 @@
...
+  // Attach project description to cart
+  await v6.post(
+    `order/cart/${cartId}/item/${orderedCloudProduct.itemId}/configuration`,
+    {
+      label: 'description',
+      value: projectDescription,
+    },
+  );
+
+  return cart;
```
seven-amid: IMO: it's better to separate the responsibility here, and create a specific hook that gather the ones you want togethers ( depending on your need ), because the function don't create the cart only, but do a lot of other stuff

packages/manager/apps/pci-project/src/data/hooks/useCart.tsx
```diff
@@ -0,0 +1,61 @@
...
+    enabled: !!cartId,
+    queryKey: ['new-cart', cartId, 'contract-agreements'],
+    queryFn: async () => {
+      // TODO: optimize with another (less costly) endpoint
+      const summary = await getSummary(cartId || '');
+      const contracts: CartContract[] = summary.contracts.map((d) => ({
+        name: d.name,
+        url: d.url,
+      }));
+      return contracts;
```
seven-amid, line 68: Better use a select to return the contracts from summary

packages/manager/apps/pci-project/src/data/hooks/useCart.tsx
```diff
@@ -0,0 +1,61 @@
...
+  return useQuery<Cart>({
+    queryKey: ['new-cart', planCode],
+    queryFn: () => createCart(ovhSubsidiary, planCode, projectDescription),
+  });
+};
+
+export const useContractAgreements = (cartId: string | null) => {
+  return useQuery<CartContract[]>({
+    enabled: !!cartId,
+    queryKey: ['new-cart', cartId, 'contract-agreements'],
```
seven-amid: better to refactor the key in a separate function, in case of we want to use it to invalidate the cache

packages/manager/apps/pci-project/src/data/hooks/useCart.tsx
```diff
@@ -0,0 +1,61 @@
...
+
+export const useFinalizeCart = ({
+  onError,
+  onSuccess,
+}: TUseFinalizeCartInterface) => {
+  const mutation = useMutation({
+    mutationFn: (cartId: string) => checkoutCart(cartId),
+    onError,
+    onSuccess: (summary: CartSummary) => {
+      if (onSuccess) {
```
seven-amid: why  you check onSuccess, and in your type is mandatory !

packages/manager/apps/pci-project/src/data/hooks/useCart.tsx
```diff
@@ -0,0 +1,61 @@
...
+        onSuccess(summary);
+      }
+    },
+  });
+  return {
+    finalizeCart: (cartId: string) => mutation.mutate(cartId),
+    ...mutation,
```
seven-amid, line 89: no need to do this, u can return all the promise and rename the mutate as you wish on use 
`const { mutate: finalizeCart } = useFinalizeCart() `

packages/manager/apps/pci-project/src/data/types/eligibility.type.ts
```diff
@@ -0,0 +1,47 @@
...
+  LTL = 'LTL',
+  MAD = 'MAD',
+  PLN = 'PLN',
+  SGD = 'SGD',
+  TND = 'TND',
+  USD = 'USD',
+  XOF = 'XOF',
+  NOT_APPLICABLE = 'N/A',
+  POINTS = 'points',
+}
```
seven-amid: There is an existing enum already exported in manager-react-components

packages/manager/apps/pci-project/src/pages/creation/Creation.page.tsx
```diff
@@ -0,0 +1,3 @@
+export default function CreationPage() {
+  return <h1>TODO: creation page</h1>;
+}
```
seven-amid, line 3: what's the difference between Creation.page and Creating.page ?

packages/manager/apps/pci-project/src/pages/onboarding/DiscoveryGuard.component.tsx
```diff
@@ -0,0 +1,107 @@
...
+    const unpaidProjects = activeProjects.filter((d) => d.isUnpaid);
+    if (unpaidProjects.length > 0) {
+      goToProjectsList();
+      return;
+    }
+
+    // If user already have projects, show discovery or default as priority.
+    const discoveryProject =
+      activeProjects.find((d) => d.planCode === PlanCode.PROJECT_DISCOVERY) ||
+      null;
```
seven-amid: there is an existing  function in manager-pci-common called isDiscoveryProject that do the job

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.spec.tsx
```diff
@@ -0,0 +1,136 @@
...
+          onChange={(e) => onOdsChange(e)}
+        />
+        {children}
+      </>
+    ),
+    OdsLink: ({ label, href }: { label: string; href: string }) => (
+      <a href={href}>{label}</a>
+    ),
+  };
+});
```
seven-amid, line 75: what's the need behind mocking Ods component ?

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
```diff
@@ -0,0 +1,244 @@
...
+    !hasContractsAgreements ||
+    (isItalySubsidiary ? !hasItalyAgreements : false);
+
+  const handleDiscoveryProjectClick = () => {
+    if (isProjectCreationDisabled) {
+      return;
+    }
+    if (!cart || !cart.cartId) {
+      return;
+    }
```
seven-amid, line 78: since uu break in the both condition, you can make group them within one condition!

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
```diff
@@ -0,0 +1,244 @@
...
+    return (
+      <DiscoveryGuard>
+        <div
+          className="pci_project_onboarding_us w-100 h-100 flex flex-row max-w-[100vw] overflow-x-hidden"
+          data-testid="discovery-page-us"
+        >
+          <img
+            className="onboarding-us-img invisible md:visible"
+            src={onboardingUsImage}
+            alt=""
```
seven-amid, line 96: empty alt!

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
```diff
@@ -0,0 +1,244 @@
...
+                  )}
+                  href={INDIAN_KYC_SUPPORT_LINK}
+                  rel="noopener"
+                  target="_blank"
+                />
+              </div>
+            </OdsMessage>
+          )}
+
+          <img className="w-100" src={onboardingImage} alt="" />
```
seven-amid, line 159: empty alt!

packages/manager/apps/pci-project/src/data/hooks/useCart.tsx
MarcMonchablon: Maybe, but most of the example I've seen in the database renamed the mutate in the hook declaration (ex: updateVolume in pci-block-storage)

packages/manager/apps/pci-project/src/pages/creation/Creation.page.tsx
MarcMonchablon: Creation page is the one to create a (regular, aka non-discovery) project. Creating page is the one with the sliders, and doing loading. On the angular app those are two different URLs, and we might need to redirect to one or the other.

packages/manager/apps/pci-project/src/pages/onboarding/DiscoveryGuard.component.tsx
MarcMonchablon: No sure it's worthwhile to add a new import just to avoid a simple equality check on planCode.

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.spec.tsx
MarcMonchablon: In my unit test I had lots of issues with ODS, where I couldn't interact with the elements to check the proper behavior, hence the mocks.

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
MarcMonchablon: ISO with prod. Having an empty alt attribute is not always an accessibility issue, especially when the image is purely decorative. From what I've heard screen readers tends to read the path of the image when no alt attribute is found, which is less useful than simply omitting it.

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
MarcMonchablon: Same as above

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
MarcMonchablon: Fixed.

packages/manager/apps/pci-project/src/pages/onboarding/DiscoveryGuard.component.tsx
```diff
@@ -0,0 +1,107 @@
+import { useState, useEffect, useMemo, useContext, ReactNode } from 'react';
+import { useNavigate } from 'react-router-dom';
+import { OdsSpinner } from '@ovhcloud/ods-components/react';
+import { ShellContext } from '@ovh-ux/manager-react-shell-client';
+import { useEligibility } from '@/data/hooks/useEligibility';
+import { PaymentMethod } from '@/data/types/eligibility.type';
+import useActiveProjects from '@/data/hooks/useActiveProjects';
+import { PlanCode } from '@/data/types/cart.type';
+import { TProjectWithService } from '@/data/types/project.type';
+import { urls } from '@/routes/routes.constant';
```
StephaneRavet, line 11: ⚠️ **Centralisation de la logique de garde**
Le composant `DiscoveryGuard.component.tsx` contient la condition d'affichage de la page onboarding. Si d'autres routes futures doivent être protégées par le même critère, envisagez plutôt un *route wrapper* ou un HOC pour ne pas dupliquer la logique.

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
```diff
@@ -0,0 +1,241 @@
...
+
```
StephaneRavet, line 160: ⚠️ **Taille des assets images**
Le fichier `onboarding.png` pèse 1,27 Mo et `onboarding-us.png` 241 Ko.

* Convertir en WebP ou AVIF pour réduire le poids.
* Ajouter l'attribut `loading="lazy"` sur les balises ``.

packages/manager/apps/pci-project/src/routes/routes.tsx
```diff
@@ -10,7 +10,9 @@ const RemovePage = lazy(() => import('@/pages/remove/Remove.page'));
 const MainPage = lazy(() => import('@/pages/home/Header.page'));
 const HomePage = lazy(() => import('@/pages/home/Home.page'));
 const SettingsPage = lazy(() => import('@/pages/home/edit/Edit.page'));
-const OnboardingPage = lazy(() => import('@/pages/onboarding'));
+const OnboardingPage = lazy(() => import('@/pages/onboarding/Onboarding.page'));
```
StephaneRavet, line 13: ⚠️ **Chargement de la route**
La page d'onboarding est importée en dur dans `routes.tsx`. Envisagez un `React.lazy` + `Suspense` pour loader ce module à la demande, réduisant le bundle initial.

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
```diff
@@ -0,0 +1,241 @@
...
+  if (isUsRegion) {
+    return (
+      <DiscoveryGuard>
+        <div
+          className="pci_project_onboarding_us w-100 h-100 flex flex-row max-w-[100vw] overflow-x-hidden"
+          data-testid="discovery-page-us"
+        >
+          <img
+            className="onboarding-us-img invisible md:visible"
+            src={onboardingUsImage}
```
StephaneRavet, line 95: 📛 **Attributs ALT manquants**
Les balises `` sur la page onboarding n'ont pas d'attribut `alt`. Toutes les images décoratives ou informatives doivent en avoir un descriptif ou être marquées `alt=""` si purement décoratives.

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
```diff
@@ -0,0 +1,241 @@
...
+          <div
+            className="contracts-agreements-block"
+            data-testid="contracts-agreements"
+          >
+            <div className="flex items-center justify-start gap-4 mb-2">
+              <OdsCheckbox
+                name={'project-contracts-agreements'}
+                inputId={'project-contracts-agreements'}
+                isChecked={hasContractsAgreements}
+                onOdsChange={(e) => setHasContractsAgreements(e.detail.checked)}
```
StephaneRavet, line 182: 📛 **Labels et focus pour la checkbox de contrat**
Le label de la case à cocher des conditions n'est pas relié via `htmlFor` à l'input, ce qui nuit à la navigation clavier et aux lecteurs d'écran. Bien que l'`inputId` soit défini, le label doit utiliser `htmlFor="project-contracts-agreements"` pour être correctement associé.

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.spec.tsx
```diff
@@ -0,0 +1,136 @@
+import { ReactNode, SyntheticEvent } from 'react';
+import { render } from '@testing-library/react';
+import { describe, expect, it, vi } from 'vitest';
+import { ShellContextType } from '@ovh-ux/manager-react-shell-client';
+import OnboardingPage from './Onboarding.page';
+import { createWrapper, shellContext } from '@/wrapperRenders';
+
+function createCustomWrapper({
+  region,
+  ovhSubsidiary,
```
StephaneRavet, line 10: ⚠️ **Couverture des cas d'erreur et de garde**
Ajouter des tests simulant les réponses `useEligibility` en erreur ou non autorisées pour s'assurer que `DiscoveryGuard` redirige ou affiche le bon fallback.

Idéalement couvrir :
* redirection vers `/creation` si pas éligible
* affichage du loader en attente
* sélecteurs robustes avec `data-testid` sur les éléments critiques (checkbox, boutons)

packages/manager/apps/pci-project/src/pages/onboarding/DiscoveryGuard.component.tsx
MarcMonchablon: A priori, seule la page onboarding a cette logique spécifique-là :)

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
MarcMonchablon: Je préfère être iso prod afin d'éviter une régression.

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
MarcMonchablon: Mais pour le coup il y a bien un attribut alt (vide, parce image décorative)

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.tsx
MarcMonchablon: Mais y'a bien le htmlFor défini pourtant ?

packages/manager/apps/pci-project/src/data/hooks/useCart.tsx
seven-amid: what was done in other applications, is not necessary an example to take in consideration, but anyway up to you.

packages/manager/apps/pci-project/src/pages/onboarding/DiscoveryGuard.component.tsx
seven-amid: I disagree here, this was specifically designed for this purpose, so it’s better to use it. If we change the logic in the future, at least we won’t have to update it everywhere

MarcMonchablon: Fixed.

packages/manager/apps/pci-project/src/pages/onboarding/Onboarding.page.spec.tsx
MarcMonchablon: J'ai ajouté un test pour s'assurer que qu'un check eligibilité en erreur n'empeche pas l'affichage de la page.

Pour info, le test n'est pas de rediriger vers création si pas éligible, mais de rediriger vers création automatiquement si l'utilisateur est éligible à un crédit. Cf le code angular que cela émule:


```
      onBoardingStateName: /* @ngInject */ (OvhApiCloud) =>
        OvhApiCloud.v6()
          .getEligibility()
          .$promise.then(({ paymentMethodsAuthorized }) =>
            paymentMethodsAuthorized.includes('credit')
              ? 'pci.projects.new'
              : 'pci.projects.onboarding',
          )
          .catch(() => 'pci.projects.onboarding'),
```

packages/manager/apps/pci-project/src/data/types/eligibility.type.ts
MarcMonchablon: Fixed.

packages/manager/apps/pci-project/src/data/hooks/useCart.tsx
MarcMonchablon: Fixed.

packages/manager/apps/pci-project/src/data/hooks/useCart.tsx
MarcMonchablon: Fixed.

packages/manager/apps/pci-project/src/data/api/cart.ts
MarcMonchablon: Fixed.

seven-amid (approved):

PR #17705, opened 2025-06-18
feat(kms): secrets listing page

mathieumousnier:
## Description

Ticket Reference: #MANAGER-18246


packages/manager/apps/key-management-service/public/translations/secret-manager/common/Messages_fr_FR.json
```diff
@@ -0,0 +1,6 @@
+{
+  "secret_manager": "Secret Manager",
+  "path": "Path",
+  "version": "Version",
+  "creation_date": "Date de création"
```
rjamet-ovh: you can replace it with `creation_date` from @ovh-ux/manager-common-translations

packages/manager/apps/key-management-service/src/modules/secret-manager/data/api/secrets.ts
```diff
@@ -0,0 +1,11 @@
...
+import { Secret } from '@secret-manager/types/secret.type';
+
+export const secretListQueryKey = (okmsId: string) => ['secrets/list', okmsId];
+
+export const getSecretList = async (okmsId: string) => {
+  const response = await apiClient.v2.get<Secret[]>(
+    `okms/resource/${okmsId}/secret`,
+  );
+  return response.data;
+};
```
rjamet-ovh:
```suggestion
export const getSecretList = async (okmsId: string) => {
  const { data } = await apiClient.v2.get<Secret[]>(
    `okms/resource/${okmsId}/secret`,
  );
  return data;
};
```
for the sake of consistency, what do you think?

mathieumousnier: For me is "bonnet blanc" / "blanc bonnet".
It does the same and is colocated in the function scope.
So you prefer it like this I can do it this way.

My main concern is to avoid returning the full axios response and manage the "data" properties in the components as it is done in kms code, this makes my eyes bleed with a lot of pain :)

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/listing.page.spec.tsx
```diff
@@ -0,0 +1,60 @@
...
+    const tableHeaders = [
+      labels.secretManager.common.path,
+      labels.secretManager.common.version,
+      labels.secretManager.common.creation_date,
+    ];
+
+    tableHeaders.forEach((header) => {
+      expect(screen.queryAllByText(header)).toHaveLength(1);
+    });
+  });
```
rjamet-ovh:
```suggestion
  it('should display the secrets listing page', async () => {
    await renderPage();
  });

  it('should display the listing table with all columns', async () => {
    await renderPage();

    const tableHeaders = [
      labels.secretManager.common.path,
      labels.secretManager.common.version,
      labels.secretManager.common.creation_date,
    ];

    tableHeaders.forEach((header) => {
      expect(screen.queryAllByText(header)).toHaveLength(1);
    });
  });
```
forgot to unskip?

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/ListingCells.component.tsx
```diff
@@ -0,0 +1,38 @@
...
+};
+
+export const DatagridCellPath = (item: DatagridItem) => {
+  const url = SECRET_MANAGER_ROUTES_URLS.secretDashboard(
+    item.okmsId,
+    item.secret.path,
+  );
+
+  return <Link href={url} label={item.secret.path} useRouter />;
+};
```
rjamet-ovh: probably better to use `useParams` to get the domainId instead of augmenting the secret data.
```suggestion
export const DatagridCellPath = (item: Secret) => {
    const { okmsId } = useParams();
    
  const url = SECRET_MANAGER_ROUTES_URLS.secretDashboard(
    okmsId,
    item.secret.path,
  );

  return <Link href={url} label={item.secret.path} useRouter />;
};
```

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/listing.page.tsx
```diff
@@ -0,0 +1,59 @@
...
+export default function SecretListingPage() {
+  const { t } = useTranslation('secret-manager/common');
+  const { domainId } = useParams<SecretListingParams>();
+  const { data: secrets, isPending } = useSecretList(domainId);
+
+  const items: DatagridItem[] =
+    secrets?.map((secret) => ({
+      okmsId: domainId,
+      secret,
+    })) || [];
```
rjamet-ovh: we can get rid of this with useParams in the cell components

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/listing.page.tsx
```diff
@@ -0,0 +1,59 @@
...
+    <BaseLayout header={{ title: t('secret_manager') }}>
+      <div className="flex flex-col gap-6">
+        <Datagrid
+          columns={columns}
+          items={items}
+          totalItems={items.length}
+          isLoading={isPending}
+          contentAlignLeft
+        />
+      </div>
```
rjamet-ovh: with only one child, this div with flex rules is not necessary.
```suggestion
        <Datagrid
          columns={columns}
          items={items}
          totalItems={items.length}
          isLoading={isPending}
          contentAlignLeft
        />
```

mathieumousnier: yup, that was copy pasta I should have re-read

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/listing.page.spec.tsx
mathieumousnier: I like to skip things to go faster

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/ListingCells.component.tsx
mathieumousnier: I guess your solution wins because it simplier.
But I am a bit bothered with those useParams that are implying a "context" where the component is used.
So I think I will add a .type file with the `SecretListingPageParams` that explicitly write the dependency.

This then opens the question about `common` components that would use useParams hook, and I think we just should avoid doing that.

packages/manager/apps/key-management-service/src/modules/secret-manager/data/api/secrets.ts
```diff
@@ -0,0 +1,11 @@
+import apiClient from '@ovh-ux/manager-core-api';
+import { Secret } from '@secret-manager/types/secret.type';
+
+export const secretListQueryKey = (okmsId: string) => ['secrets/list', okmsId];
```
rjamet-ovh:
```suggestion
export const secretListQueryKey = (okmsId: string) => ['secrets', 'list', okmsId];
```

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/listing.page.spec.tsx
rjamet-ovh, line 1: should be Listing.page.spec.tsx

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/listing.page.tsx
rjamet-ovh, line 1: should be Listing.page.tsx

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/listing.type.tsx
rjamet-ovh, line 1: should be listing.type.ts

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/listing.type.tsx
```diff
@@ -0,0 +1,3 @@
+export type SecretListingPageParams = {
```
rjamet-ovh: i'm not sure why this type is necessary?

rjamet-ovh (approved):

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/listing.type.tsx
mathieumousnier: We'll talk about this, we'll remove it if we consider it useless.
It feels crazy to me that there is nothing to type what to expect from the router params

packages/manager/apps/key-management-service/src/modules/secret-manager/data/api/secrets.ts
mathieumousnier: Maybe, I took the "/" thing from other code but it already felt weird to me.
I'll think a bit about how we should manage those keys

mathieumousnier: this is the way 👍🏼

packages/manager/apps/key-management-service/src/common/components/Link/Link.component.tsx
```diff
@@ -0,0 +1,29 @@
+import React from 'react';
+import { OdsLink } from '@ovhcloud/ods-components/react';
+import { useHref, useNavigate } from 'react-router-dom';
+
+type OdsLinkProps = React.ComponentProps<typeof OdsLink> & {
+  /** Display a OdsLink but uses the router for navigation */
+  useRouter?: boolean;
```
lionel95200x: Could you keep `use` keyword for hook and convert boolean by `should` or `is`

`shouldUseRouter`

mathieumousnier: I made the change for `isRouterLink`

PR #17708, opened 2025-06-19
feat(manager-react-components): migrate Badge to ODS v19

anooparveti:
## Description

Migrate Badge component to ODSv19
Align Badge with MRC component guidelines

Ticket Reference: #MANAGER-18612


packages/manager-react-components/src/components/badge/__tests__/__snapshots__/badge.snapshot.test.tsx.snap
```diff
@@ -0,0 +1,30 @@
+// Vitest Snapshot v1, https://vitest.dev/guide/snapshot.html
+
+exports[`Badge Snapshot Tests > Displays Badge 1`] = `
+<div>
```
helabenkhalfallah, line 4: I have a doubt about having a div and span for badge, perhaps we can inspire from these documentations: 
https://design.gccollab.ca/component/badges/
https://www.patternfly.org/components/badge/accessibility/
https://www.w3.org/WAI/WCAG21/Techniques/aria/ARIA22#example-2-updating-the-shopping-cart-status

And if it display status, I think we should add aria: 
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/status_role
https://www.w3.org/WAI/WCAG22/Techniques/aria/ARIA22

anooparveti: I believe it's for ODS as both of them are added by ODS and not MRC.

helabenkhalfallah: OK ;)

PR #17712, opened 2025-06-19
fix(iam): unit test of logs routing

tibs245:
## Description

Ticket Reference: #MANAGER-18760


rjamet-ovh (approved):

anooparveti (changes requested):
Type of commit should be `test` while updating only the tests

darsene:
> Type of commit should be `test` while updating only the tests

@anooparveti there is a fix also on routing part so it's OK for the scope, however, the message should be updated

sefiani-ovh (approved):

anooparveti (approved):

Issue #17727, opened 2025-06-19
[manager-pci-common]: <PciDiscoveryBanner component>


seven-amid:
### Have you already checked if a similar item is present on manager-components?

- [x] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

For the exported function isDiscoveryProject in this file, it may be necessary to include the undefined type, in case the passed project is undefined during the initial render

### Where do you expect to use this?

the migration of pci project

### When do you expect this to be delivered?

as soon as we can

PR #17728, opened 2025-06-19
fix(manager-pci-common): add undefined type to the project on isDisco…

seven-amid:
…veryProject

ref: #17727


packages/manager/modules/manager-pci-common/src/components/banner/discovery-banner/PciDiscoveryBanner.tsx
```diff
@@ -11,7 +11,7 @@ import '../../../translations/discovery-banner';
 
 const DISCOVERY_PROJECT_PLAN_CODE = 'project.discovery';
 
-export const isDiscoveryProject = (project: TProject) =>
+export const isDiscoveryProject = (project: TProject | undefined) =>
```
fredericvilcot:
```suggestion
export const isDiscoveryProject = (project?: TProject) =>
```

seven-amid: Is the same idea no ?

fredericvilcot: Yes, but the first one is verbose regarding how TS defines optional parameters (https://www.typescriptlang.org/docs/handbook/2/functions.html#optional-parameters)
it's up to you as of course it's non blocking ;)

lionel95200x: For me we should not have an undefined project at this step. You should catch errors at the top app level, and throughout your entire app consider that the project exists.

Because with this implementation we check multiple times in each banner and screen if a project exists.

seven-amid: It's a bit difficult in the current state to remove the check, because it's not just used for the DiscoveryBanner, it's also used for other checks, so removing it would involve changes across quite a few apps.

fredericvilcot:
@seven-amid Could you please check for reviews and merge it into develop?

frenautvh (approved):

PR #17732, opened 2025-06-19
Fix/sync master to develop

fredericvilcot:
## Description

Ticket Reference: #...


anooparveti (changes requested):
I dont see the sync commit. I see the same commits as on the original PR.
I believe you may have rebased or amended the commit and none of these operations are correct as they modify the commit hash.

PR #17739, opened 2025-06-20
feat(manager-react-components): add strict mode in new components

aboungnaseng-ovhcloud:
## Description

- Adding typescript strict mode for new components in tsconfig.new.json

Ticket Reference: #MANAGER-15000

### DIFF
--- a/packages/manager-react-components/tsconfig.new.json
+++ b/packages/manager-react-components/tsconfig.new.json
@@ -1,0 +1,16 @@
+{
+  "extends": "./tsconfig",
+  "compilerOptions": {
+    "strict": true
+  },
+  "include": [
+    // Add you components migrated here
+  ],
+  "exclude": [
+    "./src/hooks/*",
+    "./src/utils/*",
+    "./src/tailwind/*",
+    "./src/__mocks__/*",
+    "./src/enumTypes/*"
+  ]
+}
--- a/packages/manager-react-components/package.json
+++ b/packages/manager-react-components/package.json
@@ -20,7 +20,7 @@
   ],
   "scripts": {
     "build": "tsc && vite build",
-    "dev": "tsc && vite build",
+    "dev": "tsc -p tsconfig.new.json && vite build",
     "prepare": "tsc && vite build",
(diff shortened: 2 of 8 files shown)


packages/manager-react-components/tsconfig.new.json
```diff
@@ -0,0 +1,31 @@
...
+    "moduleResolution": "bundler",
+    "isolatedModules": true,
+    "declaration": true,
+    "declarationDir": "./dist/types",
+    "jsx": "react-jsx",
+    "strict": true
+  },
+  "files": ["global.d.ts"],
+  "include": [
+    // Add you components migrated here
```
lionel95200x, line 7: Why not migrate directly to strict mode ? It will be very hard to maintain with this two `tsconfig` file 
Maybe you could enable strict mode and override some error in warning ?

aboungnaseng-ovhcloud: Its only temporary, we will keep only one tsconfig file, once we we will correct all the error and the warning in strict mode

helabenkhalfallah (approved):

packages/manager-react-components/tsconfig.new.json
fredericvilcot: @aboungnaseng-ovhcloud 
Why not : 
- adding strict mode by default in `tsconfig.json`
- adding a `tsconfig.build.json` that overrides the strict mode to false to allow the lib to be built?

Something like:

```ts
// to remove after package has been cleaned up
{
  "extends": "./tsconfig",
  "compilerOptions": {
    "strict": false 
  }
}
```

aboungnaseng-ovhcloud: It can be a solution but we want to keep one tsconfig file at the end, and keep the strict mode not only on build mode

fredericvilcot: Yes, but it's the same as you did, except you don't need to copy the whole `tsconfig.json` file, juste add a build one that extends the base config, that is much more cleaner IMO. And once errors cleaned, you'll remove it in the same way you'll remove the `.new` file :)
But it's up to you, not blocking for me.

anooparveti (approved):

PR #17752, opened 2025-06-20
feat(pci-kubernetes): add cluster plan in list

lolottetheclash:
## Description

This PR aims to add the future `plan` cluster property, which will be added next week by the api.
Translations will be added at the end, as they are not stable yet.

Ticket Reference: #TAPC-3168


packages/manager/apps/pci-kubernetes/src/pages/list/useDatagridColumn.spec.tsx
```diff
@@ -29,4 +35,22 @@ describe('useDatagridColumn', () => {
...
+
+    const fakeKube = [{ plan } as TKube];
+
+    const { getByText } = render(
+      <Datagrid
+        columns={result.current}
+        items={fakeKube}
+        totalItems={fakeKube.length}
+      />,
+      { wrapper },
```
SimonChaumet: Since you are testing useDatagridColumn I think you should abstract the datagrid and test directly on the result

lolottetheclash: Not sure I understand what you mean by this, is it possible to test the text cell rendered without giving data to the `Datagrid` component?

SimonChaumet (approved):

PR #17755, opened 2025-06-20
fix(kms): manage users with no permissions to services

mathieumousnier:
## Description

When connected with NONE role (even with all the right permissions on KMS), the KMS dashboard page fails and displays an error.

Ticket Reference: #MANAGER-18260

## Fix Details

- Since there is no way to check for IAM permissions on services endpoint, the error is on the endpoint call is handle and considered as a lack of permissions.

### Use of `OutletContext` to pass data to tabs 

The `useServiceDetails` was called 1st in the `dashboard/index.tsx` file, then a 2nd time in the tab file.
The call on the tab file was conditional to the first one, when the fetch returns an error this setup triggers infinite loop calls in errors.
This can be solved by disabling the `retryOnMount` option, [but this is looks like a patch on a design flaw,](https://github.com/TanStack/query/discussions/2755) so I removed the call on the inner component, and passed the `okmsService` data via the `OutletContext` feature of react-router, which feels a lot cleaner.

Then I saw that the Okms data was also called from the top page component AND the tab component, so I used also `OutletContext`.
Then I saw that for other tabs the Okms data was passed from a ReactContext, so I used `OutletContext` everywhere and remove the use of ReactContext.

- `react-query` allows to call queries in sub-components and queries are deduplicated, but it seems to be a lot better to just use props to pass down dependencies than abuse of queries everywhere
- I could have chosen to pass `OkmsService` from `React.Context` with the `Okms` data but `OutletContext` is here out of the box, it seems better to me to use it and pass some props down to subcomponents

### DIFF
--- a/packages/manager/apps/key-management-service/src/components/Listing/ListingCells.tsx
+++ b/packages/manager/apps/key-management-service/src/components/Listing/ListingCells.tsx
@@ -1,4 +1,4 @@
-import React, { useContext } from 'react';
+import React from 'react';
 import {
   ActionMenu,
   Clipboard,
@@ -24,10 +24,8 @@ import { OkmsAllServiceKeys } from '@/types/okmsServiceKey.type';
 import { useServiceKeyTypeTranslations } from '@/hooks/serviceKey/useServiceKeyTypeTranslations';
 import { ServiceKeyStatus } from '../serviceKey/serviceKeyStatus/serviceKeyStatus.component';
 import useServiceKeyActionsList from '@/hooks/serviceKey/useServiceKeyActionsList';
-import { useOkmsServiceKeyById } from '@/data/hooks/useOkmsServiceKeys';
 import { useFormattedDate } from '@/hooks/useFormattedDate';
 import { OkmsServiceState } from '../layout-helpers/Dashboard/okmsServiceState/OkmsServiceState.component';
-import { OkmsContext } from '@/pages/dashboard';
 import { KMS_ROUTES_URLS } from '@/routes/routes.constants';
 
 export const DatagridCellId = (props: OKMS | OkmsAllServiceKeys) => {
@@ -66,16 +64,19 @@ export const DatagridCellRegion = (kms: OKMS) => {
 };
 
 export const DatagridCellStatus = (kms: OKMS) => {
-  const { data: OkmsServiceInfos, isLoading, isError } = useServiceDetails({
+  const { data: okmsService, isPending, isError } = useServiceDetails({
     resourceName: kms.id,
   });
-  if (isLoading) {
+
+  if (isPending) {
     return <OdsSpinner size={ODS_SPINNER_SIZE.sm} />;
   }
-  if (isError) {
+
+  if (!isError) {
     return <></>;
   }
-  return <OkmsServiceState state={OkmsServiceInfos.data.resource.state} />;
+
+  return <OkmsServiceState state={okmsService?.data?.resource?.state} />;
 };
 
 export const DatagridResourceKmipCountCell = (kms: OKMS) => {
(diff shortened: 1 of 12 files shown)


rjamet-ovh (approved):

rjamet-ovh (approved):

packages/manager/apps/key-management-service/src/pages/dashboard/KmsDashboard.page.spec.tsx
```diff
@@ -123,16 +124,15 @@ describe('KMS dashboard test suite', () => {
...
-              'key_management_service_service-keys_dashboard_field_name'
-            ],
-          ),
-        ).toHaveLength(2),
-      WAIT_FOR_DEFAULT_OPTIONS,
-    );
+    // Wait for modal to open
+    await waitFor(() => {
+      assertOdsModalVisibility({ container, isVisible: true });
+    }, WAIT_FOR_DEFAULT_OPTIONS);
```
pauldkn: I believe this `assertOdsModalVisibility` is already wrapped with a waitFor, so you may just use it like this:
`await assertOdsModalVisibility({ container, isVisible: true })`
(and pass another timeout as parameter if you want a different one)

packages/manager/apps/key-management-service/src/pages/dashboard/index.tsx
```diff
@@ -45,17 +43,15 @@ export default function DashboardPage() {
   } = useOkmsById(okmsId);
 
   const {
-    data: okmsServiceInfos,
-    isLoading: isOkmsServiceInfosLoading,
-    isError: isOkmsServiceInfosError,
-    error: okmsServiceInfoError,
+    data: okmsService,
+    isPending: isOkmsServiceLoading,
+    isError: hasServicesPermissions,
```
pauldkn: **question:** if the useServiceDetails query has an error, that means you have services permissions?

mathieumousnier: If the useServiceDetails query has an error, it means I DON'T HAVE services permissions.
There is no way to check for IAM permissions on services.
We chose to handle this case as a silent error.

packages/manager/apps/key-management-service/src/components/Listing/ListingCells.tsx
```diff
@@ -66,16 +64,19 @@ export const DatagridCellRegion = (kms: OKMS) => {
...
     return <OdsSpinner size={ODS_SPINNER_SIZE.sm} />;
   }
-  if (isError) {
+
+  if (!isError) {
     return <></>;
   }
-  return <OkmsServiceState state={OkmsServiceInfos.data.resource.state} />;
+
+  return <OkmsServiceState state={okmsService?.data?.resource?.state} />;
```
pauldkn, line 79: **question:** if there is no error, nothing has to be displayed as a status?

mathieumousnier: Good question. The PR is a fix relative to the dashboard page that failed to be displayed.
I reworked some code here, but there is not feature added.
We will add later : 
- displaying something here as the status
- displaying something also on the dashboard instead of the billing tile

We don't know yet what to display.

packages/manager/apps/key-management-service/src/pages/dashboard/KmsDashboard.page.spec.tsx
mathieumousnier: tip top thanks

packages/manager/apps/key-management-service/src/pages/dashboard/index.tsx
pauldkn: Okay, then maybe you should rename `hasServicesPermissions` to something like `missingServicesPermissions` or `hasNoServicesPermissions` because `isError: hasServicesPermissions` makes it look like the name is the opposite of what it actually means

mathieumousnier: Ok, in fact this is a remains of something I removed, I clean this right now

PR #17762, opened 2025-06-20
feat(ips): gamefirewall listing page

chipp972:
ref: #MANAGER-18594

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/ips/src/routes/routes.constant.ts
+++ b/packages/manager/apps/ips/src/routes/routes.constant.ts
@@ -9,7 +9,7 @@ export const subRoutes = {
   byoipOrder: 'byoip-order',
   terminate: 'terminate',
   upsertDescription: 'upsert-description',
-  configureGameFirewall: 'game-firewal',
+  configureGameFirewall: 'game-firewall',
 };
 
 export const urlDynamicParts = {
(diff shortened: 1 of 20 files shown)


packages/manager/apps/ips/src/routes/routes.constant.ts
```diff
@@ -25,7 +25,7 @@ export const urls = {
   listing: subRoutes.root,
   listingTerminate: `${subRoutes.root}/${subRoutes.terminate}/${urlDynamicParts.id}`,
   listingUpsertDescription: `${subRoutes.root}/${subRoutes.upsertDescription}/${urlDynamicParts.id}`,
-  listingConfigureGameFirewall: `${subRoutes.root}/${urlDynamicParts.id}/${subRoutes.configureGameFirewall}`,
+  configureGameFirewall: `${subRoutes.root}/${urlDynamicParts.id}/${subRoutes.configureGameFirewall}`,
```
sachinrameshn, line 28: I preferred listingConfigureGameFirewall as it make more sense as we have this action in listing page knw ? What do you think ?

chipp972: I think it's not a good idea since we can access directly this url and it opens a new page.
It makes sense for modals that appear on top of another page but configureGameFirewall is not a children of listing page so it's not a good idea to create coupling between them because tomorrow if we add a link to the gamefirewall page on a dashboard, it will be the same link.

sachinrameshn: Agreed !

PR #17763, opened 2025-06-20
feat(ips): add delete byoip action changes

sachinrameshn:
ref: #MANAGER-18501

## Description

Ticket Reference: #...


packages/manager/apps/ips/src/pages/actions/terminate/handleLink.component.tsx
```diff
@@ -0,0 +1,40 @@
...
+          {},
+        );
+        setHref(response as string);
+      } catch (error) {
+        setHref('#');
+      }
+    };
+    fetchUrl();
+  }, []);
+  return (
```
lionel95200x, line 33: Please add a blank line between effect and return :)

packages/manager/apps/ips/src/utils/links.ts
```diff
@@ -0,0 +1 @@
+export const US_API_CONSOLE_LINK = 'https://api.us.ovhcloud.com/console/';
```
lionel95200x: Seems to not be an utils but a constant, could you move it in an appropriate file

packages/manager/apps/ips/src/pages/actions/terminate/handleLink.component.tsx
sachinrameshn: okay

PR #17826, opened 2025-06-25
refactor(pci-instances): rework instance call with clean architecture

fredericvilcot:
This PR introduces a significant refactor of the `GET /instance/:instanceId` endpoint's handling and its consumption in the frontend, with a strong focus on clean code principles and hexagonal architecture patterns. Key changes include:

**🔧 Clean Architecture Alignment**

- Refactored the GET instance endpoint to better align with hexagonal architecture principles.
- Introduced a clear separation between domain entities (`Instance`) and transport layer representations (DTOs).
- Added a dedicated mapper within the adapter layer to handle transformations between Instance domain entities and DTOs. This helps isolate domain logic from infrastructure concerns and reinforces the boundaries of the domain layer.
- Stored data in `tanstack` cache layer is now filled up with entities

**💡 UI – Dumb Components & Business-Oriented Hooks**

- Introduced business-specific React hooks to encapsulate application logic, ensuring UI components remain dumb and purely presentational.
- Implemented view models close to the components to derive relevant data from the Instance entity. This enhances maintainability and clarity, as components only receive what they need in the format they expect.

**✅ Benefits**

- Better separation of concerns across layers (domain, infrastructure, application, presentation).
- Improved testability and maintainability, thanks to isolated domain logic and clearer data flow.
- Frontend components are now decoupled from domain structures, relying only on view models tailored to their needs.
- Easier to extend and evolve the system without unintended side effects, thanks to a cleaner, more modular architecture.

This refactor lays the foundation for future scalability and aligns both backend and frontend with a more robust and decoupled architectural design.


packages/manager/apps/pci-instances/src/data/api/instance.ts
```diff
@@ -137,13 +139,29 @@ export const activateMonthlyBilling = (
...
+  params,
+}: TGetInstanceArgs): Promise<TInstance> =>
   v6
-    .get(`/cloud/project/${projectId}/aggregated/instance/${instanceId}`)
-    .then((response) => response.data);
+    .get(
+      `/cloud/project/${projectId}/region/${region}/instance/${instanceId}`,
+      { params },
+    )
+    .then((response) => mapDtoToInstance(response.data));
```
lionel95200x, line 167: Think we should not map inside the `getInstance`method but use a selector oustide to keep the `single responsability`, method should just fetch a single instance, and mapper should be use from the caller

fredericvilcot: @lionel95200x 
Map data into the adapter is the core concept of clean architecture ;)

I get your point about separation of concerns, but in a hexagonal architecture, it's actually the adapter’s responsibility to perform the mapping between DTOs (from the infrastructure side) and domain Entities.

The adapter acts as a boundary between the technical world (e.g. database, API models) and the domain core — so mapping inside methods like getInstance is not only acceptable, it's expected.

Moving that mapping outside the adapter would break the architectural boundaries and expose infrastructure concerns to the domain or application layer, which we want to avoid.

fredericvilcot: And BTW, it ensures that `tanstack` cache is populated with entities, not dto :)

packages/manager/apps/pci-instances/src/data/api/mapper/instance.mapper.ts
```diff
@@ -0,0 +1,67 @@
...
+        foundAddresses.push({ ...rest, subnet: rest.subnet ?? null });
+      return acc;
+    }
+    return acc.set(type, [{ ...rest, subnet: rest.subnet ?? null }]);
+  }, new Map<TInstanceAddressType, TInstanceAddress[]>());
+
+export const mapDtoToInstance = (dto: TInstanceDto): TInstance => ({
+  ...dto,
+  region: mapRegion(dto),
+  pricings: dto.pricings?.map(mapPricing) ?? null,
```
lionel95200x, line 59: Maybe we should return an empty array [] instead of nul in our react context ?

lolottetheclash (approved):
well done !!!

packages/manager/apps/pci-instances/src/pages/instances/action/view-models/selectInstanceForActionModal.ts
```diff
@@ -0,0 +1,46 @@
...
+});
+
+const mapInstance = (instance: TInstance) => ({
+  isImageDeprecated: instance.image?.deprecated ?? false,
+  ip: instance.addresses.values().next().value?.[0].ip ?? '',
+  region: instance.region.name,
+  imageId: instance.image?.id ?? '',
+});
+
+const instanceIsAggregated = (
```
lionel95200x:
```suggestion
const isInstanceAggregated = (
```

packages/manager/apps/pci-instances/src/data/api/mapper/instance.mapper.ts
fredericvilcot: Well, it's a good question :)

Both can work, but I've chosen `null` for some reasons:

- `null` can clearly indicate that the data is not available, not loaded, or unknown, rather than just being an empty list.
- regarding semantic distinction, you can distinguish between:
`null`: "We don’t know yet" or "It hasn’t been initialized."
`[]`: "We know it's empty."

packages/manager/apps/pci-instances/src/data/api/instance.ts
SimonChaumet: I think there is a layer missing, if we generate different entities from a single call (e.g. : productAvailability has regions, products, etc) we will have multiple time the same url inside the project with each time a different adapter and there will be no cache between

packages/manager/apps/pci-instances/src/data/api/mapper/instance.mapper.ts
SimonChaumet: I think you should either have one pricing or null, or a list of pricings with a default to empty list

On block storage when getting a volume there is only one pricing (but it can be configured with a specific size passed as a parameter). The only time were there are multiple pricing is on the creation page.

These are 2 different use cases so I think the view shouldn't have a list of pricings

This is the same for other properties (like volumes line 61) : the view shouldn't know if there are loading, unkown etc at this point, it should only take the list to display it, and another property for the state

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.ts
```diff
@@ -1,23 +1,48 @@
...
+  params,
+  queryOptions,
+}: TUseInstanceArgs) => {
   const projectId = useProjectId();
 
+  const paramsQueryKey = useMemo(() => {
+    if (!params) return [];
+    return Object.keys(params).filter(
+      (key): key is keyof TGetInstanceQueryParams =>
+        !!params[key as keyof TGetInstanceQueryParams],
```
SimonChaumet: Either the keys returned value should be cast or there are some check missings

Object.keys is typed to generic string because there are other internal values that can be in the output, so it should be used in combination with : https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/hasOwn

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.ts
```diff
@@ -1,23 +1,48 @@
...
+    );
+  }, [params]);
+
   return useQuery({
-    queryKey: instancesQueryKey(projectId, ['instance', instanceId]),
-    queryFn: () => getInstance({ projectId, instanceId }),
+    queryKey: instancesQueryKey(projectId, [
```
SimonChaumet, line 29: What do you think of using this refactor to move to tanstack queryOptions for the modified calls and have the query keys typed ?

packages/manager/apps/pci-instances/src/pages/instances/action/hooks/useInstanceActionModal.tsx
```diff
@@ -97,7 +111,7 @@ export const useCachedInstanceAction = (
   }, [addError, error, instance, navigate, t]);
 
   return {
-    instance,
+    instance: selectInstanceForActionModal(instance),
```
SimonChaumet: This should be wrapped in a useMemo

packages/manager/apps/pci-instances/src/pages/instances/action/view-models/selectInstanceForActionModal.ts
```diff
@@ -0,0 +1,46 @@
...
+  'imageId' in instance && Array.isArray(instance.addresses);
+
+export const selectInstanceForActionModal = (
+  instance?: TAggregatedInstanceDto | TInstance,
+): TInstanceActionModalViewModel => {
+  if (!instance) return null;
+
+  const instanceActionForActionElement = instanceIsAggregated(instance)
+    ? mapAggregatedInstanceDto(instance)
+    : mapInstance(instance);
```
fredericvilcot: Review hint: 

This will be removed when listing page's api call will also be mapped into `Instance` entity (`TInstance[]`)

It was too much for this PR so we added this temporary 'hack'

packages/manager/apps/pci-instances/src/pages/instances/action/view-models/selectInstanceForActionModal.ts
```diff
@@ -0,0 +1,46 @@
...
+const mapAggregatedInstanceDto = (instance: TAggregatedInstanceDto) => ({
+  isImageDeprecated: instance.isImageDeprecated,
+  ip: instance.addresses[0].ip,
+  region: instance.region,
+  imageId: instance.imageId,
+});
+
+const mapInstance = (instance: TInstance) => ({
+  isImageDeprecated: instance.image?.deprecated ?? false,
+  ip: instance.addresses.values().next().value?.[0].ip ?? '',
```
SimonChaumet: Since it is a map this first value might be inconsistent, I think there is some logic missing

packages/manager/apps/pci-instances/src/pages/instances/action/view-models/selectInstanceForActionModal.ts
fredericvilcot: done thanks

packages/manager/apps/pci-instances/src/pages/instances/datagrid/hooks/useDatagridPolling.ts
```diff
@@ -0,0 +1,87 @@
...
+  const pollingData = useInstancesPolling(
+    pendingTasks,
+    {
+      onSuccess: handlePollingSuccess,
+      onError: handlePollingError,
+    },
+    { retry: shouldRetryAfter404Error },
+  );
+
+  return selectPollingDataForDatagrid(pollingData);
```
SimonChaumet: This should be in a useMemo

packages/manager/apps/pci-instances/src/types/instance/common.type.ts
```diff
@@ -57,3 +57,42 @@ export type TInstanceActionGroup =
   | 'details'
   | 'lifecycle'
   | 'shelve';
+
+export type TPrice = {
+  type: 'hour' | 'month' | 'licence' | 'savingplans';
```
SimonChaumet: Since iceberg always returns an enum shouldn't we add string as default to avoid issues ?

packages/manager/apps/pci-instances/src/data/api/instance.ts
fredericvilcot: In pure hexagonal architecture, `useQuery` function should be in the same level than this function.
Because they are both part of the adapter, or infra layer.

Here, I have to map the data inside the get function to ensure the cache will only handle entities, like we would have done using an external store as redux. 

So, the only other solution to achieve that is to move the mapper into que `queryFn` function of the `useInstance` hook (orchestrator) but it won't guarantee anymore that dto are mapped at higher adapter level.

packages/manager/apps/pci-instances/src/types/instance/common.type.ts
fredericvilcot: Yes we can, but IMO union type is much more explicit regarding the structure of the model entity (as we want to be agnostic from iceberg or another technology )

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.ts
fredericvilcot: Yes you're right, thanks for this catch :)

I'll suggest something like this with as assertion as TS engine cannot narrow type at this level:

```ts
 const paramsQueryKey = useMemo(() => {
    if (!params) return [];
    return Object.keys(params).filter(
      (key): key is keyof TGetInstanceQueryParams =>
        Object.prototype.hasOwnProperty.call(params, key) &&
        !!params[key as keyof TGetInstanceQueryParams],
    );
  }, [params]);
```

Or to prevent from type cast, we can create an utility function like `isKeyOf`, what you think?

packages/manager/apps/pci-instances/src/data/api/mapper/instance.mapper.ts
fredericvilcot: @SimonChaumet 

In our use case, we can receive several prices from API like:
- flavor hourly price 
- flavor monthly price
- window's licence price
- ...etc

So, we designed the entity to be defined with multiple prices, regarding this considerations and to make the front-end structure more extendable and easier to manipulate through view models.

fredericvilcot: And furthermore, the `intance` model entity represents the instance structure, its business properties, regardless the view layer.

SimonChaumet: Okay, I understand, it makes sense for the entity, I thought it was intended to be used by view

packages/manager/apps/pci-instances/src/types/instance/common.type.ts
fredericvilcot: Updated to ` type: 'hour' | 'month' | 'licence' | 'savingplans' | string;`

SimonChaumet (approved):

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.ts
fredericvilcot: I missed this comment! 

For sure we'll implement it for next iteration if you don't mind :)

PR #17829, opened 2025-06-25
feat(manager-react-components): migrate Clipboard component to ODS v19

anooparveti:
## Description

Migrate Clipboard to ODS v19
Add unit and snapshot tests
Add loading feature available from ODS v19
Add stories for disabled, masked and loading cases

Ticket Reference: #MANAGER-18708    

## Additional Information

BREAKING CHANGE: Props update:
 isDisabled -> disabled
 onOdsCopy -> onCopy
 isMasked -> masked


packages/manager-react-components/src/components/clipboard/Clipboard.component.tsx
```diff
@@ -0,0 +1,31 @@
...
+  ClipboardTrigger,
+} from '@ovhcloud/ods-react';
+import { useTranslation } from 'react-i18next';
+import './translations';
+import { ClipboardProps } from './Clipboard.props';
+
+export const Clipboard = ({
+  loading = false,
+  masked = false,
+  ...props
```
helabenkhalfallah: what do you think other instead of props because props includes all

packages/manager-react-components/src/components/clipboard/__tests__/Clipboard.spec.tsx
```diff
@@ -0,0 +1,39 @@
+import { waitFor } from '@testing-library/react';
+import { render } from '../../../utils/test.provider';
+import { Clipboard } from '../Clipboard.component';
+
+const setupSpecTest = async (props = {}) =>
```
helabenkhalfallah: is it possible to choose another name than `setupSpecTest`, thanks

packages/manager-react-components/src/components/clipboard/__tests__/Clipboard.spec.tsx
```diff
@@ -0,0 +1,39 @@
...
+  waitFor(() =>
+    render(<Clipboard value={'Value to copy to clipboard'} {...props} />),
```
helabenkhalfallah: why we need waitFor if we dont have async inside the Clipboard component and without shadow dom?

packages/manager-react-components/src/components/clipboard/__tests__/Clipboard.spec.tsx
anooparveti: It is from the existing code but anyways, i will rename it.

packages/manager-react-components/src/components/clipboard/__tests__/Clipboard.spec.tsx
anooparveti: It is from the existing code but you're right we no longer need it. I will remove it.

packages/manager-react-components/src/components/clipboard/__tests__/Clipboard.spec.tsx
```diff
@@ -0,0 +1,36 @@
+import { render } from '../../../utils/test.provider';
+import { Clipboard } from '../Clipboard.component';
+
+const renderClipboardComponent = async (props = {}) =>
```
helabenkhalfallah: if it's a normal react component, we dont need I think even the async, for me the test should be simply : 

```js
const { getByTestId } = render(<Clipboard value={'Value to copy to clipboard'} {...props} />)
```

this avoid async non-determinism.

helabenkhalfallah (approved):

PR #17853, opened 2025-06-27
feat(container): add a modal to invite customer to fill their information

JacquesLarique:
## Description

Created a modal to inform customer that we suspect their company information are incorrect and invite them to either correct it or let us update it for them

Ticket Reference: #MANAGER-14721


oalkabouss (approved):

packages/manager/apps/container/src/components/ModalContainer/ModalsContainer.tsx
```diff
@@ -0,0 +1,83 @@
...
+    if (id === CurrentModal.displayName || id === CurrentModal.name) {
+      const next = currentIndex + 1;
+      setCurrentIndex(next < MODALS.length ? next : null);
+    }
+  }, [currentIndex]);
+
+  useEffect(() => {
+    if (currentIndex === null) return;
+    uxPlugin.registerModalActionDoneListener(handleActionDone);
+    // TODO: implement the unregisterModalActionDoneListener in order to clean up this event listener (MANAGER-18813)
```
helabenkhalfallah, line 69: what's the risk of this TODO now?

packages/manager/apps/container/src/components/SuggestionModal/SuggestionModal.component.tsx
```diff
@@ -0,0 +1,162 @@
...
+
+  return (
+    showModal && (
+      <OsdsModal
+        dismissible={false}
+        headline={t('suggestion_modal_title')}
+        color={ODS_THEME_COLOR_INTENT.info}
+        data-testid="suggestion-modal"
+      >
+        <OsdsText
```
helabenkhalfallah, line 116: is this span? if yes then span cannot contains p or ul

JacquesLarique: It looks like it's behaving more like a div

packages/manager/apps/container/src/components/ModalContainer/ModalsContainer.tsx
JacquesLarique: Not too risky, as we can register only one listener ever (if we recall it, the previous one will be override and should be garbage collected)
The implementation of the TODO is coming soon: https://github.com/ovh/manager/pull/18827

packages/manager/apps/container/src/components/SuggestionModal/SuggestionModal.component.tsx
helabenkhalfallah: Okay thanks ! :) :)

packages/manager/apps/container/src/components/ModalContainer/ModalsContainer.tsx
helabenkhalfallah: Okay deal! thank you!

PR #17871, opened 2025-06-27
fix(pci-block-storage): use volume type for pricing

SimonChaumet:
ref: #TAPC-4448

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-block-storage/src/api/hooks/useVolume.tsx
+++ b/packages/manager/apps/pci-block-storage/src/api/hooks/useVolume.tsx
@@ -1,9 +1,14 @@
 import { useCallback, useMemo } from 'react';
 import { applyFilters, Filter } from '@ovh-ux/manager-core-api';
-import { ColumnSort, PaginationState } from '@ovh-ux/manager-react-components';
+import {
+  ColumnSort,
+  PaginationState,
+  useCatalogPrice,
+} from '@ovh-ux/manager-react-components';
 import { useMutation, useQueries, useQuery } from '@tanstack/react-query';
 import { useTranslation } from 'react-i18next';
 import { pipe } from 'lodash/fp';
+import { NAMESPACES } from '@ovh-ux/manager-common-translations';
 import {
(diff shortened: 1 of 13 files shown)


packages/manager/apps/pci-block-storage/src/api/hooks/useVolume.tsx
```diff
@@ -118,16 +126,29 @@ export const getVolumeQuery = (projectId: string, volumeId: string) => ({
   enabled: !!volumeId,
 });
 
-export type UseVolumeResult = TVolume | undefined;
-
-export const useVolume = (projectId: string, volumeId: string) => {
+type UseVolumeVolume = TAPIVolume &
```
lolottetheclash: `UseVolumeVolume`is a little redondant, no? ;)

packages/manager/apps/pci-block-storage/src/api/select/catalog.ts
```diff
@@ -167,14 +184,53 @@ export const mapVolumeModelPriceSpecs = <T extends TVolumeAddon>(
...
+  }
+
+  let pricings = models.flatMap((m) => m.pricings);
+
+  if (filter.region || filter.encryptionType || filter.volumeType) {
+    pricings = pricings.filter((p) => {
+      if (filter.region !== undefined && !p.regions.includes(filter.region))
+        return false;
+
+      if (
```
lolottetheclash: I think it could improve readability to update with some const like `matchesEncryptionType`, `matchesVolumeType`, what do you think? jut a suggestion

packages/manager/apps/pci-block-storage/src/api/hooks/useVolume.tsx
SimonChaumet: It was the concatenation of UseVolume for the hook and Volume for the type of the data, if I name it UseVolume it would match the hook name so it will be weird

I don't have any other idea so if you have a suggestion I can take it

PR #17879, opened 2025-06-30
feat(sap-features-hub): replace vdcId key by datacenterId

pauldkn:
## Description

Replace `vdcId` key with `datacenterId` to reflect API changes

Ticket Reference: #MANAGER-18621


pauldkn:
Included in #17892

PR #17881, opened 2025-06-30
Feat/manager 18922

StephaneRavet:
## Description

Add a header component, used on the home page (`/pci/project/:id`) and the settings page (`/pci/project/:id/edit`).
Route and folder names were chosen to be as clear and meaningful as possible, since legacy URLs are no longer domain-driven.

### Design deviations

image

* The home icon in the breadcrumb does not exist in MRC. While it could be implemented via ODS, we’ll likely avoid requesting changes to MRC to keep improvements low-cost.
* The project description text doesn’t exist as actual data. Although projects have `id`, `name`, and `description` fields, the stored values are currently:

  * `id`: id
  * `name`: id
  * `description`: name

**Ticket Reference**: `MANAGER-18922`


packages/manager/apps/pci-project/src/data/api/pci-project.ts
```diff
@@ -0,0 +1,18 @@
+import { apiClient } from '@ovh-ux/manager-core-api';
```
FabienHenon: Generally we don't put `pci-` in the file name, `project.ts` is enough and uses the same pattern as the other api files. (The generator generates with `pci-` but we never did that way)

packages/manager/apps/pci-project/src/pages/home/Header.page.tsx
```diff
@@ -0,0 +1,66 @@
...
+
+export default function ProjectHeader() {
+  const { t } = useTranslation(['common', 'home', 'settings']);
+
+  const { data: project, isError } = useProject();
+
+  const appHomeUrl = useProjectUrl('public-cloud') + urls.home;
+
+  if (!project) return <NotFound />;
+  if (isError) return <OdsText>{t('error')}</OdsText>;
```
FabienHenon: I think, in case of error, you can let it crash, the ErrorBoundary will catch it and display an error

packages/manager/apps/pci-project/src/pages/home/Header.page.tsx
```diff
@@ -0,0 +1,66 @@
...
+  if (!project) return <OdsText>{t('project_not_found')}</OdsText>;
```
FabienHenon: You have the same condition 2 lines above. Shouldn't you check for `isLoading` ? Because at page loading `project` will be `undefined` but we want to display a spinner instead of a `NotFound` message when the API call is still in progress

packages/manager/apps/pci-project/src/pages/home/Header.page.tsx
```diff
@@ -0,0 +1,66 @@
...
+  const projectDescription = project?.description;
+
+  const ROADMAP_CHANGELOG_LINKS = {
+    changelog: 'https://github.com/orgs/ovh/projects/16/views/6?pane=info',
+    roadmap: 'https://github.com/orgs/ovh/projects/16/views/1?pane=info',
+    'feature-request':
+      'https://github.com/ovh/public-cloud-roadmap/issues/new?assignees=&labels=&projects=&template=feature_request.md&title=',
+  };
+
+  return isError ? (
```
FabienHenon: Same, you already checked that condition

packages/manager/apps/pci-project/src/data/api/pci-project.ts
StephaneRavet: ok Done.

packages/manager/apps/pci-project/src/pages/home/Header.page.tsx
StephaneRavet: ok Done.

packages/manager/apps/pci-project/src/pages/home/Header.page.tsx
StephaneRavet: ok Done.

packages/manager/apps/pci-project/src/pages/home/Header.page.tsx
StephaneRavet: ok Done.

packages/manager/apps/pci-project/src/pages/home/Header.page.tsx
```diff
@@ -0,0 +1,61 @@
...
+    <BaseLayout
+      breadcrumb={
+        <OdsBreadcrumb>
+          <OdsBreadcrumbItem href={appHomeUrl} label={projectDescription} />
+        </OdsBreadcrumb>
+      }
+    >
+      <header className="flex items-center justify-between w-full">
+        <div className="flex items-center gap-4">
+          <Title className="">{projectDescription}</Title>
```
lionel95200x:
```suggestion
          <Title>{projectDescription}</Title>
```

packages/manager/apps/pci-project/src/pages/home/Header.page.tsx
```diff
@@ -0,0 +1,61 @@
...
+import Navigation from './Navigation.component';
+import { urls } from '@/routes/routes.constant';
+import FullPageSpinner from '@/components/FullPageSpinner';
+
+export default function ProjectHeader() {
+  const { t } = useTranslation(['common', 'home', 'settings']);
+  const appHomeUrl = useProjectUrl('public-cloud') + urls.home;
+  const { data: project, isLoading, error } = useProject();
+
+  if (isLoading) return <FullPageSpinner />;
```
lionel95200x: I think we should not display a full loader on global app, but display static data immediatly and display skeleton loader only on dynamic data (grid)

FabienHenon (approved):

PR #17884, opened 2025-06-30
refactor(zimbra): use common translations module

atefzed:
ref: #MANAGER-17403

## Description

Ticket Reference: #MANAGER-17403


scripts/check-duplicated-translations.js
```diff
@@ -0,0 +1,152 @@
+const fs = require('fs');
```
helabenkhalfallah: what do you think to do as I did here: https://github.com/ovh/manager/tree/master/packages/manager-cli

since we already have a cli dedicated package. The CLI can also be helpful for other team. :)

atefzed: This script will be included in a separate pull request.

atefzed: Here is the new pull request for the cli command: https://github.com/ovh/manager/pull/17939

ghyenne (approved):

PR #17895, opened 2025-07-01
feat(kms): secret dashboard page

mathieumousnier:
Ticket Reference: #MANAGER-17850

## Description

Add the secret dashboard page, with "general information" and "settings" tiles.

2 minor bug fixes : 
- onboarding illustration svg was broken, converted it to png
- the loading state of the secrets list was not working well


packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/Listing.page.tsx
```diff
@@ -38,8 +38,8 @@ export default function SecretListingPage() {
     <BaseLayout header={{ title: t('secret_manager') }}>
       <Datagrid
         columns={columns}
-        items={secrets}
-        totalItems={secrets.length}
+        items={secrets || []}
+        totalItems={secrets?.length}
```
rjamet-ovh, line 59: it'll be better to handle error with a dedicated return, that way `secrets` will always exist on this render.
and right now we lack error notification on this page

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/Dashboard.page.spec.tsx
```diff
@@ -0,0 +1,92 @@
...
+    const { container } = await renderTestApp(mockPageUrl);
+
+    const labelOnce = [
+      // tabs
+      labels.secretManager.common.versions,
+      // general information tile
+      labels.secretManager.common.path,
+      labels.secretManager.common.urn,
+      labels.secretManager.dashboard.last_update,
+      '15/01/2023, 09:30',
```
rjamet-ovh: Maybe we should mock `useFormatDate` to avoid an hardcoded value ?

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/generalInformation/InformationsTile.component.spec.tsx
```diff
@@ -0,0 +1,30 @@
...
+    const { container } = render(<InformationsTile secret={mockSecret} />);
+
+    const labelList = [
+      'secret-manager/dashboard:general_information',
+      'path',
+      mockSecret.path,
+      'urn',
+      'secret-manager/dashboard:last_update',
+      '15/01/2023, 09:30',
+    ];
```
rjamet-ovh: I think you forgot creation date.
and another case to mock `useFormatDate` ?

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/Listing.page.tsx
mathieumousnier: These lines won't change because they handle the "intialisation" phase, where the query data is undefined.
But you're right about the error !

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/Dashboard.page.spec.tsx
mathieumousnier: I don't see real problems with hard coded values in tests.
I think it is better than writing specific logic for tests that can hide issues.
Plus it is an integration test, it is good to keep it without mockups.

How would you mock this function ?

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/Listing.page.tsx
rjamet-ovh: you're right, i thought that <Datagrid /> handled that case with the `isLoading` prop but apparently no.

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/Dashboard.page.spec.tsx
rjamet-ovh: I would use a test constant maybe? that way we could manage both the original data and its formatted value once

mathieumousnier: We discussed this a bit, and here's what I did :

I've created a useFormatDate hook for the project.
It might look like a stupid "passe plat" but it contains the business logic that defines the date format used on the project, this is missing from the hook imported from MRC. It ensures we use the same format everywhere in the project.

On the other hand, when it comes to testing, I feel that the best way to test is to check with a hard string in the test.
This allows you to test both :
- you've got the right date in input
- that you've got the right format
All this without having to create a mock and check that it's called with the right parameters.

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/generalInformation/InformationsTile.component.spec.tsx
mathieumousnier: I added the creation date

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/listing/Listing.page.tsx
mathieumousnier: I added the missing error check

packages/manager/apps/key-management-service/src/common/hooks/useFormatDate.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+import { useFormatDate } from './useFormatDate';
+
+// Mock the manager-react-components hook
+const mockFormatDateMrc = vi.fn(({ date }) => date);
+vi.mock('@ovh-ux/manager-react-components', () => ({
+  useFormatDate: () => mockFormatDateMrc,
+}));
+
+describe('useFormatDate', () => {
+  beforeEach(() => {
```
darsene:
```suggestion
  afterEach(() => {
```

packages/manager/apps/key-management-service/src/common/hooks/useFormatDate.spec.tsx
```diff
@@ -0,0 +1,51 @@
...
+      const { result } = renderHook(() => useFormatDate());
+      const formattedDate = result.current.formatDate(input, format);
+
+      expect(mockFormatDateMrc).toHaveBeenCalledWith({
+        date: input,
+        format: expectedFormatCall,
+      });
+      expect(formattedDate).toBe(input); // The mock returns input for simplicity
+
+      vi.clearAllMocks();
```
darsene:
```suggestion
```

packages/manager/apps/key-management-service/src/common/hooks/useFormatDate.tsx
```diff
@@ -0,0 +1,23 @@
+import { useFormatDate as useFormatDateMrc } from '@ovh-ux/manager-react-components';
+
+type FormatDateType = 'short' | 'long';
+
+const formatDateTable: Record<FormatDateType, string> = {
+  short: 'P', // - 04/29/1453
+  long: 'Pp', // - 04/29/1453, 12:00 AM
```
darsene:
```suggestion
  short: 'P', // - MM/dd/yyyy
  long: 'Pp', // - MM/dd/yyyy, HH:mm AM
```

packages/manager/apps/key-management-service/src/modules/secret-manager/data/api/secrets.ts
```diff
@@ -5,13 +5,27 @@ import {
   SecretVersionDataField,
 } from '@secret-manager/types/secret.type';
 
-// LIST Secret
-export const secretListQueryKey = (okmsId: string) => [
-  'secrets',
-  'list',
-  okmsId,
-];
+// Query keys for secrets
```
darsene: niptick : comments aren't very useful as the name is quite explicit

darsene:
@mathieumousnier : tests are KO

PR #17908, opened 2025-07-01
fix(mrc.datagrid): fix flex alignment

qpavy:
## Description

Fix datagrid left topbar to make it takes entire width available and center it with right topbar

Ticket Reference: #MANAGER-15793

### DIFF
--- a/packages/manager-react-components/src/components/datagrid/datagrid-topbar.component.tsx
+++ b/packages/manager-react-components/src/components/datagrid/datagrid-topbar.component.tsx
@@ -78,9 +78,15 @@ export const DatagridTopbar = <T,>({
         filtersColumns?.length > 0 ||
         topbar ||
         hasVisibilityFeature) && (
-        <div id="container" className="flex flex-wrap justify-between pb-6">
-          <div id="left-side" className="w-full md:w-auto md:order-1">
-            {topbar && <div>{topbar}</div>}
+        <div
+          id="container"
+          className="flex flex-wrap justify-between pb-6 items-center"
+        >
+          <div
+            id="left-side"
+            className="flex-1 w-full md:w-auto md:order-1 mr-4"
+          >
+            {topbar && <>{topbar}</>}
           </div>
           <div
             id="right-side"


chipp972 (dismissed):

packages/manager-react-components/src/components/datagrid/datagrid-topbar.component.tsx
```diff
@@ -78,8 +78,14 @@ export const DatagridTopbar = <T,>({
         filtersColumns?.length > 0 ||
         topbar ||
         hasVisibilityFeature) && (
-        <div id="container" className="flex flex-wrap justify-between pb-6">
-          <div id="left-side" className="w-full md:w-auto md:order-1">
+        <div
```
helabenkhalfallah, line 81: why we need two div ?

packages/manager-react-components/src/components/datagrid/datagrid-topbar.component.tsx
```diff
@@ -78,8 +78,14 @@ export const DatagridTopbar = <T,>({
...
+          id="container"
+          className="flex flex-wrap justify-between pb-6 items-center"
+        >
+          <div
+            id="left-side"
+            className="flex-1 w-full md:w-auto md:order-1 mr-4"
+          >
             {topbar && <div>{topbar}</div>}
```
helabenkhalfallah: and why we need the empty nested div?

packages/manager-react-components/src/components/datagrid/datagrid-topbar.component.tsx
qpavy: There is 2 "columns" on the topbar. Left one is for the custom node and right one of datagrid filters. Those columns are inside 1 flex container.

packages/manager-react-components/src/components/datagrid/datagrid-topbar.component.tsx
qpavy: It has been done like that originally. I removed it

helabenkhalfallah (approved):

PR #17916, opened 2025-07-02
feat(billing): show my services datagrid for us

oalkabouss:
ref: #MANAGER-18770

## Description

Ticket Reference: #MANAGER-18770


packages/manager/modules/new-billing/src/autoRenew/autorenew.routing.js
JacquesLarique, line 1: As I told you when grooming the ticket: using the feature flipping in the `redirectTo` is no longer necessary

JacquesLarique (changes requested):

PR #17928, opened 2025-07-02
feat(billing): add dataplatform

lionel95200x:
ref: #TAPC-4222

## Description

- Add Dataplatform in billing

Ticket Reference: #TAPC-4222

### DIFF
--- a/packages/manager/apps/pci-billing/src/api/hook/useConsumption.tsx
+++ b/packages/manager/apps/pci-billing/src/api/hook/useConsumption.tsx
@@ -11,9 +11,6 @@ import {
 const roundPrice = (num: number, fractionDigits = 2) =>
   Number(num.toFixed(fractionDigits));
 
-const sumPrices = <T,>(items: T[], key: string) =>
-  roundPrice(items?.reduce((sum, item) => sum + item[key], 0));
-
 const initMonthlyInstanceList = (data: TCurrentUsage) => {
   if (!data?.monthlyUsage) {
     return {
@@ -304,7 +301,9 @@ const initResourceUsage = (data: TCurrentUsage, resourceType: string) => {
     resources = reduceColdArchiveBillingInfo(resources);
   }
 
-  const totalPrice = sumPrices(resources, 'totalPrice');
+  const totalPrice = roundPrice(
+    resources.reduce((sum, item) => sum + item.totalPrice, 0),
+  );
 
   return {
     resources: resources || [],
@@ -401,6 +400,7 @@ export type TConsumptionDetail = {
   aiEndpoints: TResourceUsage[];
   notebooks: TResourceUsage[];
   aiDeploy: TResourceUsage[];
+  dataplatform: TResourceUsage[];
   coldArchive: TResourceUsage[];
   dataProcessing: TResourceUsage[];
   databases: TResourceUsage[];
@@ -418,6 +418,7 @@ export type TConsumptionDetail = {
       snapshot: number;
       volume: number;
       rancher: number;
+      dataplatform: number;
       bandwidth: number;
       privateRegistry: number;
       kubernetesLoadBalancer: number;
@@ -458,6 +459,7 @@ export const initializeTConsumptionDetail = (): TConsumptionDetail => ({
   aiEndpoints: [],
   coldArchive: [],
   dataProcessing: [],
+  dataplatform: [],
   databases: [],
   floatingIP: [],
   publicIP: [],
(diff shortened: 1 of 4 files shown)


packages/manager/apps/pci-billing/src/api/hook/useConsumption.tsx
```diff
@@ -11,8 +11,8 @@ import {
 const roundPrice = (num: number, fractionDigits = 2) =>
   Number(num.toFixed(fractionDigits));
 
-const sumPrices = <T,>(items: T[], key: string) =>
-  roundPrice(items?.reduce((sum, item) => sum + item[key], 0));
+const sumPrices = <T,>(items: T[], key: keyof T) =>
+  roundPrice(items?.reduce((sum, item) => sum + (item[key] as number), 0) || 0);
```
fredericvilcot: This cast `as number` is pretty weak, because you suppose you will receive  a number as `item[key]` and TS will believe it. 

But, what if you try to add a string or another structure with a number?

I would have added an assertion on runtime:

```suggestion

const sumPrices = <T,>(items: T[], key: keyof T): number =>
  roundPrice(
    items.reduce((sum, item) => {
      const val = item[key];
      return typeof val === 'number' ? sum + val : sum;
    }, 0),
  );
```

packages/manager/apps/pci-billing/src/components/consumption/HourlyConsumption.component.tsx
```diff
@@ -185,7 +193,7 @@ export default function HourlyConsumption({
         slot="summary"
         className="my-2"
       >
-        {`${title} (${consumption?.totals?.hourly?.[key]?.toFixed(2)} ${
+        {`${title} (${consumption.totals?.hourly?.[key]?.toFixed(2)} ${
```
fredericvilcot:
```suggestion
        {`${title} (${consumption.totals.hourly[key].toFixed(2)} ${
```

packages/manager/apps/pci-billing/src/api/hook/useConsumption.tsx
fredericvilcot: @lionel95200x 
I didn't notice this before, but this function is only intended to be used with `Object` structure as `T` generic.

if I consider that `T infer string[]`, the `key` parameter makes no sense and the function becomes unusable.

What do you think clearly exposing the structure of the expected object instead of T generic? Or maybe we could consider constraining `T` type to ensure that you have at least an object as parameter.

Something like `<T extends Record<string, number>`

packages/manager/apps/pci-billing/src/api/hook/useConsumption.tsx
```diff
@@ -11,8 +11,14 @@ import {
...
 
-const sumPrices = <T,>(items: T[], key: string) =>
-  roundPrice(items?.reduce((sum, item) => sum + item[key], 0));
+const sumPrices = (
+  items: {
+    region: string;
+    quantity: TQuantity;
+    name: string;
+    totalPrice: number;
+  }[],
```
fredericvilcot: May you please declare the type explicitly, so we can reuse it to type function parameters?

packages/manager/apps/pci-billing/src/api/hook/useConsumption.tsx
```diff
@@ -11,8 +11,14 @@ import {
...
+) => roundPrice(items?.reduce((sum, item) => sum + item.totalPrice, 0) || 0);
```
fredericvilcot:
```suggestion
) => roundPrice(items.reduce((sum, item) => sum + item.totalPrice, 0));
```

packages/manager/apps/pci-billing/src/components/consumption/HourlyConsumption.component.tsx
```diff
@@ -185,7 +193,7 @@ export default function HourlyConsumption({
         slot="summary"
         className="my-2"
       >
-        {`${title} (${consumption?.totals?.hourly?.[key]?.toFixed(2)} ${
+        {`${title} (${consumption.totals.hourly[key]?.toFixed(2)} ${
```
fredericvilcot:
```suggestion
        {`${title} (${consumption.totals.hourly[key].toFixed(2)} ${
```

SimonChaumet (approved):
Except other comments it seems good

packages/manager/apps/pci-billing/src/components/consumption/HourlyConsumption.component.tsx
```diff
@@ -173,7 +181,7 @@ export default function HourlyConsumption({
   ];
 
   const renderAccordion = (
-    key: string,
+    key: keyof typeof consumption.totals.hourly,
```
fredericvilcot, line 191: This generates a type error in line 209, in the usage of `renderAccordion`, because key is a string in the given parameter

fredericvilcot (approved):

abullet33 (approved):

PR #17934, opened 2025-07-02
feat(billing): my services page add us description

oalkabouss:
ref: #MANAGER-18769

## Description

Ticket Reference: #MANAGER-18769

### DIFF
--- a/packages/manager/modules/new-billing/src/autoRenew/services/services.controller.js
+++ b/packages/manager/modules/new-billing/src/autoRenew/services/services.controller.js
@@ -93,12 +93,6 @@ export default class ServicesCtrl {
     this.currentCriteria = JSON.parse(JSON.stringify(this.criteria));
   }
 
-  descriptionOfHeading() {
-    return this.coreConfig.getRegion() !== 'US'
-      ? this.$translate.instant('billing_description')
-      : '';
-  }
-
   getCriterionTitle(type, value) {
     return `${this.$translate.instant(
       `billing_autorenew_criterion_${type}`,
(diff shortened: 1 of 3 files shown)


packages/manager/modules/new-billing/src/autoRenew/autorenew-translation.service.js
```diff
@@ -0,0 +1,15 @@
+export default class {
+  /* @ngInject */
+  constructor($translate, coreConfig) {
+    this.$translate = $translate;
+    this.coreConfig = coreConfig;
+  }
+
+  descriptionOfHeading() {
```
JacquesLarique: Can we have a more explicit name for this function, maybe `getMyServicesPageTitle`

packages/manager/modules/new-billing/src/autoRenew/autorenew-translation.service.js
```diff
@@ -0,0 +1,15 @@
...
+    const region = this.coreConfig.getRegion();
+    const description = this.$translate.instant('billing_description');
+    const details = this.$translate.instant('billing_description_details');
+
+    return region !== 'US' ? `${description} ${details}` : description;
```
JacquesLarique: This could be improved to avoid unnecessary computation:
```suggestion
    const region = this.coreConfig.getRegion();
    const description = this.$translate.instant('billing_description');
    if (region === 'US') return description;
    
    const details = this.$translate.instant('billing_description_details');

    return `${description} ${details}`;
```

_This change is not mandatory, since the method is called only once_

packages/manager/modules/new-billing/src/autoRenew/services/services.controller.js
```diff
@@ -94,9 +96,7 @@ export default class ServicesCtrl {
   }
 
   descriptionOfHeading() {
```
JacquesLarique: This method is not used, you can remove it from here (it seems I forgot to clean it up when slicing the services page from the autorenew "container")

PR #17944, opened 2025-07-02
fix(sap-features-hub): fix initial step select errors

pauldkn:
## Description

- add error when an expired service is selected
- fix a bug on default select value
- group hooks at the top of the component

Ticket Reference: #MANAGER-18621

### DIFF
--- a/packages/manager/apps/sap-features-hub/src/utils/selectValues.ts
+++ b/packages/manager/apps/sap-features-hub/src/utils/selectValues.ts
@@ -3,15 +3,15 @@ export const getSelectDefaultValue = <T extends string | number>(
   options: T[] | undefined,
 ): string | undefined => (options?.includes(value) ? `${value}` : undefined);
 
-export const getSelectLatestValue = ({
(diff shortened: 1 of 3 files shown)


packages/manager/apps/sap-features-hub/src/utils/selectValues.ts
```diff
@@ -3,15 +3,15 @@ export const getSelectDefaultValue = <T extends string | number>(
   options: T[] | undefined,
 ): string | undefined => (options?.includes(value) ? `${value}` : undefined);
 
-export const getSelectLatestValue = ({
+export const getSelectLatestValue = <T extends string | number>({
```
tibs245, line 6:
```suggestion
export const getSelectLatestValue = <TValue extends string | number>({
```

Just as suggestion for readability

pauldkn: I personally prefer using `T` in this kind of scenario — I find it the clearest way to indicate that it's a generic type

PR #17980, opened 2025-07-04
feat(billing): enrich storage billing with additional info

lionel95200x:
ref: #TAPC-3709

## Description

- Add bucket name and type to storage

Ticket Reference: #TAPC-3709

### DIFF
--- a/packages/manager/apps/pci-billing/src/components/consumption/useObjectStorageListColumns.tsx
+++ b/packages/manager/apps/pci-billing/src/components/consumption/useObjectStorageListColumns.tsx
@@ -86,7 +86,7 @@ export function useObjectStorageListColumns() {
       cell: (row: TStorage) => (
         <div className="flex gap-2">
           <DataGridTextCell>
-            {`${(row.retrievalFees.totalPrice.value || 0).toFixed(2)} ${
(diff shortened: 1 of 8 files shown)


fredericvilcot (approved):

packages/manager/apps/pci-billing/src/components/consumption/useObjectStorageListColumns.tsx
```diff
@@ -69,5 +81,18 @@ export function useObjectStorageListColumns() {
       ),
       label: t('cpbc_object_storage_col_output_traffic'),
     },
+    {
+      id: 'retrievalFees',
+      cell: (row: TStorage) => (
+        <div className="flex gap-2">
+          <DataGridTextCell>
+            {`${(row.retrievalFees.totalPrice.value || 0).toFixed(2)} ${
```
fredericvilcot: AFAIS, `value` is always a `number` so only `0` could be falsy here, so I'm not sure fallback value is needed:


```suggestion
            {`${(row.retrievalFees.totalPrice.value).toFixed(2)} ${
```

SimonChaumet (approved):

PR #17991, opened 2025-07-04
feat(account-creation): data usage consent modal

JacquesLarique:
## Description

- Created a new modal to ask customer consent to use their data (agnostic of the ShellContext)
- Integrated the new modal in the account-creation app
- Updated ovh-at-internet to not break when we configure it with a user without a legalform

Ticket Reference: #MANAGER-17746

### DIFF
--- a/packages/manager/core/test-utils/src/utils/ui-test-helpers-ods17.ts
+++ b/packages/manager/core/test-utils/src/utils/ui-test-helpers-ods17.ts
@@ -56,7 +56,7 @@ export const getButtonByLabel = async ({
     );
     button = Array.from(buttonList).filter((btn) =>
       [label, altLabel].includes(btn.textContent),
-    )[nth];
+    )[nth] as HTMLElement;
     return disabled
       ? expect(button).toHaveAttribute('disabled')
       : expect(button).not.toHaveAttribute('disabled');
--- a/packages/manager/core/test-utils/src/utils/ui-test-helpers-ods18.ts
+++ b/packages/manager/core/test-utils/src/utils/ui-test-helpers-ods18.ts
@@ -1,5 +1,5 @@
 import React from 'react';
-import { OsdsIcon } from '@ovhcloud/ods-components/react';
+import { OdsIcon } from '@ovhcloud/ods-components/react';
 import {
   screen,
   waitFor,
@@ -50,7 +50,7 @@ type GetOdsButtonParams = {
   container: HTMLElement;
   label?: string;
   altLabel?: string;
-  iconName?: React.ComponentProps<typeof OsdsIcon>['name'];
+  iconName?: React.ComponentProps<typeof OdsIcon>['name'];
   disabled?: boolean;
   isLink?: boolean;
   nth?: number;


packages/manager/apps/account-creation/src/context/tracking/useTracking.ts
```diff
@@ -0,0 +1,10 @@
+import { useContext } from 'react';
+import trackingContext, { TrackingContext } from './tracking.context';
+
+export const useTrackingContext = (): TrackingContext => {
+  const context = useContext(trackingContext);
+  if (context === undefined) {
+    throw new Error('useTrackingContext must be used within a TackingProvider');
```
aottr, line 7: Thanks !

packages/manager/apps/account-creation/src/index.tsx
```diff
@@ -8,10 +8,12 @@ import {
 } from '@ovh-ux/manager-config';
 import '@ovh-ux/manager-react-components/dist/style.css';
 import { defineApplicationVersion } from '@ovh-ux/request-tagger';
+import '@ovh-ux/manager-gcj-module/dist/style.css';
```
aottr, line 11: Are you sure you want to import from `/dist/` ?

packages/manager/apps/account-creation/src/pages/accountType/AccountType.page.tsx
```diff
@@ -25,28 +23,22 @@ export default function AccountType() {
...
-    null,
-  );
   const [legalFormError, setLegalFormError] = useState<boolean>(false);
 
-  const validateStep = () => {
-    if (selectedLegalForm === null) {
+  const validateStep = useCallback(() => {
+    if (!legalForm) {
       setLegalFormError(true);
+    } else if (country === 'FR' && legalForm !== 'individual') {
```
aottr, line 36: Do we also want to navigate to the company search for an administration ?

packages/manager/modules/gcj/src/components/data-usage-policy/modal/DataUsagePolicyModal.component.tsx
```diff
@@ -0,0 +1,126 @@
...
+import {
+  Links,
+  Modal,
+  OvhSubsidiary,
+  Subtitle,
+} from '@ovh-ux/manager-react-components';
+import { useDataUsageConsent } from '../../../hooks/dataUsageConsent/useDataUsageConsent';
+import { useDataUsageInfoLink } from '../../../hooks/dataUsageConsent/useDataUsageLink';
+
+import ovhCloudLogo from '../../../../public/assets/logo-ovhcloud.png';
```
aottr: maybe you want to set a path resolver in the tsconfig to not have such deep relative imports.

packages/manager/modules/gcj/src/components/data-usage-policy/modal/DataUsagePolicyModal.component.tsx
```diff
@@ -0,0 +1,126 @@
...
+  trackingPlugin,
+  onModalHidden,
+}: Props) => {
+  const { t } = useTranslation('data-usage-policy');
+  const { shouldRequestConsent, setConsent } = useDataUsageConsent(
+    region,
+    trackingPlugin,
+  );
+  const moreInfoLink = useDataUsageInfoLink(
+    subsidiary as OvhSubsidiary,
```
aottr: Not sure this is the best approach. Subsidary is potentially `undefined`. You assert it as enum into the `useDataUsageInfoLink` function. You basically "overwrite" Typescripts concerns with this assertion but the enum doesn't cover this. 

The enum has a `DEFAULT` fallback you can use if the subsidary is `undefined` and then in the hook you check for the value `DEFAULT` of the enum and return the default for the region in this case. (as an example)

packages/manager/modules/gcj/src/components/data-usage-policy/modal/DataUsagePolicyModal.component.tsx
```diff
@@ -0,0 +1,126 @@
...
+          </div>
+          <ul>
+            <li>
+              <OdsText preset={ODS_TEXT_PRESET.paragraph}>
+                {t('cookie_policy_description_3')}
+              </OdsText>
+            </li>
+          </ul>
+          <OdsText preset={ODS_TEXT_PRESET.paragraph}>
+            {t('cookie_policy_description_2')}
```
aottr, line 91: The ordering is interesting, did this change over time or why is (2) after (3) ?

packages/manager/modules/gcj/src/components/data-usage-policy/modal/translations/Messages_de_DE.json
```diff
@@ -0,0 +1,12 @@
+{
+  "cookie_policy_accept": "Annehmen",
+  "cookie_policy_refuse": "Fortfahren ohne Annahme",
```
aottr, line 3: very bad translation...

better:
```
"cookie_policy_accept": "Akzeptieren", // or "Zustimmen"
"cookie_policy_refuse": "Weiter ohne zu akzeptieren", // or "Ablehnen"
 ```

aottr: **why is it bad?**

*Annahme* means *Assumption* in most cases. And on a cookie banner we don't want to assume or continue without an assumption...

packages/manager/apps/account-creation/src/pages/accountType/AccountType.page.tsx
JacquesLarique: Yes, and associations as well, but only for french residents

packages/manager/modules/gcj/src/components/data-usage-policy/modal/translations/Messages_de_DE.json
JacquesLarique: I just took it back from existing translation :s

packages/manager/modules/gcj/src/components/data-usage-policy/modal/DataUsagePolicyModal.component.tsx
JacquesLarique: Look's like it was created this way https://github.com/ovh/manager/commit/dd2dcd75c03a6e25e69af454311d5dc49aa4d5ec#diff-d923798727f035e7d163b70503e8e0cbec45481dfbea7a67302b1588c58c4306

aottr (approved):

PR #18001, opened 2025-07-07
feat(savings-plan): enrich stats value and add strict mode

lionel95200x:
## Description

- Replace the 0 value instead of "Non disponible" info
- Add strict mode on savings plan

Ticket Reference: #TAPC-4392


packages/manager/apps/pci-savings-plan/src/hooks/useSavingsPlan.tsx
```diff
@@ -198,3 +200,13 @@ export const useSavingsPlanContract = () => {
     queryFn: () => getSavingsPlanContracts(serviceId),
   });
 };
+
+export const useSavingsPlanId = (): string => {
+  const { savingsPlanId } = useParams();
+
+  if (!savingsPlanId) {
+    throw new Error('Missing ProjectId');
```
fredericvilcot:
```suggestion
    throw new Error('Missing savingsPlanId');
```

packages/manager/apps/pci-savings-plan/src/hooks/useSavingsPlan.tsx
```diff
@@ -198,3 +200,13 @@ export const useSavingsPlanContract = () => {
     queryFn: () => getSavingsPlanContracts(serviceId),
   });
 };
+
+export const useSavingsPlanId = (): string => {
```
fredericvilcot: As you are using the same hook for another param, I'll suggest to add some generic custom hook, something like that:

```ts
const useParam = (param: string): string => {
  const { [param]: paramValue } = useParams();
  if (!paramValue) {
    throw new Error(`Missing ${param} in URL.`);
  }
  return paramValue;
}
```

packages/manager/apps/pci-savings-plan/src/hooks/useService.tsx
```diff
@@ -6,9 +6,9 @@ const getServices = async (projectId: string): Promise<number[]> => {
   return data;
 };
 
-export const useServices = ({ projectId }: { projectId: string }) => {
-  return useQuery({
+export const useServices = ({ projectId }: { projectId: string }) =>
+  useQuery({
     queryKey: ['services', projectId],
     queryFn: () => getServices(projectId),
+    enabled: !!projectId,
```
fredericvilcot: Here, do you want to test length or nullity ?

```ts
projectId.length > 0
```

packages/manager/apps/pci-savings-plan/src/pages/layout.tsx
```diff
@@ -49,7 +49,7 @@ export default function Layout() {
   } = useSavingsPlan();
 
   if (isError || isErrorSavingsPlan || error) {
-    return <ErrorBanner error={error?.response} />;
+    return <ErrorBanner error={error?.response ?? {}} />;
```
fredericvilcot, line 52: Instead of having a simple assertion, we could enhance by using some type guard to capture the error before giving it through props

packages/manager/apps/pci-savings-plan/src/hooks/useSavingsPlan.tsx
```diff
@@ -79,11 +80,12 @@ const postSavingsPlan = async ({
...
 
 export const useServiceId = () => {
-  const { projectId } = useParams();
+  const projectId = useProjectId();
   const { data: services } = useServices({
     projectId,
   });
-  return services?.[0];
+
+  return services?.[0] || 0;
```
fredericvilcot: **Question**: Do we want to return an arbitrary `0` (that can cause some unexpected errors) or an explicit `null` to be checked at higher level?

fredericvilcot (changes requested):

packages/manager/apps/pci-savings-plan/src/hooks/useSavingsPlan.tsx
lionel95200x: Yes you're right, do you think we can add it in pci-common ?

fredericvilcot: @lionel95200x For sure 👍🏻

packages/manager/apps/pci-savings-plan/src/pages/layout.tsx
fredericvilcot:
```suggestion
    return <ErrorBanner error={error.response ?? {}} />;
```

fredericvilcot (approved):

packages/manager/apps/pci-savings-plan/src/components/CreatePlanForm/CreatePlanFormDetails.tsx
SimonChaumet, line 1: Shouldn't it be removed instead of emptied ?

packages/manager/apps/pci-savings-plan/src/utils/test/test.provider.tsx
```diff
@@ -23,7 +22,17 @@ i18n.use(initReactI18next).init({
...
+  new QueryClient({
+    defaultOptions: {
+      queries: {
+        retry: false,
+      },
+    },
+  });
+
 const Wrappers = ({ children }: { children: React.ReactNode }) => {
+  const [queryClient] = useState(createTestQueryClient);
```
SimonChaumet: seems more like a ref isn't it ?

SimonChaumet (approved):
nice clean

fredericvilcot (approved):

PR #18016, opened 2025-07-07
chore(kms): improve types strictness on nullables

mathieumousnier:
ref: #MANAGER-18393

## Description

Ticket Reference: #...


packages/manager/apps/key-management-service/src/data/hooks/useIdentity.ts
```diff
@@ -29,17 +29,17 @@ const useIdentityList = <T>(
...
-    isFetched,
   } = useQuery<ApiResponse<string[]>, ApiError>({
     queryKey: idsQueryKeyFn(),
     queryFn: idsQueryFn,
   });
 
   const detailsQueries = useQueries({
     queries:
-      isFetched && listData?.data.length > 0
+      !isListPending && !isListError
```
rjamet-ovh, line 42: what happen when the list is empty?

packages/manager/apps/key-management-service/src/pages/credential/create/CreateAddIdentities.component.tsx
```diff
@@ -63,9 +63,9 @@ const CreateAddIdentities = ({
         : [
             ...userList.map((user) => user.urn),
             ...groupList.map((group) => group.urn),
-            ...serviceAccountList.map(
-              (serviceAccount) => serviceAccount.identity,
-            ),
+            ...serviceAccountList
+              .map((serviceAccount) => serviceAccount.identity)
+              .filter((identity) => identity !== null),
```
rjamet-ovh, line 68: why add this filter?

packages/manager/apps/key-management-service/src/pages/credential/create/identities/IdentitiesSelectedBase.component.tsx
```diff
@@ -14,25 +14,27 @@ import {
   IdentityOauthClient,
 } from '@/types/identity.type';
 
-type IdentitiesSelectedBaseProps = {
+type IdentityType = IdentityUser | IdentityGroup | IdentityOauthClient;
+
+type IdentitiesSelectedBaseProps<T extends IdentityType> = {
```
rjamet-ovh, line 19: is the `extends` necessary here?

packages/manager/apps/key-management-service/src/pages/credential/create/identities/tile/IdentitiesBaseTile.component.tsx
```diff
@@ -4,7 +4,7 @@ import { OdsCard, OdsText } from '@ovhcloud/ods-components/react';
 
 type IdentitiesBaseTileProps = {
   title: string;
-  urn: string;
+  urn?: string;
```
rjamet-ovh: it seems that this props is never used

packages/manager/apps/key-management-service/src/pages/dashboard/generalInformations/GeneralInformations.tsx
```diff
@@ -13,12 +13,14 @@ import RestApiTile from '@/components/layout-helpers/Dashboard/GeneralInformatio
...
-  const { okmsId } = useParams();
+  const { okmsId } = useParams() as { okmsId: string };
   const { data: okms, error, isLoading: isOkmsLoading } = useOkmsById(okmsId);
   const {
     data: okmsService,
     isLoading: isOkmsServiceLoading,
-  } = useServiceDetails({ resourceName: okms?.data.id });
+    // This has been refactored and removed by https://github.com/ovh/manager/pull/17804
+    // eslint-disable-next-line @typescript-eslint/no-non-null-asserted-optional-chain, @typescript-eslint/no-non-null-assertion
+  } = useServiceDetails({ resourceName: okms?.data.id! });
```
rjamet-ovh: don't forget to fix that after the release

packages/manager/apps/key-management-service/src/data/hooks/useIdentity.ts
mathieumousnier: The list can not be `undefined`.
This is the beauty of using `isPending` and `isError` with a proper typescript configuration.
With `react-query`, if you ensure that `isPending=false` and `isError=false`, then you can be sure that the data has been fetched. (which is not true if you use `isLoading` instead of `isPending`.

When typescript is configured with strict=true, you see the type of the query data being narrowed and going from `MyFetchedData | undefined` to `MyFetchedData`

rjamet-ovh: I meant when the list is an empty array.

you removed this rule:
`listData?.data.length > 0`

packages/manager/apps/key-management-service/src/pages/credential/create/CreateAddIdentities.component.tsx
mathieumousnier: The type of `IdentityOauthClient` specifies `indentity` as `string | null`.
And we don't want null values to be kept in the array.

```ts
export type IdentityOauthClient = {
  callbackUrls: string[];
  clientId: string;
  createdAt: string;
  description: string;
  flow: IdentityOauthClientFlow;
  identity: string | null;
  name: string;
};
```

packages/manager/apps/key-management-service/src/pages/credential/create/identities/IdentitiesSelectedBase.component.tsx
mathieumousnier: The `extends IdentityType` part  restricts T so that it must be a subtype of IdentityType.
Otherwise we can put `string` or anything else in the `items` prop .

packages/manager/apps/key-management-service/src/pages/credential/create/identities/tile/IdentitiesBaseTile.component.tsx
mathieumousnier: Mmh yes, I have seen that but it is called several times, I removed it ✅

packages/manager/apps/key-management-service/src/pages/credential/create/identities/IdentitiesSelectedBase.component.tsx
rjamet-ovh: it does not work like that ?
```
type IdentitiesSelectedBaseProps = {
 title: string;
  identityURNs: string[];
  addCallback: () => void;
  addButtonLabel: string;
  deleteCallback: () => void;
  datagridColumns: DatagridColumn<T>[];
  items: IdentityType[];
};
```

packages/manager/apps/key-management-service/src/data/hooks/useIdentity.ts
mathieumousnier: When `listData.data === []` ?

This will return `listData.data.map()` which will be an empty array, do you see something wrong ?

```ts
!isListPending && !isListError
  ? listData.data.map(
      (id): UseQueryOptions<ApiResponse<T>, ApiError> => ({
        retry: false,
        queryKey: detailsQueryKeyFn(id),
        queryFn: () => detailsQueryFn(id),
      }),
    )
  : [],
```

packages/manager/apps/key-management-service/src/pages/credential/create/identities/IdentitiesSelectedBase.component.tsx
mathieumousnier: This works well you're right (with replacing `DatagridColumn<T>`), and is simplier.
But my solution made me look smarter 😂

I'll make the change
 
```ts 
type IdentityType = IdentityUser | IdentityGroup | IdentityOauthClient;

type IdentitiesSelectedBaseProps = {
  title: string;
  identityURNs: string[];
  addCallback: () => void;
  addButtonLabel: string;
  deleteCallback: () => void;
  datagridColumns: DatagridColumn<IdentityType>[];
  items: IdentityType[];
};
```

packages/manager/apps/key-management-service/src/data/hooks/useIdentity.ts
rjamet-ovh: seems ok, i just wanted to make sure that removing this rule would not cause any regressions

packages/manager/apps/key-management-service/src/pages/credential/create/identities/IdentitiesSelectedBase.component.tsx
mathieumousnier: After reconsideration, this is needed with `"strict"="true"`

rjamet-ovh (approved):

darsene (approved):

darsene (approved):

PR #18020, opened 2025-07-07
refactor(zimbra): use generic modal from MRC

atefzed:
ref: #MANAGER-17543

## Description

The goal of this PR is to migrate all modals to the generic modal from Manager React Components (MRC)

Ticket Reference: #MANAGER-17543

## Additional Information
The Modal component was previously exported directly using React.forwardRef, like so:
```js
export const Modal = React.forwardRef(ModalComponent);
```
This caused an issue where the component was typed as any, resulting in the loss of TypeScript features like autocompletion, type checking, and intellisense support when consuming the component.

To fix this, I explicitly typed the export using React.ForwardRefExoticComponent, combining React.PropsWithoutRef<ModalProps> and React.RefAttributes<HTMLOdsModalElement>:
```js
export const Modal: React.ForwardRefExoticComponent<
  React.PropsWithoutRef<ModalProps> & React.RefAttributes<HTMLOdsModalElement>
> = React.forwardRef(ModalComponent);
```


packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -28,6 +28,8 @@ export interface ModalProps {
   isPrimaryButtonDisabled?: boolean;
   /** Action of primary button */
   onPrimaryButtonClick?: () => void;
+  /** Test id of primary button */
+  primaryTestId?: string;
```
helabenkhalfallah: since we add new props we should not cover these case especially the if, else this will impact the coverage and the code will be fragile for future change.

atefzed: done

anooparveti: **nitpick:** With ODS v19, it will be easy to find the buttons with text. So, i think these props can be redundant for MRC v3.

anooparveti (changes requested):

packages/manager-react-components/src/components/Modal/Modal.component.tsx
helabenkhalfallah: but they dont use ODS 19 actually and MRC v3 is not yet ready.

anooparveti:
```suggestion
  primaryButtonTestId?: string;
```

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -36,6 +38,8 @@ export interface ModalProps {
   isSecondaryButtonLoading?: boolean;
   /** Is disabled state for secondary button */
   onSecondaryButtonClick?: () => void;
+  /** Test id of primary button */
+  secondaryTestId?: string;
```
anooparveti:
```suggestion
  secondaryButtonTestId?: string;
```

anooparveti (changes requested):

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -36,6 +38,8 @@ export interface ModalProps {
   isSecondaryButtonLoading?: boolean;
   /** Is disabled state for secondary button */
   onSecondaryButtonClick?: () => void;
+  /** Test id of primary button */
```
anooparveti:
```suggestion
  /** Test id of secondary button */
```

ghyenne (approved):

PR #18023, opened 2025-07-07
feat(pci-instances): add distant backup

SimonChaumet:
ref: #TAPC-3053

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-instances/public/translations/actions/Messages_de_DE.json
+++ b/packages/manager/apps/pci-instances/public/translations/actions/Messages_de_DE.json
@@ -62,7 +62,7 @@
   "pci_instances_actions_backup_instance_distant_label": "Ein Remote-Backup hinzufügen (optional)",
   "pci_instances_actions_backup_instance_distant_name_label": "Name des Remote-Backups",
   "pci_instances_actions_backup_instance_distant_region_label": "Wählen Sie einen Standort aus.",
-  "pci_instances_actions_backup_instance_distant_region_price": "Für jedes Backup fallen Gebühren an: ~{{price}} € inkl. MwSt./Monat/GB",
+  "pci_instances_actions_backup_instance_distant_region_price": "Jedes Backup wird berechnet: ~{{price}}/Monat/GB",
   "pci_instances_actions_backup_instance_region_enable_warning": "Der ausgewählte Standort ist nicht aktiviert. Wenn Sie auf „Bestätigen“ klicken, wird er Ihrem Projekt hinzugefügt.",
   "pci_instances_actions_backup_instance_with_distant_success_message": "Das Backup {{ name }} wird erstellt. Sobald dieses lokale Backup erstellt wurde, erstellen wir das Remote-Backup mit dem Namen {{ distantName }}. Um sich den Status und die Details des Vorgangs anzusehen, wechseln Sie zur Seite <InstanceBackupLink>Instance Backup</InstanceBackupLink>. Wenn das Backup nicht sofort angezeigt wird, können Sie Ihre Seite aktualisieren, damit auch die Liste aktualisiert wird."
 }
--- a/packages/manager/apps/pci-instances/public/translations/actions/Messages_en_GB.json
+++ b/packages/manager/apps/pci-instances/public/translations/actions/Messages_en_GB.json
@@ -62,7 +62,7 @@
   "pci_instances_actions_backup_instance_distant_label": "Add a remote backup (Option)",
   "pci_instances_actions_backup_instance_distant_name_label": "Remote backup name",
   "pci_instances_actions_backup_instance_distant_region_label": "Select a region",
-  "pci_instances_actions_backup_instance_distant_region_price": "Each backup will be billed at: ~ £{{ price }} ex. VAT/month/GB",
+  "pci_instances_actions_backup_instance_distant_region_price": "Each backup will be billed at: ~{{price}}/month/GB",
   "pci_instances_actions_backup_instance_region_enable_warning": "The selected location is not currently active. Click ‘confirm’ to add it to your project.",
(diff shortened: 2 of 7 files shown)


packages/manager/apps/pci-instances/src/components/modal/Modal.component.tsx
```diff
@@ -52,35 +57,40 @@ const Modal = ({
...
-            hue={ODS_TEXT_COLOR_HUE._800}
-            level={ODS_TEXT_LEVEL.heading}
-          >
-            {title}
-          </OsdsText>
-          {children}
-        </div>
-        <DialogFooter className="flex justify-end p-6 pt-0">
-          <DialogClose asChild>
+        <Wrapper>
```
SimonChaumet, line 49: You can use "hide whitespace" github feature to get a better display here

packages/manager/apps/pci-instances/tsconfig.json
```diff
@@ -2,7 +2,7 @@
   "compilerOptions": {
     "target": "ES2020",
     "useDefineForClassFields": true,
-    "lib": ["ES2020", "DOM", "DOM.Iterable"],
+    "lib": ["ESNext", "DOM", "DOM.Iterable"],
```
SimonChaumet, line 4: Required for https://github.com/ovh/manager/pull/18023/files?diff=unified&w=1#diff-f1d3b50251c02f1b9e00633bf4f86085349446f4d9c189604cdaa168f26fa39fR120

(As well as the new typescript version)

packages/manager/apps/pci-instances/src/data/api/instance.ts
```diff
@@ -110,13 +110,19 @@ export const backupInstance = ({
...
   instanceId,
   snapshotName,
+  distantSnapshotName,
+  disantRegion,
 }: {
   projectId: string;
   instanceId: string;
   snapshotName: string;
+  distantSnapshotName?: string;
+  disantRegion?: string;
```
aTurmo: Should'nt it be "distantRegion" instead of "disantRegion" ? I think it's missing a 't'

packages/manager/apps/pci-instances/src/data/hooks/instance/action/useInstanceBackupPrice.ts
```diff
@@ -38,8 +54,42 @@ export const useInstanceBackupPrice = (projectId: string, region: string) => {
...
+  const currentRegion = useMemo(
+    () => currentPlan?.regions.find((r) => r.name === region),
+    [currentPlan, region],
+  );
+
+  const distantContinents = useMemo<Map<string, ContinentRegion[]>>(() => {
+    if (
+      !currentPlan ||
+      !currentRegion ||
+      (currentRegion.type !== 'region' && currentRegion.type !== 'region-3-az')
```
aTurmo: For this kind of mutiple comparison, you could do something like 
`
!['region', 'region-3-az'].includes(currentRegion.type)
`
It's optionnal, but I find it more readable and easier to add/remove cases if needs be.

SimonChaumet: I'm not a big fan of this when comparing enum types because it removes the completion so you could set a value that isn't possible without seeing it

aTurmo: you could even put the `['region', 'region-3-az'].includes(currentRegion.type)` in a function and reuse it line 75

SimonChaumet: It is a weird case, these the calls are the same but they don't check the same thing : here we do not display any distant datacenter for LZ, and below we remove LZ from being used as a distant region. In the future we might have LZ opened as a distant backup zone while it couldn't have distant backups

In fact this condition should be handled by the backend, but as now we don't have anything to check which datacenter is available for this

packages/manager/apps/pci-instances/src/data/api/instance.ts
SimonChaumet: Good catch, fixed

packages/manager/apps/pci-instances/src/data/hooks/instance/action/useInstanceBackupPrice.ts
aTurmo: Mmmh I see. So right now it's the same filter but it might not be.
But personaly I prefer to handle things as they are and not as they might be because the potential futur may never happen.

And for the typing of the array problem, you can type it if you extract it to a const first like `const specialRegions: Region[]  = ['region', 'region-3-az'];` (I dont know the enum type I just assumed it).

But anyway it's just 2 comparisaons, it does not change much one way or the other.

You can resolve the conversation.

packages/manager/apps/pci-instances/src/input-validation/index.ts
```diff
@@ -0,0 +1,75 @@
...
+import * as core from 'zod/v4/core';
+import { NAMESPACES } from '@ovh-ux/manager-common-translations';
+import { TFunction } from 'i18next';
+import { useTranslation } from 'react-i18next';
+import { useMemo } from 'react';
+
+/**
+ * This string is required with good naming value
+ */
+const string = (t: TFunction<[typeof NAMESPACES.FORM]>) => () =>
```
aTurmo, line 19: Isn't this a TODO you left for yourself ?
Anyway it's wierd to have a const named 'string'

SimonChaumet: It is not a TODO, it is the description of the descriptive call, like in zod or yup, since it is used for input validation

selm3n (approved):

PR #18030, opened 2025-07-08
feat(pci-instances): add new dashboard page

Tsiorifamonjena:
## Description

Ticket Reference:  #TAPC-4388

### DIFF
--- a/packages/manager/apps/pci-instances/src/components/tab/TabsPanel.component.tsx
+++ b/packages/manager/apps/pci-instances/src/components/tab/TabsPanel.component.tsx
@@ -1,0 +1,53 @@
+import { FC, useMemo } from 'react';
+import { NavLink, useLocation } from 'react-router-dom';
+import { Badge } from '@datatr-ux/uxlib';
+import { clsx } from 'clsx';
+
+type TTabItem = {
+  label: string;
+  to: string;
+  badge?: string;
+};
+
+type TTabsPanelProps = {
+  tabs: TTabItem[];
+};
+
+const TabsPanel: FC<TTabsPanelProps> = ({ tabs }) => {
+  const location = useLocation();
+
+  const activeTab = useMemo(
+    () => tabs.find((tab) => location.pathname === tab.to),
+    [location.pathname, tabs],
+  );
+
+  return (
+    <div className="flex border-b border-[#e6f5fc]">
+      {tabs.map((tab) => (
+        <NavLink
+          key={tab.to}
+          to={tab.to}
+          className={clsx(
+            'whitespace-nowrap w-fit text-[--ods-color-primary-500] text-base font-semibold m-0 py-2 hover:text-primary-700',
+            activeTab?.to === tab.to &&
+              'border-b-2 border-[--ods-color-primary-500]',
+          )}
+        >
+          <span className="w-full px-6 flex gap-2 items-center">
+            {tab.label}
+            {tab.badge && (
+              <Badge
+                variant="primary"
+                className="hidden md:block text-xs rounded-full"
+              >
+                {tab.badge}
+              </Badge>
+            )}
+          </span>
+        </NavLink>
+      ))}
+    </div>
+  );
+};
+
+export default TabsPanel;
--- a/packages/manager/apps/pci-instances/src/data/hooks/instance/selectors/instances.selector.ts
+++ b/packages/manager/apps/pci-instances/src/data/hooks/instance/selectors/instances.selector.ts
@@ -72,7 +72,7 @@ const getActionHrefByName = (
   { region, id }: Pick<TAggregatedInstance, 'id' | 'region'>,
 ): TAggregatedInstanceAction['link'] => {
   if (name === 'details') {
-    return { path: id, isExternal: false };
+    return { path: `${id}?region=${region}`, isExternal: false };
   }
 
   if (name === 'edit') {
--- a/packages/manager/apps/pci-instances/src/pages/instances/datagrid/components/cell/NameIdCell.component.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/datagrid/components/cell/NameIdCell.component.tsx
@@ -11,15 +11,19 @@ type TNameIdCellProps = {
   isLoading: boolean;
 };
 
-export const NameIdCell: FC<TNameIdCellProps> = ({ isLoading, instance }) => (
-  <LoadingCell isLoading={isLoading}>
-    <OsdsLink
-      color={ODS_THEME_COLOR_INTENT.primary}
-      className={'block'}
-      href={useHref(instance.id)}
-    >
-      {instance.name}
-    </OsdsLink>
-    <BaseTextCell>{instance.id}</BaseTextCell>
-  </LoadingCell>
-);
+export const NameIdCell: FC<TNameIdCellProps> = ({ isLoading, instance }) => {
+  const detailHref = useHref(`${instance.id}?region=${instance.region}`);
+
+  return (
+    <LoadingCell isLoading={isLoading}>
+      <OsdsLink
+        color={ODS_THEME_COLOR_INTENT.primary}
+        className={'block'}
+        href={detailHref}
+      >
+        {instance.name}
+      </OsdsLink>
+      <BaseTextCell>{instance.id}</BaseTextCell>
+    </LoadingCell>
+  );
+};
--- a/packages/manager/apps/pci-instances/src/pages/instances/instance/Instance.page.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/instance/Instance.page.tsx
@@ -1,13 +1,68 @@
-import { PageLayout } from '@ovh-ux/manager-react-components';
-import { FC } from 'react';
+import {
+  ChangelogButton,
+  PageLayout,
+  PciGuidesHeader,
+  Notifications,
+} from '@ovh-ux/manager-react-components';
+import { FC, useMemo } from 'react';
+import { Outlet, useResolvedPath, useRouteLoaderData } from 'react-router-dom';
+import { useTranslation } from 'react-i18next';
+import { TProject } from '@ovh-ux/manager-pci-common';
+import { Breadcrumb } from '@/components/breadcrumb/Breadcrumb.component';
 import { GoBack } from '@/components/navigation/GoBack.component';
+import TabsPanel from '@/components/tab/TabsPanel.component';
 import InstanceWrapper from './InstanceWrapper.page';
+import { CHANGELOG_LINKS } from '@/constants';
 
-const Instance: FC = () => (
-  <InstanceWrapper>
-    <PageLayout>
-      <GoBack />
-    </PageLayout>
-  </InstanceWrapper>
-);
+const Instance: FC = () => {
+  const { t } = useTranslation('dashboard');
+  const project = useRouteLoaderData('root') as TProject;
+
+  const dashboardPath = useResolvedPath('');
+  const vncPath = useResolvedPath('vnc');
+
+  const tabs = useMemo(
+    () => [
+      {
+        label: t('pci_instances_dashboard_tab_info_title'),
+        to: dashboardPath.pathname,
+      },
+      {
+        label: t('pci_instances_dashboard_tab_vnc_title'),
+        to: vncPath.pathname,
+      },
+    ],
+    [],
+  );
+
+  return (
+    <InstanceWrapper>
+      <PageLayout>
+        <Breadcrumb
+          projectLabel={project.description ?? ''}
+          items={[{ label: '' }]}
+        />
+        <div className="header mt-8">
+          <div className="flex items-center justify-between">
+            <div className="flex-[0.8]">{/* instance name */}</div>
+            <div className="flex gap-x-3">
+              <ChangelogButton links={CHANGELOG_LINKS} />
+              <PciGuidesHeader category="instances" />
+            </div>
+          </div>
+        </div>
+        <div className="mt-8 mb-8">
+          <Notifications />
+        </div>
+        <GoBack />
+        <div className="mt-8">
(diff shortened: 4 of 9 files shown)


packages/manager/apps/pci-instances/src/routes/loaders/instanceLegacy.loader.ts
```diff
@@ -0,0 +1,20 @@
...
+  const { instanceId } = params;
+  const region = searchParams.get('region') ?? null;
+
+  if (
+    pathname === `/pci/projects/${params.projectId}/instances/${instanceId}`
+  ) {
+    return redirect(`../region/${region}/instance/${instanceId}`);
+  }
+
+  return { notFoundAction: true };
```
fredericvilcot: Not sure you want to return `notFoundAction` property..

packages/manager/apps/pci-instances/src/routes/loaders/instanceLegacy.loader.ts
```diff
@@ -0,0 +1,20 @@
...
+  const url = new URL(request.url);
+  const { pathname, searchParams } = url;
+
+  const { instanceId } = params;
+  const region = searchParams.get('region') ?? null;
+
+  if (
+    pathname === `/pci/projects/${params.projectId}/instances/${instanceId}`
+  ) {
+    return redirect(`../region/${region}/instance/${instanceId}`);
```
fredericvilcot: What if `instanceId` is `undefined` ?

packages/manager/apps/pci-instances/src/routes/routes.tsx
```diff
@@ -126,6 +128,10 @@ const routes: RouteObject[] = [
           import('@/pages/instances/create/CreateInstance.page'),
         ),
       },
+      {
+        path: SECTIONS.instanceLegacy,
```
fredericvilcot, line 137: You should also declare new routes, if not, they will be considered as 404

packages/manager/apps/pci-instances/src/routes/routes.tsx
```diff
@@ -51,7 +52,8 @@ export const INSTANCE_PATH = 'instance/:instanceId';
 export const SECTIONS = {
   onboarding: 'onboarding',
   new: 'new',
-  instance: ':instanceId',
+  instance: 'region/:regionId/instance/:instanceId',
```
fredericvilcot: this is not a section, you should consider doing the same as action routes in line 95

packages/manager/apps/pci-instances/src/components/tab/TabsPanel.component.tsx
```diff
@@ -0,0 +1,49 @@
...
+import { Badge } from '@datatr-ux/uxlib';
+import { clsx } from 'clsx';
+
+type TTabItem = {
+  label: string;
+  to: string;
+  count?: number;
+};
+
+const TabsPanel: FC<{ tabs: TTabItem[] }> = ({ tabs }) => {
```
fredericvilcot: Please, declare an explicit props type :)

packages/manager/apps/pci-instances/src/components/tab/TabsPanel.component.tsx
```diff
@@ -0,0 +1,49 @@
...
+          to={tab.to}
+          className={clsx(
+            'whitespace-nowrap w-fit text-[--ods-color-primary-500] text-base font-semibold m-0 py-2 hover:text-primary-700',
+            activeTab?.to === tab.to &&
+              'border-b-2 border-[--ods-color-primary-500]',
+          )}
+        >
+          <span className="w-full px-6 flex gap-2 items-center">
+            {tab.label}
+            {!!tab.count && (
```
fredericvilcot:
```suggestion
            {tab.count && (
```

packages/manager/apps/pci-instances/src/components/tab/TabsPanel.component.tsx
```diff
@@ -0,0 +1,49 @@
+import { FC, useMemo } from 'react';
+import { NavLink, useLocation } from 'react-router-dom';
+import { Badge } from '@datatr-ux/uxlib';
+import { clsx } from 'clsx';
+
+type TTabItem = {
+  label: string;
+  to: string;
+  count?: number;
```
fredericvilcot: I would have rather named it `badge` as it's much more understandable than a count :)

packages/manager/apps/pci-instances/src/data/hooks/instance/selectors/instances.selector.ts
```diff
@@ -72,7 +72,7 @@ const getActionHrefByName = (
   { region, id }: Pick<TAggregatedInstance, 'id' | 'region'>,
 ): TAggregatedInstanceAction['link'] => {
   if (name === 'details') {
-    return { path: id, isExternal: false };
+    return { path: `region/${region}/instance/${id}`, isExternal: false };
```
fredericvilcot: Please, as we made for action urls, use legacy routes because we need to handle hybridation ;)

packages/manager/apps/pci-instances/src/pages/instances/datagrid/components/cell/NameIdCell.component.tsx
```diff
@@ -11,15 +11,21 @@ type TNameIdCellProps = {
...
-      href={useHref(instance.id)}
-    >
-      {instance.name}
-    </OsdsLink>
-    <BaseTextCell>{instance.id}</BaseTextCell>
-  </LoadingCell>
-);
+export const NameIdCell: FC<TNameIdCellProps> = ({ isLoading, instance }) => {
+  const detailHref = useHref(
+    `region/${instance.region}/instance/${instance.id}`,
```
fredericvilcot: Same here, use legacy routes :)

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/DashboardCardLayout.component.tsx
```diff
@@ -0,0 +1,30 @@
+import { FC, PropsWithChildren } from 'react';
+import { OsdsDivider, OsdsText } from '@ovhcloud/ods-components/react';
+import { ODS_TEXT_SIZE, ODS_TEXT_LEVEL } from '@ovhcloud/ods-components';
+import {
+  ODS_THEME_COLOR_HUE,
+  ODS_THEME_COLOR_INTENT,
+} from '@ovhcloud/ods-common-theming';
+import { Card, CardHeader, CardContent } from '@datatr-ux/uxlib';
+
+const DashboardCardLayout: FC<PropsWithChildren<{ title: string }>> = ({
```
fredericvilcot: Please, declare an explicit props type

fredericvilcot (changes requested):

packages/manager/apps/pci-instances/src/routes/loaders/instanceLegacy.loader.ts
Tsiorifamonjena: I'll redirect the user to an 404 then. `instanceId` is mandatory to be in the dashboard. BTW, maybe if `region` is also `null` or `undefined` this behaviour is also the right way to handle that. What do you think ?

packages/manager/apps/pci-instances/src/routes/loaders/instanceLegacy.loader.ts
Tsiorifamonjena: Yes, you're right. I won't return a property but directly redirect to an 404. How about that ?

packages/manager/apps/pci-instances/src/routes/routes.tsx
Tsiorifamonjena: I dont sure to understand. The new routes is already declare as I now. In line 136. Or you are talking about another ?

packages/manager/apps/pci-instances/src/routes/loaders/instanceLegacy.loader.ts
fredericvilcot: IMO, we should avoid redirecting to a 404 url as you lose the current page information, particularly the current URL.

I will keep the original implementation that returns a dedicated property that is used into the page to render a specific content, like an `Error` component but still ion the same page ;)

My comment was mainly related to the property itself that must not be the same for action part ;)

packages/manager/apps/pci-instances/src/routes/loaders/instanceLegacy.loader.ts
fredericvilcot: For `region`,  it's acceptable to put `null` in the url as it's well handled by the back-end if you have to handle some HTTP call with region as route parameter.

For `instanceId`, like we made for actions, I will return a router data object in the same way we do if assertion if unmet.

packages/manager/apps/pci-instances/src/routes/routes.tsx
fredericvilcot: Sorry, I didn't noticed it as it has only been replaced in the section part.

packages/manager/apps/pci-instances/src/routes/routes.tsx
Tsiorifamonjena: BTW, what do you mean in `section` ? like a section for instances page (listing) ?

packages/manager/apps/pci-instances/src/routes/loaders/instanceLegacy.loader.ts
Tsiorifamonjena: Ok I see. it's related by your previous comment https://github.com/ovh/manager/pull/18030#discussion_r2192373318 then

packages/manager/apps/pci-instances/src/routes/routes.tsx
fredericvilcot: Well this is a good question, IMO a `section` in the µapp is much more related to a piece of path, not the full path itself.

It has been designed to handle actions, appended at the last section of the url :)

So to keep consistency, I'd put put this pathname into a dedicated path constant, to be compliant with what has been done in line 95 :)

packages/manager/apps/pci-instances/src/components/tab/TabsPanel.component.tsx
Tsiorifamonjena: It will display the 0 if count is 0. It's not the goal

packages/manager/apps/pci-instances/src/routes/routes.tsx
Tsiorifamonjena: I agree with you, later it will be useful for all `actions` in the dashboard page :)

packages/manager/apps/pci-instances/src/components/tab/TabsPanel.component.tsx
fredericvilcot: Of course ;)

packages/manager/apps/pci-instances/src/routes/loaders/instanceLegacy.loader.ts
Tsiorifamonjena: After I got into the code, I think instanceId will never be undefined since this loader can be triggered only by this path route :instanceId (line 137)

packages/manager/apps/pci-instances/src/components/tab/TabsPanel.component.tsx
Tsiorifamonjena: I replace the name and the type to string to be more generic. `count` is effectively not more clear. Like that, badge can contain another information

packages/manager/apps/pci-instances/src/routes/loaders/instanceLegacy.loader.ts
fredericvilcot: Yes you're right :)

packages/manager/apps/pci-instances/src/data/hooks/instance/selectors/instances.selector.ts
```diff
@@ -72,7 +72,7 @@ const getActionHrefByName = (
   { region, id }: Pick<TAggregatedInstance, 'id' | 'region'>,
 ): TAggregatedInstanceAction['link'] => {
   if (name === 'details') {
-    return { path: id, isExternal: false };
+    return { path: `${id}?region=${region}`, isExternal: false };
```
SimonChaumet, line 75: Why not use the same redirect method as datagrid modals ?

packages/manager/apps/pci-instances/src/pages/instances/instance/Instance.page.tsx
```diff
@@ -1,13 +1,68 @@
...
+  return (
+    <InstanceWrapper>
+      <PageLayout>
+        <Breadcrumb
+          projectLabel={project.description ?? ''}
+          items={[{ label: '' }]}
+        />
+        <div className="header mt-8">
+          <div className="flex items-center justify-between">
+            <div className="flex-[0.8]">{/* instance name */}</div>
```
SimonChaumet, line 47: still WIP ?

fredericvilcot (approved):
Great work, thanks!

packages/manager/apps/pci-instances/src/data/hooks/instance/selectors/instances.selector.ts
Tsiorifamonjena: It is already handle by the same redirection logic (a react-router-dom `loader` catch this). this path is the legacy but I had to add the query params to send the `region`

packages/manager/apps/pci-instances/src/pages/instances/instance/Instance.page.tsx
Tsiorifamonjena: Ah it's an oversight sorry. I let it because it will be change by this https://github.com/ovh/manager/pull/18047

PR #18046, opened 2025-07-08
feat(manager-react-components): migrate Headers component to ODS v19

anooparveti:
## Description

Migrate Headers component to ODS v19
Add unit and snapshot test-cases
Move Headers component from content folder to components folder

Ticket Reference: #MANAGER-18748    

## Additional Information

BREAKING CHANGE: 

1. Headers component is not exposed from MRC and now only an internal component
2. Rename headerButton prop to guideButton
3. Remove description and subtitle from Header and is now directly set at BaseLayout


packages/manager-react-components/src/components/templates/base/headers/__tests__/Headers.snapshot.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+import { Headers } from '../Headers.component';
+import { IamAuthorizationResponse } from '../../../../../hooks/iam/iam.interface';
+import { useAuthorizationIam } from '../../../../../hooks/iam';
+import {
+  GuideButton,
+  ActionMenu,
+  ChangelogButton,
+} from '../../../../navigation';
+
+export const header = () => <Headers title="Example for header" />;
```
helabenkhalfallah: why an export from snapshot file?

packages/manager-react-components/src/components/templates/base/headers/__tests__/Headers.snapshot.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+export const headerWithHeaderButtons = () => (
```
helabenkhalfallah: why an export from snapshot file?

packages/manager-react-components/src/components/templates/base/headers/__tests__/Headers.snapshot.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+          changelog: 'https://ovh.com/',
+          roadmap: 'https://ovh.com/',
+          'feature-request': 'https://ovh.com/',
+        }}
+      />
+    }
+  />
+);
+
+export const headerWithActions = () => (
```
helabenkhalfallah: why an export from snapshot file?

packages/manager-react-components/src/components/templates/base/headers/__tests__/Headers.snapshot.test.tsx
```diff
@@ -0,0 +1,108 @@
...
+
+vitest.mock('../../../../../hooks/iam');
+vitest.mock('@ovh-ux/manager-react-shell-client', () => ({
+  useOvhTracking: () => ({
+    trackClick: vi.fn(),
+  }),
+}));
+vitest.mock('storybook-addon-react-router-v6', () => ({
+  withRouter: () => vi.fn(),
+}));
```
helabenkhalfallah: I think these mock can be global in setup to avoid having the same thing in each component

packages/manager-wiki/stories/manager-react-components/templates/Base/base.stories.tsx
```diff
@@ -16,8 +16,11 @@ import {
   GuideItem,
```
helabenkhalfallah, line 16: the name should be `Base.stories.tsx`

anooparveti: This is the migration for Headers component. It will be handled with the migration of Baslayout component.

packages/manager-react-components/src/components/templates/base/headers/__tests__/Headers.snapshot.test.tsx
anooparveti: Well it was a redundant mock and i have removed it.

packages/manager-wiki/stories/manager-react-components/templates/Base/base.stories.tsx
helabenkhalfallah: Okay!

PR #18047, opened 2025-07-08
feat(pci-instances): display instance name and allow updating if edit is enabled

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-4006

### DIFF
--- a/packages/manager/apps/pci-instances/src/components/input/InputCancellable.component.tsx
+++ b/packages/manager/apps/pci-instances/src/components/input/InputCancellable.component.tsx
@@ -1,0 +1,69 @@
+import { DetailedHTMLProps, InputHTMLAttributes, FC } from 'react';
+import { Input as DatatrInput } from '@datatr-ux/uxlib';
+import { OsdsButton, OsdsIcon } from '@ovhcloud/ods-components/react';
+import {
+  ODS_BUTTON_SIZE,
+  ODS_BUTTON_VARIANT,
+  ODS_ICON_NAME,
+  ODS_ICON_SIZE,
+} from '@ovhcloud/ods-components';
+import { ODS_THEME_COLOR_INTENT } from '@ovhcloud/ods-common-theming';
+import { clsx } from 'clsx';
+
+type TInputCancellableProps = {
+  onClear?: () => void;
+  onSubmit?: () => void;
+  isValid?: boolean;
+} & Omit<
+  DetailedHTMLProps<InputHTMLAttributes<HTMLInputElement>, HTMLInputElement>,
+  'ref'
+>;
+
+export const InputCancellable: FC<TInputCancellableProps> = ({
+  onClear,
+  onSubmit,
+  isValid = true,
+  className,
+  ...props
+}) => (
+  <form className="flex">
+    <DatatrInput
+      className={clsx(
+        'focus-visible:ring-transparent focus-visible:bg-primary-300 text-[var(--ods-typography-heading-500-font-size)]',
+        className,
+      )}
+      {...props}
+    />
+    <div className="flex">
+      <OsdsButton
+        aria-label="Cancel"
+        color={ODS_THEME_COLOR_INTENT.primary}
+        variant={ODS_BUTTON_VARIANT.ghost}
+        size={ODS_BUTTON_SIZE.sm}
+        onClick={onClear}
+      >
+        <OsdsIcon
+          name={ODS_ICON_NAME.CLOSE}
+          color={ODS_THEME_COLOR_INTENT.primary}
+          size={ODS_ICON_SIZE.xs}
+        />
+      </OsdsButton>
+      <OsdsButton
+        aria-label="Submit"
+        color={ODS_THEME_COLOR_INTENT.primary}
+        variant={ODS_BUTTON_VARIANT.ghost}
+        size={ODS_BUTTON_SIZE.sm}
+        {...(!isValid && { disabled: true })}
+        onClick={onSubmit}
+      >
+        <OsdsIcon
+          name={ODS_ICON_NAME.CHECK}
+          color={ODS_THEME_COLOR_INTENT.primary}
+          size={ODS_ICON_SIZE.xs}
+        />
+      </OsdsButton>
+    </div>
+  </form>
+);
+
+export default InputCancellable;
--- a/packages/manager/apps/pci-instances/src/components/input/InputCancellable.spec.tsx
+++ b/packages/manager/apps/pci-instances/src/components/input/InputCancellable.spec.tsx
@@ -1,0 +1,56 @@
+import { describe, it, vi, expect } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
+import InputCancellable from './InputCancellable.component';
+
+const handleClear = vi.fn();
+const handleValidate = vi.fn();
+const handleChange = vi.fn();
+
+const InputComponent = ({ isValid }: { isValid?: boolean }) => (
+  <InputCancellable
+    value="NAME"
+    onChange={handleChange}
+    onClear={handleClear}
+    onSubmit={handleValidate}
+    isValid={isValid}
+  />
+);
+
+describe('Input Component', () => {
+  it('should render input with its default value', async () => {
+    render(<InputComponent />);
+
+    expect(screen.getByDisplayValue('NAME')).toBeInTheDocument();
+  });
+
+  it('should call onChange props', async () => {
+    render(<InputComponent />);
+
+    await userEvent.type(screen.getByRole('textbox'), 'new_text');
+
+    expect(handleChange).toHaveBeenCalled();
+  });
+
+  it('should call handleClear when cancel', async () => {
+    render(<InputComponent />);
+
+    await userEvent.click(screen.getByLabelText('Cancel'));
+
+    expect(handleClear).toHaveBeenCalled();
+  });
+
+  it('should call handleValid', async () => {
+    render(<InputComponent />);
+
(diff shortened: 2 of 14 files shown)


packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.spec.tsx
```diff
@@ -0,0 +1,65 @@
...
+import queryClient from '@/queryClient';
+
+const projectId = '8c8c4fd6d4414aa29fc777752b00005198664';
+const instanceId = 'fake-id';
+const region = 'fake-region';
+const fakeInstance = { id: instanceId };
+const editInstanceNameMock = vi.fn();
+
+vi.mock('@/data/api/instance');
+vi.mocked(getInstance as Mock).mockResolvedValue(fakeInstance);
```
SimonChaumet: cast should be on value, not on the function

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.spec.tsx
```diff
@@ -0,0 +1,65 @@
...
+          instanceId,
+          callbacks: {
+            onSuccess: mockSuccess,
+            onError: mockError,
+          },
+        }),
+      { wrapper },
+    );
+
+    result.current.mutate({ instanceName: 'new_instanceName' });
```
SimonChaumet: Shouldn't this be wrapped in a "act" ?

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.spec.tsx
Tsiorifamonjena: The type error is still catch even cast on value

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.spec.tsx
Tsiorifamonjena: I think it's no longer necessary since during the assertion we use `await waitFor`

Tsiorifamonjena:
I put this PR on draft because we are not yet able to merge it onto the target. A refactor is in progress

packages/manager/apps/pci-instances/src/components/input/InputCancellable.component.tsx
```diff
@@ -0,0 +1,69 @@
+import { DetailedHTMLProps, InputHTMLAttributes, FC } from 'react';
+import { Input as DatatrInput } from '@datatr-ux/uxlib';
```
SimonChaumet, line 2: Doesn't migration to ODS 19 has been done ?

packages/manager/apps/pci-instances/src/components/input/InputCancellable.spec.tsx
```diff
@@ -0,0 +1,56 @@
...
+  it('should render input with its default value', async () => {
+    render(<InputComponent />);
+
+    expect(screen.getByDisplayValue('NAME')).toBeInTheDocument();
+  });
+
+  it('should call onChange props', async () => {
+    render(<InputComponent />);
+
+    await userEvent.type(screen.getByRole('textbox'), 'new_text');
```
SimonChaumet, line 30: Is there a way to have a label for accessibility ?

packages/manager/apps/pci-instances/src/components/input/editableContent/EditableContent.component.tsx
```diff
@@ -0,0 +1,63 @@
...
+          autoFocus
+          onChange={(e) => setText(e.target.value)}
+          onClear={handleCancel}
+          onSubmit={handleSubmit}
+        />
+      ) : (
+        <div className="flex">
+          {text}
+          <OsdsButton
+            aria-label="Edit"
```
SimonChaumet, line 52: no translation ?

packages/manager/apps/pci-instances/src/components/input/editableContent/EditableContent.spec.tsx
```diff
@@ -0,0 +1,40 @@
...
+
+  it('Should display input when on edit mode', async () => {
+    render(<EditableContent defaultValue={value} onSubmit={handleValidate} />);
+
+    await act(() => userEvent.click(screen.getByLabelText('Edit')));
+
+    const input = screen.getByDisplayValue(value);
+    await waitFor(() => expect(input).toBeInTheDocument());
+
+    fireEvent.change(screen.getByRole('textbox'), {
```
SimonChaumet: why not use userEvent here ?

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.spec.tsx
SimonChaumet: If there is an error it means that there is an issue with the cast

Cast is allowed when there are missing properties, if there is an error it means that both types are incompatible

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.spec.tsx
SimonChaumet: So you could wrap it in act and remove the waitFor instead ?

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.ts
```diff
@@ -37,3 +41,35 @@ export const useInstance = ({
...
+type TUpdateInstanceNameArgs = {
+  projectId: string;
+  instanceId: string;
+  callbacks: TUpdateInstanceCallbacks;
+};
+
+export const useUpdateInstanceName = ({
+  projectId,
+  instanceId,
+  callbacks = {},
```
SimonChaumet, line 65: TUpdateInstanceCallbacks is not nullable, can you update the type ?

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/InstanceName.component.tsx
```diff
@@ -0,0 +1,67 @@
...
+};
+
+const InstanceName: FC<TInstanceNameProps> = ({ instance, region }) => {
+  const { t } = useTranslation('dashboard');
+  const { addError } = useNotifications();
+  const projectId = useProjectId();
+
+  const handleSuccessUpdate = useCallback(
+    (value: string) => {
+      updateDashboardCache({
```
SimonChaumet, line 23: I think cache should be updated from the mutation, not from the view

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/useDashboard.ts
```diff
@@ -0,0 +1,54 @@
...
+
+export const useDashboard = ({ region, instanceId }: TUseDashboardArgs) => {
+  const { data: instance, isPending } = useInstance({
+    region,
+    instanceId,
+    params: ['withBackups', 'withImage', 'withNetworks', 'withVolumes'],
+  });
+
+  return {
+    instance: selectInstanceDashboard(instance),
```
SimonChaumet: should be wrapped in a useMemo

packages/manager/apps/pci-instances/src/components/input/InputCancellable.component.tsx
Tsiorifamonjena: The migration is work in progress. The plan is to let it be merged onto `master` and then rebase my epic branch when ready

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.spec.tsx
Tsiorifamonjena: You are right, my bad. I thought you were talking about casting with `Mock` interface

packages/manager/apps/pci-instances/src/components/input/editableContent/EditableContent.spec.tsx
Tsiorifamonjena: the difference here is that with `fireEvent.change` it will change(replace) `the default text` with `new_text` but yes I can also use `userEvent.type` and expect a default value + new text `(test-textnew_text)` in my case.
But I think both is correct

packages/manager/apps/pci-instances/src/components/input/InputCancellable.spec.tsx
Tsiorifamonjena: Testing accessibility for this component is already cover by `EditableContent.spec`

packages/manager/apps/pci-instances/src/components/input/editableContent/EditableContent.component.tsx
Tsiorifamonjena: Nice catch, I'll do it later in a dedicated PR with removing all duplicated translations

packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.ts
Tsiorifamonjena: Done ;)

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/InstanceName.component.tsx
Tsiorifamonjena: I think in this case, it's more flexible to let it as is because if I update cache directly there, the mutation must depend on `region`. This is not what we want, this method is generic and not regionalised. It can be called by listing for example.

PR #18071, opened 2025-07-10
feat(manager-react-components): migrate Price component

anooparveti:
## Description

Migrate Price component to ODS v19
Refactor to split child components
Enhance test-coverage

Ticket Reference: #MANAGER-18749


packages/manager-react-components/src/components/price/Price.component.tsx
```diff
@@ -0,0 +1,144 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { Text } from '@ovhcloud/ods-react';
+
+import { IntervalUnitType } from '../../enumTypes';
+import { PriceProps } from './Price.props';
```
helabenkhalfallah, line 6: mmm we don't have '.props' on our static kit, you need to add it perhaps not? https://github.com/ovh/manager/blob/2c9c77796e9038230a80cab968969beb149a9024/packages/manager/tools/static-analysis-kit/src/configs/naming-conventions-config.ts

packages/manager-react-components/src/components/price/Price.component.tsx
```diff
@@ -0,0 +1,144 @@
...
+import {
+  getPrice,
+  convertIntervalPrice,
+  getPriceTextFormatted,
+  checkAsianFormat,
+  checkGermanFormat,
+  checkFranceFormat,
+  checkUSFormat,
+} from './Price.utils';
```
helabenkhalfallah, line 15: we don't have '.utils' on our static kit, you need to add it perhaps not? https://github.com/ovh/manager/blob/2c9c77796e9038230a80cab968969beb149a9024/packages/manager/tools/static-analysis-kit/src/configs/naming-conventions-config.ts

packages/manager-react-components/src/components/price/Price.component.tsx
```diff
@@ -0,0 +1,144 @@
...
+import './translations/translations';
```
helabenkhalfallah: we defined translation like this https://github.com/ovh/manager/blob/2c9c77796e9038230a80cab968969beb149a9024/packages/manager/tools/static-analysis-kit/src/configs/naming-conventions-config.ts#L16

should we add new rule also?

packages/manager-react-components/src/components/price/Price.component.tsx
```diff
@@ -0,0 +1,144 @@
...
+
+import { PriceText, PRICE_TEXT_PRESET } from './price-text';
```
helabenkhalfallah, line 18: why the import from 'price-text' and does not follow our guideline: 
https://github.com/ovh/manager/blob/2c9c77796e9038230a80cab968969beb149a9024/packages/manager/tools/static-analysis-kit/src/configs/naming-conventions-config.ts#L16

packages/manager-react-components/src/components/price/Price.component.tsx
```diff
@@ -0,0 +1,144 @@
...
+  value,
+  intervalUnit,
+  tax,
+  ovhSubsidiary,
+  locale,
+  isConvertIntervalUnit,
+}: Readonly<PriceProps>) {
+  const { t } = useTranslation('price');
+
+  const isAsiaFormat = checkAsianFormat(ovhSubsidiary);
```
helabenkhalfallah, line 30: why a utils by language instead of an Enum and a Factory or Builder?

packages/manager-react-components/src/components/price/Price.component.tsx
```diff
@@ -0,0 +1,144 @@
...
+  const isFrenchFormat = checkFranceFormat(ovhSubsidiary);
+  const isUSFormat = checkUSFormat(ovhSubsidiary);
+
+  const convertedValue = isConvertIntervalUnit
+    ? convertIntervalPrice(value, intervalUnit)
+    : value;
+
+  const convertedTax = isConvertIntervalUnit
+    ? convertIntervalPrice(tax || 0, intervalUnit)
+    : tax || 0;
```
helabenkhalfallah: nested ternary

packages/manager-react-components/src/components/price/Price.component.tsx
```diff
@@ -0,0 +1,144 @@
...
+    locale,
+    getPrice(convertedValue, convertedTax),
+  );
+
+  const intervalUnitText =
+    intervalUnit && intervalUnit !== IntervalUnitType.none
+      ? t(`price_per_${intervalUnit}`)
+      : '';
+
+  const components = [
```
helabenkhalfallah, line 60: what do you think to create a separated utils and replace all calculation inside, and use pattern matching instead of array find: 
https://github.com/helabenkhalfallah/dsa-toolbox/blob/main/src/functional/pattern-matching/Match.ts#L38
https://github.com/helabenkhalfallah/dsa-toolbox/blob/main/src/functional/pattern-matching/Match-test.ts#L7

helabenkhalfallah: also some conditions can be merged.

packages/manager-react-components/src/components/price/Price.utils.ts
```diff
@@ -59,3 +50,19 @@ export const getPriceTextFormatted = (
     }).format(priceValue);
   }
 };
+
+export const checkAsianFormat = (subsidiary: string) => {
```
helabenkhalfallah, line 54: same as https://github.com/ovh/manager/pull/18071/files#r2207299236

packages/manager-react-components/src/components/price/__tests__/Price.spec.tsx
```diff
@@ -1,9 +1,30 @@
 import React from 'react';
-import { screen } from '@testing-library/react';
-import Price from './price.component';
-import { render } from '../../../utils/test.provider';
+import { vitest } from 'vitest';
+import { screen, render } from '@testing-library/react';
+import Price from '../Price.component';
 import { IntervalUnitType, OvhSubsidiary } from '../../../enumTypes';
 
+vitest.mock('react-i18next', () => ({
```
helabenkhalfallah, line 7: this mock can be global in setup perhaps.

packages/manager-react-components/src/components/price/__tests__/Price.utils.spec.ts
```diff
@@ -0,0 +1,126 @@
+import { vi, describe, it, expect } from 'vitest';
+import {
+  getPrice,
+  convertIntervalPrice,
+  getPriceTextFormatted,
+} from '../Price.utils';
+import { IntervalUnitType, OvhSubsidiary } from '../../../enumTypes';
+
+// Mock OVH_CURRENCY_BY_SUBSIDIARY
+vi.stubGlobal('OVH_CURRENCY_BY_SUBSIDIARY', {
```
helabenkhalfallah: you can use setup test for global mock

packages/manager-react-components/src/components/price/price-text/PriceText.component.tsx
```diff
@@ -0,0 +1,31 @@
...
+import clsx from 'clsx';
+import { PriceTextProps, PRICE_TEXT_PRESET } from './PriceText.props';
+
+export const PriceText: React.FC<PriceTextProps> = ({
+  preset = PRICE_TEXT_PRESET.BASE,
+  price,
+  intervalUnitText,
+  label,
+}) => (
+  <span
```
helabenkhalfallah: we don't have component from ODS?

packages/manager-react-components/src/components/price/Price.component.tsx
anooparveti: Not required. This was for importing a JS file and i have changed it to index.ts instead of `translations.ts`

packages/manager-react-components/src/components/price/Price.component.tsx
anooparveti: Not sure to understand you. Pricetext is a child component of Price component to simplify the text display.

packages/manager-react-components/src/components/price/Price.component.tsx
anooparveti: In your example, you're always depending on single variable which is not the case here.
But we can still combine them by making an object and if you think it's good, we can handle it with different ticket as it's probably out of scope of the ODS v19 migration.

packages/manager-react-components/src/components/price/Price.component.tsx
anooparveti: We can create a ticket to enhance the static kit to support these extensions depending on our new component development guidelines.

packages/manager-react-components/src/components/price/Price.component.tsx
anooparveti: We can create a ticket to enhance the static kit to support these extensions depending on our new component development guidelines.

packages/manager-react-components/src/components/price/Price.component.tsx
anooparveti: Agreed we can plan this refactoring outside of ODS migration.
Ticket: MANAGER-19103

packages/manager-react-components/src/components/price/__tests__/Price.spec.tsx
anooparveti: Screenshot 2025-07-15 at 7 50 53 PM

Moving this to global mock results in a failure on many other test. We can plan a separate task for that maybe.

packages/manager-react-components/src/components/price/Price.component.tsx
helabenkhalfallah: I'll add it to actual in progress PR

packages/manager-react-components/src/components/price/Price.component.tsx
helabenkhalfallah: I'll add it to actual in progress PR

packages/manager-react-components/src/components/price/Price.component.tsx
helabenkhalfallah: OK.

packages/manager-react-components/src/components/price/Price.component.tsx
helabenkhalfallah: Yes we can create another tiket and refine it ;)

packages/manager-react-components/src/components/price/__tests__/Price.spec.tsx
helabenkhalfallah: ok

helabenkhalfallah: OK

PR #18073, opened 2025-07-10
fix(ips): fix type issue for servermodel

sachinrameshn:
ref: #MANAGER-18395

## Description

Ticket Reference: #...


packages/manager/apps/ips/src/data/api/get/dedicatedServerVmac.ts
```diff
@@ -39,9 +39,12 @@ export const getdedicatedServerVmac = async (
     )}/virtualMac`,
     page: 1,
   });
-
 export type ServerModelsType = {
-  data: any;
+  apiVersion: string;
+  apis: any[];
```
qpavy: Better put something else than any

packages/manager/apps/ips/src/data/api/get/dedicatedServerVmac.ts
```diff
@@ -39,9 +39,12 @@ export const getdedicatedServerVmac = async (
...
+  basePath: string;
+  models: Record<string, any>;
```
qpavy: Better use something else than any

packages/manager/apps/ips/src/data/api/get/dedicatedServerVmac.ts
sachinrameshn: Can yiu check now

qpavy (approved):

PR #18079, opened 2025-07-10
feat(manager-react-component): add badge within the header of base la…

seven-amid:
…yout

ref: #MANAGER-18077

## Description

add a prop badge within the header of base layout

Ticket Reference: #...

### DIFF
--- a/packages/manager-react-components/src/components/content/headers/headers.component.tsx
+++ b/packages/manager-react-components/src/components/content/headers/headers.component.tsx
@@ -1,9 +1,15 @@
-import React from 'react';
-import { OdsText } from '@ovhcloud/ods-components/react';
-import { Title, Subtitle } from '../../typography';
+import React, { ReactNode } from 'react';
+import { OdsBadge, OdsText } from '@ovhcloud/ods-components/react';
+import { ODS_BADGE_COLOR, ODS_BADGE_SIZE } from '@ovhcloud/ods-components';
+import { Subtitle } from '../../typography';
 
 export interface HeadersProps {
   title?: string;
+  badge?: {
+    color: ODS_BADGE_COLOR;
+    size: ODS_BADGE_SIZE;
+    label: string;
+  };
   subtitle?: string;
   description?: string;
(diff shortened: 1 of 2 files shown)


packages/manager-react-components/src/components/content/headers/headers.component.tsx
```diff
@@ -1,9 +1,10 @@
-import React from 'react';
+import React, { ReactNode } from 'react';
 import { OdsText } from '@ovhcloud/ods-components/react';
-import { Title, Subtitle } from '../../typography';
+import { Subtitle } from '../../typography';
 
 export interface HeadersProps {
   title?: string;
+  badge?: ReactNode;
```
anooparveti: **nitpick:** Instead of directly accepting the ReactNode, accept the required info in the form of Object and use Badge component from ODS in the Header component itself.

```suggestion
  badge?:{
    color: BADGE_COLOR,
    size: BADGE_SIZE,
    label: string
  };
```

anooparveti (changes requested):

packages/manager-react-components/src/components/content/headers/headers.component.tsx
seven-amid: Fixed

packages/manager-react-components/src/components/content/headers/headers.component.tsx
```diff
@@ -20,7 +27,16 @@ export const Headers: React.FC<HeadersProps> = ({
   return (
     <div className="flex items-start justify-between">
       <div>
-        {title && <Title>{title}</Title>}
+        <div className="flex items-center gap-4 mb-[24px]">
```
aboungnaseng-ovhcloud, line 30: Why not adding badge below title without wrapping it in div with class, we have already validate the css rules with design team

aboungnaseng-ovhcloud:
can you add a story to illustrate it please ? 🙏

aboungnaseng-ovhcloud (approved):

PR #18085, opened 2025-07-10
feat(pci-common): add useParam hook & test suite

fredericvilcot:
## Description

Ticket Reference: #...


packages/manager/modules/manager-pci-common/src/hooks/url/useParam.ts
```diff
@@ -0,0 +1,16 @@
+import { useParams } from 'react-router-dom';
+
+export const useParam = <T extends string>(
+  ...params: T[]
+): Record<T, string> => {
+  const urlParams = useParams();
+
+  return params.reduce((acc, param) => {
+    const value = urlParams[param];
+    if (!value) {
```
frenautvh: what about params whose value equals 0 ?

packages/manager/modules/manager-pci-common/src/hooks/url/useParam.ts
```diff
@@ -0,0 +1,16 @@
+import { useParams } from 'react-router-dom';
+
+export const useParam = <T extends string>(
```
frenautvh, line 3: what is this hook achieving that is different from using useParams directly ?

fredericvilcot: @frenautvh The hook ensures that `params` are `string` to use all across apps, and prevent from manipulating `undefined` values

packages/manager/modules/manager-pci-common/src/hooks/url/useParam.ts
fredericvilcot: Yes you're right, I modified the assertion to handle falsy values :)

frenautvh (approved):

PR #18089, opened 2025-07-10
fix(kms): set kms datagrid page to 100

mathieumousnier:
## Description

Ticket Reference: #MANAGER-19075


mathieumousnier:
this was a temporary patch for a MRC bug that has been fix

PR #18091, opened 2025-07-10
feat(pci-private-registry): add 3AZ deployment to europe and canada

lolottetheclash:
ref: #TAPC-3197,  #TAPC-4531

## Description

Ticket Reference: ...

### DIFF
--- a/packages/manager/apps/pci-private-registry/src/pages/create/steps/PlanStep.tsx
+++ b/packages/manager/apps/pci-private-registry/src/pages/create/steps/PlanStep.tsx
@@ -129,21 +129,24 @@ export default function PlanStep({
       >
         {tUpgrade('private_registry_upgrade_plan_description')}
       </OsdsText>
-      {isPending && (
-        <div className="mt-5">
-          <OsdsSpinner inline />
-        </div>
-      )}
-      {!isPending && store.state.region && (
-        <PlanChooser
-          plan={store.state.plan}
-          plans={
-            capabilities.find((c) => c.regionName === store.state.region.name)
-              .plans
-          }
-          onInput={(value) => store.set.plan(value)}
-        />
-      )}
+      <>
+        {isPending && (
+          <div className="mt-5">
+            <OsdsSpinner inline />
+          </div>
+        )}
+        {!isPending && store.state.region && (
+          <PlanChooser
+            plan={store.state.plan}
+            plans={
+              capabilities?.find(
+                (c) => c.regionName === store.state.region.name,
+              )?.plans
+            }
+            onInput={(value) => store.set.plan(value)}
+          />
+        )}
+      </>
       <OsdsText
         color={ODS_THEME_COLOR_INTENT.text}
         level={ODS_THEME_TYPOGRAPHY_LEVEL.body}
--- a/packages/manager/apps/pci-private-registry/src/pages/create/steps/PlanStep.tsx
+++ b/packages/manager/apps/pci-private-registry/src/pages/create/steps/PlanStep.tsx
@@ -19,8 +19,9 @@ import {
 } from '@ovhcloud/ods-components';
 import { useTranslation, Translation } from 'react-i18next';
 import { useContext } from 'react';
+import { useParam } from '@ovh-ux/manager-pci-common';
 import { ShellContext } from '@ovh-ux/manager-react-shell-client';
-import { useNavigate, useParams } from 'react-router-dom';
+import { useNavigate } from 'react-router-dom';
 import { StepEnum } from '@/pages/create/types';
 import { useStore } from '@/pages/create/store';
 import { useGetCapabilities } from '@/api/hooks/useCapabilities';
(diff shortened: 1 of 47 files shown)


packages/manager/apps/pci-private-registry/src/pages/create/steps/PlanStep.tsx
```diff
@@ -138,8 +138,9 @@ export default function PlanStep({
         <PlanChooser
           plan={store.state.plan}
           plans={
-            capabilities.find((c) => c.regionName === store.state.region.name)
-              .plans
+            (capabilities || []).find(
+              (c) => c.regionName === store.state.region.name,
+            ).plans
```
SimonChaumet: If capabilities can be nullish this will throw an error

packages/manager/apps/pci-private-registry/src/pages/create/steps/RegionStep.tsx
```diff
@@ -52,12 +68,45 @@ export default function RegionStep({
     return [];
   }, [localisations, capabilities]);
 
+  const getDeploymentZones = (acc: TDeployment[], region: TLocalisation) => {
```
SimonChaumet: Either this method doesn't depend on props and should be outside the component, or if it depends on the props it should be wrapped in a useCallback and added as a dependency to deploymentZones

packages/manager/apps/pci-private-registry/src/pages/create/steps/RegionStep.tsx
```diff
@@ -52,12 +68,45 @@ export default function RegionStep({
...
+    const deploymentZone = acc.find((zone) => zone.name === region.type);
+    if (!deploymentZone)
+      acc.push({
+        name: region.type,
+      });
```
SimonChaumet: You could use Object.groupBy or a Set instead which would take less lines of code

packages/manager/apps/pci-private-registry/src/pages/create/steps/RegionStep.tsx
```diff
@@ -52,12 +68,45 @@ export default function RegionStep({
...
+        : regions,
+    [regions, selectedRegionGroup],
+  );
+
+  const handleRegionChange = useCallback((regionGroup: TDeployment) => {
+    setSelectedRegionGroup({
+      name: regionGroup.name,
+    });
+    store.set.region(null);
+  }, []);
```
SimonChaumet: There are some dependencies missing here

Also since the store is changed within the same time as the local state, are both needed ?

packages/manager/apps/pci-private-registry/src/pages/create/steps/RegionStep.tsx
```diff
@@ -90,63 +139,81 @@ export default function RegionStep({
...
+                  {t('create:private_registry_create_location')}
+                </Subtitle>
+              </div>
+            </div>
+          )}
+
+          {isLocked ? (
+            <RegionSummary region={store.state.region} />
+          ) : (
+            <TilesInputComponent
```
SimonChaumet, line 201: Why not use the new TilesInput from pci-common ?

packages/manager/apps/pci-private-registry/src/pages/create/steps/RegionStep.tsx
lolottetheclash: I was thinking that as it was a store method from zustand it was stable

SimonChaumet: It might be stable but it is more to avoid future changes when eslint rule of hooks will be enable, not much of a deal now but it will avoid future changes

packages/manager/apps/pci-private-registry/src/pages/create/steps/RegionStep.tsx
lolottetheclash: yes at the start of the feature we talked together about using this component, but in the end (if I remember correctly) it would have meant big changes, so it was decided to stick with the MRC component.

packages/manager/apps/pci-private-registry/src/pages/create/steps/RegionStep.tsx
lolottetheclash: the problem with the set is that it adds all the objects because they are compared by reference...

for groupBy, I'd have to switch to ES2024 (which could be a good thing) but I don't think it would be more readable...
If you have a code suggestion that I haven't thought of, don't hesitate, otherwise we could leave it as it is.

What do you think?

SimonChaumet (approved):

packages/manager/apps/pci-private-registry/src/pages/create/steps/RegionStep.tsx
```diff
@@ -90,63 +144,81 @@ export default function RegionStep({
...
+                label: (continent: string) => (
+                  <OsdsText
+                    break-spaces="false"
+                    size={ODS_TEXT_SIZE._600}
+                    color={ODS_THEME_COLOR_INTENT.text}
+                    level={ODS_THEME_TYPOGRAPHY_LEVEL.body}
+                    hue={ODS_THEME_COLOR_HUE._400}
+                  >
+                    <div className="whitespace-nowrap px-2 text-lg">
+                      {continent ||
```
lionel95200x: Could you move this computation outside of render

SimonChaumet (dismissed):

SimonChaumet (dismissed):

PR #18093, opened 2025-07-10
feat(pci-instances): add instance status and add polling feature

Tsiorifamonjena:
## Description

This PR add the status information for the dashboard. 
It will manage the polling when a task is pending on the instance and then update the status

Ticket Reference:  #TAPC-4037

## Additional Information

image

### DIFF
--- a/packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/InstanceGeneralInfoBlock.component.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/InstanceGeneralInfoBlock.component.tsx
@@ -8,6 +8,7 @@ import {
   Clipboard,
   Links,
   LinkType,
+  TileBlock,
   useTranslatedMicroRegions,
 } from '@ovh-ux/manager-react-components';
 import { RegionChipByType, useParam } from '@ovh-ux/manager-pci-common';
@@ -19,6 +20,8 @@ import {
   DashboardTileBlock,
   DashboardTileText,
 } from './DashboardTile.component';
+import { useDashboardPolling } from '../hooks/useDashboardPolling';
+import { TaskStatus } from '@/pages/instances/task/TaskStatus.component';
 
 const InstanceGeneralInfoBlock: FC = () => {
   const { t } = useTranslation(['dashboard', 'list', 'actions']);
@@ -29,6 +32,12 @@ const InstanceGeneralInfoBlock: FC = () => {
   const { instance, isPending: isInstanceLoading } = useDashboard({
     region: regionId,
     instanceId,
+  });
+
+  const polling = useDashboardPolling({
+    region: regionId,
+    instanceId,
+    taskStatus: instance?.task,
   });
 
   return (
@@ -61,6 +70,16 @@ const InstanceGeneralInfoBlock: FC = () => {
           )}
         </div>
       </DashboardTileBlock>
+      {instance && (
+        <TileBlock label={t('list:pci_instances_list_column_status')}>
+          <TaskStatus
+            isLoading={isInstanceLoading}
+            isPolling={polling.length > 0}
+            taskState={instance.task.status}
+            status={instance.status}
+          />
+        </TileBlock>
+      )}
       <DashboardTileBlock
         label={t('list:pci_instances_list_column_region')}
         isLoading={isInstanceLoading}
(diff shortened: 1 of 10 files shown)


packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/useDashboardPolling.ts
```diff
@@ -0,0 +1,106 @@
...
+  const pendingTasks = isPending ? [{ region, instanceId }] : [];
+
+  const handlePollingSuccess = useCallback(
+    (instance?: TInstance) => {
+      if (!instance) return;
+      const { status, task } = instance;
+      const isDeleted = !task.isPending && status === 'DELETED';
+      const newInstance = isDeleted
+        ? getPartialDeletedInstance(instance.id)
+        : getPartialInstance(instance);
```
SimonChaumet: I think there is some misconception here, either you are manipulating DTOs or you are manipulating entities

If you are manipulating DTOs the functions already exists so no need to add more
If you are manipulating entities then you should not change their properties with a builder, it should be handled by the map DTO -> entity

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/InstanceGeneralInfoBlock.component.tsx
```diff
@@ -31,6 +34,12 @@ const InstanceGeneralInfoBlock: FC = () => {
     instanceId,
   });
 
+  const polling = useDashboardPolling({
+    region: regionId,
+    instanceId,
+    isPending: instance?.task.isPending,
```
SimonChaumet: I don't understand what this isPending is used for, maybe some rename might be required

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/useDashboardPolling.ts
Tsiorifamonjena: You are right, no need to use a builder any more. It's already handle by the mapDTO so cache can be updated well.
I remove the builder

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/InstanceGeneralInfoBlock.component.tsx
Tsiorifamonjena: the isPending is to determine if there is a pending task for the current instance. But I change to pass the `taskStatus` directly into the hook. I hope this is clearer

packages/manager/apps/pci-instances/src/pages/instances/mapper/status.mapper.ts
```diff
@@ -0,0 +1,48 @@
...
+    case 'ACTIVE':
+    case 'RESCUED':
+    case 'RESIZED':
+      return 'success';
+    default:
+      return 'info';
+  }
+};
+
+export const getInstanceStatus = (status: TStatus): TStatusSeverity => ({
```
ppprevost, line 45:
```ts
const severityToStatuses: Record<TSeverity, TStatus[]> = {
  warning: [
    'BUILDING', 'REBOOT', 'REBUILD', 'REVERT_RESIZE', 'SOFT_DELETED',
    'VERIFY_RESIZE', 'MIGRATING', 'RESIZE', 'BUILD', 'SHUTOFF',
    'RESCUE', 'SHELVED', 'SHELVED_OFFLOADED', 'RESCUING', 'UNRESCUING',
    'SNAPSHOTTING', 'RESUMING', 'HARD_REBOOT', 'PASSWORD', 'PAUSED',
  ],
  error: [
    'DELETED', 'ERROR', 'STOPPED', 'SUSPENDED', 'UNKNOWN',
  ],
  success: [
    'ACTIVE', 'RESCUED', 'RESIZED',
  ],
};

const getInstanceStatusSeverity = (status: TStatus): TSeverity => {
  for (const severity in severityToStatuses) {
    if (severityToStatuses[severity as TSeverity].includes(status)) {
      return severity as TSeverity;
    }
  }
  return 'info';
};

export const getInstanceStatus = (status: TStatus): TStatusSeverity => ({
  label: status,
  severity: getInstanceStatusSeverity(status),
});
```

ppprevost: que penses tu ?

PR #18095, opened 2025-07-10
feat(pci-instances): allow activating monthly billing when enabled

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-4076

### DIFF
--- a/packages/manager/apps/pci-instances/src/routes/routes.tsx
+++ b/packages/manager/apps/pci-instances/src/routes/routes.tsx
@@ -91,7 +91,7 @@ const instanceActionLegacyRoutes: RouteObject[] = instanceActionsSections.map(
   }),
 );
 
-const instanceActionRoutes = instanceActionsSections.map((section) => ({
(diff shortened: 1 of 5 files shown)


packages/manager/apps/pci-instances/src/routes/routes.tsx
```diff
@@ -104,6 +104,13 @@ const instanceLegacyRoutes: RouteObject[] = instanceActionsSections.map(
   }),
 );
 
+const instanceActionRoutes = instanceActionsSections.map((section) => ({
```
lolottetheclash: just a detail for instancesActionRoutes & instanceActionRoutes, I think both should be plural (Action => Actions).
As instanceActionsSections is already plural

lolottetheclash (approved):

packages/manager/apps/pci-instances/src/routes/routes.tsx
Tsiorifamonjena: Done :)

PR #18111, opened 2025-07-11
fix(dedicated-servers): fix search input

mhelhali-soufien:
ref: #MANAGER-19034

## Description

Ticket Reference: #MANAGER-19034

### DIFF
--- a/packages/manager/apps/dedicated-servers/src/pages/listing/server/index.tsx
+++ b/packages/manager/apps/dedicated-servers/src/pages/listing/server/index.tsx
@@ -1,6 +1,4 @@
-import React, { useEffect, useState } from 'react';
-import { useTranslation } from 'react-i18next';
-import { ShellContext } from '@ovh-ux/manager-react-shell-client';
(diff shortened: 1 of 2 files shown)


qpavy (approved):

packages/manager/apps/dedicated-servers/src/pages/listing/server/index.tsx
```diff
@@ -18,7 +18,8 @@ import { urls } from '@/routes/routes.constant';
 import { ErrorComponent } from '@/components/errorComponent';
 
 export default function ServerListing() {
-  const [columns] = useState([]);
+  const { t } = useTranslation('dedicated-servers');
+  const [columns] = useState(getColumns(t));
```
chipp972: if you rename the function to `useColumns` (to make it a hook) you will be able to use `useTranslation` directly inside the function.

And you should use `React.useMemo` if you don't want to recalculate the columns variable on each re-render instead of `useState` I think

PR #18114, opened 2025-07-11
feat(kms): domains list page for secrets

mathieumousnier:
## Description

Ticket Reference: #MANAGER-18850


packages/manager/apps/key-management-service/src/data/api/okms.ts
```diff
@@ -1,21 +1,22 @@
...
 import { OKMS, OkmsPublicCa } from '@/types/okms.type';
 
-export const getOkmsResourceQueryKey = (okmsId: string) => [
-  `get/okms/resource/${okmsId}`,
-];
-
-export const getOkmsServicesResourceListQueryKey = ['get/okms/resource'];
+export const okmsQueryKeys = {
+  list: () => ['okms'],
+  listDatagrid: () => ['okms', 'datagrid'],
```
rjamet-ovh: you don't need functions here

packages/manager/apps/key-management-service/public/translations/key-management-service/listing/Messages_fr_FR.json
```diff
@@ -3,6 +3,7 @@
   "key_management_service_listing_name_cell": "Nom",
   "key_management_service_listing_id_cell": "ID",
   "key_management_service_listing_kmip_cell": "KMIP",
+  "key_management_service_listing_secret_cell": "Secrets",
   "key_management_service_listing_region_cell": "Région",
   "key_management_service_listing_status_cell": "Statut",
```
rjamet-ovh, line 8: should be rename as `header` or `column`

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/domains/Domains.page.spec.tsx
```diff
@@ -0,0 +1,99 @@
...
+
+    const name = labels.listing.key_management_service_listing_name_cell;
+    const id = labels.listing.key_management_service_listing_id_cell;
+    const secret = labels.listing.key_management_service_listing_secret_cell;
+    const status = labels.listing.key_management_service_listing_status_cell;
+
+    expect(await screen.findAllByText(name)).toHaveLength(1);
+    expect(await screen.findAllByText(id)).toHaveLength(1);
+    expect(await screen.findAllByText(secret)).toHaveLength(1);
+    expect(await screen.findAllByText(status)).toHaveLength(1);
```
rjamet-ovh: we need to check, either on this test or on a specific one for `<OkmsDatagrid />`, that we render the columns but also that we don't render the unwanted columns.

packages/manager/apps/key-management-service/public/translations/key-management-service/listing/Messages_fr_FR.json
mathieumousnier: You really want to rename the translation keys in the already existing kms code ?
I am not sure we should do this now...

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/domains/Domains.page.spec.tsx
```diff
@@ -0,0 +1,99 @@
...
+    ).toBeGreaterThan(0);
+
+    // Check the okms link with the first okmsMock name is present
+    const okmsNameLink = await getOdsButtonByLabel({
+      container,
+      label: okmsMock[0].iam.displayName,
+      isLink: true,
+    });
+    expect(okmsNameLink).toBeInTheDocument();
+  });
```
rjamet-ovh: you should split the `not a valid region` use-case and the `no domain on region` use-case.
for the latter one, you could mock the return okms list to not have any okms of that particular region

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/domains/Domains.page.tsx
```diff
@@ -0,0 +1,69 @@
...
+  const navigate = useNavigate();
+  const { region } = useParams() as { region: string };
+
+  const { data, error, isPending, refetch } = useOkmsDatagridList();
+  const flattenData = data?.pages.flatMap((page) => page.data);
+
+  // Filter okms by regionId
+  const okmsList = flattenData
+    ? flattenData.filter((okms) => okms.region === region)
+    : [];
```
rjamet-ovh: you can use `filterDomainsByRegion` here

packages/manager/apps/key-management-service/src/modules/secret-manager/routes/routes.constants.ts
```diff
@@ -11,16 +12,17 @@ const URLS = {
   secretManagerRoot: `/${URIS.root}`,
   secretManagerOnboarding: `/${URIS.root}/${URIS.onboarding}`,
   secretCreate: `/${URIS.root}/${URIS.create}`,
+  secretDomains: (regionId: string) =>
+    `/${URIS.root}/${URIS.region}/${regionId}`,
   secretListing: (domainId: string) => `/${URIS.root}/${domainId}`,
   secretDashboard: (domainId: string, secretPath: string) =>
-    `/${URIS.root}/${domainId}/${encodeSecretPath(secretPath)}`,
+    `${URLS.secretListing(domainId)}/${encodeSecretPath(secretPath)}`,
```
rjamet-ovh: i'm not sure about referencing other url functions inside an url function. It creates adherence and makes it more difficult to read.

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/domains/Domains.page.spec.tsx
mathieumousnier: You're probably right, it would be better to have 2 tests.
I wrote only one because "mock the return okms list to not have any okms of that particular region" consists just to point to a region on which the okmsMock has no domain.

So the only difference between both tests would be that on the first one the url points to `notValidRegionName`, and on the second one the url points to `ca-east-bhs`

I guess I will factorize the rest of the code to separate both tests

packages/manager/apps/key-management-service/src/modules/secret-manager/routes/routes.constants.ts
mathieumousnier: I'm not sure either, I see the "readability" argument (even if I feel like both are hard to read), but not the "adherence", it is has to be the same thing, the adherence is on purpose.

Both are ok in my opinion

image

packages/manager/apps/key-management-service/public/translations/key-management-service/listing/Messages_fr_FR.json
rjamet-ovh: they're only used on the already updated okms list so it should not cause more issues i think

packages/manager/apps/key-management-service/src/common/components/okmsDatagrid/ListingCells.component.tsx
```diff
@@ -0,0 +1,91 @@
...
+export const DatagridCellName = (okms: OKMS, type: OkmsDatagridType) => {
+  const { trackClick } = useOvhTracking();
+
+  const urls: Record<OkmsDatagridType, string> = {
+    kms: KMS_ROUTES_URLS.kmsDashboard(okms.id),
+    'secret-manager': SECRET_MANAGER_ROUTES_URLS.secretListing(okms.id),
+  };
+  const tracking: Record<OkmsDatagridType, string> = {
+    kms: 'go-to-kms',
+    'secret-manager': '',
```
darsene, line 38: Are you waiting for further tracking implementation ?

packages/manager/apps/key-management-service/src/common/components/okmsDatagrid/ListingCells.component.tsx
```diff
@@ -0,0 +1,91 @@
...
+  };
+
+  return (
+    <Link
+      href={urls[type]}
+      label={okms.iam.displayName}
+      isRouterLink
+      onClick={() => {
+        if (tracking[type].length > 0) {
```
darsene, line 47: What about if `tracking[type]` is `null` ?

packages/manager/apps/key-management-service/src/common/components/okmsDatagrid/ListingCells.component.tsx
```diff
@@ -0,0 +1,91 @@
...
+export const DatagridCellStatus = (okms: OKMS) => {
+  const { data: OkmsServiceInfos, isLoading, isError } = useServiceDetails({
+    resourceName: okms.id,
+  });
+  if (isLoading) {
+    return <OdsSpinner size={ODS_SPINNER_SIZE.sm} />;
+  }
+  if (isError) {
+    return <></>;
+  }
```
darsene, line 80: No error label displayed ?

packages/manager/apps/key-management-service/src/components/layout-helpers/Dashboard/GeneralInformationsTiles/InformationsTile.spec.tsx
```diff
@@ -28,6 +28,7 @@ describe('InformationsTile component tests suite', () => {
     swaggerEndpoint: 'https://swagger-endpoint',
     kmipObjectCount: 1,
     serviceKeyCount: 2,
+    secretCount: 3,
```
darsene, line 31: Why do you need to add this property if the assert part is not updated ?

packages/manager/apps/key-management-service/src/components/layout-helpers/Dashboard/GeneralInformationsTiles/KmipTile.spec.tsx
```diff
@@ -35,6 +35,7 @@ describe('KmipTile component tests suite', () => {
     swaggerEndpoint: 'https://swagger-endpoint',
     kmipObjectCount: 1,
     serviceKeyCount: 2,
+    secretCount: 3,
```
darsene, line 38: Why do you need to add this property if the assert part is not updated ?

packages/manager/apps/key-management-service/src/components/layout-helpers/Dashboard/GeneralInformationsTiles/RestApiTile.spec.tsx
```diff
@@ -35,6 +35,7 @@ describe('RestApiTile component tests suite', () => {
     swaggerEndpoint: 'https://swagger-endpoint',
     kmipObjectCount: 1,
     serviceKeyCount: 2,
+    secretCount: 3,
```
darsene, line 38: Why do you need to add this property if the assert part is not updated ?

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/root/Root.page.tsx
```diff
@@ -9,26 +10,35 @@ export default function SecretManagerRootPage() {
...
+      const firstRegionOkmsList = groupedOkms[0];
+      const firstRegionOkms = firstRegionOkmsList[0];
+
+      if (firstRegionOkmsList.length > 1) {
+        // If there are multiple OKMS in the first region, navigate to the domains list
+        url = SECRET_MANAGER_ROUTES_URLS.secretDomains(firstRegionOkms.region);
+      } else {
+        // If there is only one OKMS in the first region, navigate to the secret listing
+        url = SECRET_MANAGER_ROUTES_URLS.secretListing(firstRegionOkms.id);
+      }
```
darsene: no need to test that `firstRegionOkmsList.length === 1` ? Could not be 0 ?

packages/manager/apps/key-management-service/src/types/api.type.ts
```diff
@@ -4,3 +4,25 @@ export type ErrorResponse = {
...
+    return false;
+  }
+
+  const { status, data } = response as { status?: unknown; data?: unknown };
+  if (typeof status !== 'number' || typeof data !== 'object' || data === null) {
+    return false;
+  }
+
+  return 'message' in data;
+};
```
darsene: It's not really a type, makes more sense to move it into a util file

packages/manager/apps/key-management-service/src/common/components/okmsDatagrid/ListingCells.component.tsx
mathieumousnier: Yes, there will be tracking for secret-manager at the end

packages/manager/apps/key-management-service/src/common/components/okmsDatagrid/ListingCells.component.tsx
mathieumousnier: This is the point of typescript to ensure that we do not set `type=null` when using the `DatagridCellName` component.

packages/manager/apps/key-management-service/src/common/components/okmsDatagrid/ListingCells.component.tsx
mathieumousnier: Good question, this is the current state in the kms datagrid, I only copy pasted this code in this new file.
Today we don't display anything because it is either an error, or the sign that the user does have the permissions on the `services` endpoint, usecase where we do not want to display an error.

packages/manager/apps/key-management-service/src/components/layout-helpers/Dashboard/GeneralInformationsTiles/InformationsTile.spec.tsx
mathieumousnier: This field has bee added to the data returned by the okms endpoint.
I then added it to the `OKMS` type, which led me to update all the local mocks in different tests files

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/root/Root.page.tsx
mathieumousnier: It can't be 0 because `groupedOkms` is an array of array built with kms.
So the function only returns regions with kms.

Then when I write the code below, `firstRegionOkmsList.length === 1` can only be the else.

```ts
if (firstRegionOkmsList.length > 1) {
  // If there are multiple OKMS in the first region, navigate to the domains list
  url = SECRET_MANAGER_ROUTES_URLS.secretDomains(firstRegionOkms.region);
} else {
  // If there is only one OKMS in the first region, navigate to the secret listing
  url = SECRET_MANAGER_ROUTES_URLS.secretListing(firstRegionOkms.id);
}
```

This works for sure but I see your point, maybe I should do something like below.


```ts
const groupedOkms = Object.values(groupDomainsByRegion(okmsList));
const firstRegionOkmsList = groupedOkms[0];
const firstRegionOkms = firstRegionOkmsList[0];

if (firstRegionOkmsList.length > 0) {
  if (firstRegionOkmsList.length === 1) {
    // If there is only one OKMS in the first region, navigate to the secret listing
    url = SECRET_MANAGER_ROUTES_URLS.secretListing(firstRegionOkms.id);
  } else {
    // If there are multiple OKMS in the first region, navigate to the domains list
    url = SECRET_MANAGER_ROUTES_URLS.secretDomains(firstRegionOkms.region);
  }
```

packages/manager/apps/key-management-service/src/types/api.type.ts
mathieumousnier: Mmh your right, I like to put typeguard with types but it is a different from projects to projects.
I'll move it

packages/manager/apps/key-management-service/src/common/components/okmsDatagrid/ListingCells.component.tsx
mathieumousnier: This typescript "strict"="false" is the worst fuckery I have ever seen 😭
I guess I will add a default value for type.

image

mathieumousnier: done

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/root/Root.page.tsx
mathieumousnier: I updated the code here.
I do a useless `firstRegionOkmsList.length > 0` but its more readable and more robust to change

packages/manager/apps/key-management-service/src/types/api.type.ts
mathieumousnier: done

packages/manager/apps/key-management-service/src/common/components/okmsDatagrid/ListingCells.component.tsx
darsene: even if type is not null, tracking[type] could be null as soon as the definition of type is missing inside tracking property

darsene (approved):

packages/manager/apps/key-management-service/src/common/components/okmsDatagrid/ListingCells.component.tsx
mathieumousnier: No because `tracking` is define with this type `Record<OkmsDatagridType, string>` !
This ensures that `tracking` has all the `OkmsDatagridType` keys well declared.
I used this pattern a hundred times, any mismatch is catched by typescript.

PR #18125, opened 2025-07-11
feat(manager-react-components): migrate RedirectionGuard to ODS v19

anooparveti:
## Description

Migrate RedirectionGuard component to ODS v19
Align with component development guidelines
Add documentation for the component

Ticket Reference: #MANAGER-18965


packages/manager-react-components/src/components/redirection-guard/RedirectionGuard.component.tsx
```diff
@@ -0,0 +1,25 @@
...
+    return (
+      <Spinner data-testid="redirectionGuard_spinner" size={SPINNER_SIZE.md} />
+    );
+  }
+
+  if (isError && errorComponent) {
+    return <>{errorComponent}</>;
+  }
+
+  return condition ? <Navigate to={route} /> : <>{children}</>;
```
lionel95200x, line 24:
```suggestion
  return condition ? <Navigate to={route} /> : {children};
```

anooparveti: It results in an error we cannot return without fragment.

lionel95200x: You should update output component type to React.ReactNode

anooparveti: Screenshot 2025-07-11 at 8 59 34 PM

Still have the issue after updating it.

PR #18128, opened 2025-07-11
feat(datagrid): add row select feature

qpavy:
## Description

Ticket Reference:  #MANAGER-15795

### DIFF
--- a/packages/manager-react-components/package.json
+++ b/packages/manager-react-components/package.json
@@ -72,7 +72,7 @@
     "babel-loader": "9.1.3",
     "babel-preset-react-app": "^10.0.1",
     "date-fns": "~4.1.0",
-    "element-internals-polyfill": "^1.3.10",
+    "element-internals-polyfill": "^3.0.2",
     "eslint": "^9.22.0",
     "eslint-config-prettier": "^10.1.1",
     "eslint-plugin-import": "^2.31.0",
--- a/packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
+++ b/packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
@@ -11,6 +11,8 @@ import {
   Row,
   OnChangeFn,
   VisibilityState,
+  RowSelectionState,
+  Table,
 } from '@tanstack/react-table';
 import {
   ODS_ICON_NAME,
@@ -36,6 +38,7 @@ import { DataGridTextCell } from './text-cell.component';
 import { defaultNumberOfLoadingRows } from './datagrid.constants';
 import { DatagridTopbar } from './datagrid-topbar.component';
 import './translations';
+import { IndeterminateCheckbox } from './indeterminate-checkbox';
 
 export type ColumnSort = TanstackColumnSort;
 export type PaginationState = TanstackPaginationState;
@@ -87,6 +90,15 @@ export interface SearchProps {
   searchInput: string;
   setSearchInput: React.Dispatch<React.SetStateAction<string>>;
   onSearch: (search: string) => void;
+}
+
+export interface RowSelectionProps<T> {
+  rowSelection: RowSelectionState;
+  setRowSelection: React.Dispatch<React.SetStateAction<RowSelectionState>>;
+  /** This callback is called every time 1 or multiple rows are selected */
+  onRowSelectionChange?: (selectedRows: T[]) => void;
+  /** when used, for each row if expression is false, the row is disabled */
+  enableRowSelection?: (row: Row<T>) => boolean;
 }
 
 export interface DatagridProps<T> {
@@ -151,6 +163,10 @@ export interface DatagridProps<T> {
   tableLayoutFixed?: boolean;
   /** To use if tag column is present and filter is enabled. This allows to fetch all tags from iam only for this resource type */
   resourceType?: string;
+  /** Enable and configure row selection */
+  rowSelection?: RowSelectionProps<T>;
+  /** Use to overwrite row id */
+  getRowId?: (originalRow: T, index: number) => string;
 }
 
 export const Datagrid = <T,>({
@@ -182,6 +198,8 @@ export const Datagrid = <T,>({
   hideHeader,
   tableLayoutFixed,
   resourceType,
+  rowSelection,
+  getRowId,
 }: DatagridProps<T>) => {
   const { t } = useTranslation('datagrid');
   const pageCount = pagination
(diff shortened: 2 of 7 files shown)


packages/manager-react-components/package.json
```diff
@@ -72,7 +72,7 @@
     "babel-loader": "9.1.3",
     "babel-preset-react-app": "^10.0.1",
     "date-fns": "~4.1.0",
-    "element-internals-polyfill": "^1.3.10",
+    "element-internals-polyfill": "^3.0.2",
```
aboungnaseng-ovhcloud, line 75: No yarn.lock to update ?

packages/manager-wiki/stories/manager-react-components/components/Datagrid/documentation.mdx
```diff
@@ -320,4 +320,127 @@ Exemple of column definition
...
+    items.forEach(({label}) => {
+        selectedItems[label] = !isAllRowSelected;
+    });
+    setRowSelection(selectedItems);
+};
+
+return
+  <Datagrid
+    ...
+    manageRowSelection={{
```
aboungnaseng-ovhcloud: I don't see manageRow selection in Datagrid props

packages/manager-react-components/src/components/datagrid/indeterminate-checkbox.tsx
```diff
@@ -0,0 +1,33 @@
...
+  id: string;
+  name: string;
+  label: string;
+  onChange: () => void;
+  isChecked?: boolean;
+  isIndeterminate?: boolean;
+  isDisabled?: boolean;
+};
+
+export const IndeterminateCheckbox = ({
```
atefzed, line 13: **Question:** Why not use `OdsCheckbox` directly

qpavy: In next Ods version, OdsCheckbox props will be changed. Using a wrapper component allow to handle the fix of the breaking change in 1 place

aboungnaseng-ovhcloud (dismissed):

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
```diff
@@ -319,6 +380,27 @@ export const Datagrid = <T,>({
     [columns],
   );
 
+  // Handle onRowSelectionChange callback
+  useEffect(() => {
+    const selectedRows =
+      table.getSelectedRowModel()?.rows?.map(({ original }) => original) || [];
+    rowSelection?.onRowSelectionChange?.(selectedRows);
+  }, [JSON.stringify(rowSelection?.rowSelection)]);
```
helabenkhalfallah, line 388: we cannot only listen to rowSelection?.rowSelection changes? instead of JSON.stringify, it consumes memory and time.

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
```diff
@@ -319,6 +380,27 @@ export const Datagrid = <T,>({
     [columns],
   );
 
+  // Handle onRowSelectionChange callback
+  useEffect(() => {
+    const selectedRows =
```
helabenkhalfallah, line 385: it seems that the map does not do anything, it returns the same object `original`, we cannot simply read `table.getSelectedRowModel()?.rows` without map?

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
```diff
@@ -319,6 +380,27 @@ export const Datagrid = <T,>({
...
+   */
+  useEffect(() => {
+    if (table.getSelectedRowModel()?.rows) {
+      const newSelectionState = {};
+      table.getSelectedRowModel().rows.forEach(({ id }) => {
+        newSelectionState[id] = true;
+      });
+      table.setRowSelection(newSelectionState);
+    }
+  }, [JSON.stringify(items)]);
```
helabenkhalfallah, line 406: same issue, why JSON.stringfy instead of items simply?

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
```diff
@@ -319,6 +380,27 @@ export const Datagrid = <T,>({
...
+  }, [JSON.stringify(rowSelection?.rowSelection)]);
+
+  /**
+   * Update internal rowSelection state when items list change:
+   * if items selected but is remove from list, remove it from row selection state
+   */
+  useEffect(() => {
+    if (table.getSelectedRowModel()?.rows) {
+      const newSelectionState = {};
+      table.getSelectedRowModel().rows.forEach(({ id }) => {
```
helabenkhalfallah: I'm wondering about the performance of forEach, this is a linear path. Can we compare the performance of this to the performance of using `reduce` or a simple `for ... of`: 
https://jsperf.app
https://developer.mozilla.org/en-US/docs/Web/API/Performance

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
qpavy: I had issue of multiple re-renders when only listen to rowSelection changes because it rerendered everytime the reference changed even if the selected rows didn't changed.

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
qpavy: it does not return the row but only the row value (which is call original)

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
qpavy: Same answer than before

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
helabenkhalfallah: strange, ok.

packages/manager-react-components/src/components/datagrid/datagrid.component.tsx
helabenkhalfallah: ok, I'm afraid this may cause performance issues if we select many row or all check. What do you think?

qpavy: I think performance issue will occur while trying to render all the dom (as the datagrid is not virtualized) before we can see perfomance issue from this map :innocent:

PR #18134, opened 2025-07-15
feat(ips): manage ip mitigation

sachinrameshn:
ref: #MANAGER-18510

## Description

Ticket Reference: #...    

## Additional Information
Scrubbing center
- Actions - Scrubbing center: Auto
- Actions - Scrubbing center: Permanent

### DIFF
--- a/packages/manager/apps/ips/src/pages/actions/manageIpMitigation/manageIpMitigation.page.tsx
+++ b/packages/manager/apps/ips/src/pages/actions/manageIpMitigation/manageIpMitigation.page.tsx
@@ -1,0 +1,108 @@
+import React from 'react';
+import { useTranslation } from 'react-i18next';
+import { useNavigate, useParams } from 'react-router-dom';
+import { NAMESPACES } from '@ovh-ux/manager-common-translations';
+import { useNotifications } from '@ovh-ux/manager-react-components';
+import { OdsButton, OdsModal, OdsText } from '@ovhcloud/ods-components/react';
+import {
+  ODS_BUTTON_VARIANT,
+  ODS_SPINNER_SIZE,
+  ODS_TEXT_PRESET,
+} from '@ovhcloud/ods-components';
+import { useGetIpMitigation, useUpdateIpMitigation } from '@/data/hooks/ip';
+import { fromIdToIp, ipFormatter } from '@/utils';
+import Loading from '../../listing/manageOrganisations/components/Loading/Loading';
+
+export default function ManageIpMitigationModal() {
+  const navigate = useNavigate();
+  const { id } = useParams();
+  const { ip, ipGroup } = ipFormatter(fromIdToIp(id));
+  const { ipMitigation, isLoading: isMitigationLoading } = useGetIpMitigation({
+    ip,
+  });
+  const { addSuccess, addError } = useNotifications();
+  const { t } = useTranslation(['listing', NAMESPACES.ACTIONS, 'error']);
+
+  // If there is no mitigation on IP then its a default mitigation.
+  const isDefaultMitigation = (ipMitigation?.length ?? 0) === 0;
+  const title = isDefaultMitigation
+    ? t('listingManageMitigation_permanent_title')
+    : t('listingManageMitigation_auto_title');
+  const question = isDefaultMitigation
+    ? t('listingManageMitigation_permanent_question', { t0: ip })
+    : t('listingManageMitigation_auto_question', { t0: ip });
+
+  // Determine the new mitigation type to set for ip upon confirmation
+  // If isDefaultMitigation the new mitigation will be permanent and vice versa
+  const mitigation = isDefaultMitigation ? 'PERMANENT' : 'DEFAULT';
+
+  const ipBlock = ipGroup;
+
+  const {
+    mutate: updateIpMitigation,
+    isPending: updateIpMitigationPending,
+  } = useUpdateIpMitigation({
+    ipBlock,
+    ip,
+    mitigation,
+    onSuccess: () => {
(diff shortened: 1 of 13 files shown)


sachinrameshn:
> I think we will need to do a refactor of action cell to split it into 1 file / action or 1 file per action for the same feature. It become really big adding more and more actions

Agree !!  I will see it in coming PRs.

packages/manager/apps/ips/src/data/hooks/ip/useIpHasServicesNotAttached.ts
```diff
@@ -0,0 +1,33 @@
...
+  serviceName?: string;
+};
+
+export const useIpHasServicesNotAttached = ({
+  serviceName,
+}: UseIpHasServicesNotAttachedParams) => {
+  // Fetch all product services for all paths/categories
+  const productServicesList = PRODUCT_PATHS_AND_CATEGORIES.map((entry) =>
+    useGetProductService(entry),
+  );
```
chipp972: We should have a useQueries for multiple queries, not use a hook inside a `.map` I think

chipp972: Even though the array is static, I don't think it's a good idea

packages/manager/apps/ips/src/data/hooks/ip/useIpHasServicesNotAttached.ts
```diff
@@ -0,0 +1,33 @@
...
+  { path: '/vrack', category: 'VRACK' },
+  { path: '/dedicatedCloud', category: 'PRIVATE_CLOUD' },
+  { path: '/ipLoadbalancing', category: 'IP_LOAD_BALANCING' },
+];
+
+export type UseIpHasServicesNotAttachedParams = {
+  serviceName?: string;
+};
+
+export const useIpHasServicesNotAttached = ({
```
chipp972: I think it would be better to have a hook to get the list of attached service and if the list is empty then we know that we don't have any attached service

If we need to exploit the attached services later on we will have to create another hook that does exactly the same things but returns the data instead of a boolean

packages/manager/apps/ips/src/pages/listing/ipListing/components/DatagridCells/IpActionsCell/IpActionsCell.tsx
```diff
@@ -84,148 +92,195 @@ export const IpActionsCell = ({ parentIpGroup, ip }: IpActionsCellParams) => {
...
+      hasDedicatedServiceAttachedToIp,
+      vmacsWithIp,
+      isVmacAlreadyExist,
+      vrackPage,
+      gameMitigationDetails,
+      hasForcedMitigation,
+      ipMitigation,
+      isMitigationLoading,
+      isDefaultMitigation,
+    ],
```
chipp972: Well, you observed less re-renders of the column ?
Because if you put every variables and props in the dependency array, it won't be different compared to not using useMemo

packages/manager/apps/ips/src/pages/actions/manageIpMitigation/manageIpMitigation.page.tsx
```diff
@@ -0,0 +1,108 @@
...
+          <OdsText className="block mb-4" preset={ODS_TEXT_PRESET.paragraph}>
+            {question}
+          </OdsText>
+          <OdsButton
+            slot="actions"
+            type="button"
+            variant={ODS_BUTTON_VARIANT.ghost}
+            label={t('cancel', { ns: NAMESPACES.ACTIONS })}
+            onClick={cancel}
+            data-testid="cancel-button"
```
chipp972, line 88: why do we need testids ?

packages/manager/apps/ips/src/data/hooks/ip/useIpHasServicesNotAttached.ts
sachinrameshn: Okay I will modify it

chipp972: Maybe we should have another hook useGetProductService`s` that accept an array and returns the result of useQueries

packages/manager/apps/ips/src/data/hooks/ip/useIpHasServicesNotAttached.ts
sachinrameshn: Okay will modify the hook to get the list of attached service to reuse in other cases.

packages/manager/apps/ips/src/pages/actions/manageIpMitigation/manageIpMitigation.page.tsx
sachinrameshn: I need it for test case which I will add in another commit. I will use it in the getAllByTestId bound function.

chipp972: ok

packages/manager/apps/ips/src/pages/listing/ipListing/components/DatagridCells/IpActionsCell/IpActionsCell.tsx
sachinrameshn: I tried to verify the recalculation and performance by adding console.log statement with and without useMemo. 
Even though building the item array is slightly expensive I dint see much improvement with memoization.
Updated the code.

chipp972 (approved):

PR #18209, opened 2025-07-18
fix(pci-kubernetes): let the edit value unchanged

ppprevost:
ref: #TAPC-4453,  TAPC-4662
I took the opportunity to move the types into the types file.
I also ensured that Scale, Node Pool creation in the cluster, and the Node Pool page now all share the same validation mechanism for creation.
I fixed the nullability of the types.
I facilitated rules from Scale component. 
Lastly, I made sure the minNode and maxNode values are no longer modified when toggling autoscaling from true to false.

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-kubernetes/src/helpers/node-pool.ts
+++ b/packages/manager/apps/pci-kubernetes/src/helpers/node-pool.ts
@@ -1,0 +1,47 @@
+import { ANTI_AFFINITY_MAX_NODES, NODE_RANGE } from '@/constants';
+import { isMonoDeploymentZone } from '.';
+import { NodePoolState } from '@/types';
+import { TRegionInformations } from '@/types/region';
+
+export const exceedsMaxNodes = (quantity: number) => quantity > NODE_RANGE.MAX;
+
+export const zoneAZisChecked = (
+  regionInformations: TRegionInformations,
+  nodePoolState: NodePoolState,
+) =>
+  isMonoDeploymentZone(regionInformations?.type) ||
+  !!nodePoolState.selectedAvailabilityZone;
+
+export const isScalingValid = (
+  nodePoolState: Pick<NodePoolState, 'scaling'>,
+) => {
+  if (!nodePoolState.scaling) return true;
+
+  const { isAutoscale, quantity } = nodePoolState.scaling;
+  const { desired, min, max } = quantity;
+
+  if (!isAutoscale) {
+    return !exceedsMaxNodes(desired);
+  }
+
+  const isMinValid = min >= 0 && min <= max;
+  const isMaxValid = max <= NODE_RANGE.MAX;
+  const isDesiredInRange = min <= desired && max >= desired;
+
+  return isMinValid && isMaxValid && isDesiredInRange;
+};
+
+export const hasMax5NodesAntiAffinity = (nodePoolState: NodePoolState) =>
+  !nodePoolState.antiAffinity ||
+  (nodePoolState.antiAffinity &&
+    nodePoolState.scaling &&
+    nodePoolState.scaling.quantity.desired <= ANTI_AFFINITY_MAX_NODES) ||
+  false;
+
+export const hasInvalidScalingOrAntiAffinityConfig = (
+  regionInformations: TRegionInformations,
+  nodePoolState: NodePoolState,
+) =>
+  !isScalingValid(nodePoolState) ||
+  !hasMax5NodesAntiAffinity(nodePoolState) ||
+  !zoneAZisChecked(regionInformations, nodePoolState);
--- a/packages/manager/apps/pci-kubernetes/src/pages/detail/nodepools/new/store.ts
+++ b/packages/manager/apps/pci-kubernetes/src/pages/detail/nodepools/new/store.ts
@@ -1,9 +1,11 @@
 import { create } from 'zustand';
 import { createRef, RefObject } from 'react';
 import { StepsEnum } from '@/pages/detail/nodepools/new/steps.enum';
-import { AutoscalingState } from '@/components/Autoscaling.component';
+
 import { isNodePoolNameValid } from '@/helpers/matchers/matchers';
 import { TComputedKubeFlavor } from '@/components/flavor-selector/FlavorSelector.component';
+import { NODE_RANGE } from '@/constants';
+import { TAutoscalingState } from '@/types';
 
 type TStep = {
   isOpen: boolean;
@@ -20,7 +22,7 @@ export type TFormStore = {
   };
   flavor?: TComputedKubeFlavor;
   selectedAvailabilityZone: string;
-  autoScaling: AutoscalingState | null;
+  autoScaling: TAutoscalingState;
   antiAffinity: boolean;
   isMonthlyBilling: boolean;
   steps: Map<StepsEnum, TStep>;
@@ -28,7 +30,7 @@ export type TFormStore = {
     name: (val: string) => void;
     flavor: (val?: TComputedKubeFlavor) => void;
     selectedAvailabilityZone: (selectedZone: string) => void;
-    autoScaling: (val: AutoscalingState | null) => void;
+    autoScaling: (val: TAutoscalingState) => void;
     antiAffinity: (val: boolean) => void;
     isMonthlyBilling: (val: boolean) => void;
   };
@@ -88,6 +90,11 @@ const initialSteps = () =>
     ],
   ]);
 
+const initAutoscale = {
+  quantity: { desired: NODE_RANGE.MIN, min: 0, max: NODE_RANGE.MAX },
+  isAutoscale: false,
+};
+
 export const useNewPoolStore = create<TFormStore>()((set, get) => ({
   name: {
     value: '',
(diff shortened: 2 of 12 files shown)


fredericvilcot (approved):

packages/manager/apps/pci-kubernetes/src/pages/new/steps/NodePoolStep.component.tsx
```diff
@@ -178,12 +178,15 @@ const NodePoolStep = ({
         desiredNodes: nodePoolState.scaling.quantity.desired,
         minNodes: nodePoolState.scaling.quantity.min,
         flavorName: flavor.name,
+        // You cannot have antiAffinity and autoscale at the same time,
```
lionel95200x: I think you can reflect your comment in the type of state to improve control

```
type AffinityEnabled = {
antiAffinity: true
isAutoscale: false 
} 

type AutoScaleEnabled = {
antiAffinity: false
isAutoscale: true 
} 

type MyCustomType = AutoScaleEnabled | AutoScaleEnabled
```

SimonChaumet (approved):

packages/manager/apps/pci-kubernetes/src/pages/detail/nodepools/new/store.ts
```diff
@@ -4,6 +4,7 @@ import { StepsEnum } from '@/pages/detail/nodepools/new/steps.enum';
 import { AutoscalingState } from '@/components/Autoscaling.component';
```
fredericvilcot: Should be great that the store does not depend on a presentational component.

So let's consider moving types to a dedicated folder :)

packages/manager/apps/pci-kubernetes/src/helpers/node-pool.ts
```diff
@@ -0,0 +1,58 @@
...
+  scaling: { isAutoscale: false };
+  antiAffinity: boolean;
+} => state.scaling?.isAutoscale === false;
+
+export const canToggleAutoScale = (
+  state: NodePoolState,
+): state is NodePoolState & {
+  scaling: { isAutoscale: boolean };
+  antiAffinity: false;
+} => state.antiAffinity === false;
```
fredericvilcot: These type guards are misleading: they assert properties without actually checking for them. This can lead to unsafe assumptions in downstream code.

Example: `canToggleAntiAffinity` only checks if `isAutoscale === false` but never if `antiAffinity` is defined in type.

The runtime condition must fully reflect the asserted type to be valid.

packages/manager/apps/pci-kubernetes/src/helpers/node-pool.ts
```diff
@@ -0,0 +1,58 @@
...
+  return isMinValid && isMaxValid && isDesiredInRange;
+};
+
+export const hasMax5NodesAntiAffinity = (nodePoolState: NodePoolState) =>
+  !nodePoolState.antiAffinity ||
+  (nodePoolState.antiAffinity &&
+    nodePoolState.scaling &&
+    nodePoolState.scaling.quantity.desired <= ANTI_AFFINITY_MAX_NODES);
+
+export const errorWithScalingOrAntiAffinity = (
```
fredericvilcot: Naming could be improved, looks like it returns an error instead ;)

packages/manager/apps/pci-kubernetes/src/pages/detail/nodepools/new/store.ts
```diff
@@ -46,7 +47,7 @@ export type TFormStore = {
   reset: () => void;
 
   scrollToStep: (step: StepsEnum) => void;
-};
+} & IsAntiAffinityOrAutoScaling;
```
fredericvilcot: Where does this type come from?

packages/manager/apps/pci-kubernetes/src/pages/new/steps/NodePoolStep.component.tsx
```diff
@@ -268,10 +233,25 @@ const NodePoolStep = ({
             )}
           <div className="mb-8">
             <NodePoolSize
+              scaling={nodePoolState.scaling}
               isMonthlyBilled={isMonthlyBilled}
-              onScaleChange={(scaling: AutoscalingState) =>
-                setNodePoolState((state) => ({ ...state, scaling }))
-              }
+              onScaleChange={(scaling) => {
```
fredericvilcot: Please, move this code out of JSX to a dedicated function :)

packages/manager/apps/pci-kubernetes/src/pages/new/steps/NodePoolStep.component.tsx
```diff
@@ -281,9 +261,21 @@ const NodePoolStep = ({
                 <NodePoolAntiAffinity
                   isChecked={nodePoolState.antiAffinity}
                   isEnabled={!nodePoolState.scaling?.isAutoscale}
-                  onChange={(antiAffinity: boolean) =>
-                    setNodePoolState((state) => ({ ...state, antiAffinity }))
-                  }
+                  onChange={(antiAffinity) => {
```
fredericvilcot: Same here ;)

packages/manager/apps/pci-kubernetes/src/types/index.ts
```diff
@@ -198,4 +199,29 @@ export enum DeploymentMode {
...
+export type TAntiAffinityOrAutoScaling =
+  | TAutoScalingOn
+  | TAntiAffinityOn
+  | TAutoScalingAndAntiAffinityOff;
+
+export type NodePoolState = {
+  name: string;
+  isTouched: boolean;
+  selectedAvailabilityZone?: string;
+} & TAntiAffinityOrAutoScaling;
```
fredericvilcot: Globally, this type is hard to discriminate as you have to declare type guards to check properties of sub-types.

Furthermore, TS engine does not make exclusive inference over union types, so the code below will not throw any issue ;)
But if we look into it, business rule is broken as `antiAffinity` is truthy so `isAutoscale` cannot be also thruthy 

```ts
const state: NodePoolState = {
  name: 'pool-1',
  isTouched: true,
  antiAffinity: true,
  scaling: { quantity: { desired: 1, max: 0, min: 0 }, isAutoscale: true }, <--- OK for TS engine 😱
};
```

As an improvement, I'll suggest to add a discriminant like a `mode` or something like that to easily handle type discrimination: 

```ts
type NodePoolMode =
  | { mode: 'autoscale'; scaling: AutoscalingState & { isAutoscale: true }; antiAffinity: false }
  | { mode: 'antiAffinity'; scaling: AutoscalingState & { isAutoscale: false }; antiAffinity: true }
  | etc...

type NodePoolState = {
  name: string;
  isTouched: boolean;
  selectedAvailabilityZone?: string;
} & NodePoolMode;
```

fredericvilcot (changes requested):

packages/manager/apps/pci-kubernetes/src/types/index.ts
```diff
@@ -2,6 +2,7 @@ import { z } from 'zod';
 import { OvhSubsidiary } from '@ovh-ux/manager-react-components';
 import { pluginData } from '@/api/data/plugins';
 import { isBase64 } from '@/helpers';
+import { AutoscalingState } from '@/components/Autoscaling.component';
```
lionel95200x: Think you should not call a component type from your type file
Maybe add this type in the file

packages/manager/apps/pci-kubernetes/src/helpers/node-pool.ts
```diff
@@ -0,0 +1,87 @@
...
+
+export const hasInvalidScalingOrAntiAffinityConfig = (
+  regionInformations: TRegionInformations,
+  nodePoolState: NodePoolState,
+) =>
+  !isScalingValid(nodePoolState) ||
+  !hasMax5NodesAntiAffinity(nodePoolState) ||
+  !zoneAZisChecked(regionInformations, nodePoolState);
+
+const autoScaleObject = (scaling: AutoscalingState): NodePoolMode => ({
```
lionel95200x: New mode implementation seems to be very complex, i think you can simplify it and remove all this file by just moving the scaling part oustide of NodePoolState object 

```suggestion
export type NodePoolState = {
  name: string;
  ... 
  scaling: AutoScale } & { isAutoscale: true; antiAffinity: false} |  { isAutoscale: false; antiAffinity: false} |  { isAutoscale: false; antiAffinity: true}
```

What do you think ?

fredericvilcot: @ppprevost @lionel95200x 
This works as `isAutoscale` property is now out of the first intersection of type, so TS can infer the union part.

But, removing this property from `AutoscalingState` is a non sense in the current architecture as nodes quantity is impacted by autoscaling. So we should consider reworking the state with this option ;)

packages/manager/apps/pci-kubernetes/src/pages/detail/nodepools/new/store.ts
fredericvilcot: Still here and I can find nowhere this type, could you please have a look?

SimonChaumet (dismissed):

fredericvilcot (approved):

SimonChaumet (approved):

PR #18227, opened 2025-07-18
feat(billing): my services datagrid add renew frequency column

oalkabouss:
ref: #MANAGER-18780

## Description

Ticket Reference: #MANAGER-18780

### DIFF
--- a/packages/manager/modules/billing/src/autoRenew/services/services.controller.js
+++ b/packages/manager/modules/billing/src/autoRenew/services/services.controller.js
@@ -14,6 +14,7 @@ import {
   TRACKING_PAGE_CATEGORY,
   TRACKING_PAGE,
   TRACKING_ACTIONS_PREFIX,
+  SERVICE_RENEW_MODES,
 } from '../autorenew.constants';
 
 export default class ServicesCtrl {
@@ -27,6 +28,7 @@ export default class ServicesCtrl {
     billingRenewHelper,
     coreConfig,
     ouiDatagridService,
+    billingPeriodTranslatorHelper,
   ) {
     this.$filter = $filter;
     this.$q = $q;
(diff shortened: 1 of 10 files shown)


packages/manager/modules/billing/src/autoRenew/services/services.controller.js
```diff
@@ -36,6 +39,8 @@ export default class ServicesCtrl {
     this.renewHelper = billingRenewHelper;
     this.coreConfig = coreConfig;
     this.ouiDatagridService = ouiDatagridService;
+    this.ovhUpdatePeriodTranslationService = ovhUpdatePeriodTranslationService;
```
JacquesLarique: This injection seems to be unused, I think you can remove it

packages/manager/modules/billing/src/autoRenew/services/services.html
```diff
@@ -179,9 +179,7 @@
             data-title="::'billing_autorenew_service_frequency' | translate"
             data-type="options"
         >
-            <span
-                data-ng-bind-html="$row.hasAutomaticRenew() && !$row.isResiliated() && $row.renewPeriod ? 'billing_autorenew_service_every_' + $row.renew.period + '_month': 'billing_autorenew_service_none' | translate"
-            >
+            <span data-ng-bind-html="$ctrl.getRenewPeriodTranslation($row)">
```
JacquesLarique: Can you use one time binding here ?

JacquesLarique (changes requested):

packages/manager/modules/billing/src/autoRenew/services/services.html
oalkabouss: yes

packages/manager/modules/billing/src/autoRenew/services/services.controller.js
oalkabouss: I'd forgotten about it. Thank you so much.

PR #18230, opened 2025-07-18
feat(*): static kit enhancement after v1 publish

helabenkhalfallah:
## Description

Enhance kit V1 after publishing (developers feedbacks).

Ticket Reference: #MANAGER-17954

### DIFF
--- a/packages/manager/tools/static-analysis-kit/src/adapters/eslint/rules/eslint-react.ts
+++ b/packages/manager/tools/static-analysis-kit/src/adapters/eslint/rules/eslint-react.ts
@@ -7,7 +7,6 @@ import { jsTsFiles } from '../../../configs/file-globs-config';
 
 /**
  * ESLint Flat Config for React and React Hooks.
- *
  * This configuration applies:
  * - Recommended React rules from `eslint-plugin-react`
  * - Recommended rules for hooks from `eslint-plugin-react-hooks`


packages/manager-cli/static-analysis-migrate/steps/eslint-config/addEslintStaticKitConfig.mjs
```diff
@@ -144,7 +147,27 @@ const addEslintStaticKitConfig = async () => {
     }
   }
 
-  // 2. Update package.json
+  // 2. Run the build script using yarn workspace command
```
fredericvilcot, line 150: We could prevent from building if versions are different between `package.json` and monolithe, as it will be automatically downloaded from npm registry.

packages/manager/tools/static-analysis-kit/src/adapters/eslint/rules/eslint-react.ts
```diff
@@ -43,6 +43,9 @@ export const reactEslintConfig: Linter.FlatConfig = {
     ...reactPlugin.configs.recommended.rules,
     ...reactHooksPlugin.configs.recommended.rules,
 
+    // https://github.com/jsx-eslint/eslint-plugin-react/blob/master/docs/rules/react-in-jsx-scope.md#when-not-to-use-it
+    'react/react-in-jsx-scope': 'off', // Disable React import rule for JSX
```
fredericvilcot: It could be great to extend from `'plugin:react/jsx-runtime'` , because it turns off 2 rules, as explained in the official documentation:
https://github.com/jsx-eslint/eslint-plugin-react?tab=readme-ov-file#configuration-legacy-eslintrc-

https://github.com/jsx-eslint/eslint-plugin-react/blob/c8917b0885094b5e4cc2a6f613f7fb6f16fe932e/index.js#L163-L176

I think it's much cleaner than turning-off rules manually, what do you think.

helabenkhalfallah: DONE!

packages/manager-cli/static-analysis-migrate/steps/eslint-config/addEslintStaticKitConfig.mjs
helabenkhalfallah: I'll do this enhancement with the next version, actually it prepare the dist locally, yarn will use it for versions >= local version else it'll fetch from registry for versions < local, I need to optimize this case, but I need to compare the two time:
- t1: actual time turbo build, it'll take time only when package change or we dont have dist.
- t2: to add the step of comparison.

I noted the enhancement.

fredericvilcot (approved):
LGTM, thx a lot! 🙏

PR #18232, opened 2025-07-18
fix(sap-features-hub): QA wave fixes

pauldkn:
## Description

QA wave fixes : 
- use odsPassword instead of odsText (ods fix)
- fix overflow in summary field
- fix toggle with not detected changes
- step 7: fix invalid state by default
- step 1: prevent selection while loading queries
- always initialize a clear form on wizard modal

Ticket Reference: #MANAGER-19010

### DIFF
--- a/packages/manager/apps/sap-features-hub/src/components/Fields/RhfToggleField.component.tsx
+++ b/packages/manager/apps/sap-features-hub/src/components/Fields/RhfToggleField.component.tsx
@@ -17,15 +17,26 @@ type RhfToggleFieldProps<TFieldName extends string> = {
   tooltip?: string;
 };
 
+// OdsToggle web component seems to have an issue where the new value is not always detected
+// As a temporary solution, the value is kept in sync with an internal state.
+// TODO: ODS migration
 export const RhfToggleField = <TFieldName extends string>({
   field,
   label,
   tooltip,
 }: RhfToggleFieldProps<TFieldName>) => {
+  const { name, value, onChange, onBlur } = field;
+  const [internalChecked, setInternalChecked] = React.useState(!!value);
+
   const toggleId = `${field.name}-toggle`;
+
+  React.useEffect(() => {
+    setInternalChecked(!!value);
+  }, [value]);
+
   return (
     <OdsFormField
-      key={toggleId}
+      key={`${toggleId}-${internalChecked}`}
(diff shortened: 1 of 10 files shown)


packages/manager/apps/sap-features-hub/src/components/Fields/RhfToggleField.component.tsx
```diff
@@ -17,15 +17,26 @@ type RhfToggleFieldProps<TFieldName extends string> = {
...
+// OdsToggle web component seems to have an issue where the new value is not always detected
+// As a temporary solution, the value is kept in sync with an internal state.
+// TODO: ODS migration
 export const RhfToggleField = <TFieldName extends string>({
   field,
   label,
   tooltip,
 }: RhfToggleFieldProps<TFieldName>) => {
+  const { name, value, onChange, onBlur } = field;
+  const [internalChecked, setInternalChecked] = React.useState(!!value);
```
darsene, line 29: should only be initialized as it's updated in the useEffect part
```suggestion
  const [internalChecked, setInternalChecked] = React.useState(false);
```

pauldkn: Actually it's necessary to initialize it with the default value passed as props, as we may launch the wizard with a file, it may be initialized as `true` (from previous form data)

packages/manager/apps/sap-features-hub/src/pages/installation/initialStep/InstallationInitialStep.page.tsx
```diff
@@ -55,16 +61,27 @@ export default function InstallationInitialStep() {
...
-    serviceName,
-    datacenterId: datacenterId?.toString(),
-  });
+  } = useClusters(
+    {
+      serviceName,
+      datacenterId: datacenterId?.toString(),
+    },
+    {
+      enabled: !!serviceDisplayName && !!datacenterName,
```
tibs245, line 80: Why not keep defautl behavior for enabled ?

pauldkn: It's because we want to be sure the user has selected a valid service, therefore has a `serviceDisplayName`.
Otherwise it could be an invalid `serviceName` inserted in a json file used to bootstrap the form and in this case we don't have to trigger the query

PR #18240, opened 2025-07-21
chore(ips): fixing the unit tests for the IPActionCell component

aderghamov:
## Description

Ticket Reference: #...


aderghamov:
Not needed anymore since it will be fixed by another PR

PR #18266, opened 2025-07-22
feat(pci-object-storage): add tags to replication

selm3n:
ref: #TAPC-4080

## Description

Ticket Reference: #...


packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/useTagValidation.ts
```diff
@@ -0,0 +1,137 @@
+import { TTagMap } from './ReplicationRuleTag.component';
```
lionel95200x: Your component name `useTagValidation` but there is no hook inside, maybe move it to an utils folder and add some unit test to check common case

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ManageReplicationPage.form.tsx
```diff
@@ -146,6 +152,27 @@ export function ManageReplicationForm({
     shouldEditRule ? existingRule.priority : 1,
   );
   const [priorityError, setPriorityError] = useState<boolean>(false);
+
+  const getInitialTags = (): TTagMap => {
```
SimonChaumet: Either this function doesn't depend on the props and should be outside, or if it depends on props it should be wrapped in a useCallback

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ManageReplicationPage.form.tsx
```diff
@@ -302,16 +333,55 @@ export function ManageReplicationForm({
     },
   });
 
+  const handleValidateTags = useCallback(() => {
```
SimonChaumet: `handle` name should be reserved to side-effect functions. Here it should be hasErrors or something near

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ManageReplicationPage.form.tsx
```diff
@@ -302,16 +333,55 @@ export function ManageReplicationForm({
...
+          if (key !== '') {
+            acc[key] = value;
+          }
+          return acc;
+        },
+        {} as Record<string, string>,
+      );
+      tagsToPass = {
+        ...transformedTags,
+        ...(newTag.key && !newTagErrors.key && !newTagErrors.value
```
SimonChaumet: How can newTagErrors be filled if you checked in handleValidateTags that it isn't ?

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ManageReplicationPage.form.tsx
```diff
@@ -302,16 +333,55 @@ export function ManageReplicationForm({
...
+    return !hasTagErrors && !hasNewTagErrors;
+  }, [tags, newTag, t]);
+
   const onSubmitReplicationRule = () => {
     if (!isButtonActive) return;
 
+    const hasTagsErrors = !handleValidateTags();
+
+    let tagsToPass = {};
+    if (!hasTagsErrors) {
```
SimonChaumet, line 361: It seems weird to accept the submit if there are errors inside the form

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ManageReplicationPage.form.tsx
```diff
@@ -302,16 +333,55 @@ export function ManageReplicationForm({
...
+      const transformedTags = Object.values(tags).reduce(
```
SimonChaumet, line 362: There are a lot of computations on these tags, maybe another data-structure would suit it more ? Maybe a Map ?

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ManageReplicationPage.form.tsx
```diff
@@ -396,13 +471,30 @@ export function ManageReplicationForm({
                   : undefined
               }
             />
+
+            <ReplicationRuleTag
+              tags={tags}
+              setTags={setTags}
+              t={t}
```
SimonChaumet: Why pass the t function ? You should get it from useTranslation isn't it ?

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/NewTagForm.component.tsx
```diff
@@ -0,0 +1,156 @@
...
+    <div className="w-full md:w-auto mt-4 mb-2 md:mb-0">
+      <OdsButton
+        label={t(
+          'containers/replication/add:pci_projects_project_storages_containers_replication_add_tags_add_new_tag',
+        )}
+        variant={ODS_BUTTON_VARIANT.outline}
+        onClick={onAddNewTag}
+        isDisabled={
+          deleteMarkerReplication === STATUS_ENABLED ||
+          (tagsLength > DEFAULT_MAX_TAGS - 2 && showNewTagForm)
```
SimonChaumet: Why 2 ?

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ManageReplicationPage.form.tsx
selm3n: the business demand is to validate the tags when clicking on the submit button

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ManageReplicationPage.form.tsx
selm3n: the tags business rule is to accept maximum 10 tags including the tags  and the newtag  inputs
so we have to merge them and adapt them before sending them to the api

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/NewTagForm.component.tsx
selm3n: The buisness rule is to disable the button addTag when attending 9 tags

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ManageReplicationPage.form.tsx
selm3n: Done

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ManageReplicationPage.form.tsx
SimonChaumet: I am talking about the data structure used in front here, not the business logic

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/NewTagForm.component.tsx
SimonChaumet: Here you say 9 tags so I don't see where this "2" comes from

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ManageReplicationPage.form.tsx
SimonChaumet: That is not what I am saying

selm3n: I added the if (!hasTagsErrors)  to do not submit if there are tagsEroors

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/NewTagForm.component.tsx
selm3n: the DEFAULT_MAX_TAGS = 10 and we have to disable the add button when we have 9 tags so the condition is 
tagsLength > DEFAULT_MAX_TAGS - 2. the DEFAULT_MAX_TAGS - 2 = 8

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/useTagValidation.ts
selm3n: Done

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ManageReplicationPage.form.tsx
selm3n: Current Approach: Record<number, { key: string; value: string }>
Pros:
Direct indexing: Easy to access/modify tags by their ID
Order preservation: Maintains insertion order (in modern JavaScript engines)
Simple CRUD operations: Easy to add/remove tags by ID
Compatibility: Works well with React state management

Cons:
Conversion needed: Requires transformation before sending to API
ID management: Need to track nextId separately

Alternative: Using a Map
Pros:
Built-in methods: Has convenient methods like .entries(), .values()
Better typing: Can enforce key-value types more strictly
Direct iteration: Can be iterated directly without Object conversion
Potentially better performance for frequent additions/removals

Cons:
Conversion still needed: Would still need to convert to plain object for API
Less React-friendly: React state works better with plain objects
More verbose: Requires new Map() and specific Map methods

Recommendation:
The current approach using Record<number, { key: string; value: string }> is actually quite good for this use case because:

Using React state management which works better with plain objects
The transformation to API format is straightforward (as shown in your code)
The ID-based access pattern fits well with the UI where tags can be individually edited/removed
The conversion cost is minimal since it only happens on form submission

If we were doing more complex tag manipulations or needed more advanced collection methods, a Map might be worth considering. But for our current needs, the Record type is simpler and works well.

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ReplicationRuleDeleteMarker.component.tsx
```diff
@@ -2,28 +2,46 @@ import { useTranslation } from 'react-i18next';
...
   deleteMarkerReplication,
   setDeleteMarkerReplication,
   asyncReplicationLink,
+  tags,
+  newTag,
+  isReplicationApplicationLimited,
 }: TReplicationRuleDeleteMarkerProps) {
   const { t } = useTranslation(['containers/replication/add']);
 
+  const disabledReplicateDeleteMarker = useMemo(() => {
```
lionel95200x: Better to prefix all boolean using `is` or `should`

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/NewTagForm.component.tsx
```diff
@@ -0,0 +1,156 @@
...
+  OdsText,
+} from '@ovhcloud/ods-components/react';
+import { ODS_BUTTON_VARIANT } from '@ovhcloud/ods-components';
+import clsx from 'clsx';
+import { TReplicationStatus } from './ManageReplicationPage.form';
+import { STATUS_ENABLED } from '@/constants';
+import { DEFAULT_MAX_TAGS } from './ReplicationRuleTag.component';
+
+export type TNewTagFormProps = {
+  newTag: { key: string; value: string };
```
lionel95200x: Could you update it everywhere

```suggestion
  newTag: TNewTag;
```

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/NewTagForm.component.tsx
```diff
@@ -0,0 +1,156 @@
...
+              errors?.key && !errors?.value && 'mb-[21px]',
+            )}
+            placeholder={t(
+              'containers/replication/add:pci_projects_project_storages_containers_replication_add_value_placeholder',
+            )}
+            value={newTag.value}
+            name="new-tag-value"
+            color="primary"
+            hasError={!!errors?.value}
+            onFocus={() => onNewValueFocus()}
```
lionel95200x:
```suggestion
            onFocus={onNewValueFocus}
```

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ManageReplicationPage.form.tsx
```diff
@@ -146,6 +152,27 @@ export function ManageReplicationForm({
...
+      (acc, [key, value], index) => {
+        acc[index + 1] = { key, value };
+        return acc;
+      },
+      {},
+    );
+  }, [shouldEditRule, existingRule?.filter?.tags]);
+
+  const [tags, setTags] = useState<TTagMap>(getInitialTags);
+  const [newTag, setNewTag] = useState({ key: '', value: '' });
```
lionel95200x: you could declare an empty tag  in utils file
```suggestion
  const emptyTag = { key: '', value: '' }
  
```

An reuse it in your component where needed
```
const [newTag, setNewTag] = useState(emptyTag);

...
setNewTag(emptyTag)
```

selm3n: Done

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/NewTagForm.component.tsx
selm3n: Done

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/NewTagForm.component.tsx
selm3n: Done

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/ReplicationRuleDeleteMarker.component.tsx
selm3n: Done

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/add-replication/NewTagForm.component.tsx
SimonChaumet: What do you think of tagsLength >= DEFAULT_MAX_TAGS - 1 instead ?

selm3n: Done

PR #18267, opened 2025-07-22
feat(manager-react-components): migrate tile to ODS v19

anooparveti:
## Description

Refactor Tile component to better semantic structure with section element 
Enclose all dt and dd in only only dl element
Introduce new development guidline to export hierarchy of components 
Add unit and Integrate Tests
Add documentation to explain Tile component

Ticket Reference: #MANAGER-18746    

## Additional Information

BREAKING CHANGE: 

1. Rename ManagerTile component to Tile 
2. Rename ManagerTile.Item.label to Manager.Item.Term 
3. Mandatory Title prop for Tile component


packages/manager-wiki/stories/manager-react-components/content/tile/tile.stories.tsx
```diff
@@ -0,0 +1,161 @@
...
+  },
+  {
+    id: 2,
+    onClick: () => window.open('https://ovhcloud.com', '_blank', 'noopener'),
+    label: 'Action 2',
+  },
+];
+
+export const SimpleTile = () => (
+  <Tile.Root title="Sample Tile Header">
```
aboungnaseng-ovhcloud, line 28: `
<Tile title="Sample Tile Header">
   <TileItem>
      <TileItem.Term label="Term" />
      <TileItem.Description label="Description" />
    </TileItem>
    <TileItem>
      <TileItem.Term label="Last Term" />
      <TileItem.Description divider={false}>
        <Text>Description without Divider</Text>
      </TileItem.Description>
    </TileItem>
  </Tile>
`

Is it possible to have only Tile component and TileItem ?

anooparveti: No, to provide more autonomy to devs for using `dl..dt..dd` with all possible ways, this is the best split.

packages/manager-react-components/src/components/tile/__tests__/Tile.test.tsx
```diff
@@ -0,0 +1,98 @@
...
+import { render, screen } from '@testing-library/react';
+import { userEvent } from '@testing-library/user-event';
+import { Tile } from '../index';
+import { ActionMenu } from '../../navigation/menus/action/action.component';
+import { Link } from '@ovhcloud/ods-react';
+
+describe('Tile Snapshot tests', () => {
+  it('renders simple tile', () => {
+    const { container } = render(
+      <Tile.Root title="Simple Tile">
```
helabenkhalfallah, line 11: I like the namespace approach! nice! is not possible to do: 
```
<Tile
   Tile.Item
       ...
```

Without `Root`?

packages/manager-react-components/src/components/tile/tile-root/TileRoot.component.tsx
```diff
@@ -0,0 +1,27 @@
...
+  ...props
+}: TileRootProps) => {
+  return (
+    <Card
+      className={clsx('w-full flex-col p-[1rem]', className ?? className)}
+      color={color}
+      {...props}
+    >
+      <section className="flex flex-col w-full">
+        <Text preset={TEXT_PRESET.heading4}>{title}</Text>
```
helabenkhalfallah, line 21: the issue of fixing heading to 4 is that it depends on the page organization. It can be wrong. If we say to developers to pay attention, I'm afraid to not be sufficient.

packages/manager-react-components/src/components/tile/__tests__/__snapshots__/Tile.test.tsx.snap
```diff
@@ -0,0 +1,316 @@
...
+        <div
+          class="flex flex-col gap-1"
+        >
+          <dt
+            class="flex justify-between"
+          >
+            <div
+              class="flex items-center gap-1 text-[var(--ods-color-text)]"
+            >
+              <label
```
helabenkhalfallah: why label?

packages/manager-react-components/src/components/tile/__tests__/__snapshots__/Tile.test.tsx.snap
```diff
@@ -0,0 +1,316 @@
...
+          <dd
+            class="m-0"
+          >
+            <a
+              class="_link_udbet_2"
+              href="https://www.ovhcloud.com"
+            >
+              Link
+            </a>
+            <hr
```
helabenkhalfallah, line 56: I have a doubt about this hr:  
each ```<dd>``` should describe the ```<dt>``` above it. If the ```<hr>``` is visual and not logically tied to the term-description structure, move it outside the ```<dd>```, after the list item.

I think we should do:
```js
<dd class="m-0">
  <a class="_link_udbet_2" href="https://www.ovhcloud.com">Link</a>
</dd>
<hr class="_divider..." />
```

packages/manager-react-components/src/components/tile/__tests__/__snapshots__/Tile.test.tsx.snap
```diff
@@ -0,0 +1,316 @@
...
+        <div
+          class="flex flex-col gap-1"
+        >
+          <dt
+            class="flex justify-between"
+          >
+            <div
+              class="flex items-center gap-1 text-[var(--ods-color-text)]"
+            >
+              <label
```
helabenkhalfallah: same https://github.com/ovh/manager/pull/18267/files#r2242786439

packages/manager-react-components/src/components/tile/__tests__/__snapshots__/Tile.test.tsx.snap
```diff
@@ -0,0 +1,316 @@
...
```
helabenkhalfallah: https://github.com/ovh/manager/pull/18267/files#r2242786439

packages/manager-react-components/src/components/tile/tile-divider/TileDivider.component.tsx
```diff
@@ -0,0 +1,5 @@
+import { Divider, DIVIDER_SPACING, DividerProp } from '@ovhcloud/ods-react';
+
+export const TileDivider = ({ ...rest }: DividerProp) => (
```
helabenkhalfallah: why not: 

```
export const TileDivider = (props: DividerProp) => (
         <Divider spacing={DIVIDER_SPACING._24} {...props} />
 );
```

Since we use all the props.

packages/manager-react-components/src/components/tile/tile-divider/__tests__/TileDivider.spec.tsx
```diff
@@ -0,0 +1,30 @@
+import { describe, it, expect } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import { TileDivider } from '../TileDivider.component';
+
+const renderComponent = ({ ...props }) => {
```
helabenkhalfallah: I commented this approach in Alex PR: https://github.com/ovh/manager/pull/18342/files#r2242158612

packages/manager-react-components/src/components/tile/tile-item/tile-item-description/__tests__/TileItemDescription.spec.tsx
```diff
@@ -0,0 +1,68 @@
+import { describe, it, expect, vitest } from 'vitest';
+import { render, screen } from '@testing-library/react';
+import { TileItemDescription } from '../TileItemDescription.component';
+
+vitest.mock('../../../tile-divider/TileDivider.component', () => ({
```
helabenkhalfallah, line 7: I need an argument to accept the mock :) :)

packages/manager-react-components/src/components/tile/tile-item/tile-item-term/TileItemTerm.component.tsx
```diff
@@ -0,0 +1,37 @@
...
+export const TileItemTerm = ({
+  label,
+  tooltip,
+  actions,
+  ...rest
+}: TileItemTermProps) => {
+  return (
+    <dt className="flex justify-between" {...rest}>
+      <div className="flex items-center gap-1 text-[var(--ods-color-text)]">
+        <Text preset={TEXT_PRESET.label} className="font-bold">
```
helabenkhalfallah: I think this should be span not a label, what do you think?

packages/manager-react-components/src/components/tile/tile-root/TileRoot.component.tsx
```diff
@@ -0,0 +1,27 @@
...
+        <TileDivider className="w-full" />
+        <dl className="flex flex-col m-0">{children}</dl>
```
helabenkhalfallah, line 23: is not possible to use TS to limit acceptable children type:
https://developer.mozilla.org/en-US/docs/Web/HTML/Reference/Elements/dl#technical_summary

packages/manager-react-components/src/components/tile/__tests__/Tile.test.tsx
anooparveti: No, since we will export the root component as `Tile` itself, it is a composition. So, root cannot be omitted.

packages/manager-react-components/src/components/tile/__tests__/__snapshots__/Tile.test.tsx.snap
anooparveti:
```
<dt></dt>
<hr />
<dd></dd>
``` 
Above pattern is semantically wrong. So, i have implemented `Tile.Item.Description` in such a way that `hr` element is part of `dd`.

Resulting structure is,

```
<dt></dt>
<dd>
   ....
   <hr />
</dd>
```

packages/manager-react-components/src/components/tile/tile-root/TileRoot.component.tsx
anooparveti: this component is always used in Dashboard page which already has a `h1`.
And the resulted pattern does not have any accessibility violation.

Screenshot 2025-07-30 at 8 22 32 PM

packages/manager-react-components/src/components/tile/tile-item/tile-item-term/TileItemTerm.component.tsx
anooparveti: In some cases it can be beneficial to have label when we display clipboard. But span is better suitable for all cases.
I will update it.

packages/manager-react-components/src/components/tile/tile-root/TileRoot.component.tsx
anooparveti: I found that it's not possible purely with Typescript - https://www.totaltypescript.com/type-safe-children-in-react-and-typescript
However, we can enforce it with eslint plugin https://www.npmjs.com/package/eslint-plugin-enforce-children.

If you like this idea, maybe we can plan a different ticket to integrate it with static toolkit.

packages/manager-react-components/src/components/tile/__tests__/__snapshots__/Tile.test.tsx.snap
helabenkhalfallah: Perfect, thank you!

helabenkhalfallah (approved):

PR #18281, opened 2025-07-22
fix(manager-pci-common): add showTooltip to RegionTile

xobaba:
## Description
Added the ability to display a tooltip for the RegionChipByType component. A new showTooltip parameter has been added to the RegionTileProps interface to enable/disable this feature. The RegionTile component can now display a tooltip for the region type if showTooltip is set to true.

Ticket Reference: #TAPC-4008

### DIFF
--- a/packages/manager/modules/manager-pci-common/src/components/region-selector/RegionTile.tsx
+++ b/packages/manager/modules/manager-pci-common/src/components/region-selector/RegionTile.tsx
@@ -9,12 +9,14 @@ export interface RegionTileProps {
   region: TLocalisation;
   isSelected: boolean;
   isCompact?: boolean;
+  showTooltip?: boolean;
 }
 
 export const RegionTile = ({
   region,
   isSelected,
   isCompact,
+  showTooltip,
 }: Readonly<RegionTileProps>) => (
   <div className="flex flex-col w-full items-center">
     <div className={isCompact ? 'my-4' : ''}>
@@ -30,7 +32,7 @@ export const RegionTile = ({
       <>
         <hr className="w-full border-solid border-0 border-b border-ods-primary-200" />
         <div>
-          <RegionChipByType type={region.type} />
+          <RegionChipByType showTooltip={!!showTooltip} type={region.type} />
         </div>
       </>
     )}


packages/manager/modules/manager-pci-common/src/components/region-selector/RegionTile.tsx
```diff
@@ -9,12 +9,14 @@ export interface RegionTileProps {
   region: TLocalisation;
   isSelected: boolean;
   isCompact?: boolean;
+  showTooltip;
```
ppprevost: No definition ?

fredericvilcot (commented):
@xobaba 
Thanks for then squash :)
Could you please enhance the commit message by adding the component impacted?
Vecause here, in the changelog we'll see you added some `showTooltip` property , but where? ;)

fredericvilcot (dismissed):
👍🏻

packages/manager/modules/manager-pci-common/src/components/region-selector/RegionTile.tsx
```diff
@@ -30,7 +32,7 @@ export const RegionTile = ({
       <>
         <hr className="w-full border-solid border-0 border-b border-ods-primary-200" />
         <div>
-          <RegionChipByType type={region.type} />
+          <RegionChipByType showTooltip={showTooltip} type={region.type} />
```
ppprevost: showToolTip could be undefined--> showTooltip={!!showTooltip}

fredericvilcot: @ppprevost @xobaba 
AFAIS, `showTooltip ` is optional in the `RegionChipByType` props definition, so this change is useless ;)
BTW, please, I'd like you to check how the `RegionChipByType` consumes this props, because right now it's not used anywhere: 

```tsx
export const RegionChipByType = ({ type }: RegionChipByTypeProps) => {
  const Chip = CHIP_BY_TYPE[type];
  return Chip ? <Chip /> : null;
};
```

xobaba:
See with Nicolas Lacroux and Pierre-Philippe Prevost 
No modification pci-common
No remove tooltip in "Select a location"

PR #18302, opened 2025-07-23
feat(communication): add communication listing page

aottr:
## Description

This change adds the communications listing tab with the current implementation of the v2 `/notification/history` api.

Ticket Reference: #MANAGER-18606

## Additional Information
- We might want to replace the `useNotificationHistory` hook with `useResourcesIcebergV2` if the api changes it's filtering
- There is a wrapper component for the `OdsLink` that enables the usage of a slot which might be replaced by `OdsLink` in ODS v19


packages/manager/apps/communication/src/pages/communications/Communications.page.tsx
```diff
@@ -0,0 +1,145 @@
...
+        <DataGridTextCell>{notification.title}</DataGridTextCell>
+      ),
+    },
+    {
+      id: 'status',
+      label: '',
+      isFilterable: true,
+      cell: (notification) => (
+        <DataGridTextCell>
+          <DataGridTextCell>
```
JacquesLarique: Why two `<DataGridTextCell>` ?

packages/manager/apps/communication/src/components/notificationContactStatus/NotificationContactStatus.component.tsx
```diff
@@ -0,0 +1,80 @@
...
+    () =>
+      contacts.reduce((highest, current) => {
+        if (
+          !highest ||
+          STATUS_PRIORITY[current.status] > STATUS_PRIORITY[highest.status]
+        ) {
+          return current;
+        }
+        return highest;
+      }, null as Contact | null),
```
JacquesLarique, line 35: Wouldn't it be simpler / more performant to just find a contact for which the status is not 'SENT' ?

_I'm not sure to understand the need correctly_

packages/manager/apps/communication/src/pages/communications/Communications.page.tsx
```diff
@@ -0,0 +1,145 @@
...
+      label: t('table_column_categories'),
+      isFilterable: true,
+      type: FilterTypeCategories.Options,
+      comparator: [FilterComparator.IsEqual],
+      filterOptions: reference?.categories?.map(({ name }) => ({
+        label: name,
+        value: name,
+      })),
+      cell: (notification) => (
+        <DataGridTextCell>
```
JacquesLarique: Shouldn't we add a class to handle max-width ?
I'm afraid in some case the list might be long and result in a multi line display

packages/manager/apps/communication/src/pages/dashboardLayout.tsx
```diff
@@ -60,18 +56,22 @@ export default function DashboardPage() {
...
-                {tab.title}
-              </NavLink>
+            <OdsTab
+              key={`osds-tab-bar-item-${tab.name}`}
+              className="select-none"
+              isSelected={tab.name === panel}
+              onOdsTabSelected={(event) => {
+                event.preventDefault();
+                navigate(tab.to);
+              }}
```
JacquesLarique: Is it necessary ?
I see on other apps that it is just used for tracking purposes

packages/manager/apps/communication/src/components/notificationContactStatus/NotificationContactStatus.component.tsx
aottr: I can change this if the demand changes but I got asked to display a red triangle on `DROPPED` and a yellow triangle on `QUEUED` with `DROPPED` being the higher severity (`DROPPED` > `QUEUED`) So if I exist early on a `QUEUED`, a `DROPPED` try might be overlooked.

packages/manager/apps/communication/src/pages/dashboardLayout.tsx
aottr: It is the non-generic event for the change handling. Other apps use either the `onOdsTabsSelected` of the parent component or a generic `onClick`.

So it is necessary, unless we want to change it to the generic handler.

packages/manager/apps/communication/src/pages/communications/Communications.page.tsx
aottr: The amount of categories is limited by the API but we can add a word-wrap for the categories and maybe a truncate for the title of the message.

packages/manager/apps/communication/src/pages/communications/Communications.page.tsx
aottr: Good find, it's a mistake from moving cells ^^

JacquesLarique (approved):

MaximeBajeux (approved):

PR #18312, opened 2025-07-23
feat(pci-instances): add distant backup creation from details page

SimonChaumet:
ref: #TAPC-4432

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/modules/pci/src/components/project/images-list/images-list.controller.js
+++ b/packages/manager/modules/pci/src/components/project/images-list/images-list.controller.js
@@ -8,13 +8,20 @@ import some from 'lodash/some';
 import groupBy from 'lodash/groupBy';
 
 import { IMAGE_ASSETS } from './images.constants';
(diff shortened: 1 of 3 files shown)


packages/manager/modules/pci/src/projects/project/instances/instance/backup/backup.controller.js
```diff
@@ -86,4 +176,18 @@ export default class PciInstanceBackupController {
...
   }
+
+  getRegionGroup(region) {
+    return region.continent;
+  }
+
+  shouldEnableRegion(regionName) {
+    if (!regionName) return false;
+
+    const region = !this.distantBackupRegions.find(
```
lolottetheclash: If I understand correctly, this means it's a boolean, not a region? 
if it's the case, maybe change the naming?

SimonChaumet: The exclamation mark should not be here, good catch

selm3n (approved):

lolottetheclash (approved):

PR #18316, opened 2025-07-23
feat(manager-react-components): delete title subtitle

aboungnaseng-ovhcloud:
ref: #MANAGER-18713

## Description

- delete title subtitle from typography

Ticket Reference: #MANAGER-18713


packages/manager-react-components/src/components/templates/base/headers/__tests__/Headers.snapshot.test.tsx
```diff
@@ -9,6 +9,25 @@ import {
   ChangelogButton,
 } from '../../../../navigation';
 
+vitest.mock('../../../../../hooks/iam');
```
anooparveti, line 12: **question:** Is it still needed?

aboungnaseng-ovhcloud: Yes we need it cause in Headers we pass Link with Iam authorization, and we need to mock it

PR #18326, opened 2025-07-24
feat(sap-features-hub): µapp shadow prod with auto trad

pauldkn:
### This branch has been restored due to an incorrect commit reference that need to be rewritten.

This branch is a copy of branch #14719, with automatic translation (shadow prod).
The base branch is still open, waiting for the real translation process.

Every change has already been validated on base branch.

Ticket Reference: #MANAGER-19043


pauldkn:
Copy not needed anymore (of #18282)

PR #18336, opened 2025-07-24
feat(core): add filter by tag feature to iam route

qpavy:
## Description

To be able to filter on iamTags, all resource can be filtered using iamTags query param (ex: /dedicated/server endpoint).
And the filter is already working for all those cases.

But for the endpoint /iam/resource (and only this one) the filterd should be made using tags query param instead of iamTags. To fix filter by tags on /iam/resource endpoint, this PR is needed.

Ticket Reference: #MANAGER-15795

### DIFF
--- a/packages/manager/core/api/src/iceberg.ts
+++ b/packages/manager/core/api/src/iceberg.ts
@@ -149,11 +149,11 @@ export async function fetchIcebergV2<T>({
     .build();
 
   const params = new URLSearchParams();
-
   if (filters?.length) {
     const tagsFilterParams = transformTagsFiltersToQuery(filters);
     if (tagsFilterParams) {
-      params.append('iamTags', tagsFilterParams);
+      const paramName = route.includes('/iam/resource') ? 'tags' : 'iamTags';
+      params.append(paramName, tagsFilterParams);
     }
   }
 


packages/manager/core/api/src/iceberg.ts
```diff
@@ -146,10 +146,12 @@ export async function fetchIcebergV2<T>({
     .build();
 
   const params = new URLSearchParams();
-
   if (filters?.length) {
     const tagsFilterParams = transformTagsFiltersToQuery(filters);
-    if (tagsFilterParams) {
+    if (tagsFilterParams && route.includes('/iam/resource')) {
```
helabenkhalfallah: this a common module, I may need the help of @aboungnaseng-ovhcloud  and @anooparveti .

packages/manager/core/api/src/iceberg.ts
```diff
@@ -146,10 +146,12 @@ export async function fetchIcebergV2<T>({
...
+      params.append('tags', tagsFilterParams);
+    }
+    if (tagsFilterParams && !route.includes('/iam/resource')) {
       params.append('iamTags', tagsFilterParams);
     }
```
anooparveti:
```suggestion
    if (tagsFilterParams) {
      const paramName =  route.includes('/iam/resource') ? 'tags' : 'iamTags';
      params.append(paramName, tagsFilterParams);
    }
```

anooparveti (approved):

aderghamov (approved):

PR #18380, opened 2025-07-28
feat(manager-react-components): migrate TagsList to ODS v19

anooparveti:
## Description

Refactor TagsList component to tackle number of tags to display 
Migrate TagsList component to ODS v19
Add unit and integration tests

Ticket Reference: #MANAGER-19045    

## Additional Information

BREAKING CHANGE: Rename lineNumber prop to maxLines


packages/manager-react-components/src/components/tags-list/TagsList.component.tsx
```diff
@@ -0,0 +1,109 @@
...
+    return null;
+  }
+
+  return (
+    <div
+      ref={containerRef}
+      className="w-full h-full flex flex-wrap min-w-[85px]"
+    >
+      <Badge
+        className="mr-1 mb-1 truncate inline-block"
```
aboungnaseng-ovhcloud, line 71:
```suggestion
        className="mr-1 mb-1 truncate inline-block"
```

packages/manager-react-components/src/components/tags-list/TagsList.component.tsx
```diff
@@ -0,0 +1,109 @@
...
+        color={BADGE_COLOR.neutral}
+        ref={tagRef}
+        style={{
+          maxWidth: `${maxWidth}px`,
```
aboungnaseng-ovhcloud, line 75: we can use `max-w-[${maxWidth}px]`from tailwind ?

anooparveti: Actually, i dont have the same behaviour in UI. So, I'll keep it this way.

helabenkhalfallah (approved):

PR #18402, opened 2025-07-28
feat(common-translations): add deploy translation

ppprevost:
ref:#TAPC-4420

## Description

Ticket Reference: #...


ppprevost:
It's about deployment zone (mono zone or multi zone), any advice ?

PR #18403, opened 2025-07-28
feat(billing): update actions menu in my services page for us manager

oalkabouss:
ref: #MANAGER-18779

## Description

Ticket Reference: #MANAGER-18779


packages/manager/modules/billing-components/src/components/services-actions/services-actions.controller.js
```diff
@@ -132,7 +134,8 @@ export default class ServicesActionsCtrl {
           this.autorenewLink &&
           (this.service.canBeDeleted() || this.service.isSuspendedHostingWeb());
         this.canDisplaySmsSpecificMenuEntries =
-          this.service.serviceType === this.SERVICE_TYPE.SMS;
+          this.service.serviceType === this.SERVICE_TYPE.SMS &&
+          !this.isUSRegion;
```
JacquesLarique: This is unecessary: SMS (and globally telecom services) are not available in US (so the check on service type is already enough)

packages/manager/modules/billing-components/src/components/services-actions/services-actions.controller.js
```diff
@@ -92,27 +90,31 @@ export default class ServicesActionsCtrl {
...
+          this.service.serviceType === this.SERVICE_TYPE.EXCHANGE &&
+          !this.isUSRegion;
         this.canDisplayXdslSpecificResiliationMenuEntry =
           this.service.serviceType === this.SERVICE_TYPE.PACK_XDSL &&
           !this.service.shouldDeleteAtExpiration() &&
           !this.service.isResiliated() &&
           !this.service.hasDebt() &&
-          !this.service.hasPendingResiliation();
+          !this.service.hasPendingResiliation() &&
+          !this.isUSRegion;
```
JacquesLarique: This is unecessary: XDSL pack services are not available in US (so the check on service type is already enough)

packages/manager/modules/billing/src/autoRenew/autorenew.service.js
```diff
@@ -154,37 +168,42 @@ export default class {
...
+      this.updateService(service),
+    );
+
+    return this.$q.all(updatePromises).then((results) => {
+      const errors = results.filter((res) => res.type === 'ERROR');
+
+      return {
+        messages: errors,
+        state: errors.length > 0 ? 'PARTIAL' : 'OK',
+      };
```
JacquesLarique: I don't think this give the same result as previously: here you'll always resolve the promise, when previously we were rejecting if there was any error

packages/manager/modules/billing/src/autoRenew/autorenew.service.js
```diff
@@ -154,37 +168,42 @@ export default class {
...
+    return this.OvhHttp.put(url, {
+      rootPath: 'apiv6',
+      data: {
+        renew: service.renew,
+      },
+    }).catch((err) => ({
+      id: service.serviceId,
+      message: err,
+      type: 'ERROR',
+    }));
```
JacquesLarique: I don't think this give the same result as previously: here you'll always resolve the promise, when previously we were rejecting if there was any error

packages/manager/modules/billing-components/src/components/services-actions/services-actions.controller.js
```diff
@@ -92,27 +90,31 @@ export default class ServicesActionsCtrl {
...
         this.canDisplayManageCommitmentMenuEntry =
           this.service.canBeEngaged &&
           !this.service.hasPendingEngagement &&
           !this.service.isSuspended();
         this.canDisplayCancelCommitmentMenuEntry = this.service.hasPendingEngagement;
         this.canDisplayExchangeSpecificMenuEntries =
-          this.service.serviceType === this.SERVICE_TYPE.EXCHANGE;
+          this.service.serviceType === this.SERVICE_TYPE.EXCHANGE &&
+          !this.isUSRegion;
```
JacquesLarique: This is unecessary: EXCHANGE services are not available in US (so the check on service type is already enough)

JacquesLarique (changes requested):

PR #18409, opened 2025-07-29
Feat/tapc 3830

aTurmo:
It's possible that I will add another commit to change some tracking page keys.


packages/manager/apps/pci-load-balancer/src/components/create/InstanceTable.component.tsx
```diff
@@ -116,9 +116,9 @@ export function InstanceTable({
               ],
             },
           ]);
-          tracking?.trackClick({
-            name: `${TRACKING_PREFIX}::add-listener`,
-            type: 'action',
+          trackClick({
+            actions: [`${TRACKING_PREFIX}`, 'add-listener'],
```
SimonChaumet: I don't think TRACKING_PREFIX should be kept when using useOvhTracking, can you verify ?

packages/manager/apps/pci-load-balancer/src/hooks/useTranslatedLinkReference.ts
```diff
@@ -23,9 +26,10 @@ export const useTranslatedLinkReference = () => {
           anchor.setAttribute('data-handled', 'true');
           if (trackOn === 'click') {
             anchor.addEventListener('click', () => {
-              tracking.trackClick({
-                name: trackName,
-                type: trackType || 'action',
+              trackClick({
+                actions: [trackName],
+                actionType:
+                  (trackType as TrackingClickParams['actionType']) || 'action',
```
SimonChaumet: What happens if the actionType isn't a valid param ? Could we add a predicate to verify the cast ?

packages/manager/apps/pci-load-balancer/src/pages/create/hooks/useCreateActions.tsx
```diff
@@ -17,24 +22,33 @@ export const useCreateActions = () => {
...
+        location: PageLocation.funnel,
       });
 
       store.create(flavor, onSuccess, onError);
     },
     cancel: () => {
       trackClick({
-        name: LOAD_BALANCER_CREATION_TRACKING.CANCEL,
-        type: 'action',
+        actions: ['cancel', `${storeTracingInfo}`],
```
SimonChaumet: template string doesn't do anything here

packages/manager/apps/pci-load-balancer/src/pages/create/steps/region/RegionStep.tsx
```diff
@@ -131,7 +134,12 @@ export const RegionStep = ({
       order={1}
       next={{
         action: () => {
-          trackStep(1);
+          trackClick({
+            actions: ['select_location', `add_${store.region?.name}`],
```
SimonChaumet: When clicking next the region shouldn't be null isn't it ?

packages/manager/apps/pci-load-balancer/src/pages/delete/Delete.page.spec.tsx
```diff
@@ -14,6 +16,11 @@ vi.mock('@/api/hook/useLoadBalancer', async () => {
   };
 });
 
+const mockedTrackClick = vi.fn();
+vi.mocked(useOvhTracking).mockReturnValue(({
+  trackClick: mockedTrackClick,
+} as unknown) as ReturnType<typeof useOvhTracking>);
```
SimonChaumet: the cast as unknown shouldn't be required, please fix the typing

packages/manager/apps/pci-load-balancer/src/pages/delete/Delete.page.tsx
```diff
@@ -64,11 +71,24 @@ export default function DeletePage() {
     },
   });
 
+  const trackClickButton = useCallback(
+    (button: 'confirm' | 'cancel') =>
+      trackClick({
+        actions: ['delete_loadbalancer', button, loadBalancer?.region],
```
SimonChaumet: can loadBalancer be null here ?

packages/manager/apps/pci-load-balancer/src/pages/create/steps/region/RegionStep.tsx
aTurmo: The region should not be null on click, but it can be when the component is intanciated

SimonChaumet: So if at this point it cannot be null the '?' shouldn't be set, you can use '!' instead to assert the nun-null

packages/manager/apps/pci-load-balancer/src/hooks/useTranslatedLinkReference.ts
aTurmo: If trackType is not correct it will still take it as a valid 'actionType'
I wish I could easily but tracking param is define in useOvhTracking :
```typescript
export type TrackingClickParams = {
  location?: PageLocation;
  buttonType?: ButtonType;
  actions?: string[];
  actionType?: 'action' | 'navigation' | 'download' | 'exit';
};
```
The only thing I could see doing is add a 'isActionType' function to check is the string is in the list of possible values. It would be better to add it in the useOvhTracking file directly but I wanted to avoid touching 'manager-react-shell-client'.

I don't know if there is a better solution

packages/manager/apps/pci-load-balancer/src/pages/delete/Delete.page.spec.tsx
aTurmo: I want to avoid mocking the all useOvhTracking exports. I'll do something better

packages/manager/apps/pci-load-balancer/src/pages/delete/Delete.page.tsx
aTurmo: it can be when the component is loaded

packages/manager/apps/pci-load-balancer/src/hooks/useTranslatedLinkReference.ts
SimonChaumet: I think the isActionType solution is the best, but it's up to you if it is worth it to do the change

packages/manager/apps/pci-load-balancer/src/pages/delete/Delete.page.spec.tsx
SimonChaumet: The cast `as ReturnType<typeof xxx>` is sufficient to do this, the "as unknown" is only required if the two types are incompatible, so either it is required and you have a mismatch, or it's not required

packages/manager/apps/pci-load-balancer/src/pages/delete/Delete.page.tsx
SimonChaumet: But the trackClick isn't call at this point ?

Type of the variable is not the same as the type at this point : if at this point it cannot be null you should use "!" to assert it

packages/manager/apps/pci-load-balancer/src/setupTests.tsx
```diff
@@ -15,7 +15,7 @@ vi.mock('react-router-dom', async () => {
     }),
     useHref: vi.fn(),
     useLocation: vi.fn(),
-    useNavigate: vi.fn(),
+    useNavigate: vi.fn().mockReturnValue(vi.fn()),
```
ppprevost, line 18: why return a vi.fn()?

packages/manager/apps/pci-load-balancer/src/hooks/useTranslatedLinkReference.ts
fredericvilcot: Yes, please add a type guard here to safely prevent from unexpected errors, because type cast only shunts `ts` engine and makes the code weaker.

packages/manager/apps/pci-load-balancer/src/setupTests.tsx
aTurmo: useNavigate returns a 'navigate' function.
If I don't do that I get a 'TypeError: navigate is not a function'

packages/manager/apps/pci-load-balancer/src/hooks/useTranslatedLinkReference.ts
aTurmo: still waiting for PR https://github.com/ovh/manager/pull/18440 to be approved and merged

aTurmo: done

Tsiorifamonjena (dismissed):

packages/manager/apps/pci-load-balancer/src/pages/delete/Delete.page.spec.tsx
```diff
@@ -38,4 +43,42 @@ describe('DeletePage', () => {
...
+  });
+
+  it('should track clicking on cancel', async () => {
+    vi.mocked(useLoadBalancer).mockReturnValue({
+      data: { id: 'lb1', name: 'Load Balancer 1', region: 'region' },
+      isPending: false,
+    } as UseQueryResult<TLoadBalancer>);
+
+    const { getByText } = render(<DeletePage />, { wrapper });
+    await act(() =>
```
ppprevost: todo(if-minor): do not use act with userevent

As it already wrap inside UserEvent, instead use await userEvent

ppprevost (dismissed):

PR #18433, opened 2025-07-30
ci(shell-client): add test configuration

fredericvilcot:
This PR adds test configuration inherited from `@ovh-ux/manager-tests-setup` shared config.

BYW, it adds the package in the workspace main configuration that was causing the module to be treated out of the mono repository, as a standalone package. 

It should also fix the auto-versionning that was not working anymore due to workspace configuration.

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,24 +8,25 @@
     "packages": [
       "docs",
       "packages/components/*",
+      "packages/manager-cli",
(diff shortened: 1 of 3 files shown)


fredericvilcot:
I close the PR because it doesn't meet yet the conditions of putting back the module into the monolithe workspaces.

PR #18434, opened 2025-07-30
feat(dedicated): remove backupftp 2api call

sachinrameshn:
ref: #MANAGER-17540

## Description

Ticket Reference: #...    

## Additional Information

Replace /sws/dedicated/server/:serviceName/backupFtp 2api call by 2 calls that can be done in parallel:

1. /dedicated/server/{serviceName}/features/backupFTP
2. /order/dedicated/server/{serviceName}


packages/manager/modules/bm-server-components/src/server/server.service.js
```diff
@@ -395,15 +395,47 @@ export default class Server {
...
+    return this.$q
+      .all({
+        backupFtp: this.getFtpBackupFeatures(serviceName),
+        order: this.getOrder(serviceName),
+      })
+      .then(({ backupFtp, order }) => {
+        return {
+          activated: !!backupFtp,
+          login: serviceName,
+          canOrder: order?.includes('backupStorage'),
```
aderghamov: This value will be undefined if order is undefined.
Is it OK?

sachinrameshn: Thanks will update it to cosider false.

qpavy (approved):

qpavy (approved):

PR #18440, opened 2025-07-30
feat(shell-client): add type gard to actionType for TrackingClickParams in useOvhTracking

aTurmo:
I'm adding a type gard in useOvhTracking for the 'actionType' field of TrackingClickParams.
The use case is I need to know if a string has a value included in actionType (which is a union type of strings), to then cast it and used it for useOvhTracking.trackCkick.


aboungnaseng-ovhcloud (dismissed):

packages/manager/core/shell-client/src/hooks/useOvhTracking.ts
```diff
@@ -73,11 +73,20 @@ export const getPageProps = ({
   level2,
 });
 
+const actionTypes = ['action', 'navigation', 'download', 'exit'] as const;
+type ActionType = typeof actionTypes[number];
```
fredericvilcot: Please export type also ;)

SimonChaumet (dismissed):

fredericvilcot (dismissed):

aboungnaseng-ovhcloud (dismissed):

fredericvilcot (approved):

acavalloni (approved):

PR #18445, opened 2025-07-30
fix(ips): refreshing the ip-listing display when deleting a vmac

aderghamov:
ref: #MANAGER-19266

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
+++ b/packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
@@ -29,7 +29,6 @@ export const useGetIpVmacWithIp = ({
   serviceName,
   enabled = true,
 }: UseGetIpVmacWithIpParams) => {
-  const [isInvalidated, setIsInvalidated] = useState(false);
   const [invalidatedData, setInvalidatedData] = useState(undefined);
 
   const { data: dedicatedServerVmacResponse, isLoading, isError } = useQuery<
@@ -45,7 +44,6 @@ export const useGetIpVmacWithIp = ({
       const macAdresses = queryData.map(({ macAddress }) => macAddress);
 
       if (query.state.isInvalidated) {
-        setIsInvalidated(true);
         setInvalidatedData(macAdresses);
         return INVALIDATED_REFRESH_PERIOD;
       }
@@ -54,11 +52,10 @@ export const useGetIpVmacWithIp = ({
         !!invalidatedData &&
         !hasSameStringElements(invalidatedData, macAdresses)
       ) {
-        setIsInvalidated(false);
         setInvalidatedData(undefined);
       }
 
-      return isInvalidated ? INVALIDATED_REFRESH_PERIOD : undefined;
+      return invalidatedData ? INVALIDATED_REFRESH_PERIOD : undefined;
     },
   });
 
@@ -82,7 +79,7 @@ export const useGetIpVmacWithIp = ({
   const formattedResult = {
     isLoading:
       isLoading ||
-      !!results.find((result) => !!result.isLoading || isInvalidated),
+      !!results.find((result) => !!result.isLoading || !!invalidatedData),
     isError: isError || !!results.find((result) => !!result.isError),
     vmacsWithIp: results?.map(({ data }, index) => ({
       ...dedicatedServerVmacResponse?.data?.[index],


packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
```diff
@@ -21,11 +22,32 @@ export type VmacWithIpType = DedicatedServerVmacType & {
   ip: string[];
 };
 
+function arraysEqual(arr1: string[], arr2: string[]): boolean {
```
helabenkhalfallah: the name of the function and variables are not good.

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
```diff
@@ -21,11 +22,32 @@ export type VmacWithIpType = DedicatedServerVmacType & {
...
+  if (arr1.length !== arr2.length) {
```
helabenkhalfallah: the code lack same safety, what happens if arrays are null, undefined at runtime?

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
```diff
@@ -21,11 +22,32 @@ export type VmacWithIpType = DedicatedServerVmacType & {
...
+    return false;
+  }
+
+  const sortedArr1 = arr1.slice().sort();
+  const sortedArr2 = arr2.slice().sort();
```
helabenkhalfallah: to create a copy of the array you should use spread instead of slice.

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
```diff
@@ -21,11 +22,32 @@ export type VmacWithIpType = DedicatedServerVmacType & {
...
+
+  return sortedArr1.reduce((acc, item, index) => {
```
helabenkhalfallah: slice, sort and reduce seems to be a cost operations, we cannot optimise?

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
```diff
@@ -21,11 +22,32 @@ export type VmacWithIpType = DedicatedServerVmacType & {
   ip: string[];
 };
 
+function hasSameElements(
```
helabenkhalfallah: perhaps we can extract this is utils

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
```diff
@@ -21,11 +22,32 @@ export type VmacWithIpType = DedicatedServerVmacType & {
...
+
+  const arrayAsSet = new Set(anArray);
+  if (arrayAsSet.size !== otherArray.length) {
+    return false;
+  }
+
+  return otherArray.reduce((acc, item) => acc && arrayAsSet.has(item), true);
+}
+
+const INVALIDATED_REFRESH_PERIOD = 5000;
```
helabenkhalfallah: same as https://github.com/ovh/manager/pull/18445/files#r2293059808

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
aderghamov: Where should I extract it exactly?

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
```diff
@@ -21,11 +22,32 @@ export type VmacWithIpType = DedicatedServerVmacType & {
...
+  if (!anArray || !otherArray || anArray.length !== otherArray.length) {
+    return false;
+  }
+
+  const arrayAsSet = new Set(anArray);
+  if (arrayAsSet.size !== otherArray.length) {
+    return false;
+  }
+
+  return otherArray.reduce((acc, item) => acc && arrayAsSet.has(item), true);
```
chipp972:
```suggestion
  return otherArray.every(item => arrayAsSet.has(item));
```

Like that it will stop at the first false instead of going through the whole array when it encounters a false

aderghamov: Good Idea!

sachinrameshn:
Hello, Do we need map here for bulk delete ? Is it possible to delete multiple vmacs ? 

https://github.com/ovh/manager/blob/fix/MANAGER-19266/packages/manager/apps/ips/src/data/api/delete/virtualMac.ts#L14

aderghamov:
> Hello, Do we need map here for bulk delete ? Is it possible to delete multiple vmacs ?
> 
> https://github.com/ovh/manager/blob/fix/MANAGER-19266/packages/manager/apps/ips/src/data/api/delete/virtualMac.ts#L14


According to the schema, there can be multiple Vmac adresses for a single IP so deleting should remove all of them.
According to the API, I have not found any away to trigger a bulk update with multiple Vmac Adresses.
The `useMutation` hook should requires a mutation function single promise so I chose to synchronize all the calls with a `Promise.all` (which requires a map)

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
aderghamov: Ok, I have just found out

sachinrameshn:
> > Hello, Do we need map here for bulk delete ? Is it possible to delete multiple vmacs ?
> > https://github.com/ovh/manager/blob/fix/MANAGER-19266/packages/manager/apps/ips/src/data/api/delete/virtualMac.ts#L14
> 
> According to the schema, there can be multiple Vmac adresses for a single IP so deleting should remove all of them. According to the API, I have not found any away to trigger a bulk update with multiple Vmac Adresses. The `useMutation` hook should requires a mutation function single promise so I chose to synchronize all the calls with a `Promise.all` (which requires a map)


I am not sure if we can add more than one vmac for a ip. Based on the document I see below requirement to disable "Add virtual Mac" option. So if we dont show add option if we already have vmac exist for a ip, its sure it wont have multiple vmac attchaed to a ip. 



Confluence rule for Add Vmac button - This button is enable if the vMAC status is not ok or if the server linked to the ip doesn't have vMAC for the current ip. If it is disable because of the previous rules, we display a tooltip saying a vMAC already exist.

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
```diff
@@ -26,6 +29,9 @@ export const useGetIpVmacWithIp = ({
   serviceName,
   enabled = true,
 }: UseGetIpVmacWithIpParams) => {
+  const [isInvalidated, setIsInvalidated] = useState(false);
+  const [invalidatedData, setInvalidatedData] = useState(undefined);
```
qpavy: Can we remove isInvalidated ? I think only use !!invalidatedData is doing same

qpavy (approved):

PR #18456, opened 2025-07-31
fix(pci-block-storage): fix sorting of volumes listing

aTurmo:
ref: #TAPC-4409

### DIFF
--- a/packages/manager/apps/pci-block-storage/src/api/select/volume.ts
+++ b/packages/manager/apps/pci-block-storage/src/api/select/volume.ts
@@ -17,30 +17,46 @@ import {
   getVolumeModelPricings,
   TModelPrice,
 } from '@/api/select/catalog';
+import {
+  BlockStorageListColumn,
+  isBlockStorageListColumn,
+} from '@/hooks/useDatagridColumn';
+
+const volumeComparator = <T>(extractor: (_: TVolume) => T) => (
+  t1: TVolume,
+  t2: TVolume,
+) => (extractor(t1) > extractor(t2) ? 1 : -1);
 
 export const sortResults = (
   items: TVolume[],
   sorting: ColumnSort | undefined,
 ) => {
-  if (!sorting) return items;
+  if (!sorting || items.length === 0) return items;
+  const sortingColumn = sorting.id;
+  if (!isBlockStorageListColumn(sortingColumn)) return items;
 
-  let sortFn: (a: TVolume, b: TVolume) => number;
-
-  switch (sorting.id) {
-    case 'status':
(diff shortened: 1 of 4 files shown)


SimonChaumet (dismissed):

lolottetheclash (approved):

packages/manager/apps/pci-block-storage/src/api/select/volume.ts
```diff
@@ -17,38 +17,56 @@ import {
...
+      case BlockStorageListColumn.ATTACHED:
+        return volumeComparator(
+          (t: TVolume) => t.attachedTo[0].toLocaleLowerCase() || '',
+        );
+      default: {
+        const typeOfColumn = typeof items[0][sortingColumn];
+        if (typeOfColumn === 'string') {
+          return volumeComparator((t: TVolume) =>
+            (t[sortingColumn] as string).toLocaleLowerCase(),
+          );
```
SimonChaumet: What happens when it's not a string or if the first item doesn't have the value filled ?

packages/manager/apps/pci-block-storage/src/api/select/volume.ts
```diff
@@ -17,38 +17,56 @@ import {
...
+        return volumeComparator((t: TVolume) => t[sortingColumn]);
+      }
+    }
+  })();
 
   return [...items].sort(sorting.desc ? (a, b) => -sortFn(a, b) : sortFn);
 };
 
-export const paginateResults = <T>(
+export const paginateResults: <T>(
```
SimonChaumet: I don't understand why this new way is necessary, couldn't you add ": myType" after the parenthesis ?

packages/manager/apps/pci-block-storage/src/api/select/volume.ts
aTurmo: If it's not a string it's using the comparator line 54

And yes there is a problème if the first item has no value. Working on it

packages/manager/apps/pci-block-storage/src/api/select/volume.ts
aTurmo: yes you're right

packages/manager/apps/pci-block-storage/src/api/select/volume.ts
aTurmo: fixed

packages/manager/apps/pci-block-storage/src/api/select/volume.ts
aTurmo: fixed

lolottetheclash (approved):

PR #18203, opened 2025-07-17
Add Project Dashboard/Home Page

StephaneRavet:
Add Project Dashboard/Home Page
#MANAGER-18604

For clarity, the tests are in another PR: [https://github.com/ovh/manager/pull/18451](https://github.com/ovh/manager/pull/18451)

PR #18500, opened 2025-08-04
refactor(pci-project): refactor useProject 

StephaneRavet:
Refactor following @lionel95200x 's PR comment : https://github.com/ovh/manager/pull/18203#discussion_r2244921752


packages/manager/apps/pci-project/src/routes/loader/validateProjectIdLoader.ts
```diff
@@ -0,0 +1,12 @@
...
+export const projectIdLoader = async (args: LoaderFunctionArgs) => {
+  const { projectId } = args.params;
+
+  if (!projectId) {
+    throw new Error('No project id found');
+  }
+
+  // Return the validated projectId to be used in route data
+  return { projectId };
+};
```
seven-amid: You have the same logic in the file above `packages/manager/apps/pci-project/src/routes/loader/projectIdLoader.ts` !!

packages/manager/apps/pci-project/src/hooks/useProjectIdFromRoute.ts
```diff
@@ -0,0 +1,33 @@
...
+import { ROUTE_IDS, ProjectRouteData } from '@/routes/routes.constant';
+
+export const useProjectIdFromRoute = (): string => {
+  // Try to get projectId from route data first
+  let routeProjectId: string | undefined;
+
+  try {
+    routeProjectId = (useRouteLoaderData(ROUTE_IDS.PROJECT) as ProjectRouteData)
+      ?.projectId;
+  } catch {
```
seven-amid: Which error are you encountering here? We shouldn't be getting an undefined ID, since we already perform a check in the loader.

packages/manager/apps/pci-project/src/routes/loader/projectIdLoader.ts
```diff
@@ -0,0 +1,12 @@
...
+export const projectIdLoader = async (args: LoaderFunctionArgs) => {
+  const { projectId } = args.params;
+
+  if (!projectId) {
+    throw new Error('No project id found');
+  }
+
+  // Return the validated projectId to be used in route data
+  return { projectId };
+};
```
seven-amid: Think this check isn't necessary! because the projectId is part of our routes. If it's not present, the route won't be rendered at all. Maybe you should check here whether the appropriate project is defined, rather than checking the ID !

packages/manager/apps/pci-project/src/routes/loader/validateProjectIdLoader.ts
StephaneRavet: Exact 👍

packages/manager/apps/pci-project/src/hooks/useProjectIdFromRoute.ts
StephaneRavet: The try/catch block in `useProjectIdFromRoute.ts` does not handle a projectId validation error, but rather a technical error from React Router. Here are the cases where `useRouteLoaderData` can fail:

* **Tests with MemoryRouter**: As mentioned in the comment
* **Route not found**: If the route is not properly configured
* **Loader not executed**: If the loader has not been called yet
* **Type error**: If the returned data does not match the expected type

### Why the loader does not guarantee a defined projectId

Even if the loader validates the projectId, there are several scenarios where `useRouteLoaderData` may return `undefined`:

* **Direct navigation**: The user directly accesses a URL without going through the loader
* **Routing error**: Route configuration issues
* **Tests**: Testing environments without a complete router

packages/manager/apps/pci-project/src/routes/loader/projectIdLoader.ts
StephaneRavet: Exact 👍

packages/manager/apps/pci-project/src/hooks/useProjectIdFromRoute.ts
MarcMonchablon: I'm not sure there's really a need for the whole 'loader' part. It's usually clear-cut whether there is a 'projectId' available in a route or not (and for other micro-app runtime conditions make it so we know it's always defined).

Usually we have the following process:
 - We use react-router 'useParams' to get a projectId from route
 - We cast it as a string, since it's typescripted to be potentially undefined (even though we have strong runtime garantees)
 - Since it's usually bad form to cast something in Typescript, during PR we are told to extract it in a way that thrown in case of 'undefined', which is fair I think as an approach.
 - We end up with a well-typed projectId, behind strong run-time garantees instead of casting it 'as string'.

Given all that, we usually end-up with a small hook that wraps around the 'useParams' but throw if the expected value is undefined. I'm not sure about the need for a loader function applied to the route, but maybe I'm missing something ?

StephaneRavet: I agree. I've simplified the code, without loader.

seven-amid (approved):

PR #18516, opened 2025-08-05
feat(container): handle nin type in suggestion modal logic

oalkabouss:
ref: #MANAGER-14721

## Description

Ticket Reference: #MANAGER-14721

### DIFF
--- a/packages/manager/apps/container/src/components/SuggestionModal/SuggestionModal.spec.tsx
+++ b/packages/manager/apps/container/src/components/SuggestionModal/SuggestionModal.spec.tsx
@@ -12,7 +12,8 @@ const mocks = vi.hoisted(() => ({
   user: {
     legalform: 'corporation',
     country: 'FR',
-    companyNationalIdentificationNumber: 'suggestedNIN',
+    companyNationalIdentificationNumber: 'suggestedSIRET',
+    nationalIdentificationNumber: 'suggestedNIN',
     vat: 'suggestedVAT',
   },
   data: [] as Suggestion[],
@@ -161,11 +162,18 @@ describe('SuggestionModal', () => {
     mocks.data = [
       {
         type: 'SIRET',
-        id: 'suggestedNIN',
+        id: 'suggestedSIRET',
+        isActive: true,
       },
       {
         type: 'VAT',
         id: 'suggestedVAT',
+        isActive: true,
+      },
+      {
+        type: 'NIN',
+        id: 'suggestedNIN',
+        isActive: true,
       },
     ];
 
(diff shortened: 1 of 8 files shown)


packages/manager/apps/container/src/components/SuggestionModal/SuggestionModal.spec.tsx
JacquesLarique, line 1: Can you add some tests with inactive suggestions (and one where NIN is active)

JacquesLarique (changes requested):
`useSuggestionsCheck` in `packages/manager/apps/container/src/hooks/suggestion/useSuggestion.tsx`, should check that the NIN suggestion is not `active` (meaning the SIRET filled by the user is already valid)

packages/manager/apps/container/src/components/SuggestionModal/SuggestionModal.spec.tsx
```diff
@@ -180,11 +188,18 @@ describe('SuggestionModal', () => {
...
       },
       {
         type: 'VAT',
         id: 'differentVAT',
+        isActive: true,
+      },
+      {
+        type: 'NIN',
+        id: 'differentNIN',
+        isActive: true,
```
JacquesLarique: Having the NIN suggestion active should not allow the display of the modal

packages/manager/apps/container/src/helpers/suggestions/suggestionsHelper.spec.ts
```diff
@@ -42,29 +42,29 @@ describe('suggestionsHelpers', () => {
...
       const user = {};
       // TODO: remove unnecessary `as unknown` when User type is fixed
       expect(isSuggestionRelevant(suggestion, user as unknown as User)).toEqual(true);
     });
 
     it('should consider suggestion relevant if its value is not the same as the user\'s', () => {
-      const suggestion: Suggestion = { type: 'SIRET', id: '11111' };
-      const user = { companyNationalIdentificationNumber: '11112' };
+      const suggestion: Suggestion = { type: 'SIRET', id: '11111', isActive: true };
+      const user = { companyNationalIdentificationNumber: '11112', isActive: true };
```
JacquesLarique: There is no `isActive` property on the user, you can remove this line

packages/manager/apps/container/src/hooks/suggestion/useSuggestion.tsx
```diff
@@ -29,7 +29,14 @@ export const useSuggestions = (enabled: boolean) => useQuery({
...
-    suggestions.some((suggestion) => isSuggestionRelevant(suggestion, user)),
+  (suggestions: Suggestion[]) =>
+    suggestions.some((suggestion) => {
+      return suggestions.some(suggestion => {
+        if (suggestion.type === 'NIN' && !suggestion.isActive) {
+          return false;
+        }
+        return isSuggestionRelevant(suggestion, user);
+      });
+    }),
```
JacquesLarique: This does not work in the expected way, you should do
```suggestion
  (suggestions: Suggestion[]) => {
    const ninSuggestion = suggestions.find((suggestion) => suggestion.type === 'NIN');
    if (ninSuggestion?.isActive) {
      return false;
    }
    return suggestions.some((suggestion) => isSuggestionRelevant(suggestion, user)),
  },
```

PR #18567, opened 2025-08-07
fix(dedicated): monitoring status in listing page

carsso:
ref: #MANAGER-19387

## Description

In the new dedicated servers listing page, if you enable the monitoring column, the value is inverted between proactive monitoring and non-proactive monitoring (colors are also inverted).

## Additional Information

Ticket Reference: MANAGER-19387

### DIFF
--- a/packages/manager/apps/dedicated-servers/src/components/monitoringStatus.tsx
+++ b/packages/manager/apps/dedicated-servers/src/components/monitoringStatus.tsx
@@ -4,18 +4,25 @@ import { OdsBadge } from '@ovhcloud/ods-components/react';
 import { useTranslation } from 'react-i18next';
 import { DedicatedServer } from '@/data/types/server.type';
 
+const getMonitoringStatusLabel = (
+  isMonitoring: boolean,
+  hasNoIntervention: boolean,
+) => {
+  if (!isMonitoring) return 'disabled';
+  if (hasNoIntervention) return 'no-proactive';
+  return 'proactive';
+};
+const getMonitoringStatusColor = (
+  isMonitoring: boolean,
+  hasNoIntervention: boolean,
+) => {
+  if (!isMonitoring) return ODS_BADGE_COLOR.warning;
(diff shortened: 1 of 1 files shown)


qpavy (dismissed):

packages/manager/apps/dedicated-servers/src/components/monitoringStatus.tsx
```diff
@@ -6,15 +6,20 @@ import { DedicatedServer } from '@/data/types/server.type';
...
-    return _noIntervention ? 'proactive' : 'no-proactive';
+    return _noIntervention ? 'no-proactive' : 'proactive';
+  };
+  const getMonitoringStatusColor = (
+    _monitoring: boolean,
+    _noIntervention: boolean,
+  ) => {
+    if (!_monitoring) return ODS_BADGE_COLOR.warning;
+    return _noIntervention ? ODS_BADGE_COLOR.warning : ODS_BADGE_COLOR.success;
   };
```
tibs245: I suggest to export function outside the component. This avoid to redeclare during each re-rending. And it's easier to test without use a component render.

And rename props of function to respect nomenclature of boolean
And use "if () return" for each case allow better readibility

```suggestion
const getMonitoringStatusLabel = (
isMonitoring: boolean,
hasNoIntervention: boolean,
) => {
    if (!isMonitoring) return 'disabled';
    if (hasNoIntervention) return 'no-proactive';
    return 'proactive';
  };
  const getMonitoringStatusColor = (
    isMonitoring: boolean,
    hasNoIntervention: boolean,
  ) => {
    if (!isMonitoring) return ODS_BADGE_COLOR.warning;
    if (hasNoIntervention) return ODS_BADGE_COLOR.warning;
    return ODS_BADGE_COLOR.success;
};

export const MonitoringStatusChip = (server: DedicatedServer) => {
const { t } = useTranslation('dedicated-servers');


```

carsso: Done :)

qpavy (approved):

PR #18569, opened 2025-08-07
refactor(pci-instances): rework readibility for getInstanceStatusSeverity function

Tsiorifamonjena:
## Description

- This PR makes `getInstanceStatusSeverity` more readable and expressive as suggested [here](https://github.com/ovh/manager/pull/18093#discussion_r2257597824). And BTW unit test was added
- It fix also the way updating cache after a polling on the Dashboard. Only some relevant properties are updated

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/useDashboardPolling.ts
+++ b/packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/useDashboardPolling.ts
@@ -31,13 +31,13 @@ export const useDashboardPolling = ({
   const handlePollingSuccess = useCallback(
     (instance?: TInstance) => {
       if (!instance) return;
-      const { task } = instance;
+      const { task, actions, addresses, volumes, status } = instance;
 
       updateDashboardCache({
         projectId,
         region,
         instanceId,
-        payload: { ...instance },
+        payload: { actions, addresses, volumes, status, task },
       });
 
       if (!task.isPending) {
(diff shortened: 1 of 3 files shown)


packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/useDashboardPolling.ts
```diff
@@ -31,13 +31,13 @@ export const useDashboardPolling = ({
...
       if (!instance) return;
-      const { task } = instance;
+      const { task, actions, addresses, volumes, status } = instance;
 
       updateDashboardCache({
         projectId,
         region,
         instanceId,
-        payload: { ...instance },
+        payload: { actions, addresses, volumes, status, task },
```
SimonChaumet, line 40: why not put the instance directly ?

packages/manager/apps/pci-instances/src/pages/instances/mapper/mapper.spec.tsx
```diff
@@ -22,4 +24,55 @@ describe('Mapper functions', () => {
       ]);
     });
   });
+
+  describe('Considering the getInstanceStatus function', () => {
+    it.each([
```
ppprevost: ``
Object.entries(severityToStatus).flatMap(([key, values]) => values.map(value => [value, key]))
``

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/useDashboardPolling.ts
Tsiorifamonjena: the instance from the polling here does not have all the properties because the call does not have all the filters that the dashboard uses. The polling is common
Then we update only needed property from cache

ppprevost (approved):

PR #18572, opened 2025-08-07
Build/pci instances dashboard ods19

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-4685


packages/manager/apps/pci-instances/src/data/hooks/instance/useInstance.spec.tsx
```diff
@@ -57,6 +57,7 @@ describe('useUpdateInstanceName', () => {
       { wrapper },
     );
 
+    // eslint-disable-next-line @typescript-eslint/require-await
     await act(async () =>
```
SimonChaumet: Here the issue is that you are using "async" while not awaiting anything inside the call, either you need to await the mutate call or delete both "await" and "async" keywords

Tsiorifamonjena: I can't delete both await and async. I have to await the mutate
There expression are not allowed

```js
await act(
      async () =>
        await result.current.mutate({ instanceName: 'new_instanceName' }),
    );
```

=> Redundant use of await

```js
await act(() =>
      result.current.mutate({ instanceName: 'new_instanceName' }),
    );
```
=> 'await' has no effect on the type of this expression 
=> `await` of a non-Promise (non-"Thenable") value

SimonChaumet: You have to put "void" if you don't want to await the mutate

PR #18575, opened 2025-08-07
feat(pci-instances): add instance name input with form

lolottetheclash:
ref: #TAPC-4541

## Description

Ticket Reference: #...


packages/manager/apps/pci-instances/src/components/cart/genericCart/GenericCart.tsx
```diff
@@ -0,0 +1,70 @@
...
+    (accordion: AccordionChangeDetail) =>
+      setExpandedAccordions(accordion.value),
+    [setExpandedAccordions],
+  );
+  const isExpanded = useCallback(
+    (accordion: string) => expandedAccordions.includes(accordion),
+    [expandedAccordions],
+  );
+
+  useEffect(() => {
```
lionel95200x:
```suggestion
 You don't need effect here
 
 const expandedAccordions = products
      .filter((product) => product.expanded)
      .map((product) => product.title);
```

packages/manager/apps/pci-instances/src/components/cart/Cart.tsx
```diff
@@ -1,7 +1,24 @@
+import { useFormContext, useWatch } from 'react-hook-form';
+import { useTranslation } from 'react-i18next';
+import { useMemo } from 'react';
+import { GenericCart, TCartProduct } from './genericCart/GenericCart';
+import { TInstanceCreationForm } from '@/pages/instances/create/CreateInstance.page';
```
Tsiorifamonjena, line 5: Maybe it's time to move the `TInstanceCreationForm` to a common file because it's weird to have component import from page file

lolottetheclash: Yes we have to refactor but it will be into another PR as Manon is using this one if it's ok for you :)

Tsiorifamonjena (approved):

PR #18576, opened 2025-08-07
Feat/add instance quantity tapc 4680

manoncarbonnel:
## Description

Static form field for instance quantity.

Ticket Reference: #4680

### DIFF
--- a/packages/manager/apps/pci-instances/src/components/cart/Cart.tsx
+++ b/packages/manager/apps/pci-instances/src/components/cart/Cart.tsx
@@ -3,21 +3,31 @@ import { useTranslation } from 'react-i18next';
 import { useMemo } from 'react';
 import { GenericCart, TCartProduct } from './genericCart/GenericCart';
 import { TInstanceCreationForm } from '@/pages/instances/create/CreateInstance.page';
+import { TInstance } from '@/types/instance/entity.type';
+import { quantityDefaultValue } from '@/pages/instances/create/components/QuantitySelector';
 
 export const Cart = () => {
   const { control } = useFormContext<TInstanceCreationForm>();
-  const instanceName = useWatch({ control, name: 'name' });
+  const instance: Partial<TInstance> = {
+    name: useWatch({ control, name: 'name' }),
+    quantity: useWatch({
+      control,
+      name: 'quantity',
+      defaultValue: quantityDefaultValue,
+    }),
+  };
   const { t } = useTranslation();
 
   const products: TCartProduct[] = useMemo(
     () => [
       {
         title: t('pci_instances_common_instance_title'),
-        designation: instanceName,
+        designation: instance.name,
+        quantity: instance.quantity,
         expanded: true,
       },
     ],
-    [instanceName, t],
+    [instance.name, instance.quantity, t],
   );
 
   return <GenericCart products={products} />;
--- a/packages/manager/apps/pci-instances/src/pages/instances/create/CreateInstance.page.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/create/CreateInstance.page.tsx
@@ -2,19 +2,35 @@ import { FC } from 'react';
 import { useRouteLoaderData } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 import { TProject } from '@ovh-ux/manager-pci-common';
-import { FormField, FormFieldLabel, Input, Text } from '@ovhcloud/ods-react';
+import { Text } from '@ovhcloud/ods-react';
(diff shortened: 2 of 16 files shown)


packages/manager/apps/pci-instances/src/pages/instances/create/CreateInstance.page.tsx
```diff
@@ -14,6 +22,10 @@ import { instanceNameRegex } from '@/constants';
 export type TInstanceCreationForm = z.infer<typeof instanceCreationSchema>;
 export const instanceCreationSchema = z.object({
   name: z.string().regex(instanceNameRegex),
+  quantity: z
+    .number()
+    .min(1)
+    .max(793),
```
Tsiorifamonjena: What does this number 789 represent ? Is this limit random or defined by the business ? :)

Tsiorifamonjena (approved):
Perfect 👌🏽

packages/manager/apps/pci-instances/src/components/cart/Cart.tsx
```diff
@@ -7,17 +7,21 @@ import { TInstanceCreationForm } from '@/pages/instances/create/CreateInstance.p
 export const Cart = () => {
   const { control } = useFormContext<TInstanceCreationForm>();
   const instanceName = useWatch({ control, name: 'name' });
+  const instanceQuantity = String(
+    useWatch({ control, name: 'quantity', defaultValue: 1 }),
```
lolottetheclash: I think we can get all properties like `const { name: instanceName, quantity: instanceQuantity }`with an name array param, no?

packages/manager/apps/pci-instances/src/pages/instances/create/CreateInstance.page.tsx
```diff
@@ -85,6 +100,45 @@ const CreateInstance: FC = () => {
...
+              </Text>
+              <div className="pt-3 pb-4">
+                <FormField>
+                  <FormFieldLabel>
+                    {t('creation:pci_instances_creation_quantity_label')}
+                  </FormFieldLabel>
+                  <Quantity
+                    {...register('quantity', { required: true })}
+                    onValueChange={({ value }) => {
+                      setValue('quantity', Number(value));
```
lolottetheclash: For `Quantity`, I'm not sure that it's the way to handle it as it's a custom component.
I think we have to deal with RHF `Controller` (https://react-hook-form.com/docs/usecontroller/controller), maybe you could check with our RHF experts ;) (@SimonChaumet and @ppprevost  if I'm correct?)

packages/manager/apps/pci-instances/src/pages/instances/create/CreateInstance.page.tsx
manoncarbonnel: Fair question: this is totally random (and static) until I have a proper business rule to implement :) In the near future, this should come from a quota calculated based on the selected options (not yet implemented)

packages/manager/apps/pci-instances/src/pages/instances/create/CreateInstance.page.tsx
SimonChaumet: Yes, either you use the register only, or you use the Controller component to wrap it. setValue should only be used when controlling values with side effects

Also I think it is a good idea to create pre-wrapped react-hook-form generic fields in a separate folder (like I did here : https://github.com/ovh/manager/pull/18023/files#diff-51c94ff4882234eee80e8d1f435360f8da2c925cca72abddf5c2df41f7920df1)

packages/manager/apps/pci-instances/src/components/cart/Cart.tsx
```diff
@@ -3,21 +3,26 @@ import { useTranslation } from 'react-i18next';
 import { useMemo } from 'react';
 import { GenericCart, TCartProduct } from './genericCart/GenericCart';
 import { TInstanceCreationForm } from '@/pages/instances/create/CreateInstance.page';
+import { TInstance } from '@/types/instance/entity.type';
 
 export const Cart = () => {
   const { control } = useFormContext<TInstanceCreationForm>();
-  const instanceName = useWatch({ control, name: 'name' });
+  const instance: Partial<TInstance> = {
```
Tsiorifamonjena, line 11: I think you break react hook rules here by doing this. Hook must be called at the top level

```suggestion
  const name = useWatch({ control, name: 'name' });
const quantity = useWatch({ control, name: 'quantity', defaultValue: 1 });

const instance: Partial<TInstance> = {
  name,
  quantity,
};
```

PR #18620, opened 2025-08-12
feat(pci-kubernetes): enable ldp

ppprevost:
The main purpose of this branch is to Enable those features for 3-az regions : 
OIDC
LDP
Upgrade the Cluster 

ref: #TAPC-4367, #TAPC-4366, #TAPC-4369

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-kubernetes/src/components/service/ClusterAccessAndSecurity.component.spec.tsx
+++ b/packages/manager/apps/pci-kubernetes/src/components/service/ClusterAccessAndSecurity.component.spec.tsx
@@ -33,7 +33,7 @@ describe('ClusterAccessAndSecurity', () => {
     vi.spyOn(useKubernetesModule, 'useClusterRestrictions').mockReturnValue({
       isPending: false,
       data: [],
-    } as UseQueryResult<string[]>);
+    } as UseQueryResult<[]>);
 
     const { getByText } = render(
       <ClusterAccessAndSecurity kubeDetail={{ status: 'READY' } as TKube} />,
--- a/packages/manager/apps/pci-kubernetes/src/components/service/ClusterManagement.component.tsx
+++ b/packages/manager/apps/pci-kubernetes/src/components/service/ClusterManagement.component.tsx
@@ -100,13 +100,14 @@ export default function ClusterManagement({
           isDisabled={isProcessing(kubeDetail?.status)}
           href={hrefCreateNodePool}
         />
-        {!isMultiDeploymentZones(regionInformations?.type) && (
-          <TileButton
-            title={t('kube_service_reset_kubeconfig')}
-            isDisabled={isProcessing(kubeDetail?.status)}
-            href={hrefResetClusterConfig}
-          />
-        )}
+        {regionInformations?.type &&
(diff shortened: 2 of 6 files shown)


packages/manager/apps/pci-kubernetes/src/api/hooks/useKubernetes.ts
```diff
@@ -299,7 +300,7 @@ export const useOidcProvider = (projectId: string, kubeId: string) =>
 type KubeConfigProps = {
   projectId: string;
   kubeId: string;
-  onError: (cause: Error) => void;
+  onError: (cause: ResponseAPIError) => void;
```
SimonChaumet: I think it's better to have a generic error and then use isApiError function to check if it's really an api error

packages/manager/apps/pci-kubernetes/src/components/service/ClusterManagement.component.tsx
```diff
@@ -100,13 +100,14 @@ export default function ClusterManagement({
...
         />
-        {!isMultiDeploymentZones(regionInformations?.type) && (
-          <TileButton
-            title={t('kube_service_reset_kubeconfig')}
-            isDisabled={isProcessing(kubeDetail?.status)}
-            href={hrefResetClusterConfig}
-          />
-        )}
+        {regionInformations?.type &&
+          !isMultiDeploymentZones(regionInformations?.type) && (
```
SimonChaumet:
```suggestion
          !isMultiDeploymentZones(regionInformations.type) && (
```

SimonChaumet (approved):

selm3n (approved):

packages/manager/apps/pci-kubernetes/src/components/service/ClusterAccessAndSecurity.component.spec.tsx
```diff
@@ -33,7 +33,7 @@ describe('ClusterAccessAndSecurity', () => {
     vi.spyOn(useKubernetesModule, 'useClusterRestrictions').mockReturnValue({
       isPending: false,
       data: [],
-    } as UseQueryResult<string[]>);
+    } as UseQueryResult<[] | string[]>);
```
Tsiorifamonjena: `string[]` is not enough ?

packages/manager/apps/pci-kubernetes/src/pages/detail/Detail.page.tsx
```diff
@@ -78,13 +74,13 @@ export default function DetailPage() {
       ),
       to: useResolvedPath('logs').pathname,
```
Tsiorifamonjena: The usage of `useResolvedPath` in the array violates React’s hook rules. It must be called at the top level of the function component.

ppprevost: It is not a strict violation because the object is rendered each time into the component. 
The object is recreated each time and the hooks too. 
So the hooks will always be called in the same order and React does not loose the hook cycle
However, i understand that the libility of react hook order can be confusing

SimonChaumet (dismissed):

Tsiorifamonjena (dismissed):

PR #18622, opened 2025-08-12
feat(manager-react-components): migrate link-card component to ODS v19

anooparveti:
## Description

- Migrate LinkCard component to ODS v19
- Movement of component from navigations to components folder for flat organisation Align with component development guidelines
- Add stories for uncovered use-cases
- Add more unit and snapshot tests

Ticket Reference: #MANAGER-18868    

## Additional Information

BREAKING CHANGE: 

1. Rename Card component to LinkCard component 
2. Rename isExternalHref prop to externalHref


packages/manager-react-components/src/components/link-card/LinkCard.component.tsx
```diff
@@ -0,0 +1,83 @@
...
+      target="_blank"
+      href={href}
+      className="no-underline"
+      onClick={onClick}
+      {...props}
+    >
+      <Card
+        className="w-full h-full p-[1rem]"
+        color={CARD_COLOR.neutral}
+        data-tracking={trackingLabel}
```
aboungnaseng-ovhcloud, line 35: should we keep data-tracking ? it's only for catalog apps ?

anooparveti: I think we can still use this method in future with more generalised tracking no?
So i prefer keeping it.

aboungnaseng-ovhcloud: Ok lets merge like this 👍

PR #18640, opened 2025-08-12
feat(pci-instances): move to regionalized call to create instance

SimonChaumet:
ref: #TAPC-4688

## Description

Ticket Reference: #...


SimonChaumet:
Waiting for backend update

PR #18645, opened 2025-08-13
feat(web-domains): modification page - nameservers update

kb-ovh:
ref: #MANAGER-18979

### DIFF
--- a/packages/manager/apps/web-domains/public/translations/domain/Messages_fr_FR.json
+++ b/packages/manager/apps/web-domains/public/translations/domain/Messages_fr_FR.json
@@ -93,5 +93,18 @@
   "domain_tab_general_information_pending_deletion": "Suppression en cours",
   "domain_tab_general_information_deleted": "Supprimé",
   "domain_tab_general_not_suspended": "Actif",
-  "domain_tab_general_suspended": "Suspendu"
+  "domain_tab_general_suspended": "Suspendu",
+  "domain_tab_DNS_modification_form_dns_number": "Veuillez renseigner entre {{min}} et {{max}} DNS.",
+  "domain_tab_DNS_modification_form_server_field": "Serveur DNS",
+  "domain_tab_DNS_modification_form_ip_field": "IP associée",
+  "domain_tab_DNS_modification_form_placeholder": "ns1.my-domain-is-gre.at",
+  "domain_tab_DNS_modification_form_placeholder_ip": "198.27.92.1",
+  "domain_tab_DNS_modification_form_server_field_mandatory": "obligatoire",
+  "domain_tab_DNS_modification_form_server_wrong_format": "Veuillez saisir un nom d'hôte valide.",
+  "domain_tab_DNS_modification_form_server_already_used": "Ce nom d'hôte est déjà renseigné.",
+  "domain_tab_DNS_modification_form_ip_wrong_format": "Veuillez saisir une IP valide.",
+  "domain_tab_DNS_modification_form_internal_server_error": "Pour utiliser les DNS internes d'OVHcloud, veuillez sélectionner un autre type de configuration.",
+  "domain_tab_DNS_modification_modal_body": "Après validation, les DNS utilisés seront les suivants :",
+  "domain_tab_DNS_modification_update_error": "La mise à jour de vos DNS a échoué pour la raison suivante : ",
+  "domain_tab_DNS_modification_update_success": "Votre demande de mise à jour a bien été prise en considération et sera traitée dans les meilleurs délais."
 }
--- a/packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsConfigurationForm.tsx
+++ b/packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsConfigurationForm.tsx
@@ -1,0 +1,170 @@
+import {
+  Button,
+  BUTTON_SIZE,
+  BUTTON_VARIANT,
+  Divider,
+  ICON_NAME,
+  Message,
+  MESSAGE_COLOR,
+  MessageBody,
+  MessageIcon,
+} from '@ovhcloud/ods-react';
+import React, { useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import ipaddr from 'ipaddr.js';
+import { useNavigate } from 'react-router-dom';
+import { NAMESPACES } from '@ovh-ux/manager-common-translations';
+import { TDomainZone } from '@/domain/types/domainZone';
+import { TDomainResource, TNameServer } from '@/domain/types/domainResource';
+import { ActiveConfigurationTypeEnum } from '@/domain/enum/dnsConfigurationType.enum';
+import DnsLineInput from '@/domain/components/ModifyNameServer/DnsLineInput';
+import {
+  canSaveNewDnsConfig,
+  computeDisplayNameServers,
+} from '@/domain/utils/dnsUtils';
+import NewDnsConfigModal from '@/domain/components/ModifyNameServer/NewDnsConfigModal';
+import { useGenerateUrl } from '@/domain/hooks/generateUrl/useGenerateUrl';
+import { urls } from '@/domain/routes/routes.constant';
+
+interface DnsConfigurationFormProps {
+  readonly domainZone: TDomainZone;
+  readonly selectedConfig: ActiveConfigurationTypeEnum;
+  readonly currentState: TDomainResource['currentState'];
+  readonly serviceName: string;
+}
+
+export default function DnsConfigurationForm({
+  selectedConfig,
+  domainZone,
+  serviceName,
+  currentState,
+}: DnsConfigurationFormProps) {
+  const { t } = useTranslation('domain');
+  const { t: tCommon } = useTranslation(NAMESPACES.ACTIONS);
+  const navigate = useNavigate();
+  const backUrl = useGenerateUrl(urls.domainTabDns, 'path', {
+    serviceName,
+  });
+  const initialServers = computeDisplayNameServers(
+    currentState.dnsConfiguration,
+    domainZone,
+    selectedConfig,
+  );
+
+  const [nameServers, setNameServers] = useState<TNameServer[]>(initialServers);
+  const [isSavingModalOpen, setIsSavingModalOpen] = useState(false);
+  const [feedbackMessage, setFeedbackMessage] = useState<string>(null);
+
+  const handleAdd = ({ server, ip }: { server: string; ip: string }) => {
+    if (!server) return;
+
+    const newEntry: TNameServer = { nameServer: server.replace('/ /g', '') };
+
+    if (ip && ipaddr.isValid(ip)) {
+      const parsed = ipaddr.parse(ip);
+      const kind = parsed.kind();
+      if (kind === 'ipv4') {
+        newEntry.ipv4 = ip;
+      } else if (kind === 'ipv6') {
+        newEntry.ipv6 = ip;
+      }
+    }
+    setNameServers((prev) => [...prev, newEntry]);
+  };
+
+  const handleRemove = (indexToRemove: number) => {
+    setNameServers((prev) =>
+      prev.filter((_: TNameServer, i: number) => i !== indexToRemove),
+    );
+  };
+
+  const canSave = canSaveNewDnsConfig(
+    initialServers,
+    nameServers,
+    currentState.dnsConfiguration,
+  );
+
+  return (
+    <div className="mt-6 ml-8">
+      <Message dismissible={false} className="mb-6 w-full">
+        <MessageIcon name={ICON_NAME.circleInfo} />
+        <MessageBody>
+          {t('domain_tab_DNS_modification_form_dns_number', {
+            min: currentState.dnsConfiguration.minDNS,
+            max: currentState.dnsConfiguration.maxDNS,
+          })}
+        </MessageBody>
+      </Message>
+      <div>
+        {nameServers.map((ns, index) => (
+          <React.Fragment key={ns.nameServer}>
+            <DnsLineInput
+              server={ns.nameServer.trim()}
+              showLabels={index === 0}
+              onRemove={() => handleRemove(index)}
+              editable={false}
+              allServers={nameServers}
+            />
+            <Divider className="my-6" data-testid={'dns-divider'} />
+          </React.Fragment>
+        ))}
+      </div>
+      <DnsLineInput
+        server=""
+        showLabels={true}
+        onAdd={handleAdd}
+        editable={true}
+        allServers={nameServers}
+      />
+      <div className="flex gap-4 mt-6">
+        <Button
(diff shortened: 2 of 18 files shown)


packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsConfigurationForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+              editable={false}
+              allServers={nameServers}
+            />
+            <Divider className="my-6" data-testid={'dns-divider'} />
+          </React.Fragment>
+        ))}
+      </div>
+      <DnsLineInput
+        server=""
+        ip=""
```
nfoufeovh:
```suggestion
```

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsConfigurationRadio.spec.tsx
```diff
@@ -7,26 +7,22 @@ import {
   baseDomainResource,
   domainZoneMock,
 } from '@/domain/__mocks__/dnsDetails';
-import { computeActiveConfiguration } from '@/domain/utils/dnsUtils';
-import { ActiveConfigurationTypeEnum } from '@/domain/enum/dnsConfigurationType.enum';
 
-vi.mock('@/domain/utils/dnsUtils', () => ({
-  computeActiveConfiguration: vi.fn(),
+vi.mock('@/domain/components/ModifyNameServer/DnsConfigurationForm', () => ({
+  default: () => <div>Form DNS</div>,
```
nfoufeovh:
```suggestion
  default: () => <div>DNS form</div>,
```

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsConfigurationForm.tsx
```diff
@@ -0,0 +1,170 @@
...
+
+  const initialServers = computeDisplayNameServers(
+    currentState.dnsConfiguration,
+    domainZone,
+    selectedConfig,
+  );
+
+  const [nameServers, setNameServers] = useState<TNameServer[]>(initialServers);
+  const [isSavingModalOpen, setIsSavingModalOpen] = useState(false);
+  const [feedbackMessage, setFeedbackMessage] = useState<{
```
nfoufeovh: Why is this an object and not a string containing directly the message? The type does not seem to be used to make custom display (I thought it would be used to define the type of banner to use)

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsLineInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+}: DnsLineInputProps) {
+  const { t } = useTranslation('domain');
+
+  const [serverValue, setServerValue] = useState(server);
+  const [ipValue, setIpValue] = useState(ip);
+
+  const [serverError, setServerError] = useState<'format' | 'duplicate' | null>(
+    null,
+  );
+  const [ipError, setIpError] = useState(false);
```
nfoufeovh:
```suggestion
  const [isIpError, setIsIpError] = useState(false);
```

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsLineInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+  const [isInternalDnsError, setIsInternalDnsError] = useState(false);
+
+  const dnsRegex = /^([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}$/;
+
+  const validateServer = () => {
```
nfoufeovh, line 52: I think we should do something like:
```
if (!serverValue) { # 1: we check if the input is empty -> it is legit not to return any error
      setServerError(null);
      setIsInternalDnsError(false);
      return;
}

const trimmed = serverValue.trim(); # 2: then, we try to trim it: if the !trimmed condition is verified, it would be due to an error (wrong input type for instance), so we handle it
    if (!trimmed || !dnsRegex.test(trimmed)) {
      setServerError('format');
      setIsInternalDnsError(false);
      return;
    }
```
or you can wrap all the checks in a `if (serverValue) {...}` condition, so the case when `serverValue` is empty is handled by line 73 (so we avoid duplication)

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsLineInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+    }
+
+    const isDuplicate = allServers.some((ns) => ns.nameServer === trimmed);
+    if (isDuplicate) {
+      setServerError('duplicate');
+      setIsInternalDnsError(false);
+      return;
+    }
+    if (isInternalDns(trimmed)) {
+      setIsInternalDnsError(true);
```
nfoufeovh, line 74:
```suggestion
      setServerError(null);
      setIsInternalDnsError(true);
```

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsLineInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+      return;
+    }
+    setServerError(null);
```
nfoufeovh, line 78:
```suggestion

    setServerError(null);
```

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsLineInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+    }
+    const isValid = ipaddr.isValid(ipValue);
+    setIpError(!isValid);
+  };
+
+  const handleAdd = () => {
+    validateServer();
+    validateIp();
+
+    if (serverError !== null || ipError || isInternalDnsError) return;
```
nfoufeovh:
```suggestion
    if (serverError || ipError || isInternalDnsError) return;
```

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsLineInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+  const [serverError, setServerError] = useState<'format' | 'duplicate' | null>(
+    null,
+  );
+  const [ipError, setIpError] = useState(false);
+  const [isInternalDnsError, setIsInternalDnsError] = useState(false);
+
+  const dnsRegex = /^([a-zA-Z0-9-]+\.)+[a-zA-Z]{2,}$/;
+
+  const validateServer = () => {
+    const trimmed = serverValue.trim();
```
nfoufeovh: Can't we store the result globally so we do not have to trim again when we call the `onAdd()` function (line 93)?

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsLineInput.tsx
```diff
@@ -0,0 +1,209 @@
...
+    setServerError(null);
+    setIsInternalDnsError(false);
+  };
+
+  const validateIp = () => {
+    if (!ipValue) {
+      setIpError(false);
+      return;
+    }
+    const isValid = ipaddr.isValid(ipValue);
```
nfoufeovh: Shouldn't we trim the ip before using it to check the validity (and, just as suggested for the nameserver, store the trimmedValue globally)?

packages/manager/apps/web-domains/src/domain/data/api/domainResources.ts
```diff
@@ -15,6 +15,30 @@ export const getDomainResource = async (
...
 ): Promise<TDomainOption> => {
-  const { data } = await v6.get(`/domain/${serviceName}/option/${OptionEnum.DNS_ANYCAST}`);
+  const { data } = await v6.get(
+    `/domain/${serviceName}/option/${OptionEnum.DNS_ANYCAST}`,
+  );
+  return data;
+};
+
+/**
+ *  : Update DNS configuration of resource
```
nfoufeovh:
```suggestion
 *  : Update a domain resource
```
If this function has to be used to update all the fields of the resource, update the comment; if this function should only be used to update the DNS (which mean that other ones should be defined to update each other field of the resource), rename the function (`updateDomainDnsConfiguration` for instance)

packages/manager/apps/web-domains/src/domain/utils/dnsUtils.ts
```diff
@@ -128,3 +129,69 @@ export function computeActiveConfiguration(
...
+    return initialServers.some((initial) => {
+      const idx = unmatched.findIndex((ns) => areServersEqual(initial, ns));
+      if (idx === -1) return true;
+      unmatched.splice(idx, 1);
+      return false;
+    });
+  };
+
+  const count = newServers.length;
+  if (selectedConfig === ActiveConfigurationTypeEnum.EXTERNAL) {
```
nfoufeovh: These 2 conditions (line 190 and 193) are weird, I think we should just check if the count is `>=`  to the `minDNS` and `<=` to the `maxDNS` values of the resource (do we display an error or a tooltip somewhere if they are not?)

nfoufeovh (changes requested):

packages/manager/apps/web-domains/public/translations/domain/Messages_fr_FR.json
```diff
@@ -65,6 +65,18 @@
...
   "domain_tab_DNS_modification_activate_dns_link": "Activer votre zone DNS",
   "domain_tab_DNS_modification_mixed_activate_dns": "Pour utiliser une configuration mixte incluant les DNS d’OVHcloud, veuillez d’abord activer une zone DNS.",
-  "domain_tab_DNS_modification_mixed_dns_number": "Veuillez renseigner entre 2 et 14 DNS.",
-  "domain_tab_DNS_modification_external_dns_number": "Veuillez renseigner entre 2 et 10 DNS."
+  "domain_tab_DNS_modification_form_dns_number": "Veuillez renseigner entre {{min}} et {{max}} DNS.",
+  "domain_tab_DNS_modification_form_server_field": "Serveur DNS",
+  "domain_tab_DNS_modification_form_ip_field": "IP associée",
+  "domain_tab_DNS_modification_form_placeholder": "ns1.my-domain-is-gre.at",
+  "domain_tab_DNS_modification_form_placeholder_ip": "198.27.92.1",
+  "domain_tab_DNS_modification_form_add": "Ajouter",
```
louisbensiovh: Maybe use common translation ?

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsConfigurationForm.tsx
```diff
@@ -0,0 +1,164 @@
...
+  const navigate = useNavigate();
+
+  const initialServers = computeDisplayNameServers(
+    currentState.dnsConfiguration,
+    domainZone,
+    selectedConfig,
+  );
+
+  const [nameServers, setNameServers] = useState<TNameServer[]>(initialServers);
+  const [isSavingModalOpen, setIsSavingModalOpen] = useState(false);
```
louisbensiovh, line 55:
```suggestion
  const [isSavingModalOpen, setIsSavingModalOpen] = useState<boolean>(false);
```

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsConfigurationForm.tsx
```diff
@@ -0,0 +1,164 @@
...
+    initialServers,
+    nameServers,
+    selectedConfig,
+    currentState.dnsConfiguration
+  );
+
+  return (
+    <div className="mt-6 ml-8">
+      <Message dismissible={false} className="mb-6 w-full">
+        <MessageIcon name="circle-info" />
```
louisbensiovh:
```suggestion
        <MessageIcon name={ICON_NAME.circleInfo} />
```

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsLineInput.tsx
```diff
@@ -0,0 +1,214 @@
...
+  readonly showLabels?: boolean;
+  readonly onRemove?: () => void;
+  readonly onAdd?: (values: { server: string; ip: string }) => void;
+  readonly editable: boolean;
+  readonly allServers: TNameServer[];
+}
+
+export default function DnsLineInput({
+  server,
+  ip,
```
louisbensiovh, line 31: Maybe this
```suggestion
  ip = '',
```

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsLineInput.tsx
```diff
@@ -0,0 +1,214 @@
...
+  showLabels = false,
+  onRemove,
+  onAdd,
+  editable,
+  allServers,
+}: DnsLineInputProps) {
+  const { t } = useTranslation('domain');
+
+  const [serverValue, setServerValue] = useState(server);
+  const [ipValue, setIpValue] = useState(ip ?? '');
```
louisbensiovh, line 41:
```suggestion
  const [ipValue, setIpValue] = useState(ip);
```

packages/manager/apps/web-domains/src/domain/pages/domainTabs/dns/dnsConfiguration.tsx
```diff
@@ -31,16 +31,37 @@ export default function DnsConfigurationTab({
...
   return (
     <div data-testid="datagrid">
+      {localSuccessMessage && (
+        <Message
+          color={MESSAGE_COLOR.success}
+          className="w-full mt-4"
+          dismissible={true}
+          onRemove={() => setLocalSuccessMessage(null)}
+        >
+          <MessageIcon name="circle-check" />
```
louisbensiovh:
```suggestion
        <MessageIcon name={ICON_NAME.circleCheck} />
```

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsLineInput.tsx
SBrendan: We are obliged to set a default value, otherwise we go from an uncontrolled input to a controlled one, because the IP is an optional value in the props.

packages/manager/apps/web-domains/src/domain/components/ModifyNameServer/DnsConfigurationForm.tsx
SBrendan: It is not mandatory to type it, because we pass the default value false, so useState inherits the type, which is boolean.

PR #18667, opened 2025-08-14
feat(pci-block-storage): use ods 19 components

SimonChaumet:
## Description

Ticket Reference: #...


SimonChaumet:
Will use bundler version instead

PR #18728, opened 2025-08-20
feat(*): add optional step label to mrc modal

chipp972:
ref: #ISSUE-18727

## Description

Image

Optional step label on the top right of the MRC modals

### DIFF
--- a/packages/manager-react-components/src/components/Modal/Modal.component.tsx
+++ b/packages/manager-react-components/src/components/Modal/Modal.component.tsx
@@ -12,10 +12,19 @@ import {
   ODS_TEXT_PRESET,
   ODS_SPINNER_SIZE,
 } from '@ovhcloud/ods-components';
+import { NAMESPACES } from '@ovh-ux/manager-common-translations';
+import { useTranslation } from 'react-i18next';
 
 export interface ModalProps {
   /** Title of modal */
   heading?: string;
+  /** Properties for the step number displayed on the top right of the modal */
+  step?: {
+    /** Current step displayed on the modal (must define heading and total) */
+    current?: number;
+    /** Total number of steps in the modal (must defined heading and current) */
+    total?: number;
+  };
   /** Type of modal. It can be any of `ODS_MODAL_COLOR` */
   type?: ODS_MODAL_COLOR;
   /** Is loading state for display a spinner */
@@ -59,6 +68,7 @@ export const ModalComponent: React.ForwardRefRenderFunction<
 > = (
   {
     heading,
+    step,
     type = ODS_MODAL_COLOR.information,
     isLoading,
     primaryLabel,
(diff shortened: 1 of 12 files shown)


atefzed:
I think it would be better to either create a new ModalStepper component or change the heading prop type from string to ReactNode so you can pass a custom header with the stepper. This way, the global modal stays flexible and doesn’t get tied directly to the stepper.

chipp972:
Control tower didn't want to change the title type to ReactNode @anooparveti

atefzed:
> Control tower didn't want to change the title type to ReactNode @anooparveti

If changing the title type to ReactNode isn't possible, could you at least add an optional slot to insert a custom header? That way, we can still support the stepper without affecting the existing title.

chipp972:
Well if they don't want to change title to ReactNode type which is compatible with string type, they probably don't want an optional slot to insert a custom header.

This is basically the same thing but with an additional prop.

I think if they don't want to change title to ReactNode it's because they want to limit what we can do with the title part of the modal.

packages/manager-wiki/stories/manager-react-components/components/Modal/Modal.stories.tsx
```diff
@@ -11,6 +11,8 @@ import {
 
 const ModalStory = ({
   heading,
+  currentStep,
+  totalStepNumber,
```
anooparveti: **nitpick:** We can group the related props into single object.
```suggestion
  step: {
    current: number,
    total: number,
  },
```

packages/manager-react-components/src/components/Modal/Modal.component.tsx
```diff
@@ -89,17 +98,36 @@ export const ModalComponent: React.ForwardRefRenderFunction<
...
+              {heading}
+            </OdsText>
+            <OdsText className="ml-auto" preset={ODS_TEXT_PRESET.caption}>
+              {t('stepPlaceholder', {
+                current: currentStep,
+                total: totalStepNumber,
+              })}
+            </OdsText>
+          </div>
+        )}
```
anooparveti: **nitpick:** Can be combined into only one statement.
```suggestion
      {heading &&
       (
          <div className="flex items-center mb-4">
            <OdsText
              className="block mr-3 flex-1"
              preset={ODS_TEXT_PRESET.heading4}
            >
              {heading}
            </OdsText>
            {Number.isInteger(step.current) && Number.isInteger(step.total) && <OdsText className="ml-auto" preset={ODS_TEXT_PRESET.caption}>
              {t('stepPlaceholder', {
                current: currentStep,
                total: totalStepNumber,
              })}
            </OdsText>
          </div>
        )}
```

anooparveti (changes requested):

PR #18749, opened 2025-08-21
fix(vrack-services): fix sentry issue

mhelhali-soufien:
ref: #MANAGER-16981

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/vrack-services/src/main.tsx
+++ b/packages/manager/apps/vrack-services/src/main.tsx
@@ -9,7 +9,7 @@ import { App } from './App';
 import './index.scss';
 import './vite-hmr';
 import { getTrackingContext } from './utils/tracking';
-import { TRANSLATION_NAMESPACES } from './utils/constants';
+import { LEVEL2, TRANSLATION_NAMESPACES } from './utils/constants';
 
 const init = async ({ appName }: { appName: string }) => {
   const context = await initShellContext(appName, getTrackingContext(appName));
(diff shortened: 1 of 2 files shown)


qpavy (approved):

packages/manager/apps/vrack-services/src/main.tsx
```diff
@@ -20,6 +20,14 @@ const init = async ({ appName }: { appName: string }) => {
     ns: [TRANSLATION_NAMESPACES.common, TRANSLATION_NAMESPACES.listing],
   });
 
+  const region = context.environment.getRegion();
+  context.shell.tracking.setConfig(region, LEVEL2);
+  try {
+    await import(`./config-${region}.js`);
+  } catch (error) {
```
aderghamov, line 27: Shouldn't we log an error when the config file is missing instead of silently failing?

mhelhali-soufien: I make it like all others micro apps

chipp972: it catches if the file is missing or if the script throws an error
But where do you want to log an error and for who ?

At this point in the app we don't have anything displayed and if there is a missing file in prod we shouldn't display any information to the user on implementation details and errors

So maybe we should just not try/catch this file execution

chipp972 (approved):

packages/manager/apps/vrack-services/src/main.tsx
aderghamov: Ok, the log is a bad idea.
Maybe, removing the try/catch and letting the error management redirect the user to a proper page is the good solution.

PR #18778, opened 2025-08-25
feat(ips): export ips to csv file functionality

sachinrameshn:
ref: #MANAGER-18928

## Description
This PR includes exporting ips into CSV file. 

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/ips/public/translations/export-ip-to-csv/Messages_fr_FR.json
+++ b/packages/manager/apps/ips/public/translations/export-ip-to-csv/Messages_fr_FR.json
@@ -1,0 +1,14 @@
+{
+  "exportIpToCsvTitle": "Exporter mes adresses IP dans un fichier CSV",
+  "exportIpToCsvExplanation": "L'export de vos IP dure quelques secondes (en fonction du nombre d'IP que vous possédez).",
+  "exportIpToCsvConfirm": "Etes-vous sûr de vouloir exporter les IP en format CSV ?",
+  "exportIpToCsvSuccess": "L'export de vos IP est terminé.",
+  "exportIpToCsvError": "Une erreur est survenue lors de l'export.",
+  "exportIpToCsvDownloadingFile": "Le fichier est en cours de création, Veuillez patienter svp",
+  "exportIpToCsvHeaderIpBlock": "Bloque d'IP",
+  "exportIpToCsvHeaderVersion": "Version",
+  "exportIpToCsvHeaderType": "Type",
+  "exportIpToCsvHeaderCountry": "Pays",
+  "exportIpToCsvHeaderServiceName": "Service Attaché",
+  "exportIpToCsvHeaderDescription": "Description"
+}
--- a/packages/manager/apps/ips/src/data/hooks/useExportIpToCsv.ts
+++ b/packages/manager/apps/ips/src/data/hooks/useExportIpToCsv.ts
@@ -1,0 +1,68 @@
+import { useTranslation } from 'react-i18next';
+import { mkConfig, generateCsv, download } from 'export-to-csv';
+import { useMutation } from '@tanstack/react-query';
+import { TRANSLATION_NAMESPACES } from '@/utils';
+import { getIpExport, GetIpListParams } from '../api';
+
+export interface ExportIpToCsvData {
+  ipBlock: string;
+  version: string;
+  type: string;
+  country: string | null;
+  routedTo: {
+    serviceName: string;
+  };
+  description: string | null;
+}
+
+export type ExportIpMutationData = {
+  downloadFn?: () => void;
+};
+
+export const useExportIpToCsv = ({
+  apiFilter,
+  onSuccess,
+  onError,
+}: {
+  apiFilter: GetIpListParams;
+  onSuccess?: (data: ExportIpMutationData) => void;
+  onError?: () => void;
+}) => {
+  const { t } = useTranslation(TRANSLATION_NAMESPACES.exportIpToCsv);
+  const handleExportToCsv = async (): Promise<ExportIpMutationData> => {
+    const result = await getIpExport({
+      ...apiFilter,
+      serviceName: apiFilter['routedTo.serviceName'],
+    });
+
+    const resultLength = result?.data?.length ?? 0;
+    if (resultLength > 0) {
+      const csvConfig = mkConfig({
+        filename: 'export_ips',
+        fieldSeparator: ',',
+        quoteStrings: true,
+        decimalSeparator: '.',
+        useKeysAsHeaders: true,
+      });
+
+      const csvData = result.data.map((item: ExportIpToCsvData) => ({
+        [t('exportIpToCsvHeaderIpBlock')]: item.ipBlock,
+        [t('exportIpToCsvHeaderVersion')]: item.version,
+        [t('exportIpToCsvHeaderType')]: item.type,
+        [t('exportIpToCsvHeaderCountry')]: item.country || '',
+        [t('exportIpToCsvHeaderServiceName')]: item.routedTo?.serviceName || '',
+        [t('exportIpToCsvHeaderDescription')]: item.description || '',
+      }));
+
+      const csv = generateCsv(csvConfig)(csvData);
+      return { downloadFn: () => download(csvConfig)(csv) };
+    }
+    throw new Error('No valid CSV could be downloaded');
+  };
+
+  return useMutation({
+    mutationFn: handleExportToCsv,
+    onSuccess,
+    onError,
+  });
+};
--- a/packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
+++ b/packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
@@ -1,0 +1,82 @@
+import React, { useContext, useRef } from 'react';
+import { useTranslation } from 'react-i18next';
+import { useNavigate } from 'react-router-dom';
+import { ODS_TEXT_PRESET } from '@ovhcloud/ods-components';
+import { NAMESPACES } from '@ovh-ux/manager-common-translations';
+import { OdsText } from '@ovhcloud/ods-components/react';
+import { Modal, useNotifications } from '@ovh-ux/manager-react-components';
+import Loading from '@/components/Loading/Loading';
+import { TRANSLATION_NAMESPACES } from '@/utils';
+import { useExportIpToCsv } from '@/data/hooks';
+import { ListingContext } from '@/pages/listing/listingContext';
+
+export default function ExportIpToCsv() {
+  const { t } = useTranslation([
+    TRANSLATION_NAMESPACES.exportIpToCsv,
+    NAMESPACES.ACTIONS,
+  ]);
+  const { apiFilter } = useContext(ListingContext);
+  const { addSuccess, addError } = useNotifications();
+  const navigate = useNavigate();
+
+  /*
+    We need this ref to prevent a double navigation
+    when closing the modale before the end of the download.
+    A useState or useLocation would have required a rerender 
+    that we cannot have upon closing the modale
+  */
+  const isOpen = useRef<boolean>(true);
+
+  const closeModal = () => {
+    if (isOpen.current) {
+      isOpen.current = false;
+      navigate('..');
+    }
+  };
+
+  const {
+    mutateAsync: handleExportToCsv,
+    isPending: isCSVLoading,
+  } = useExportIpToCsv({
+    apiFilter,
+    onSuccess: (data) => {
+      if (isOpen.current) {
+        data.downloadFn();
+        addSuccess(t('exportIpToCsvSuccess'));
+        closeModal();
+      }
+    },
+    onError: () => addError(t('exportIpToCsvError')),
+  });
+
+  return (
+    <Modal
+      isOpen
+      onDismiss={closeModal}
+      heading={t('exportIpToCsvTitle')}
+      primaryLabel={t('validate', { ns: NAMESPACES.ACTIONS })}
+      onPrimaryButtonClick={handleExportToCsv}
+      isPrimaryButtonLoading={isCSVLoading}
+      secondaryLabel={t('cancel', { ns: NAMESPACES.ACTIONS })}
+      onSecondaryButtonClick={closeModal}
+    >
+      {isCSVLoading ? (
+        <>
+          <OdsText className="mb-6" preset={ODS_TEXT_PRESET.paragraph}>
+            {t('exportIpToCsvDownloadingFile')}
+          </OdsText>
+          <Loading />
+        </>
+      ) : (
+        <>
+          <OdsText className="mb-2" preset={ODS_TEXT_PRESET.paragraph}>
+            {t('exportIpToCsvExplanation')}
+          </OdsText>
+          <OdsText className="mb-6" preset={ODS_TEXT_PRESET.paragraph}>
+            {t('exportIpToCsvConfirm')}
+          </OdsText>
+        </>
+      )}
+    </Modal>
+  );
+}
--- a/packages/manager/apps/ips/src/pages/listing/ipListing/ip.listing.page.tsx
+++ b/packages/manager/apps/ips/src/pages/listing/ipListing/ip.listing.page.tsx
@@ -59,23 +59,32 @@ export default function IpListingPage() {
         </div>
 
         <div className="flex flex-col items-end ml-auto gap-2">
-          <ActionMenu
-            id="settings"
-            icon={ODS_ICON_NAME.cog}
-            isCompact
-            items={[
-              {
-                id: 0,
-                label: t('listingSettingsImportIpFromSys'),
-                isDisabled: environment.user.ovhSubsidiary === 'US',
-                onClick: () => {
-                  navigate(
-                    `${urls.listingImportIpFromSys}?${search.toString()}`,
-                  );
+          <div className="flex items-end ml-auto gap-2">
+            <ActionMenu
+              id="settings"
+              icon={ODS_ICON_NAME.cog}
+              isCompact
+              items={[
+                {
+                  id: 0,
+                  label: t('listingSettingsImportIpFromSys'),
+                  isDisabled: environment.user.ovhSubsidiary === 'US',
+                  onClick: () => {
+                    navigate(
+                      `${urls.listingImportIpFromSys}?${search.toString()}`,
+                    );
+                  },
                 },
-              },
-            ]}
-          />
+              ]}
+            />
+            <OdsButton
+                variant={ODS_BUTTON_VARIANT.outline}
+                icon={ODS_ICON_NAME.download}
+                size={ODS_BUTTON_SIZE.sm}
+                onClick={() => navigate(urls.listingExportIpToCsv)}
+                label=""
+              />
+          </div>
           <QuickFilter />
         </div>
       </div>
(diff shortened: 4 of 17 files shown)


packages/manager/apps/ips/public/translations/export-ip-to-csv/Messages_fr_FR.json
```diff
@@ -0,0 +1,8 @@
+{
+  "exportIpToCsvTitle": "Exporter mes adresses IP dans un fichier CSV",
+  "exportIpToCsvExplanation": "L'export de vos IP dure quelques secondes (en fonction du nombre d'IP que vous possédez).",
+  "exportIpToCsvConfirm": "Etes-vous sûr de vouloir exporter les IP en format CSV ?",
+  "exportIpToCsvSuccess": "L'export de vos IP est terminé.",
+  "exportIpToCsvError": "Une erreur est survenue lors de l'export.",
+  "exportIpToCsvDownloadingFile": "Le fichier est en cours de creation, Veuillez patienter svp"
```
qpavy:
```suggestion
  "exportIpToCsvDownloadingFile": "Le fichier est en cours de création, Veuillez patienter svp"
```

packages/manager/apps/ips/src/data/hooks/useExportToFile.ts
```diff
@@ -0,0 +1,92 @@
...
+  const { t } = useTranslation([TRANSLATION_NAMESPACES.exportIpToCsv]);
+  const { apiFilter, ipToSearch } = useContext(ListingContext);
+  const { addSuccess, addError } = useNotifications();
+  const [isCSVLoading, setIsCSVLoading] = useState(false);
+
+  const handleExportToCsv = async () => {
+    setIsCSVLoading(true);
+    try {
+      const result = await aapi
+        .get('/ipsexportcsv', {
```
qpavy: Why not using react query ?

packages/manager/apps/ips/src/data/hooks/useExportToFile.ts
```diff
@@ -0,0 +1,92 @@
...
+
+        const csvData = result.data.map((item: ExportIpToCsvData) =>
+          exportColumns.reduce(
+            (acc, column) => ({
+              ...acc,
+              [column.label]: column.getValue(item),
+            }),
+            {},
+          ),
+        );
```
qpavy: It feels complex compute for what it does. Maybe we can simplify it and also remove the reduce inside the map :arrow_down: 

```suggestion
        const csvData = result.data.map((item: ExportIpToCsvData) =>
          ({
            'IP Block': item.ipBlock,
            'Version': item.version,
            'Type': item.type,
            'Country': item.country || '',
            'Service Name': item.routedTo?.serviceName || '',
            'Description': item.description || ''
          )}
        );
```

packages/manager/apps/ips/src/data/api/get/ipList.ts
```diff
@@ -18,7 +18,7 @@ export type GetIpListParams = {
   /** Filter the value of isAdditionalIp property (&#x3D;) (alpha) */
   isAdditionalIp?: boolean;
   /** Filter the value of routedTo.serviceName property (like) */
-  routedToserviceName?: string;
+  'routedTo.serviceName'?: string;
```
chipp972: Is this really how the property is named in the api repsonse ?

packages/manager/apps/ips/src/data/hooks/useExportToFile.ts
```diff
@@ -0,0 +1,92 @@
...
+import { ExportIpToCsvData } from '@/data/api';
+import { TRANSLATION_NAMESPACES } from '@/utils';
+
+const exportColumns = [
+  {
+    label: 'IP Block',
+    getValue: (item: ExportIpToCsvData) => item.ipBlock,
+  },
+  {
+    label: 'Version',
```
chipp972: don't forget to add translation labels if it's relevant

packages/manager/apps/ips/src/data/hooks/useExportToFile.ts
```diff
@@ -0,0 +1,92 @@
...
+        .get('/ipsexportcsv', {
+          params: {
+            ...apiFilter,
+            serviceName: apiFilter['routedTo.serviceName'],
+            ip: ipToSearch,
+          },
+        })
+        .then(({ data }) => data);
+
+      if (result && result?.data && result?.data.length > 0) {
```
b-fritz-ovh: We can simplify `result?.data?.length` or `result?.data?.length > 0` ?

packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
```diff
@@ -0,0 +1,78 @@
...
+    <OdsModal
+      isOpen
+      onOdsClose={closeModal}
+      color={ODS_MODAL_COLOR.information}
+      isDismissible
+    >
+      <OdsText className="mb-4" preset={ODS_TEXT_PRESET.heading4}>
+        {t('exportIpToCsvTitle')}
+      </OdsText>
+      {!isCSVLoading && (
```
b-fritz-ovh: May be use a ternary?

```
{ isCSVLoading ? (
<>
          <OdsText className="mb-6" preset={ODS_TEXT_PRESET.paragraph}>
            {t('exportIpToCsvDownloadingFile')}
          </OdsText>
          <Loading />
        </>) : (
        <>
          <OdsText className="mb-2" preset={ODS_TEXT_PRESET.paragraph}>
            {t('exportIpToCsvExplanation')}
          </OdsText>
          <OdsText className="mb-6" preset={ODS_TEXT_PRESET.paragraph}>
            {t('exportIpToCsvConfirm')}
          </OdsText>
        </>) }
        ```

File: packages/manager/apps/ips/src/data/api/get/ipList.ts (hunk 4 above)
Comment (sachinrameshn) [reply packages/manager/apps/ips/src/data/api/get/ipList.ts]: yes

File: packages/manager/apps/ips/src/data/hooks/useExportToFile.ts (hunk 2 above)
Comment (aderghamov) [reply packages/manager/apps/ips/src/data/hooks/useExportToFile.ts]: IMHO, using react-query would add much complexity for nothing.

File: packages/manager/apps/ips/src/data/hooks/useExportToFile.ts [hunk 8]
```diff
@@ -0,0 +1,66 @@
...
+import { useTranslation } from 'react-i18next';
+import { useNotifications } from '@ovh-ux/manager-react-components';
+import { aapi } from '@ovh-ux/manager-core-api';
+import { mkConfig, generateCsv, download } from 'export-to-csv';
+import { ListingContext } from '@/pages/listing/listingContext';
+import { ExportIpToCsvData } from '@/data/api';
+import { TRANSLATION_NAMESPACES } from '@/utils';
+
+export const useExportToCsv = () => {
+  const { t } = useTranslation([TRANSLATION_NAMESPACES.exportIpToCsv]);
```
chipp972:
```suggestion
  const { t } = useTranslation(TRANSLATION_NAMESPACES.exportIpToCsv);
```

packages/manager/apps/ips/src/pages/listing/ipListing/ip.listing.page.tsx
```diff
@@ -72,6 +72,13 @@ export default function IpListingPage() {
             ]}
           />
           <QuickFilter />
+          <OdsButton
+            variant={ODS_BUTTON_VARIANT.outline}
+            icon={ODS_ICON_NAME.download}
+            size={ODS_BUTTON_SIZE.sm}
+            onClick={() => navigate(urls.listingExportIpToCsv)}
+            label=""
+          />
```
chipp972: I think we need to fix the design to have the same as the angular app
The button should be next to the settings button, on the same line so I think you should wrap the ActionMenu above and this OdsButton in a div with a className `flex gap-2`



It should look like this (the 2 buttons on the top right)

packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
```diff
@@ -0,0 +1,77 @@
...
+    navigate('..');
+  };
+
+  return (
+    <OdsModal
+      isOpen
+      onOdsClose={closeModal}
+      color={ODS_MODAL_COLOR.information}
+      isDismissible
+    >
```
chipp972: Why not use Modal component from MRC ?

packages/manager/apps/ips/src/data/hooks/useExportToFile.ts
```diff
@@ -0,0 +1,66 @@
...
+
+  const handleExportToCsv = async () => {
+    setIsCSVLoading(true);
+    try {
+      const result = await aapi
+        .get('/ipsexportcsv', {
+          params: {
+            ...apiFilter,
+            serviceName: apiFilter['routedTo.serviceName'],
+            ip: ipToSearch,
```
chipp972:
```suggestion
```

I think we will use the ip field from apiFilter directly in the future

packages/manager/apps/ips/src/data/api/get/ipList.ts
```diff
@@ -62,3 +62,14 @@ export const getIcebergIpList = (params: GetIpListParams) =>
...
+export interface ExportIpToCsvData {
+  ipBlock: string;
+  version: string;
+  type: string;
+  country: string | null;
+  routedTo: {
+    serviceName: string;
+  };
+  description: string | null;
+}
```
chipp972: I think this type should be put next to the hook instead
It's not related to the GET /ip call

packages/manager/apps/ips/src/data/hooks/useExportToFile.ts
aderghamov: We will deal with it in the future, then 😜

packages/manager/apps/ips/src/data/api/get/ipList.ts
chipp972: You moved the type at the end of the file
It should be in packages/manager/apps/ips/src/data/hooks/useExportToFile.ts

Steffy29 (approved):

packages/manager/apps/ips/src/data/hooks/useExportToFile.ts
```diff
@@ -0,0 +1,76 @@
...
+  country: string | null;
+  routedTo: {
+    serviceName: string;
+  };
+  description: string | null;
+}
+
+export const useExportToCsv = () => {
+  const { t } = useTranslation(TRANSLATION_NAMESPACES.exportIpToCsv);
+  const { apiFilter, ipToSearch } = useContext(ListingContext);
```
chipp972: I'll delete ipToSearch also so don't use it : use apiFilter.ip instead

If it's coupled to ListingContext it should probably be in Listing folder. If you want to keep it in the data folder, you should take apiFilter as a parameter of the hook.

packages/manager/apps/ips/src/data/hooks/useExportToFile.ts
chipp972: The future is my next PR and I don't have this feature yet
Just remove it to avoid conflict
There is already an ip field in apiFilter

packages/manager/apps/ips/src/data/hooks/useExportToFile.ts
chipp972: Actually it would make more sense to decouple the display part (addSuccess, addError and loading) and just manage the data part, IF it's in the data folder and not in the Listing Page folder, and return a useMutation to manage the success/error/loading.

```
export const useExportCsv = ({apiFilter, onSuccess, onError }: { apiFilter: GetIpListParams; onSuccess?: () => void; onError?: () => void }) => {
  
  const handleExportToCsv = async () => {
    // No need to manage loading here
    const result = await aapi...
    // etc.
    return window.download(csvConfig)(csv);
  };

  return useMutation({
    mutationFn: handleExportToCsv,
    onSuccess,
    onError
  });
};
```

packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
```diff
@@ -0,0 +1,60 @@
...
+
+  const confirm = async () => {
+    await handleExportToCsv();
+    closeModal();
+  };
+
+  return (
+    <Modal
+      isOpen
+      onDismiss={!isCSVLoading ? closeModal : undefined}
```
chipp972:
```suggestion
      onDismiss={closeModal}
```

I don't think we should disable the modal closing when it's loading

packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
```diff
@@ -0,0 +1,60 @@
...
+      heading={t('exportIpToCsvTitle')}
+      primaryLabel={t('validate', { ns: NAMESPACES.ACTIONS })}
+      onPrimaryButtonClick={() => confirm()}
```
chipp972:
```suggestion
      onPrimaryButtonClick={confirm}
```

packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
```diff
@@ -0,0 +1,60 @@
...
+      isPrimaryButtonLoading={isCSVLoading}
+      secondaryLabel={t('cancel', { ns: NAMESPACES.ACTIONS })}
+      isSecondaryButtonDisabled={isCSVLoading}
```
chipp972:
```suggestion
```

User should still be able to cancel/get out of the modal if he doesn't want a csv anymore

Steffy29 (approved):

packages/manager/apps/ips/src/data/hooks/useExportIpToCsv.ts
```diff
@@ -0,0 +1,61 @@
...
+        [t('exportIpToCsvHeaderType')]: item.type,
+        [t('exportIpToCsvHeaderCountry')]: item.country || '',
+        [t('exportIpToCsvHeaderServiceName')]: item.routedTo?.serviceName || '',
+        [t('exportIpToCsvHeaderDescription')]: item.description || '',
+      }));
+
+      const csv = generateCsv(csvConfig)(csvData);
+      return { downloadFn: () => download(csvConfig)(csv) };
+    }
+    return {};
```
chipp972: Maybe we should throw instead to handle this error (no csv) ?

packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
```diff
@@ -0,0 +1,81 @@
...
+    TRANSLATION_NAMESPACES.exportIpToCsv,
+    NAMESPACES.ACTIONS,
+  ]);
+  const { apiFilter } = useContext(ListingContext);
+  const { addSuccess, addError } = useNotifications();
+  const navigate = useNavigate();
+  const isOpen = useRef<boolean>(true);
+
+  const {
+    mutateAsync: handleExportToCsv,
```
chipp972: add an `onSuccess` function and execute the downloadFn in the `onSuccess` function instead of inside an useEffect (and use `mutate` instead of `mutateAsync`)

packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
```diff
@@ -0,0 +1,81 @@
...
+
+export default function ExportIpToCsv() {
+  const { t } = useTranslation([
+    TRANSLATION_NAMESPACES.exportIpToCsv,
+    NAMESPACES.ACTIONS,
+  ]);
+  const { apiFilter } = useContext(ListingContext);
+  const { addSuccess, addError } = useNotifications();
+  const navigate = useNavigate();
+  const isOpen = useRef<boolean>(true);
```
chipp972, line 28: Why do you use a ref instead of a state ?
What's the purpose of this variable ?

packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
```diff
@@ -0,0 +1,81 @@
...
+  const closeModal = () => {
+    if (isOpen.current) {
+      isOpen.current = false;
+      navigate('..');
+    }
+  };
+
+  const confirm = async () => {
+    await handleExportToCsv();
+  };
```
chipp972: `confirm = handleExportToCsv` so just use `handleExportToCsv`, no need to create a function here

packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
aderghamov: when cancelling/dismissing the modal, we navigate to '/..'
when the download of the csv is finished, we navigate to '/..'
If I trigger a download and I cancel the modal, the navigation is triggered twice so I need to keep track of the Open/close modale state to prevent non-necessary navigation. 

The value managed by useState is refreshed only when the component is re-rendered so it is not properly updated when the modal is closed.

packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
aderghamov: oups!

I missed it.

packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
aderghamov: I tried to make it work this way before ending up with a useRef.
Maybe the useRef will make it work.

aderghamov: I confirm it works.
thank you :)

packages/manager/apps/ips/src/pages/actions/exportIpToCsv/exportIpToCsv.page.tsx
chipp972: Ok I understand
But it would be easier to just check the current location when we want to close the modal and check if it contains the url of the export-csv instead of managing a stateful variable

```
const isModalOpen = location.pathname.includes(urls.exportIpToCsv);
```

aderghamov: ok. I will try this way.

aderghamov: That won't work 😢 
`useLocation` has the same problem as `useState`.
If I pass its `location` const into the onSuccess function, the function will use its value when first rendered but not the value when route has changed.

chipp972: Ok I don't see any other simpler way that would work for now

Let's keep the ref but add a comment about why we have this ref (double navigate('..') if we cancel while the download is executing)

chipp972 (approved):

Steffy29 (approved):

PR #18790, opened 2025-08-25
feat(manager-legacy-tools): delete ods 17

aboungnaseng-ovhcloud:
ref: #MANAGER-19506

## Description

- Correct import ods icon v18


qpavy (dismissed):

packages/manager/apps/dedicated-servers/package.json
```diff
@@ -26,6 +26,7 @@
     "@ovh-ux/manager-react-core-application": "^0.11.36",
     "@ovh-ux/manager-react-shell-client": "^0.9.1",
     "@ovh-ux/request-tagger": "^0.4.0",
+    "@ovhcloud/ods-common-core": "^17.2.2",
```
tibs245: Why not 18 ?

tibs245 (dismissed):
If you are sure `"@ovhcloud/ods-common-core": "^17.2.2",` should be 17 and not 18 it's ok for me

packages/manager/apps/dedicated-servers/package.json
aboungnaseng-ovhcloud: yes good catch why  dedicated servers use ods 17 setup for msw test and ods 18, there is a mistake in this app 😄

aottr:
> If you are sure `"@ovhcloud/ods-common-core": "^17.2.2",` should be 17 and not 18 it's ok for me

@tibs245 `@ovhcloud/ods-common-core` doesn't exist for 18. Since the app is using ODS18 I don't think this package is necessary in there.

packages/manager/apps/hpc-vmware-managed-vcd/package.json
```diff
@@ -30,6 +30,7 @@
     "@ovh-ux/manager-react-core-application": "^0.11.36",
     "@ovh-ux/manager-react-shell-client": "^0.9.1",
     "@ovh-ux/request-tagger": "^0.4.1",
+    "@ovhcloud/ods-common-core": "^17.2.2",
```
aottr: Same here. App seems to be using ODS18 but we're still importing ODS17 to run `odsSetup` to init the assets.

packages/manager-tools/manager-legacy-tools/test-utils/src/utils/ui-test-helpers-ods18.ts
```diff
@@ -50,7 +50,7 @@ type GetOdsButtonParams = {
   container: HTMLElement;
   label?: string;
   altLabel?: string;
-  iconName?: React.ComponentProps<typeof OsdsIcon>['name'];
+  iconName?: React.ComponentProps<typeof OdsIcon>['name'];
```
tibs245, line 53: I recommend to use type compatible but avoid to import ODS

packages/manager/apps/dedicated-servers/package.json
tibs245: And it's not a devDependencies ?

helabenkhalfallah (approved):

PR #18799, opened 2025-08-25
fix(pci-instances): add region params to delete action

Tsiorifamonjena:
## Description

This PR fix the loader to handle legacy route and redirect on the right React page.

Ticket Reference: #TAPC-4850

### DIFF
--- a/packages/manager/apps/pci-instances/src/routes/loaders/instanceLegacy.loader.ts
+++ b/packages/manager/apps/pci-instances/src/routes/loaders/instanceLegacy.loader.ts
@@ -16,7 +16,7 @@ type RedirectPattern = {
 
 const REDIRECT_ROUTES: RedirectPattern[] = [
   {
-    expectedPath: '/pci/projects/:projectId/instances/:instanceId',
+    expectedPath: '/pci/projects/:projectId/instances/:instanceId/*',
     redirectPath:
       '/pci/projects/:projectId/instances/region/:region/instance/:instanceId',
     validate: (match: PathMatch) =>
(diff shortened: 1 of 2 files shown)


aTurmo (approved):

packages/manager/apps/pci-instances/src/routes/loaders/instanceLegacy.loader.ts
```diff
@@ -16,7 +16,7 @@ type RedirectPattern = {
 
 const REDIRECT_ROUTES: RedirectPattern[] = [
   {
-    expectedPath: '/pci/projects/:projectId/instances/:instanceId',
+    expectedPath: '/pci/projects/:projectId/instances/:instanceId/*',
```
fredericvilcot: Why do we need wildcard here?

Tsiorifamonjena: We need wildcard because this kind of route `/pci/projects/:projectId/instances/:instanceId/billing/monthly/activate` does not match so it redirect to a 404

We need to handle other sub route from `/pci/projects/:projectId/instances/:instanceId`

selm3n (approved):

packages/manager/apps/pci-instances/src/routes/loaders/instanceLegacy.loader.ts
fredericvilcot: Yes, but wildcard should be used carefully as it redirects all possible paths under root one

Tsiorifamonjena: As discussed I add a new entry in the `REDIRECT_ROUTES` for the billing/monthly/activate to avoid redirecting other paths that do not need to be redirected

fredericvilcot (approved):
To be strongly tested :)

PR #18813, opened 2025-08-26
feat(okms): use pagination on secret version list

rjamet-ovh:
ref: #MANAGER-18951

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/versions/Versions.page.tsx
+++ b/packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/versions/Versions.page.tsx
@@ -76,20 +76,6 @@ const SecretVersionsPage = () => {
 
   return (
     <>
-      <OdsButton
-        className="mb-4"
-        onClick={() =>
-          navigate(
-            SECRET_MANAGER_ROUTES_URLS.secretVersionsDrawerCreateVersion(
-              domainId,
-              secretPathDecoded,
-            ),
-          )
-        }
-        label={t('add_new_version')}
-        icon={'plus'}
-      />
-
       <Datagrid
         columns={columns}
         items={versions || []}
@@ -100,6 +86,20 @@ const SecretVersionsPage = () => {
         sorting={sorting}
         onSortChange={setSorting}
         contentAlignLeft
+        topbar={
+          <OdsButton
+            label={t('add_new_version')}
+            onClick={() =>
+              navigate(
+                SECRET_MANAGER_ROUTES_URLS.secretVersionsDrawerCreateVersion(
+                  domainId,
+                  secretPathDecoded,
+                ),
+              )
+            }
+            icon={'plus'}
+          />
+        }
       />
       <Suspense>
         <Outlet />
--- a/packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/versions/Versions.page.tsx
+++ b/packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/versions/Versions.page.tsx
@@ -14,6 +14,7 @@ import {
   VersionCellID,
   VersionCellState,
 } from './VersionCells.component';
+import { isErrorResponse } from '@/utils/api/api';
 
 const SecretVersionsPage = () => {
   const navigate = useNavigate();
@@ -24,11 +25,6 @@ const SecretVersionsPage = () => {
   ]);
   const { domainId, secretPath } = useParams<SecretDashboardPageParams>();
   const secretPathDecoded = decodeSecretPath(secretPath);
-
-  const { data: versions, isPending, error, refetch } = useSecretVersions(
-    domainId,
-    secretPathDecoded,
-  );
 
   const columns = [
     {
@@ -54,16 +50,29 @@ const SecretVersionsPage = () => {
     },
   ];
 
+  const {
+    data,
+    error,
+    hasNextPage,
+    fetchNextPage,
+    sorting,
+    isPending,
+    setSorting,
+    refetch,
+  } = useSecretVersions({ domainId, path: secretPathDecoded });
+
   if (error)
     return (
       <ErrorBanner
-        error={error}
+        error={isErrorResponse(error) ? error.response : {}}
         onRedirectHome={() =>
           navigate(SECRET_MANAGER_ROUTES_URLS.secretManagerOnboarding)
         }
         onReloadPage={refetch}
       />
     );
+
+  const versions = data?.pages.flatMap((page) => page.data);
 
   return (
     <>
@@ -86,6 +95,10 @@ const SecretVersionsPage = () => {
         items={versions || []}
         totalItems={versions?.length}
         isLoading={isPending}
+        hasNextPage={hasNextPage}
+        onFetchNextPage={fetchNextPage}
+        sorting={sorting}
+        onSortChange={setSorting}
         contentAlignLeft
       />
       <Suspense>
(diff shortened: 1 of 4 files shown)


packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/versions/Versions.page.tsx
```diff
@@ -54,10 +49,25 @@ const SecretVersionsPage = () => {
...
+
   if (error)
     return (
       <ErrorBanner
-        error={error}
+        error={{
+          data: {
+            message: error.message,
+          },
+        }}
```
mathieumousnier: `useResourcesIcebergV2` screws up the error type, but you can use a typegard and get back on feet like I did here: https://github.com/ovh/manager/blob/0e0a7eb8113477a841502c2506a05338e26cdd20/packages/manager/apps/key-management-service/src/modules/secret-manager/pages/domains/Domains.page.tsx#L61

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/drawers/valueDrawer/ValueDrawer.page.tsx
```diff
@@ -40,14 +40,15 @@ const VersionSelector = ({
...
-    // isFetching forces to display the spinner when the version has changed
-    // otherwise when the ods-select component is refreshed, react breaks
+  const {
+    flattenData: versions,
+    isLoading,
+    isFetching,
+    error,
+  } = useSecretVersions({ domainId, path, pageSize: 100 });
+
+  if (isLoading || isFetching)
```
mathieumousnier: I wouldn't remove a comment that explains why we use isFetching.
If I see this with no comments, I remove isFetching, because it looks silly

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/drawers/valueDrawer/ValueDrawer.page.tsx
```diff
@@ -40,14 +40,15 @@ const VersionSelector = ({
...
-    domainId,
-    path,
-  );
 
-  if (isPending || isFetching)
-    // isFetching forces to display the spinner when the version has changed
-    // otherwise when the ods-select component is refreshed, react breaks
+  const {
+    flattenData: versions,
+    isLoading,
```
mathieumousnier: **isLoading -> isPending** 

useQuery default isLoading state must be `isPending` variable.
If you check for `isPending` and `isError`, there a guaranty than the data from the query is available, this results in a narrowed type for the data.
https://github.com/TanStack/query/discussions/6297

----

**flattenData** 

Also, I wouldn't touch this broken `flattenData` and it is better to make one for yourself like this: https://github.com/ovh/manager/blob/0e0a7eb8113477a841502c2506a05338e26cdd20/packages/manager/apps/key-management-service/src/modules/secret-manager/pages/domains/Domains.page.tsx#L31

_Why_: Because the flattenData from the hook is calculated in a useEffect, it adds a render between isPending=false and the avaibility of the data

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/versions/Versions.page.tsx
```diff
@@ -54,10 +49,25 @@ const SecretVersionsPage = () => {
     },
   ];
 
+  const {
+    flattenData: versions,
+    error,
+    hasNextPage,
+    fetchNextPage,
+    sorting,
+    isLoading,
```
mathieumousnier: Same comment with isLoading and flattenData

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/versions/Versions.page.tsx
rjamet-ovh: done

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/drawers/valueDrawer/ValueDrawer.page.tsx
rjamet-ovh: must have been crushed during rebase, put them back.

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/drawers/valueDrawer/ValueDrawer.page.tsx
rjamet-ovh: done

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/versions/Versions.page.tsx
rjamet-ovh: done

mathieumousnier (approved):

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/versions/Versions.page.tsx
```diff
@@ -54,39 +50,56 @@ const SecretVersionsPage = () => {
...
+  const {
+    data,
+    error,
+    hasNextPage,
+    fetchNextPage,
+    sorting,
+    isPending,
+    setSorting,
+    refetch,
+  } = useSecretVersions({ domainId, path: secretPathDecoded });
```
tibs245, line 62:
```suggestion
  const {
    data,
    flattenData: versions,
    error,
    hasNextPage,
    fetchNextPage,
    sorting,
    isPending,
    setSorting,
    refetch,
  } = useSecretVersions({ domainId, path: secretPathDecoded });
```

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/versions/Versions.page.tsx
```diff
@@ -54,39 +50,56 @@ const SecretVersionsPage = () => {
...
-        error={error}
+        error={isErrorResponse(error) ? error.response : {}}
         onRedirectHome={() =>
           navigate(SECRET_MANAGER_ROUTES_URLS.secretManagerOnboarding)
         }
         onReloadPage={refetch}
       />
     );
 
+  const versions = data?.pages.flatMap((page) => page.data);
```
tibs245, line 75:
```suggestion
```

rjamet-ovh: flattendata is not well implemented in `packages/manager-react-components/src/hooks/datagrid/useIcebergV2.tsx` and it causes a re-render due to the use of useEffect.

that's why we chose to not use it.

rjamet-ovh: there's a comment about it FYI
```
  // @TODO: deprecate flattenData and let user do it via select
```

packages/manager/apps/key-management-service/src/modules/secret-manager/pages/dashboard/versions/Versions.page.tsx
mathieumousnier: `flattenData` is bugged in MRC, it is processed in a `useEffect` which unsyncs `isPending` and `data`.

Normally with `isPending` goes from `false` to `true`, data goes from `undefined` to its value. But `flattenData` takes another render. This brought me to a "data index is undefined" error in another component.

So now we declare the flattenData in our components.

We didn't make the change in MRC because : 
- this change can break some tests
- flattenData will be deprecated soon

PR #18817, opened 2025-08-26
feat(manager-legacy-tools): delete defaly by default in msw

aboungnaseng-ovhcloud:
## Description

- Delete delay by default in msw


sefiani-ovh (approved):

yjaaouane (approved):

packages/manager-tools/manager-legacy-tools/test-utils/src/utils/msw.ts
```diff
@@ -1,4 +1,4 @@
-import { http, RequestHandler, HttpResponse, delay } from 'msw';
+import { http, RequestHandler, HttpResponse } from 'msw';
```
tibs245, line 1: I think it's may be good to keep delay but with time optionnal.

It's not a argument required and this insure the async behavior : https://mswjs.io/docs/api/delay

aboungnaseng-ovhcloud: Fixed !

aottr (approved):

PR #18823, opened 2025-08-27
feat(pci-project): keep project creation state on reload

FabienHenon:
ref: #MANAGER-19526

## Description

Ticket Reference: #...


packages/manager/apps/pci-project/src/components/payment/RegisterPaymentMethod.tsx
```diff
@@ -77,6 +75,25 @@ const RegisterPaymentMethod: React.FC<RegisterPaymentMethodProps> = ({
...
+      const method = availablePaymentMethods.find(
+        (m) => m.paymentType === preselectedPaymentType,
+      );
+      if (method) {
+        onHandlePaymentMethodChange(method);
+      }
+    }
+  }, [preselectedPaymentType, selectedPaymentMethod, availablePaymentMethods]);
+
+  if (isLoading || isFFLoading || !availablePaymentMethods || !features) {
```
StephaneRavet, line 98: ⚠️ **Loading state issue**: The component returns a spinner before the useEffect hook runs, which could prevent the preselection logic from executing properly. Consider moving the useEffect before the loading check or adding a flag to track if preselection has been attempted.

packages/manager/apps/pci-project/src/components/payment/RegisterPaymentMethod.tsx
```diff
@@ -77,6 +75,25 @@ const RegisterPaymentMethod: React.FC<RegisterPaymentMethodProps> = ({
...
+      !selectedPaymentMethod
+    ) {
+      const method = availablePaymentMethods.find(
+        (m) => m.paymentType === preselectedPaymentType,
+      );
+      if (method) {
+        onHandlePaymentMethodChange(method);
+      }
+    }
+  }, [preselectedPaymentType, selectedPaymentMethod, availablePaymentMethods]);
```
StephaneRavet: ⚠️ **Missing dependency**: The useEffect dependency array is missing `onHandlePaymentMethodChange`. This could cause stale closure issues. Consider using useCallback for the handler or add it to the dependencies.

packages/manager/apps/pci-project/src/pages/creation/Creation.page.tsx
```diff
@@ -42,8 +45,36 @@ export default function ProjectCreation() {
...
+  const cart = hasInitialCart ? existingCart : createdCart;
+
+  const {
+    mutate: orderProjectItem,
+    data: createdProjectItem,
+  } = useOrderProjectItem();
+  const { data: existingProjectItem } = useGetOrderProjectId(
+    hasInitialCart ? searchParams.get('cartId') || undefined : undefined,
+  );
+
```
StephaneRavet, line 75: 🔧 **Complex state logic**: The Creation page has grown quite complex with multiple state management paths (hasInitialCart, existing vs created items). Consider extracting this logic into a custom hook like `useCartState` to improve readability and testability.

packages/manager/apps/pci-project/src/pages/creation/Creation.page.tsx
```diff
@@ -60,17 +91,36 @@ export default function ProjectCreation() {
...
-        onSuccess: (createdCart) => {
+        onSuccess: (justCreatedCart) => {
           orderProjectItem({
-            cartId: createdCart.cartId,
+            cartId: justCreatedCart.cartId,
           });
+          setSearchParams({ cartId: justCreatedCart.cartId });
         },
       },
     );
```
StephaneRavet, line 122: ⚠️ **Missing error handling**: The useEffect hook that creates the cart doesn't handle potential errors. If cart creation fails, the user will be stuck without feedback. Consider adding error handling and user feedback.

packages/manager/apps/pci-project/src/components/payment/RegisterPaymentMethod.spec.tsx
```diff
@@ -506,6 +506,107 @@ describe('RegisterPaymentMethod', () => {
...
+          {...defaultProps}
+          preselectedPaymentType={TPaymentMethodType.PAYPAL}
+          handlePaymentMethodChange={handlePaymentMethodChange}
+        />,
+      );
+
+      expect(handlePaymentMethodChange).toHaveBeenCalledWith(
+        mockAvailablePaymentMethods[1],
+      );
+    });
```
StephaneRavet, line 607: ✅ **Good testing**: Excellent test coverage for the preselection functionality, including edge cases like waiting for data to load and preventing multiple preselections.

packages/manager/apps/pci-project/src/pages/creation/Creation.page.tsx
FabienHenon: Errors are handled by the ErrorBoundary.
Also, this component is not finished yet. The goal of this PR was only to add state recovery

packages/manager/apps/pci-project/src/pages/creation/Creation.page.tsx
FabienHenon: That's something I plan on doing as soon as we finish the project creation process. Doing it now will require more work as we will have to do it again later when the project creation component is finished

packages/manager/apps/pci-project/src/components/payment/RegisterPaymentMethod.tsx
FabienHenon: I don't think it prevents the hooks from running and calling the necessary callback

packages/manager/apps/pci-project/src/pages/creation/Creation.page.tsx
```diff
@@ -42,8 +45,36 @@ export default function ProjectCreation() {
...
 
-  const { mutate: createAndAssignCart, data: cart } = useCreateAndAssignCart();
-  const { mutate: orderProjectItem, data: projectItem } = useOrderProjectItem();
+  const {
+    mutate: createAndAssignCart,
+    data: createdCart,
+  } = useCreateAndAssignCart();
+  const { data: existingCart } = useGetCart(
+    hasInitialCart ? searchParams.get('cartId') || undefined : undefined,
+  );
```
seven-amid, line 54: `  const { data: existingCart } = useGetCart(searchParams.get('cartId'));
`
Think u can write it this way, since the getCart is enabled only when cartId is defined

packages/manager/apps/pci-project/src/pages/creation/Creation.page.tsx
```diff
@@ -42,8 +45,36 @@ export default function ProjectCreation() {
...
+
+  const {
+    form,
+    setForm,
+    isConfigFormValid,
+    isLoading: isLoadingConfigForm,
+  } = useConfigForm(
+    ovhSubsidiary,
+    hasInitialCart ? searchParams.get('cartId') || undefined : undefined,
```
seven-amid, line 63: same remark here

packages/manager/apps/pci-project/src/pages/creation/Creation.page.tsx
```diff
@@ -60,17 +91,36 @@ export default function ProjectCreation() {
   } = useStepper();
 
   useEffect(() => {
+    if (isLoadingConfigForm) {
+      return;
+    }
+
+    if (!isConfigFormValid()) {
+      return;
+    }
```
seven-amid, line 100:
```
if(isLoadingConfigForm || !isConfigFormValid()) {
 return;
}
```

seven-amid (approved):

packages/manager/apps/pci-project/src/pages/creation/Creation.page.tsx
FabienHenon: Yes but after cart creation its id is added to the query params, thus `searchParams.get('cartId')` becomes non undefined and the cart is retrieved (which is not what we want here)

packages/manager/apps/pci-project/src/pages/creation/Creation.page.tsx
FabienHenon: Same answer ;)

packages/manager/apps/pci-project/src/pages/creation/Creation.page.tsx
FabienHenon: We must not check if the config form is valid if it's loading. That's why I separated the ifs.
With your solution, I think, indeed that if the first part is true it won't execute the second part, but can we rely on this?

PR #18830, opened 2025-08-27
fix(communication): fix description position and help link

aottr:
## Description

Ticket Reference: #MANAGER-19466


packages/manager/apps/communication/public/translations/common/Messages_fr_FR.json
```diff
@@ -1,6 +1,7 @@
 {
   "title": "Mes communications",
-  "description": "Retrouvez l'ensemble des communications qu'OVH vous a envoyé, exceptées les demandes d'assistance.",
+  "description": "Retrouvez l'ensemble des communications qu'OVH vous a envoyé, exceptées les demandes d'assistance. Vous pouvez les consulter via la section: <anchor>Assistance</anchor>",
+  "assistance_link_label": "Assistance",
```
JacquesLarique, line 4: You could have achieved the same result with this (avoiding to have the "assistance_link_label"):
```suggestion
  "description": "Retrouvez l'ensemble des communications qu'OVH vous a envoyé, exceptées les demandes d'assistance. Vous pouvez les consulter via la section: <anchor label=\"Assistance\"></anchor>",
```

packages/manager/apps/communication/src/pages/communications/communications.spec.tsx
```diff
@@ -37,6 +39,11 @@ vi.mock('@ovh-ux/manager-react-components', () => ({
   }),
 }));
 
+vi.mock('react-i18next', () => ({
+  Trans: vi.fn(({ children }) => children),
+  useTranslation: vi.fn().mockReturnValue({ t: (key: string) => key }),
+}));
+
```
JacquesLarique: Since this is pretty transversal, you could add it in packages/manager/apps/communication/src/setupTests.ts

packages/manager/apps/communication/public/translations/common/Messages_fr_FR.json
aottr: I'll leave it as is. When migrating to ODS19 we'd have to remove the label attribute again. But it was a cool idea
*Label is still listed as optional attribute, but children will be the preferred way*

There is also the risk of the translation team not recognizing the label as a translatable field.

PR #18841, opened 2025-08-27
feat(pci-block-storage): add retyping for non-classic 3AZ volumes

aTurmo:
ref: #TAPC-4566

### DIFF
--- a/packages/manager/apps/pci-block-storage/src/api/hooks/useCatalogWithPreselection.spec.tsx
+++ b/packages/manager/apps/pci-block-storage/src/api/hooks/useCatalogWithPreselection.spec.tsx
@@ -44,21 +44,23 @@ const PROJECT_ID = '123';
 const VOLUME_ID = '1';
 
 describe('useCatalogWithPreselection', () => {
-  it.each(['region', 'localzone'])(
-    'returns the mapped volume data, the encryption type and not pending given queries are done and region is %s',
+  it.each(['region', 'localzone', 'region-3-az'])(
+    'returns the mapped volume data, the encryption type and pending to false, given the volume is not multi-attached and region is "%s"',
     async (non3azRegionType: TRegion['type']) => {
-      const volumeMock = {
+      const notMultiAttachedVolumeMock = {
         region: 'volume region',
         type: 'type',
       } as TAPIVolume;
 
       const mockCatalog = {
-        regions: [{ name: volumeMock.region, type: non3azRegionType }],
+        regions: [
+          { name: notMultiAttachedVolumeMock.region, type: non3azRegionType },
+        ],
       } as TVolumeCatalog;
       const mockMappedCatalog = [{}] as TVolumeRetypeModel[];
 
       vi.mocked(useQueries).mockReturnValue([
-        { data: volumeMock, isPending: false },
+        { data: notMultiAttachedVolumeMock, isPending: false },
         { data: mockCatalog, isPending: false },
       ]);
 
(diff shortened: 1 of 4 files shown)


packages/manager/apps/pci-block-storage/src/api/select/catalog.ts
```diff
@@ -21,17 +21,22 @@ export type TModelName = Readonly<{
 
 type TVolumeModelWithName<T extends TVolumeAddon> = T & TModelName;
 
+const is3az = (catalogRegions: TRegion[], region: string) =>
+  catalogRegions.find((r) => r.name === region).type === 'region-3-az';
```
Tsiorifamonjena:
```suggestion
  catalogRegions.find((r) => r.name === region)?.type === 'region-3-az';
```
find can return `undefined` no ?

packages/manager/apps/pci-block-storage/src/api/hooks/useCatalogWithPreselection.spec.tsx
```diff
@@ -80,10 +80,46 @@ describe('useCatalogWithPreselection', () => {
     },
   );
 
-  it('should return empty array and null encryption if region is region-3-az', () => {
+  it.each(['region', 'localzone'])(
+    'returns the mapped volume data, the encryption type and not pending given volume is multi attach and region is "%s"',
```
lolottetheclash: I think translation is not easy to read, is it possible to update "and not pending given volume is multi attach " with "and indicates that the given volume is not pending,that it is multi-attached," or something like this?

packages/manager/apps/pci-block-storage/src/api/select/catalog.spec.ts
```diff
@@ -159,3 +165,159 @@ describe('mapVolumeCatalog', () => {
...
+describe('mapRetypingVolumeCatalog', () => {
+  const catalogPriceFormatter = (price: number) => `price: ${price}`;
+  const translator = ((keyValue: string) => keyValue) as TFunction;
+
+  const catalog = {
+    regions: [region, region3AZ],
+    models: [classicModel(), model3AZ],
+  } as TVolumeCatalog;
+
+  it('should filter out element that dont have pricing in the input region', () => {
```
lolottetheclash, line 178: is it normal that two tests have the same title? (l.71 too)

packages/manager/apps/pci-block-storage/src/api/select/catalog.spec.ts
```diff
@@ -24,23 +27,24 @@ const region3AZ = {
...
-  pricings: [
-    {
-      price: 20,
-      regions: [region.name],
-      specs: {
-        name: 'classic spec name',
-        volume: {
-          iops: { level: 10 },
-          capacity: { max: 11 },
+const classicModel = (specName = 'classic spec name') =>
```
lolottetheclash: as you changed object to function, I think you should add a verb inside ;)

packages/manager/apps/pci-block-storage/src/api/select/catalog.spec.ts
aTurmo: Yes since it's in another describe, testing another method, and both do the same thing for that case

packages/manager/apps/pci-block-storage/src/api/hooks/useCatalogWithPreselection.spec.tsx
aTurmo: I reworded, tell me if it's better :)

Tsiorifamonjena (approved):

packages/manager/apps/pci-block-storage/src/api/hooks/useCatalogWithPreselection.spec.tsx
lolottetheclash: yes, much better thanks !

PR #18845, opened 2025-08-28
feat(nutanix): fetch server tasks with iceberg and adding filter stat…

sefiani-ovh:
…us in datagrid

ref: #MANAGER-17410

### DIFF
--- a/packages/manager/modules/bm-server-components/src/tasks/controller.js
+++ b/packages/manager/modules/bm-server-components/src/tasks/controller.js
@@ -1,25 +1,67 @@
 import isFunction from 'lodash/isFunction';
 import snakeCase from 'lodash/snakeCase';
 import toUpper from 'lodash/toUpper';
+import map from 'lodash/map';
 
 import Task from './tasks.class';
 
 export default class BmServerComponentsTasksController {
   /* @ngInject */
-  constructor($http) {
+  constructor(
+    $http,
+    ouiDatagridService,
+    $translate,
+    Server,
+    iceberg,
+    icebergUtils,
+  ) {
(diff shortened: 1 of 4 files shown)


qpavy (approved):

yjaaouane (approved):

packages/manager/modules/bm-server-components/src/tasks/controller.js
```diff
@@ -44,17 +66,48 @@ export default class BmServerComponentsTasksController {
...
+
+    return request
+      .execute(urlParams, isCacheDisabled)
+      .$promise.then(({ data, headers }) => ({
+        data,
+        meta: {
+          totalCount: headers['x-pagination-elements'],
+        },
+      }));
+  }
```
tibs245: May be useful to put it in service file

packages/manager/modules/bm-server-components/src/tasks/controller.js
```diff
@@ -1,25 +1,50 @@
...
+          ),
+        }));
+      })
+      .catch((err) => {
+        this.handleError(err);
+      })
+      .finally(() => {
+        this.loading = false;
+      });
+  }
```
tibs245: Possibility add this on function or may be in service fil ? I not understand on simple read why this function is now necessary

sefiani-ovh: i added filters on task operations so i need the list of task type

qpavy (approved):

tibs245 (approved):

yjaaouane (approved):

PR #18846, opened 2025-08-28
feat(pci-instances): add flag component

lolottetheclash:
This PR implements a new flag component which will be mandatory to create localization future card.

ref: #TAPC-4555

## Description

Ticket Reference: #...


lolottetheclash:
@aTurmo For now, it has been decided to do without a third-party library so as not to weigh down the bundle, and given that ODS worked this way.
Actually, I wasn't the one who calculated the coordinates ^^
In terms of maintainability, it's true that it would potentially be more reliable, but I think we'll stick with this... What do you think, @fredericvilcot ?

fredericvilcot:
> @aTurmo For now, it has been decided to do without a third-party library so as not to weigh down the bundle, and given that ODS worked this way. Actually, I wasn't the one who calculated the coordinates ^^ In terms of maintainability, it's true that it would potentially be more reliable, but I think we'll stick with this... What do you think, @fredericvilcot ?

Yes, this code is the exact replication of what have been done in ODS library, because unfortunately the component is not exported from their lib and is not intended to be in a next future.

But I agree that using absolute postions of a single SVG sprite is not very easy to maintain if we add to add or remove some flags, so it can do the trick for now but we should think about using a tiny external lib in a next future that will be compliant with manager whitelist :)

fredericvilcot (approved):

aTurmo (approved):

PR #18847, opened 2025-08-28
fix(pci-instances): avoid maximum update depth exceeded in render

Tsiorifamonjena:
## Description

Ticket Reference: #...


packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/useDashboard.ts
```diff
@@ -22,9 +29,33 @@ export const useDashboard = ({ region, instanceId }: TUseDashboardArgs) => {
...
   );
 };
+
+type TUpdateInstanceFromAllCacheArgs = {
+  projectId: string;
+  instance: TPartialInstance;
+  region?: string;
+};
+
+export const updateInstanceFromAllCaches = (
```
fredericvilcot: I'm not sure this method should be part of a hook dedicated to the dashboard view.

As it's a method that handles entities (acts like a store), I'll suggest to take the opportunity to create an `infra` folder under `src`, containing a `cache` folder:

```md
infra/
└── cache/
    └── instances.ts
```

And in `instances.ts`, we could export all update methods, like we could do with reducers with redux.

What do you think?

yarn.lock
```diff
@@ -7035,7 +7035,7 @@
     date-fns "^3.6.0"
     lodash.isequal "^4.5.0"
 
-"@ovh-ux/manager-pci-common@^0.18.3":
+"@ovh-ux/manager-pci-common@^0.18.2", "@ovh-ux/manager-pci-common@^0.18.3":
```
fredericvilcot: Could you please update version so that we have no changes in `yarn.lock` ?

PR #18860, opened 2025-08-28
feat(okms): add secret manager breadcrumb

rjamet-ovh:
ref: #MANAGER-18759

## Description

Add a dynamic breadcrumb that fulfill all requirements:
- define once for the whole app
- dynamically generated based on url
- we can choose which url part to add to the breadcrumb, in a white list fashion.
- we can customize a breadcrumb item (mostly useful for the label : translation, format part, data from API etc...)

Ticket Reference: #...


packages/manager/apps/key-management-service/src/modules/secret-manager/utils/locationParams.ts
```diff
@@ -0,0 +1,9 @@
+export const isLocationParamsDefined = (params: string[]) => {
+  const allDefined = params.every((param) => param !== undefined);
+
+  if (!allDefined) {
+    console.warn('One or more parameters are undefined');
```
mathieumousnier:
```suggestion
    console.warn('One or more url parameters are undefined');
```

mathieumousnier (approved):

packages/manager/apps/key-management-service/src/modules/secret-manager/components/breadCrumb/index.ts
pauldkn, line 1: **nitpick:** `breadcrumb` in folder name to be consistent

pauldkn (approved):

PR #18885, opened 2025-08-29
Feat/add localization card 

lolottetheclash:
## Description

Capture d’écran 2025-08-29 à 15 24 55

Ticket Reference: #TAPC-4555

### DIFF
--- a/packages/manager/apps/pci-instances/src/components/localizationCard/LocalizationCard.component.tsx
+++ b/packages/manager/apps/pci-instances/src/components/localizationCard/LocalizationCard.component.tsx
@@ -1,0 +1,70 @@
+import {
+  Badge,
+  Card,
+  Checkbox,
+  CheckboxCheckedChangeDetail,
+  CheckboxCheckedState,
+  CheckboxControl,
+  Text,
+} from '@ovhcloud/ods-react';
+import { useState } from 'react';
+import clsx from 'clsx';
+import { Flag } from '../flag/Flag';
+import { TCountryIsoCode } from '../flag/country-iso-code';
+import { TDeploymentMode } from '@/types/instance/common.type';
+
+export type TLocalizationCardProps = {
+  title: string;
+  subtitle: string;
+  countryCode: TCountryIsoCode;
+  deploymentMode: TDeploymentMode;
+};
+
+const getBadgeClassName = (mode: TDeploymentMode) => {
+  switch (mode) {
+    case '1AZ':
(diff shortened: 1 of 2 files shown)


Rom123Soleil (approved):

packages/manager/apps/pci-instances/src/components/localizationCard/LocalizationCard.component.tsx
```diff
@@ -0,0 +1,72 @@
...
+        <Flag isoCode={countryCode} />
+        <Text preset="heading-6" className="text-nowrap">
+          {title}
+        </Text>
+      </div>
+      <div className="flex justify-between gap-4">
+        <Text className="text-[--ods-color-heading] text-nowrap">
+          {subtitle}
+        </Text>
+        <Badge className={getBadgeClassName(deploymentMode)}>
```
lionel95200x, line 64: Maybe just move to a const  to avoid recomputing method to each render

```suggestion
       const badgeClassName = getBadgeClassName(deploymentMode)
```

packages/manager/apps/pci-instances/src/components/localizationCard/LocalizationCard.component.tsx
```diff
@@ -0,0 +1,72 @@
...
+};
+export const LocalizationCard = ({
+  title,
+  subtitle,
+  countryCode,
+  deploymentMode,
+}: TLocalizationCardProps) => {
+  const [isSelected, setIsSelected] = useState<CheckboxCheckedState>(false);
+
+  const handleCheckChange = useCallback(
```
aTurmo: Is it really necessary to put this in an useCallback ?
Personnally I never do this for simple handleChange like that but maybe I'm in the wrong.

packages/manager/apps/pci-instances/src/components/localizationCard/LocalizationCard.component.tsx
lolottetheclash: @lionel95200x the `const` won't change anything without a `useMemo`, but I think `useMemo` would be overkill here...

packages/manager/apps/pci-instances/src/components/localizationCard/LocalizationCard.component.tsx
lolottetheclash: yes you're right, it's useless here !

PR #18889, opened 2025-08-29
feat(communication): add contacts listing page

aottr:
## Description

Ticket Reference: #MANAGER-18679

## Additional Information

For Lot 2

### DIFF
--- a/packages/manager/apps/communication/public/translations/contacts/Messages_fr_FR.json
+++ b/packages/manager/apps/communication/public/translations/contacts/Messages_fr_FR.json
@@ -1,0 +1,14 @@
+{
+  "table_column_description": "Nom",
+  "table_column_type": "Type",
+  "table_column_contact_mean": "Titre de Contact",
+  "table_column_status": "Statut",
+  "table_action_edit": "Renommer",
+  "table_action_deactivate": "Désactiver",
+  "table_action_delete": "Supprimer",
+  "table_action_resend_verification_code": "Renvoyer le code de validation",
+  "default_badge_label": "Par défaut",
+  "add_contact_button": "Ajouter un contact",
+  "type_email": "Email",
+  "type_phone": "Téléphone"
+}
--- a/packages/manager/apps/communication/src/components/contactMeanStatus/contactMeanStatus.component.tsx
+++ b/packages/manager/apps/communication/src/components/contactMeanStatus/contactMeanStatus.component.tsx
@@ -1,0 +1,25 @@
+import { ODS_BADGE_COLOR } from '@ovhcloud/ods-components';
+import { OdsBadge } from '@ovhcloud/ods-components/react';
+import { useTranslation } from 'react-i18next';
+import { ContactMeanStatus } from '@/data/types/contact-mean.type';
+
+const STATUS_COLOR: Record<ContactMeanStatus, ODS_BADGE_COLOR> = {
+  ERROR: ODS_BADGE_COLOR.critical,
+  DISABLED: ODS_BADGE_COLOR.critical,
+  VALID: ODS_BADGE_COLOR.success,
+  TO_VALIDATE: ODS_BADGE_COLOR.information,
+};
+
+export default function ContactMeanStatusChip({
+  status,
+}: {
+  status: ContactMeanStatus;
+}) {
+  const { t } = useTranslation('common');
+  return (
+    <OdsBadge
+      color={STATUS_COLOR[status]}
+      label={t(`status_${status.toLowerCase()}`)}
+    />
+  );
+}
--- a/packages/manager/apps/communication/src/pages/contacts/Contacts.page.tsx
+++ b/packages/manager/apps/communication/src/pages/contacts/Contacts.page.tsx
@@ -1,0 +1,173 @@
+import { useTranslation } from 'react-i18next';
+import { OdsBadge, OdsMessage } from '@ovhcloud/ods-components/react';
+import {
+  ActionMenu,
+  ActionMenuItem,
+  Datagrid,
+  DatagridColumn,
+  DataGridTextCell,
+  ManagerButton,
+  useResourcesIcebergV2,
+} from '@ovh-ux/manager-react-components';
+import { ODS_BUTTON_VARIANT } from '@ovhcloud/ods-components';
+import { useMemo } from 'react';
+import { useAuthorization } from '@/hooks';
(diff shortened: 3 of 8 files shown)


packages/manager/apps/communication/public/translations/contacts/Messages_fr_FR.json
```diff
@@ -0,0 +1,14 @@
+{
+  "table_column_description": "Nom",
+  "table_column_type": "Type",
+  "table_column_contact_mean": "Titre de Contact",
+  "table_column_status": "Statut",
+  "table_action_edit": "Renommer",
+  "table_action_deactivate": "Desactiver",
```
JacquesLarique:
```suggestion
  "table_action_deactivate": "Désactiver",
```

packages/manager/apps/communication/src/pages/contacts/Contacts.page.tsx
```diff
@@ -0,0 +1,167 @@
...
+
+function ContactsPage() {
+  const { t } = useTranslation('contacts');
+  const { t: tCommon } = useTranslation('common');
+  const { data: accountUrn } = useAccountUrn();
+  const { isAuthorized, isLoading: isLoadingAuthorization } = useAuthorization([
+    'account:apiovh:notification/contactMean/get',
+  ]);
+
+  const columns: DatagridColumn<ContactMean>[] = [
```
oalkabouss, line 68: You could wrap the columns definition in a useMemo to avoid unnecessary recalculations on each render

`const columns = useMemo(() => [
], [t, tCommon]);`

packages/manager/apps/communication/src/pages/contacts/Contacts.page.tsx
```diff
@@ -0,0 +1,167 @@
...
+      label: t('table_column_type'),
+      cell: ({ type }) => (
+        <DataGridTextCell>{t(`type_${type.toLowerCase()}`)}</DataGridTextCell>
+      ),
+    },
+    {
+      id: 'contact-mean',
+      label: t('table_column_contact_mean'),
+      cell: (contactMean) => (
+        <DataGridTextCell>{contactMean.email || '-'}</DataGridTextCell>
```
JacquesLarique, line 97: Is the contact mean property named `email` even if the type is not `email` ?

packages/manager/apps/communication/src/components/contactMeanStatus/contactMeanStatus.component.tsx
```diff
@@ -0,0 +1,25 @@
...
+import { OdsBadge } from '@ovhcloud/ods-components/react';
+import { useTranslation } from 'react-i18next';
+import { ContactMeanStatus } from '@/data/types/contact-mean.type';
+
+const STATUS_COLOR: Record<ContactMeanStatus, ODS_BADGE_COLOR> = {
+  ERROR: ODS_BADGE_COLOR.critical,
+  DISABLED: ODS_BADGE_COLOR.critical,
+  VALID: ODS_BADGE_COLOR.success,
+  TO_VALIDATE: ODS_BADGE_COLOR.information,
+};
```
JacquesLarique, line 11: This could be extracted in a constant file

packages/manager/apps/communication/src/pages/contacts/Contacts.page.tsx
```diff
@@ -0,0 +1,167 @@
...
+      id: 1,
+      label: t('table_action_edit'),
+      onClick: () => {},
+    },
+    contactMean.type !== 'EMAIL' &&
+      contactMean.status === 'TO_VALIDATE' && {
+        id: 2,
+        label: t('table_action_resend_verification_code'),
+        onClick: () => {},
+      },
```
oalkabouss: https://github.com/ovh/manager/pull/18889/files/4ed956eb2e2dda840facc153e8fb2305dc4ec2a2#r2349491908

packages/manager/apps/communication/src/pages/contacts/Contacts.page.tsx
aottr: It adds overhead to a pretty much static array assignment without any computation. The real performance is minimal while we're adding complexity. It's a big maybe from my side.

packages/manager/apps/communication/src/pages/contacts/Contacts.page.tsx
aottr: email is only filled if the type is email. At the same time there is no other value field currently available. Once I know the other field names I planned to make it a custom component that dynamically fills the value depending on the type ^^

packages/manager/apps/communication/src/pages/contacts/Contacts.page.tsx
aottr: Here I can understand the request more since it's a sub-component in the page file, but the dependency array would be including all properties so it would re-render either way with minimal computational cost. 

tldr: why not, but I don't think the gain is high ^^

packages/manager/apps/communication/src/components/contactMeanStatus/contactMeanStatus.component.tsx
aottr: could be, but I don't see a benefit of creating an export for something that is only used inside the component. The constant file should be used for constants that are used on different places and therefore exported, or for central configuration. Here it's bound to the component directly.

packages/manager/apps/communication/src/pages/contacts/Contacts.page.tsx
oalkabouss: Makes sense 👍 I just wanted to ensure columns has a stable reference in case Datagrid depends on it

packages/manager/apps/communication/src/pages/contacts/Contacts.page.tsx
JacquesLarique: I'm not convinced this is a good pattern, since they will introduce a new property each time they introduce a new contact mean.
Anyway this feedback is not really for you then, but it could be nice to share this with API team

aottr: yeah it should have a generic value field. It's overseeable since there is only SMS, email planned (for now) but once there are multiple options, it gets very messy.

JacquesLarique (approved):

oalkabouss (approved):

PR #18908, opened 2025-09-02
feat(ips): migrating an unblock antihack action to the ip listing

aderghamov:
ref: #MANAGER-18914

## Description

This PR introduces a new unblock anti-hack modal available from the revamped ip listing page

Ticket Reference: #MANAGER-18914


packages/manager/apps/ips/src/data/hooks/dateFormat/useDateFormat.tsx
```diff
@@ -0,0 +1,16 @@
...
+
+const useDateFormat = ({ options }: Props): Intl.DateTimeFormat => {
+  const context = useContext(ShellContext);
+  const { environment } = context;
+  const userLanguage = environment.getUserLocale().replace('_', '-');
+
+  return new Intl.DateTimeFormat(userLanguage, options);
+};
+
+export default useDateFormat;
```
chipp972: There is already `useFormatDate` from `'@ovh-ux/manager-react-components'`
It doesn't fit your needs ?

packages/manager/apps/ips/src/data/hooks/index.ts
```diff
@@ -1,4 +1,5 @@
 export * from './ip';
+export * from './dateFormat/useDateFormat';
```
chipp972: If we need to keep this hook, it should be in utils folder, not data

packages/manager/apps/ips/src/data/hooks/dateFormat/useDateFormat.tsx
aderghamov: My bad, I did not know about this hook.
It looks like it fits perfectly my need.

packages/manager/apps/ips/src/data/hooks/index.ts
aderghamov: no need to keep it.

PR #18915, opened 2025-09-02
feat(okms): add back to domains button

rjamet-ovh:
ref: #MANAGER-18851

## Description

Ticket Reference: #...


packages/manager/apps/key-management-service/src/modules/secret-manager/hooks/useShouldDisplayBackToDomainListButton.tsx
```diff
@@ -0,0 +1,26 @@
...
+  const currentRegionId = useCurrentRegion(OkmsList || []);
+
+  // filter all domain by current domain's
+  const regionOkmsList =
+    OkmsList && currentRegionId
+      ? filterDomainsByRegion(OkmsList, currentRegionId)
+      : [];
+
+  // return false if there is an error.
+  // useRegionSelector will send an error notification
```
mathieumousnier: Nitpick but as useRegionSelector is not called here, I won't refer to it in this place.

I feel like this is acceptable to consider that while the data is `fetching` or in `error` state, the hooks returns `false`. (Some errors can sometimes just be quiet, it's a back button after all)
And as the `loading` state is not managed, I would consider the `error` state in the same way.

This would mean either not exporting loading and error, and just mentioning them in a comment like "loading and error state both return false".
Or exporting them and write the condition explicitly.

rjamet-ovh: done

PR #18917, opened 2025-09-02
fix(pci-instances): display ellipsis when volume name is too long

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-4857


aTurmo (approved):

lolottetheclash (commented):
don't we need a `tooltip` to let the user see the entire volume title?

Tsiorifamonjena:
> don't we need a `tooltip` to let the user see the entire volume title?

Good point 👏🏽

I do it in a separate [PR](https://github.com/ovh/manager/pull/18967). I think it would be cool to talk to the designers too.

PR #18924, opened 2025-09-02
fix(pci-instances): prevent submit form on hiting Enter key

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-4869


Tsiorifamonjena:
> I'm not sure that `FormField`should contain buttons... It should only contain an input and a label, no?

@lolottetheclash you are right, I let the component to be wrapped by a form. I let you to see the diff ;)

Tsiorifamonjena:
I took advantage of this PR to remove the `EditableContent` component which is no longer used

lolottetheclash:
> > I'm not sure that `FormField`should contain buttons... It should only contain an input and a label, no?
> 
> @lolottetheclash you are right, I let the component to be wrapped by a form. I let you to see the diff ;)

@Tsiorifamonjena I think that if you don't want the enter to dispatch the submit event and control submission, maybe you can just replace the `form` by a `div`, what do you think?

Tsiorifamonjena:
> > > I'm not sure that `FormField`should contain buttons... It should only contain an input and a label, no?
> > 
> > 
> > @lolottetheclash you are right, I let the component to be wrapped by a form. I let you to see the diff ;)
> 
> @Tsiorifamonjena I think that if you don't want the enter to dispatch the submit event and control submission, maybe you can just replace the `form` by a `div`, what do you think?

@lolottetheclash Maybe I'm wrong but IMO it's a good practice to let it wrap by form for accessibility. And now when user type Enter, it submit and update name without reloading all page.

lolottetheclash (approved):

packages/manager/apps/pci-instances/src/components/input/InputCancellable.component.tsx
```diff
@@ -12,16 +14,35 @@ export const InputCancellable: FC<TInputCancellableProps> = ({
...
-    </Button>
-    <Button aria-label="Submit" variant="ghost" size="sm" onClick={onSubmit}>
-      <Icon name="check" className="text-xl" />
-    </Button>
-  </form>
-);
+}) => {
+  const { t } = useTranslation(NAMESPACES.ACTIONS);
+
+  const handleSubmit = (e: FormEvent) => {
```
aTurmo:
```suggestion
  const handleSubmit = (event: FormEvent) => {
```

PR #18949, opened 2025-09-03
fix(ips): improving the ip-reverse error display

aderghamov:
ref: #MANAGER-18757

## Description

This PR replaces the default error message display for IPReverse update by a dedicated error display.

Ticket Reference: #MANAGER-18757


chipp972:
The 2 tests broke though

PR #18956, opened 2025-09-03
chore(vrack-services): completing the SWC migration by removing the vitest dependency

aderghamov:
ref: #MANAGER-18540

## Description

Ticket Reference: #...


chipp972 (dismissed):

helabenkhalfallah (dismissed):

helabenkhalfallah (approved):

helabenkhalfallah (dismissed):

chipp972 (dismissed):

helabenkhalfallah (approved):

.eslintrc.js
anooparveti, line 1: **question:** What is the need to exclude test-utils from linting?

aderghamov: To me, `test-utils` is just like any other test related files/directories of this list.
The current `test-utils` directory has setupIntegrationTests that uses functions that return errors from the linter.

I will investigate to find a work-around

aderghamov: I have ended up with another solution by refactoring the content of the test-utils directory.
This should work as expected.

helabenkhalfallah (approved):

PR #18964, opened 2025-09-03
fix(pci-instances): dashboard - rework warning modal text

aTurmo:
ref: #TAPC-4860

Before:
image

After:
image

### DIFF
--- a/packages/manager/apps/pci-instances/src/pages/instances/action/components/modal/ActionModalContent.component.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/action/components/modal/ActionModalContent.component.tsx
@@ -1,14 +1,13 @@
 import { FC, useMemo } from 'react';
 import {
-  OsdsMessage,
-  OsdsSkeleton,
-  OsdsText,
-} from '@ovhcloud/ods-components/react';
-import { ODS_MESSAGE_TYPE, ODS_TEXT_SIZE } from '@ovhcloud/ods-components';
-import {
-  ODS_THEME_COLOR_INTENT,
-  ODS_THEME_TYPOGRAPHY_LEVEL,
-} from '@ovhcloud/ods-common-theming';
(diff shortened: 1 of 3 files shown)


packages/manager/apps/pci-instances/src/pages/instances/action/components/modal/ActionModalContent.component.tsx
```diff
@@ -86,40 +85,32 @@ export const ActionModalContent: FC<TActionModalProps> = ({
     return null;
   }, [type, t]);
 
+  const [displayWarning, setDisplayWarning] = useState(true);
```
lolottetheclash: do we really need an extra state here? can't we deal with a dismissable property on the `Message` or something like this?

aTurmo: Sadly there is no native way in ODS 19 Message to remove it when clicking on the cross.
The "dismissible" prop is used only to display (or not) the cross.
And since the figmas show the cross i'm gessing we have to remove the message when clicking on the cross.

aTurmo: After discussing with Léo, you're right and in fact it's better to not display the cross

lolottetheclash: great, thanks for your feedback ;)

aTurmo: It's me that thanks you 😊

aTurmo: done

lolottetheclash (approved):
great !

PR #18969, opened 2025-09-04
feat(billing): rework commitment page

MaximeBajeux:
ref: #MANAGER-17576

## Description

Ticket Reference: #...


packages/manager/modules/billing-components/src/components/commitment/commitment.controller.js
```diff
@@ -32,15 +33,18 @@ export default class {
       duration: null,
       engagement: null,
     };
-    return this.$q
+    this.isPaymentStepLoading = true;
+    this.$q
```
JacquesLarique, line 54: We should first fetch the default payment method and then:
- if it fails fill `this.error` with the API error (so that we display it to the user), maybe check we with PO to add a redirect link to payment registration page
- if it succeed then we can load the service and its options then get commitment and pricing data

packages/manager/modules/billing-components/src/components/commitment/commitment.controller.js
```diff
@@ -83,10 +89,70 @@ export default class {
...
+      const { savings } = this.duration;
+      const totalSavings = savedAmount + (savings?.value || 0);
+      this.upfrontSavings = {
+        amountSaved: this.getPriceAsText(
+          totalSavings,
+          upfront.pricing.price.currencyCode,
+        ),
+        amountToPay: upfront.totalPrice.text,
+      };
+    }
```
JacquesLarique, line 156: This should be moved to a dedicated service (using the `Price` class instead of `getPriceAsText` to avoid duplication)

JacquesLarique (dismissed):

sachinrameshn (dismissed):

JacquesLarique (approved):

sachinrameshn (approved):

packages/manager/modules/models/service/Pricing.class.js
```diff
@@ -25,6 +25,11 @@ export default class Pricing {
     }).format(price);
   }
 
+  static roundToTwo(num) {
+    // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/toFixed
+    return Number.parseFloat(num).toFixed(2);
+  }
```
anooparveti: **question:** What will be the type of the `num`, string or number?? If it's string better to specify it. Otherwise you can directly use `.toFixed(2)` and avoid parsing the number.

PR #18970, opened 2025-09-04
feat(*): include migration status for particular apps

helabenkhalfallah:
## Description

The goal of this PR is to include special apps to migration cli report:

```
Hello everyone !
Regarding this documentation : https://[internal_url]/digital-tools-manager/manager/manager-migration-monitor-cli/manager/migration-status-reports/migration-status-full-report.html our project web-domains isn't mark as done for route migration.

But if you check this file : https://github.com/ovh/manager/blob/master/packages/manager/apps/web-domains/src/alldoms/routes/routes.tsx you can see we use the last route manager version.

I think it is because our path is that way : /web-domains/src/alldoms/routes


Hello team, "quick question": how is it determined if an app has migrated its routes definition ? (https://[internal_url]/digital-tools-manager/manager/manager-migration-monitor-cli/manager/migration-status-reports/migration-status-full-report.html)
On container, which is marked as TODO, the routes are already defined as components, but they are not defined in the common way
```

Ticket Reference: #MANAGER-19013

### DIFF
--- a/packages/manager-tools/manager-cli/utils/AppUtils.mjs
+++ b/packages/manager-tools/manager-cli/utils/AppUtils.mjs
@@ -44,6 +44,11 @@ export const resolveRoutePath = (appName, { verbose = false } = {}) => {
     join(applicationsBasePath, appName, 'src/routes/routes.tsx'),
     join(applicationsBasePath, appName, 'src/routes.tsx'),
     join(applicationsBasePath, appName, 'src/routes/index.tsx'),
+
+    // special projects cases
+    join(applicationsBasePath, appName, 'src/alldoms/routes/routes.tsx'),
+    join(applicationsBasePath, appName, 'src/core/routing/redirections.tsx'),
+    join(applicationsBasePath, appName, 'src/core/routing/iframe-app-router.tsx'),
   ];
 
   for (const path of candidates) {


aboungnaseng-ovhcloud (approved):

packages/manager-tools/manager-cli/utils/AppUtils.mjs
```diff
@@ -44,6 +44,11 @@ export const resolveRoutePath = (appName, { verbose = false } = {}) => {
     join(applicationsBasePath, appName, 'src/routes/routes.tsx'),
     join(applicationsBasePath, appName, 'src/routes.tsx'),
     join(applicationsBasePath, appName, 'src/routes/index.tsx'),
+
+    // special projects cases
+    join(applicationsBasePath, appName, 'src/alldoms/routes/routes.tsx'),
```
anooparveti, line 49: **nitpick:** Instead of hard-coding the path of a specific application, i think we can accept it as a param and use it to make CLI more scalable

helabenkhalfallah: I love so much the idea! do you agree that I create another ticket to review this part to have the possibility to use path as args in addition to default settings?

helabenkhalfallah: Jira ticket created!

anooparveti (approved):

PR #18976, opened 2025-09-04
feat(okms): add CAS parameter support for creating secret versions

mathieumousnier:
## Description

Ticket Reference: #MANAGER-19608

### DIFF
--- a/packages/manager/apps/key-management-service/src/modules/secret-manager/data/api/secretVersions.ts
+++ b/packages/manager/apps/key-management-service/src/modules/secret-manager/data/api/secretVersions.ts
@@ -38,16 +38,27 @@ export type CreateSecretVersionParams = {
   okmsId: string;
   path: string;
   data: CreateSecretVersionBody;
+  cas?: number;
 };
 
 export const createSecretVersion = async ({
   okmsId,
   path,
   data,
+  cas,
 }: CreateSecretVersionParams) => {
+  const queryParams = new URLSearchParams();
+  if (cas) {
+    queryParams.set('cas', cas.toString());
+  }
+  const queryString = queryParams.toString();
+  const url = `okms/resource/${okmsId}/secret/${encodeURIComponent(
+    path,
+  )}/version${queryString ? `?${queryString}` : ''}`;
+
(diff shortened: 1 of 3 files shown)


packages/manager/apps/key-management-service/src/modules/secret-manager/data/api/secretVersions.ts
```diff
@@ -38,18 +38,29 @@ export type CreateSecretVersionParams = {
...
-    data,
-  });
+  >(
+    `okms/resource/${okmsId}/secret/${encodeURIComponent(
+      path,
+    )}/version?${queryParams.toString()}`,
+    {
+      data,
+    },
+  );
```
rjamet-ovh: with this implementation, we have a trailing `?` when we don't need cas queryParam|

you can do something like that
```
  const searchParams = new URLSearchParams();
  let queryParams = '';
  if (cas) {
    searchParams.set('cas', cas.toString());
    queryParams = `?${searchParams.toString()}`;
  }

  const { data: response } = await apiClient.v2.post<
    CreateSecretVersionResponse
  >(
    `okms/resource/${okmsId}/secret/${encodeURIComponent(
      path,
    )}/version${queryParams}`,
    {
      data,
    },
  );
  ```

rjamet-ovh:
can you update `packages/manager/apps/key-management-service/src/modules/secret-manager/pages/drawers/createVersionDrawer/CreateVersionDrawer.page.spec.tsx` to test this use-case?

packages/manager/apps/key-management-service/src/modules/secret-manager/data/api/secretVersions.ts
mathieumousnier: A url ending with a "?" a is a valid url so it does not harm anything to keep it.
But because I'm extremely nice, I changed it :)

mathieumousnier:
I managed to add the most important case to the integration test (add cas param to the request).
Testing the other case (not add cas param if the secret does not require it) would mean some complications for a small added value so I didn't do it

sefiani-ovh (approved):

PR #18986, opened 2025-09-04
fix(ips): skipping the organisation selection when ordering non-US IP as an US user

aderghamov:
ref: #MANAGER-19460

## Description

This PR introduces a new condition for displaying the organisation select step when ordering an IPv4 service.
The step should not be displayed for US users that are selection non-US located (parking or vrack) services.

Ticket Reference: #MANAGER-19460

### DIFF
--- a/packages/manager/apps/ips/src/test-utils/render-test.tsx
+++ b/packages/manager/apps/ips/src/test-utils/render-test.tsx
@@ -19,6 +19,7 @@ import {
   toMswHandlers,
   WAIT_FOR_DEFAULT_OPTIONS,
 } from '@ovh-ux/manager-core-test-utils';
+import { Subsidiary } from '@ovh-ux/manager-config';
 import { translations, labels } from './test-i18n';
 import { TestApp } from './TestApp';
 import {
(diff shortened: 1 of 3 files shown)


chipp972:
Could you add test cases in the order.spec.ts file since you added a complex domain rule ?

aderghamov:
> Could you add test cases in the order.spec.ts file since you added a complex domain rule ?

I will try

aderghamov:
> Could you add test cases in the order.spec.ts file since you added a complex domain rule ?

Done!

chipp972 (approved):

packages/manager/apps/ips/src/test-utils/render-test.tsx
```diff
@@ -88,6 +94,9 @@ export const renderTest = async ({
     context = await initShellContext(APP_NAME);
   }
 
+  context.environment.getUser().ovhSubsidiary =
+    mockParams.ovhSubsidiary ?? 'FR';
```
chipp972, line 104: Very nice 👍🏾

PR #19040, opened 2025-09-08
fix(*): fixing the iceberg pagination filter processing

aderghamov:
ref: #MANAGER-19352

## Description

Ticket Reference: #...


packages/manager/core/api/src/iceberg.ts
```diff
@@ -93,8 +93,10 @@ export const buildHeaders = () => {
         const filtersJoin = filters
           .filter(({ type }) => type !== FilterTypeCategories.Tags)
           .map(
-            ({ comparator, key, value }) =>
-              `${encodeURIComponent(key)}:${icebergFilter(comparator, String(value || ''))}`,
+            ({ comparator, key, value }) => {
+              const correctedValue = value instanceof Array ? value : String(value || '');
```
helabenkhalfallah: We should be careful with the use of instanceof it can throw exception https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/instanceof#exceptions

[TypeError](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/TypeError)
Thrown if constructor is not an object. If constructor doesn't have a [[Symbol.hasInstance]()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Symbol/hasInstance) method, it must also be a function.

I think it'll be more safe to use typeof instead:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof 

Array and Object give typeof object, string gives string.

helabenkhalfallah (approved):

packages/manager/core/api/src/iceberg.ts
```diff
@@ -92,10 +92,10 @@ export const buildHeaders = () => {
...
         const filtersJoin = filters
           .filter(({ type }) => type !== FilterTypeCategories.Tags)
-          .map(
-            ({ comparator, key, value }) =>
-              `${encodeURIComponent(key)}:${icebergFilter(comparator, String(value || ''))}`,
-          )
+          .map(({ comparator, key, value }) => {
+            const correctedValue = typeof value === 'object' ? value : String(value || '');
+            return `${encodeURIComponent(key)}:${icebergFilter(comparator, correctedValue)}`;
+          })
```
anooparveti, line 98: Actually this is not the right way to fix the issue. According to Iceberg documentation, the filter with `isIn` needs to be `filedName:in=val1,val2`. So the actual fix will result in an Iceberg API call failure.

anooparveti (changes requested):

helabenkhalfallah (approved):

PR #19042, opened 2025-09-08
Add Credit card payment integration

FabienHenon:
## Description

Ticket Reference: #MANAGER-18802

### DIFF
--- a/packages/manager/apps/pci-project/src/components/payment/RegisterPaymentMethod.tsx
+++ b/packages/manager/apps/pci-project/src/components/payment/RegisterPaymentMethod.tsx
@@ -19,17 +19,32 @@ import { useFilteredAvailablePaymentMethods } from '@/data/hooks/payment/useFilt
 import PaymentIcon from './PaymentIcon';
 import {
   TAvailablePaymentMethod,
+  TPaymentMethodRegisterRef,
   TPaymentMethodType,
 } from '@/data/types/payment/payment-method.type';
 import SetAsDefaultPaymentMethod from './SetAsDefaultPaymentMethod';
 import ExplanationTexts from './ExplanationTexts';
 import SepaInformationModal from './SepaInformationModal';
+import PaymentMethodIntegration from './integrations/PaymentMethodIntegration';
 
 export type RegisterPaymentMethodProps = {
   eligibility: TEligibility;
   handlePaymentMethodChange?: (method: TAvailablePaymentMethod) => void;
   handleSetAsDefaultChange?: (value: boolean) => void;
   preselectedPaymentType?: string | null;
+  paymentHandler: React.Ref<TPaymentMethodRegisterRef>;
+  cartId: string;
+  itemId: number;
+  handleCustomSubmitButton?: (btn: string | JSX.Element) => void;
+  onPaymentSubmit: ({
+    paymentMethodId,
+    skipRegistration,
+  }: {
+    paymentMethodId?: number;
+    skipRegistration?: boolean;
+  }) => Promise<unknown>;
+  onPaymentError: (err: string | undefined) => void;
+  handleValidityChange: (isValid: boolean) => void;
 };
 
 const RegisterPaymentMethod: React.FC<RegisterPaymentMethodProps> = ({
@@ -37,6 +52,13 @@ const RegisterPaymentMethod: React.FC<RegisterPaymentMethodProps> = ({
   handlePaymentMethodChange = () => {},
   handleSetAsDefaultChange = () => {},
   preselectedPaymentType,
+  paymentHandler,
+  cartId,
+  itemId,
+  handleCustomSubmitButton,
+  onPaymentSubmit,
+  onPaymentError,
+  handleValidityChange,
 }) => {
   const { t } = useTranslation([
     'payment/add',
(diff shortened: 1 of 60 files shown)


packages/manager/apps/pci-project/src/components/payment/integrations/adyen/hooks/useAdyenSDK.ts
```diff
@@ -0,0 +1,244 @@
...
+import { ShellContext } from '@ovh-ux/manager-react-shell-client';
+import {
+  ADYEN_CONFIG,
+  ADYEN_RESULT_CODE,
+  PAYMENT_METHOD_BRANDS,
+} from '../constants';
+import { TAvailablePaymentMethod } from '@/data/types/payment/payment-method.type';
+import { TPaymentFeaturesState } from '@/data/hooks/payment/usePaymentFeatureAvailabilities';
+import { convertToBase64, parseFormSessionId } from '../utils';
+import '../styles/index.less';
```
StephaneRavet, line 22: Importing CSS inside a React hook is problematic. Here are the side effects of `import '../styles/index.less'` inside a hook:

1. **Global style leakage**
   CSS imports are global. Once the hook’s module is loaded, styles are injected and remain active across the app—even after the component unmounts. This can unintentionally affect other screens.

2. **Unpredictable load order & FOUC**
   With code-splitting, the CSS loads when the hook’s module is first imported. That timing can cause Flash Of Unstyled Content or unexpected overrides vs. ODS/styles loaded earlier.

3. **Tight coupling of logic and presentation**
   Hooks should be pure (logic only). Importing CSS in a hook mixes concerns, reduces reusability, and makes the hook harder to use in contexts where those styles are not desired.

4. **Testing & DX friction (clearer explanation)**
Unit tests (Jest/JSDOM): importing .css/.less inside a hook means every test that imports the hook must also handle CSS. In Node/JSDOM, style files aren’t understood by default, so tests either fail (Cannot find module '*.less') or need extra config/mocks. Even with mocks, you’re paying for a side-effect you don’t need when you only want to test logic.

5. **Bundle & perf side effects**
   The hook’s chunk may pull CSS along even when only the logic is needed, complicating preloading and potentially bloating bundles.

**Suggested fix (keep “plug-and-play” without side effects):**

* Move CSS imports to the **UI layer** (the component/screen that renders the Adyen container), or
* Create a tiny **Styles component** that you render alongside the integration, or
* Make styles opt-in via a parameter (e.g., `autoStyles`) and load them in the component, not the hook.

**Minimal diff (remove CSS from hook):**

```diff
// useAdyenSDK.ts
- import '../styles/index.less';
+ // Keep the hook pure; import styles in the UI component
```

**Example Styles component:**

```tsx
// AdyenStyles.tsx
import '@adyen/adyen-web/dist/adyen.css';
import '../styles/index.less';
export const AdyenStyles = () => null;
```

**Usage:**

```tsx
// CreditCardPaymentMethodIntegration.tsx
import { AdyenStyles } from './AdyenStyles';

export function CreditCardPaymentMethodIntegration() {
  return (
    <>
      <AdyenStyles />
      {/* component using useAdyenSDK */}
    </>
  );
}
```

StephaneRavet (approved):

packages/manager/apps/pci-project/src/components/payment/RegisterPaymentMethod.tsx
```diff
@@ -19,24 +19,46 @@ import { useFilteredAvailablePaymentMethods } from '@/data/hooks/payment/useFilt
...
+    skipRegistration,
+  }: {
+    paymentMethodId?: number;
+    skipRegistration?: boolean;
+  }) => Promise<unknown>;
+  onPaymentError: (err: string | undefined) => void;
+  handleValidityChange: (isValid: boolean) => void;
 };
 
 const RegisterPaymentMethod: React.FC<RegisterPaymentMethodProps> = ({
```
StephaneRavet, line 50: **Many teams avoid `React.FC`** because it adds little value and historically caused confusion (especially around `children`). CRA even removed it from its TS template, and most cheatsheets rarely recommend it. (https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/)

### Why it’s commonly avoided

1. **Implicit `children` (historical confusion)**
   Before React 18, `React.FC` implicitly added a `children` prop. In React 18, **that’s no longer true**—if your component accepts children, **declare them explicitly**. This reduces the main reason people used `React.FC`. (https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/) (https://github.com/DefinitelyTyped/DefinitelyTyped/discussions/59992) (https://stackoverflow.com/questions/71788254/react-18-typescript-children-fc)

2. **Less flexible for advanced cases**
   Generics, `forwardRef`, HOCs, etc. are **simpler to type** without `React.FC` by typing the function’s props directly. (https://blog.variant.no/a-better-way-to-type-react-components-9a6460a1d4b7)

3. **Few concrete benefits**
   TS already **infers the return type**, and long-standing issues like `defaultProps` on functions are better solved with **default values in the parameter list**. (https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/)

> Note: Some recent posts argue that “`React.FC` is fine now” (TS 5.1/React 18 fixed earlier rough edges). It’s **usable**, but the common recommendation remains: **type props directly** instead of annotating with `React.FC`. (https://www.totaltypescript.com/you-can-stop-hating-react-fc) (https://react-typescript-cheatsheet.netlify.app/docs/basic/getting-started/function_components/)

### Practical, modern recommendation

**Avoid:**

```ts
const RedirectPaymentMethodIntegration: React.FC<RedirectPaymentMethodIntegrationProps> = (props) => {
  return <div />;
};
```

**Prefer:**

```ts
type RedirectPaymentMethodIntegrationProps = {
  // …your props
  children?: React.ReactNode; // add only if needed
};

export function RedirectPaymentMethodIntegration(props: RedirectPaymentMethodIntegrationProps) {
  return <div />;
}
```

**With `forwardRef`:**

```ts
import { forwardRef } from 'react';

type Props = { /* … */ };

export const RedirectPaymentMethodIntegration = forwardRef<HTMLDivElement, Props>(
  (props, ref) => <div ref={ref} />
);
```

(These patterns stay simple and robust long-term.) (https://blog.variant.no/a-better-way-to-type-react-components-9a6460a1d4b7)

PR #19048, opened 2025-09-08
refactor(pci-instances): add an expanded button to show more private network

Tsiorifamonjena:
## Description
Add a button "Voir plus" when there is more private networks than 2

Ticket Reference: #...    

## Additional Information
image

### DIFF
--- a/packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/InstanceNetworkingBlock.component.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/InstanceNetworkingBlock.component.tsx
@@ -3,12 +3,10 @@ import { useTranslation } from 'react-i18next';
 import { useProjectUrl } from '@ovh-ux/manager-react-components';
 import { Text, TEXT_PRESET } from '@ovhcloud/ods-react';
 import DashboardCardLayout from './DashboardCardLayout.component';
-import { useProjectId } from '@/hooks/project/useProjectId';
 import { useDashboard } from '../hooks/useDashboard';
 import { ActionsMenu } from '@/components/menu/ActionsMenu.component';
 import NetworkItem from './network/NetworkItem.component';
 import { DashboardTileBlock } from './DashboardTile.component';
-import { useDedicatedUrl } from '@/hooks/url/useDedicatedUrl';
 import { useInstanceParams } from '@/pages/instances/action/hooks/useInstanceActionModal';
 import ReverseDNS from './network/ReverseDNS.component';
 import ExpandedNetwork from './network/ExpandedNetwork.component';
@@ -16,8 +14,6 @@ import ExpandedNetwork from './network/ExpandedNetwork.component';
 const InstanceNetworkingBlock: FC = () => {
   const { t } = useTranslation(['dashboard', 'list', 'actions']);
   const { region, instanceId } = useInstanceParams();
-  const dedicatedUrl = useDedicatedUrl();
-  const projectId = useProjectId();
   const projectUrl = useProjectUrl('public-cloud');
 
   const { instance, isPending: isInstanceLoading } = useDashboard({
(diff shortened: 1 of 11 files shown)


packages/manager/apps/pci-instances/src/components/menu/ActionMenuItem.component.tsx
```diff
@@ -2,7 +2,15 @@ import { FC } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useHref } from 'react-router-dom';
 import { Icon, Link } from '@ovhcloud/ods-react';
-import { TAction } from '@/pages/instances/instance/dashboard/view-models/selectInstanceDashboard';
+
```
aTurmo: TAction is a view model so it seems to me that it should stay where it was.
Why did you move it ?

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/network/ExpandedNetwork.component.tsx
```diff
@@ -0,0 +1,31 @@
...
+import { Button, Icon } from '@ovhcloud/ods-react';
+import { useTranslation } from 'react-i18next';
+import NetworkItem from './NetworkItem.component';
+import { TInstanceAddress } from '@/types/instance/entity.type';
+
+const ExpandedNetwork: FC<{ networks: TInstanceAddress[] }> = ({
+  networks,
+}) => {
+  const { t } = useTranslation('common');
+  const [isExpand, setIsExpand] = useState(false);
```
aTurmo:
```suggestion
  const [isExpanded, setIsExpanded] = useState(false);
```

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/network/ExpandedNetwork.component.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  const handleExpand = () => setIsExpand(!isExpand);
```
aTurmo:
```suggestion
  const handleExpand = () => setIsExpand(prev => !prev);
```

It's better to use the lambda with the previous value instead of the exposedValue (isExpand).

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/InstanceNetworkingBlock.component.tsx
```diff
@@ -102,20 +95,22 @@ const InstanceNetworkingBlock: FC = () => {
     [projectUrl, t, instance?.isEditEnabled],
   );
 
+  const expandedNetworks = instance?.privateNetwork.slice(2) ?? null;
```
aTurmo:
```suggestion
  const expandedNetworks = instance?.privateNetwork.slice(2);
```

Empty array and undefined do the same with '&&' line 130. No nedd to add null

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/network/ExpandedNetwork.component.tsx
```diff
@@ -0,0 +1,31 @@
...
+
+  return (
+    <>
+      <div className="flex justify-end">
+        <Button variant="ghost" onClick={handleExpand}>
+          {t('pci_instances_common_show_more')}{' '}
```
aTurmo:
```suggestion
          {t('pci_instances_common_show_more')}
```

packages/manager/apps/pci-instances/src/components/menu/ActionMenuItem.component.tsx
Tsiorifamonjena: I move it because I think there is a circular import. It's weird to import a type from a page to a component. 
And also, ActionMenuItem is a generic Component not only for the dashboard. The view-model here is just for Dashboard.
What about creating a common type that can be aliased by the props and the view-model. What do you think ?

aTurmo: I'm not so sure about that.

fredericvilcot: Yes we can create a shared type like we do for entities / DTO 👍🏻

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/network/ExpandedNetwork.component.tsx
Tsiorifamonjena: nice catch 😃

packages/manager/apps/pci-instances/src/components/menu/ActionMenuItem.component.tsx
Tsiorifamonjena: For now I move type to `@/types/instance/action/action.type` then. I let you check if it's good for you. I can adjust if needed

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/InstanceNetworkingBlock.component.tsx
```diff
@@ -129,9 +124,10 @@ const InstanceNetworkingBlock: FC = () => {
             items={privateIpActionsLinks}
           />
         </div>
-        {privateIps?.map((privateIp) => (
-          <NetworkItem key={privateIp.ip} address={privateIp} />
+        {instance?.privateNetwork.slice(0, 2).map((network) => (
```
fredericvilcot: `Slice` method should be moved to the view model, view component should not re-compute in any way the selected data. 

By the way, putting it in a dedicated method in your view model will enhance documentation as I have no idea why you need to slice `privateNetwork`

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/InstanceNetworkingBlock.component.tsx
```diff
@@ -24,14 +25,10 @@ const InstanceNetworkingBlock: FC = () => {
...
 
-  const publicIPs = useMemo(
-    () =>
-      instance?.addresses.get('floating') ?? instance?.addresses.get('public'),
-    [instance?.addresses],
-  );
-
   const publicIpActionsLinks = useMemo(() => {
-    const ipV4 = publicIPs?.find((publicIp) => publicIp.version === 4)?.ip;
+    const ipV4 = instance?.publicNetwork.networks.find(
```
fredericvilcot: Should be part of the view model, no computations in the view

PR #19050, opened 2025-09-08
feat(iam): add tag condition for custom policy

darsene:
resolves: #MANAGER-19326

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/modules/iam/src/components/conditions/conditionList/conditionList.template.html
+++ b/packages/manager/modules/iam/src/components/conditions/conditionList/conditionList.template.html
@@ -1,0 +1,4 @@
+<p>/!\ Condition List placeholder /!\</p>
+<p data-ng-repeat="condition in $ctrl.conditions track by $index">
+    {{condition.values}}
+</p>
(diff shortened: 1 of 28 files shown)


yjaaouane (approved):

packages/manager/modules/iam/src/components/conditions/conditionList/conditionList.template.html
```diff
@@ -0,0 +1,4 @@
+<p>/!\ Condition List placeholder /!\</p>
+<p data-ng-repeat="condition in $ctrl.conditions track by $index">
+    {{condition.values}}
+</p>
```
tibs245, line 4: If it's a list why not use ul / li ?

tibs245 (approved):

packages/manager/modules/iam/src/components/conditions/conditionList/conditionList.template.html
darsene: It's temporary, will disappear after

PR #19056, opened 2025-09-09
fix(*): fix incorrect pagination params in iceberg fetch

atefzed:
ref: #PRDCOL-182

## Description

Ticket Reference: #PRDCOL-182

## Additional Information

This PR fixes a bug in the Zimbra app caused by a regression introduced during the recent refactor of the core-api module. The issue was related to the **icebergeV2** service, where setting **cursorNext** to an empty string ("") caused **TanStack Query**'s useInfiniteQuery to enter an infinite re-render loop. The fix ensures that cursorNext is only set when a valid value is present, preventing this behavior and restoring correct pagination.

### DIFF
--- a/packages/manager/core/api/src/iceberg.ts
+++ b/packages/manager/core/api/src/iceberg.ts
@@ -78,7 +78,7 @@ export const buildHeaders = () => {
       return builder;
     },
     setDisabledCache: (disableCache: boolean | undefined) => {
-      if (disableCache) headers['Pragma'] = 'no-cache';
+      if (disableCache) headers.Pragma = 'no-cache';
       return builder;
     },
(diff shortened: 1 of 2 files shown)


packages/manager/core/api/src/iceberg.ts
```diff
@@ -153,7 +153,7 @@ export async function fetchIcebergV2<T>({
 
   return {
     data: response.data,
-    cursorNext: headers['x-pagination-cursor-next'] ?? '',
+    cursorNext: headers['x-pagination-cursor-next'],
```
helabenkhalfallah, line 156: thank you :)

anooparveti (approved):

ghyenne (approved):

PR #19058, opened 2025-09-09
feat(pci-instances): improve card genericity

lolottetheclash:
## Description
The purpose of this PR is to modify the Cart to make it more generic.
However, it is important to keep in mind that this remains a component that must be generic but must also allow for a v.0 to be created. 
Therefore, not everything has been addressed and will be improved as it is used in instance creation.

Capture d’écran 2025-09-16 à 10 02 38

Ticket Reference: #...


packages/manager/apps/pci-instances/src/components/localizationCard/LocalizationCard.component.tsx
```diff
@@ -1,9 +1,9 @@
 import { Badge, Card, Radio, RadioControl, Text } from '@ovhcloud/ods-react';
 import clsx from 'clsx';
+import { Controller, useFormContext } from 'react-hook-form';
 import { Flag } from '../flag/Flag';
 import { TCountryIsoCode } from '../flag/country-iso-code';
 import { TDeploymentMode } from '@/types/instance/common.type';
-import { Controller, useFormContext } from 'react-hook-form';
 
 export type TLocalizationCardProps = {
```
ppprevost, line 9: go avoid a switch case 
`
const badgeStyles: Record<TDeploymentMode | 'default', [string, string]> = {
  PRD: ['700', '000'],
  
  default: ['400', '000'],
};

const getBadgeClassName = (mode: TDeploymentMode) => {
  const [bg, text] = badgeStyles[mode] ?? badgeStyles.default;
  return `bg-[--ods-color-information-${bg}] text-[--ods-color-information-${text}]`;
};
`

packages/manager/apps/pci-instances/src/components/localizationCard/LocalizationCard.component.tsx
```diff
@@ -13,16 +13,22 @@ export type TLocalizationCardProps = {
   onSelect?: (region: string) => void;
 };
 
+type TBadgeColors = {
+  background: string;
+  text: string;
+};
+
+const badgeStyle: Record<TDeploymentMode, TBadgeColors> = {
```
fredericvilcot: Maybe this part (badge) could be more generic, as we will need it in several cards in the creation page?

packages/manager/apps/pci-instances/src/components/cart/genericCart/GenericCart.tsx
```diff
@@ -4,16 +4,21 @@ import {
...
   AccordionItem,
   AccordionTrigger,
+  Divider,
   Text,
 } from '@ovhcloud/ods-react';
 import clsx from 'clsx';
-import { useCallback, useState } from 'react';
+import { PropsWithChildren, useCallback, useState } from 'react';
+
+export type TProductDetail = PropsWithChildren<{
```
fredericvilcot: What is a product regarding the cart?

lolottetheclash: the cart can contain many products (here it's an instance) which display the product details (here an instance is made of: image, ssh, network...), does it sound good for you?

fredericvilcot: I see, but IMO a product may be decorrelated from the cart itself 🧐

Maybe we could imagine being more generic by using a `Cart` that contains some `CartItems` that are described by `cartItemDetails`, `quantity`, etc..

What do you think?

packages/manager/apps/pci-instances/src/components/cart/Cart.tsx
```diff
@@ -1,33 +1,51 @@
...
 import { useTranslation } from 'react-i18next';
 import { useMemo } from 'react';
-import { GenericCart, TCartProduct } from './genericCart/GenericCart';
+import { NAMESPACES } from '@ovh-ux/manager-common-translations';
+import { Text } from '@ovhcloud/ods-react';
+import {
+  GenericCart,
+  TCartProduct,
+  TProductDetail,
+} from './genericCart/GenericCart';
```
ppprevost: GenericCart.component ?

fredericvilcot: Maybe we could enhance using `BaseCart.component.tsx` :)

aTurmo (approved):

packages/manager/apps/pci-instances/src/components/cart/Cart.tsx
lolottetheclash: I originally put `genericCart` so that we would remember that this element would then need to be migrated out of our app, into `pci-common` or elsewhere, because it should be normally be named `Cart`

packages/manager/apps/pci-instances/src/components/cart/Cart.component.tsx
```diff
@@ -0,0 +1,81 @@
...
+  expanded: boolean;
+};
+
+export type TCartItemDetail = {
+  name: string;
+  description?: JSX.Element;
+  price?: number;
+};
+
+const getTotalPrice = (items: TCartItemDetail[]) =>
```
aTurmo, line 33: Suggestion (not-blocking): this could be extracted in a service

packages/manager/apps/pci-instances/src/components/cart/components/Cart.component.tsx
```diff
@@ -0,0 +1,14 @@
+import { ReactNode } from 'react';
+
+export type TCartProps = {
+  children: ReactNode;
+};
+
+export const Cart = ({ children }: TCartProps) => (
```
aTurmo: issue (blocking) : component dupplication

There are two Cart.components files, and they have the same exported name. This will be confusing to use.
And since this one is just a div with a className, maybe it can be removed ? What do you think ?

packages/manager/apps/pci-instances/src/components/cart/components/CartActions.component.tsx
```diff
@@ -0,0 +1,13 @@
+import clsx from 'clsx';
+import { ReactNode } from 'react';
+
+type CartActionsProps = {
+  children: ReactNode;
+  className?: string;
+};
+
+export const CartActions = ({ className, children }: CartActionsProps) => (
```
aTurmo, line 8: suggestion (non-blocking): Could this also include the action buttons ?

packages/manager/apps/pci-instances/src/components/cart/components/CartItem.component.tsx
```diff
@@ -0,0 +1,12 @@
+import { AccordionItem } from '@ovhcloud/ods-react';
+import { PropsWithChildren } from 'react';
+
+type TCartItem = PropsWithChildren<{
+  value: string;
+}>;
+
+export const CartItem = ({ value, children }: TCartItem) => (
+  <AccordionItem value={value} data-testid={`cart-item-${value}`}>
```
aTurmo, line 9: Thought (non-blocking): What's the team rules with data-testid (if any)?

Because I think it's often not necessary and for testing, "getByRole" and "getByText" should cover almost everything. I'm of course opened to dicuss this :)

packages/manager/apps/pci-instances/src/components/cart/components/CartItemDetails.component.tsx
```diff
@@ -0,0 +1,29 @@
...
+          {price && (
+            <Text preset="heading-6" data-testid="cart-item-details-price">
+              {price}
+            </Text>
+          )}
+        </div>
+
+        {index !== details.length - 1 && (
+          <Divider spacing="6" data-testid="cart-item-details-divider" />
+        )}
```
aTurmo, line 25: Suggestion (non-blocking): you could put this first and do

```typescript
        {index && (
          <Divider spacing="6" data-testid="cart-item-details-divider" />
        )}
```

and thus avoid checking details length

packages/manager/apps/pci-instances/src/components/cart/components/CartTotalPrice.component.tsx
```diff
@@ -0,0 +1,46 @@
...
+import { Text } from '@ovhcloud/ods-react';
+
+type TCartTotalPrice = {
+  price: number;
+  text?: string;
+  displayHourlyPrice?: boolean;
+  displayMonthlyPrice?: boolean;
+};
+
+// TODO: add "total" translation in common-translations
```
aTurmo, line 11: todo 😅

packages/manager/apps/pci-instances/src/pages/instances/create/components/CreationCart.component.tsx
```diff
@@ -0,0 +1,63 @@
...
+export type TCartItemDetail = {
+  name: string;
+  description?: JSX.Element;
+  price?: number;
+};
+
+export const CreationCart = () => {
+  const { t } = useTranslation('common');
+  const { control } = useFormContext<TInstanceCreationForm>();
+  const [name, region] = useWatch({
```
aTurmo, line 15: Suggestion (non-blocking): useFormContext provide a 'watch' method, so using 'useWatch' could be avoided

packages/manager/apps/pci-instances/src/components/cart/__tests__/Cart.component.spec.tsx
```diff
@@ -0,0 +1,120 @@
...
+    getTextPrice: vi.fn((price: number) => `${price}`),
+    getFormattedHourlyCatalogPrice: vi.fn(),
+    getFormattedMonthlyCatalogPrice: vi.fn(),
+  }),
+}));
+
+const renderCart = (items: TCartItem[]) => {
+  render(<Cart items={items} />);
+  return screen.getByTestId('cart');
+};
```
aTurmo: Question (non-blocking): Is it necessary to return a screen.getByTestId ?

It's is used only to expect(cart).toBeInTheDocument(); on every test and it feels useless.
The render method does render the component you pass, so thoses kind of test feels like a library test.

Component testing should test for what's displayed and the behavior when interracting with it.

packages/manager/apps/pci-instances/src/components/cart/__tests__/Cart.component.spec.tsx
```diff
@@ -0,0 +1,120 @@
...
+    expect(instanceLocalizationElt).toHaveTextContent('eu-west-par');
+    expect(instanceLocalizationElt).toHaveTextContent('50');
+
+    const instanceImageElt = screen.getByTestId(`cart-item-details-image`);
+    expect(instanceImageElt).toHaveTextContent('Linux');
+    expect(instanceImageElt).toHaveTextContent('10');
+
+    const kubeElt = screen.getByTestId(`cart-item-details-kube localization`);
+    expect(kubeElt).not.toBeVisible();
+  });
```
aTurmo: Though (non-blocking) : overuse of data-testid

The problems with this is that it adds difficulties to refactoring. If i'm changing the name of the components or break/encapsulate it, data-testids have to change.

Furethermore data-testid is not something visible to the user, and the aim of component testing is to verify what's an user could expect from this component.

As an exemple instead of:
```typescript
    const instanceImageElt = screen.getByTestId(`cart-item-details-image`);
    expect(instanceImageElt).toHaveTextContent('Linux');
    expect(instanceImageElt).toHaveTextContent('10');
```

we could do:
```
const instanceImageElt = screen.getByText('Linux 10')
expect(instanceImageElt).toBeVisible()
```

If 'Linux 10' is not what's exactuly displayed you can use regex instead.

Maybe i'm wring about all of this but the issue I think is that data-testid is not a meaningful representation of what's displayed, whereas 'getByText' and 'getByRole' are.

No problem discussing this of course ^^

aTurmo:
Praise : love the global clarity of all the new components :)

packages/manager/apps/pci-instances/src/components/cart/components/CartItem.component.tsx
manoncarbonnel: note: _User facing attributes_ provided by Testing Library, like `getByLabelText` or `getByText`,  are the best practice for selectors because they mimic a user's analysis of the page content. However, when this is not possible (for example, when it is a collection with unpredictable content), it is best to avoid using additional classes, especially with utility classes like Tailwind, and instead use a `data-testid=""`.

packages/manager/apps/pci-instances/src/components/cart/components/CartTotalPrice.component.tsx
lolottetheclash: yes, it will be done after ;)

packages/manager/apps/pci-instances/src/components/cart/components/Cart.component.tsx
lolottetheclash: Yes, I understand what you mean but here, this Cart is the generic one, and the other one is the instance one. the purpose is to move it into pci-common or something else.
So I would like to keep "Cart" name as "Modal" name or something else.
 Don't hesitate to ping if you need more explanations :)

packages/manager/apps/pci-instances/src/components/cart/components/CartItem.component.tsx
lolottetheclash: I don't know exactly if there is an established way of doing things within the team, but I find that data-testid offers a more precise way of retrieving the elements we are interested in, and is possibly more robust over time. 
So I naturally went with that method... But I'm also open to discussion, especially during a team meetin.

packages/manager/apps/pci-instances/src/pages/instances/create/CreateInstance.page.tsx
```diff
@@ -91,7 +91,7 @@ const CreateInstance: FC = () => {
             <LocalizationSelection />
           </section>
           <aside className="min-w-[320px] w-full max-w-[640px]">
```
ppprevost, line 93: are you sure about that ? I mean, if we want to be more responsive on a near future ?

packages/manager/apps/pci-instances/src/components/cart/components/CartItemDetails.component.tsx
lolottetheclash: Problem is that it will display 0 instead of the divider for the first one and also display divider after the last element :'(

packages/manager/apps/pci-instances/src/pages/instances/create/components/CreationCart.component.tsx
lolottetheclash: first: thank you for this feedback, it allowed me to look a little deeper on the differences between the 2.
according to the docs and blogs, `useWatch` is recommended whenever you retrieve a value from the form in a child of the form, because if you use `watch`, it triggers a re-render at the root.
With `useWatch`, there is only a re-render in the component level.
In our cases, it is therefore better to use `useFormContext` + `useWatch`.

lolottetheclash:
> Praise : love the global clarity of all the new components :)

Thank you very much for your feedback and encouragement!

packages/manager/apps/pci-instances/src/pages/instances/create/CreateInstance.page.tsx
lolottetheclash: This one will be addressed by Manon in another PR ;)

packages/manager/apps/pci-instances/src/components/cart/__tests__/Cart.component.spec.tsx
lolottetheclash: yes, good point thank you!

packages/manager/apps/pci-instances/src/components/cart/components/Cart.component.tsx
aTurmo: So maybe the instance one could be renammed to "InstanceCart". I think it makes sens and it remove the confusion while the generic cart is still in pci-instance

packages/manager/apps/pci-instances/src/components/cart/components/CartActions.component.tsx
lolottetheclash: discussed directly with you 👌

packages/manager/apps/pci-instances/src/components/cart/components/CartItemDetails.component.tsx
aTurmo: oh yes sorry ^^
you need to do "index !==0"

and I said to put this first meaning before the 
```
     <div className="flex justify-between">
          {description}"
```          
part
          But this is really nitpicky don't bother

packages/manager/apps/pci-instances/src/pages/instances/create/components/CreationCart.component.tsx
aTurmo: Oh ok. I guess I didn't go deep enough ^^
It's wierd that they propose this option then if it's less efficient.

packages/manager/apps/pci-instances/src/components/cart/Cart.component.tsx
lolottetheclash: mmmh, I'm not sure as this function is really specific to the `Cart`, and also it's a really simple one.
If it was used by several components, or if it's becoming more complex and needs to be tested, I will move it elsewhere. Is it ok for you?

packages/manager/apps/pci-instances/src/components/cart/__tests__/Cart.component.spec.tsx
lolottetheclash: regarding the `data-testid` discussion, I suggest that we keep this subject in mind to discuss about it next friday 📣

packages/manager/apps/pci-instances/src/components/cart/components/CartActions.component.tsx
```diff
@@ -0,0 +1,13 @@
+import clsx from 'clsx';
+import { ReactNode } from 'react';
+
+type CartActionsProps = {
```
fredericvilcot: You can use `PropsWithChildren`

packages/manager/apps/pci-instances/src/components/cart/components/CartContent.component.tsx
```diff
@@ -0,0 +1,39 @@
+import { Accordion, AccordionChangeDetail } from '@ovhcloud/ods-react';
+import { ReactNode, useCallback, useState } from 'react';
+import { TCartItem } from '../Cart.component';
+
+export type TRenderItemParams = {
```
fredericvilcot:
```suggestion
export type TRenderCartItemParams = {
```

packages/manager/apps/pci-instances/src/components/cart/Cart.component.tsx
aTurmo: The logic here is to regroup the cart logic outside the disply component as mush as possible and test even the simplest things ^^
The method here is more a business logic thing than a display thing (the price of a cart makes sense outside frontend needs).
And having a file with all cart related operations (like price calculation) allow for an easy way yo see what's already there and avoid recoding something that's hidden inside a component.

packages/manager/apps/pci-instances/src/pages/instances/create/components/CreationCart.component.tsx
lolottetheclash: it's efficient if you're at the root of the form, less if you're at the childs level
and it's more appropriate if you're watching several values also

packages/manager/apps/pci-instances/src/components/cart/components/Cart.component.tsx
```diff
@@ -0,0 +1,14 @@
+import { ReactNode } from 'react';
+
+export type TCartProps = {
```
fredericvilcot: You can use `PropsWithChildren`

packages/manager/apps/pci-instances/src/components/cart/components/CartItem.component.tsx
aTurmo: As dicussed, for now i'll resolve this but we need a team discussion to clarify all this

packages/manager/apps/pci-instances/src/components/cart/components/CartContent.component.tsx
```diff
@@ -0,0 +1,39 @@
...
+};
+
+export type TCartContentProps = {
+  items: TCartItem[];
+  renderCartItem: (params: TRenderItemParams) => ReactNode;
+};
+
+export const CartContent = ({ items, renderCartItem }: TCartContentProps) => {
+  const [expandedAccordions, setExpandedAccordions] = useState<string[]>(
+    items.filter((item) => item.expanded).map((item) => item.title),
```
fredericvilcot: Hope that `titles` will be unique :D

packages/manager/apps/pci-instances/src/components/cart/components/CartContent.component.tsx
```diff
@@ -0,0 +1,39 @@
...
+  const handleAccordionChange = useCallback(
+    (accordion: AccordionChangeDetail) =>
+      setExpandedAccordions(accordion.value),
+    [setExpandedAccordions],
+  );
+
+  const isExpanded = useCallback(
+    (accordionName: string) => expandedAccordions.includes(accordionName),
+    [expandedAccordions],
+  );
```
fredericvilcot: IMO there is no need to memoize with `useCallback`

packages/manager/apps/pci-instances/src/components/cart/components/Cart.component.tsx
lolottetheclash: I thought about it before but I changed my mind as it's in components folder: as we already know that all components are instance ones...

packages/manager/apps/pci-instances/src/components/cart/components/CartTotalPrice.component.tsx
```diff
@@ -0,0 +1,46 @@
+import { useCatalogPrice } from '@ovh-ux/manager-react-components';
+import { Text } from '@ovhcloud/ods-react';
+
+type TCartTotalPrice = {
+  price: number;
+  text?: string;
+  displayHourlyPrice?: boolean;
+  displayMonthlyPrice?: boolean;
```
fredericvilcot, line 8: TODO (in generic implementation) : let the caller formats the price using a formatting function :)

lolottetheclash: haha I was expecting this one! As we said, v.0 ;)

packages/manager/apps/pci-instances/src/components/cart/components/CartContent.component.tsx
lolottetheclash: clearly not!

packages/manager/apps/pci-instances/src/components/cart/components/Cart.component.tsx
lolottetheclash: oupsy, yes cleaner

packages/manager/apps/pci-instances/src/components/cart/components/Cart.component.tsx
aTurmo: mmmh... Ok but something like "Cart" seems really generic when in the case of "src/component/cart/Cart.component.tsx" it's not. I understand why you did it like that but it still doesn't feels right for me ^^
Another point of vue would be nice but anyway I accept that. It's not such a problem.

packages/manager/apps/pci-instances/src/components/cart/components/CartContent.component.tsx
lolottetheclash: agreed!

fredericvilcot (approved):
Great work, that's a good start for this brand new component :)

PR #19073, opened 2025-09-10
feat(object-storage): add offiste replication destination

lionel95200x:
ref: #TAPC-4148

## Description

Ticket Reference: #...


packages/manager/apps/pci-object-storage/public/translations/containers/add/Messages_fr_FR.json
```diff
@@ -63,7 +63,7 @@
   "pci_projects_project_storages_containers_project_banner_message": "Pour accéder aux étapes suivantes, vous devez",
   "pci_projects_project_storages_containers_project_banner_link_text": "activer votre projet.",
   "pci_projects_project_storages_containers_offsite_replication_title": "Offsite Replication",
-  "pci_projects_project_storages_containers_offsite_replication_description": "Répliquez automatiquement vos données sur un site distant au meilleur prix et pour une résilience accrue.",
+  "pci_projects_project_storages_containers_offsite_replication_description": "Répliquez automatiquement vos données sur un site distant dans une classe de stockage Infrequent Access. <1>En savoir plus </1>",
```
Tsiorifamonjena, line 66: Is not there another way to avoid using these number tag ? like `<link>Savoir plus</link>`

packages/manager/apps/pci-object-storage/src/domain/regions/regionFilter.spec.ts
```diff
@@ -0,0 +1,572 @@
...
+      expect(result).toEqual([]);
+    });
+
+    it('should filter only storage plans', () => {
+      const result = filterAllowedRegions(
+        mockProductAvailability,
+        mockProjectRegions,
+      );
+
+      // Should include regions from storage plans only
```
aTurmo: If you need a comment for an expect that means the tests is not clear and/or should be renamed.

packages/manager/apps/pci-object-storage/src/domain/regions/regionFilter.spec.ts
```diff
@@ -0,0 +1,572 @@
...
+      expect(result.length).toBeGreaterThan(0);
```
aTurmo: Since you controls your inputs in a test the lenght should be a precise number.

```suggestion
      expect(result.length).toEqual(X);
```

packages/manager/apps/pci-object-storage/src/domain/regions/regionFilter.spec.ts
```diff
@@ -0,0 +1,572 @@
...
+      // All returned regions should be from storage plans
+      result.forEach((region) => {
+        const isFromStoragePlan = mockProductAvailability.plans
+          ?.filter((plan) => plan.code.startsWith('storage'))
+          .some((plan) => plan.regions.includes(region));
+        expect(isFromStoragePlan).toBe(true);
+      });
```
aTurmo: Way too much calculations for a tests. You should explicitly write the result. The best way is extract each region in const like 'regionWithStoragePlan' and 'regionWithoutStoragePlan' or something like that and du expect(result).toContain(...)

packages/manager/apps/pci-object-storage/src/domain/regions/regionFilter.spec.ts
```diff
@@ -0,0 +1,572 @@
...
+        deploymentMode: 'localzone',
+      };
+
+      const result = filterAllowedRegions(
+        mockProductAvailability,
+        mockProjectRegions,
+        criteria,
+      );
+
+      result.forEach((region) => {
```
aTurmo, line 140: you should add a size check to verify that  there are no missing region.

packages/manager/apps/pci-object-storage/src/domain/regions/regionFilter.spec.ts
```diff
@@ -0,0 +1,572 @@
...
+
+      const result = filterAllowedRegions(
+        mockProductAvailability,
+        mockProjectRegions,
+        criteria,
+      );
+
+      // Should include both enabled and disabled regions
+      const disabledRegions = result.filter((region) => !region.enabled);
+      expect(disabledRegions.length).toBeGreaterThan(0);
```
aTurmo: should be a precise number

packages/manager/apps/pci-object-storage/src/domain/regions/regionFilter.spec.ts
```diff
@@ -0,0 +1,572 @@
...
+import {
+  filterAllowedRegions,
+  groupRegionsByContinent,
+  ProductAvailability,
+  ProjectRegion,
+  RegionFilterCriteria,
+} from './regionFilter';
+import { OBJECT_CONTAINER_OFFER_STORAGE_STANDARD } from '@/constants';
+
+describe('filterAllowedRegions', () => {
```
aTurmo, line 12: General comments for all the file :
There should not be comment explaining what you are testing inside your tests.
There should not be complicated calculation for your expect. It looks like you are rewriting the implementation of each method when you do things like `result.forEach(region => truc.filter(...))`. Unit tests for fonction should be a much as possible a simple `expect(result).toEqual(expected)`and I think it's possible in most cases here.

packages/manager/apps/pci-object-storage/src/domain/regions/regionFilter.ts
```diff
@@ -0,0 +1,109 @@
...
+      if (
+        projectRegion &&
+        !projectRegion.services.some(({ name }) => name === offer)
+      ) {
+        return false;
+      }
+    }
+
+    return true;
+  });
```
aTurmo, line 71: A method could be extracted here

packages/manager/apps/pci-object-storage/src/pages/objects/container/new/step/OffsiteReplicationStep.component.test.tsx
```diff
@@ -59,23 +61,48 @@ vi.mock('react-router-dom', async (importOriginal) => {
...
+    isPending: false,
+  }),
 }));
 
 vi.mock('@ovh-ux/manager-react-components', async () => {
   const actual = await vi.importActual('@ovh-ux/manager-react-components');
   return {
     ...actual,
-    useMe: vi.fn(),
+    useMe: vi.fn().mockReturnValue({
```
ppprevost: you could migrate it in setupTest

packages/manager/apps/pci-object-storage/src/hooks/useAllowedRegions.ts
```diff
@@ -0,0 +1,46 @@
...
+  }, [
+    availability,
+    projectRegions,
+    criteria.deploymentMode,
+    criteria.offer,
+    criteria.excludeRegion,
+    criteria.onlyEnabled,
+    criteria.removeMacroIfMicroAvailable,
+    isPending,
+  ]);
```
aTurmo:
```suggestion
  }, [
    availability,
    projectRegions,
    criteria,
    isPending,
  ]);
```

packages/manager/apps/pci-object-storage/src/domain/regions/regionFilter.ts
lionel95200x: I just move the actual function, and as i don't know exactly which business rule is involved prefer to keep like that for the moment

aTurmo (approved):

PR #19091, opened 2025-09-10
refactor(object-storage): use global enum for ContainerMode

lionel95200x:
ref: #TAPC-4972

## Description

- Use a common enum for region container zone
- Rename variable to coherent one

Ticket Reference: #TAPC-4972

### DIFF
--- a/packages/manager/apps/pci-object-storage/src/pages/objects/container/object/show/Show.page.tsx
+++ b/packages/manager/apps/pci-object-storage/src/pages/objects/container/object/show/Show.page.tsx
@@ -8,6 +8,8 @@ import {
 } from 'react-router-dom';
 import {
   BaseLayout,
+  Links,
+  LinkType,
   Notifications,
   PciGuidesHeader,
   PciMaintenanceBanner,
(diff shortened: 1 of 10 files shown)


packages/manager/apps/pci-object-storage/src/constants.ts
```diff
@@ -3,37 +3,34 @@ export const TRACKING_PREFIX =
 
 export const MEGA_BYTES = 1024;
 
-export const OBJECT_CONTAINER_MODE_MULTI_ZONES = 'region-3-az';
-export const OBJECT_CONTAINER_MODE_MONO_ZONE = 'region';
-export const OBJECT_CONTAINER_MODE_LOCAL_ZONE = 'localzone';
+export enum ObjectContainerMode {
```
fredericvilcot, line 6: **Question**: are the zones a mode or a location ? 
In other projects, regions are more considered as locations.

lionel95200x: Good question, just keep the actual naming but it could be a full renaming also in file if mode is not appropriate

fredericvilcot: I'll suggest to rename it to harmonize same semantic around locations :)

fredericvilcot (approved):

packages/manager/apps/pci-object-storage/src/pages/objects/container/object/show/Show.page.tsx
```diff
@@ -253,11 +249,12 @@ export default function ObjectPage() {
     t,
   });
 
+  console.log({ tabs });
```
Tsiorifamonjena, line 241: **todo**: remove the console.log

PR #19100, opened 2025-09-10
feat(pci-kubernetes): enable 3az default

ppprevost:
ref:#tapc-4979

## Description

Ticket Reference: #...


packages/manager/apps/pci-kubernetes/src/pages/new/steps/LocationStep.component.spec.tsx
```diff
@@ -0,0 +1,115 @@
...
+  beforeEach(() => {
+    vi.clearAllMocks();
+  });
+
+  it('renders deployment mode and region selector', () => {
+    const props = {
+      ...baseProps,
+      step: { isLocked: false, isOpen: true, isChecked: true },
+    };
+    render(<LocationStep {...props} />, { wrapper });
```
aTurmo, line 73: Don't do this now but a remark for later :
Instead of importing a wrapper everywhere and using it like this, it's better to define a custom render and use that instead.
And furthermore 'wrapper' doesn't mean much and we don't know what it does.

packages/manager/apps/pci-kubernetes/src/pages/new/steps/LocationStep.component.spec.tsx
```diff
@@ -0,0 +1,115 @@
...
+
+    // Deployment mode section title
```
aTurmo: not a fan of comments

packages/manager/apps/pci-kubernetes/src/pages/new/steps/LocationStep.component.tsx
```diff
@@ -210,9 +235,13 @@ export function LocationStep({
...
-          onClick={() =>
-            !region.enabled ? addRegion(region.name) : onSubmit(region)
-          }
+          onClick={() => {
+            if (region) {
+              return !region.enabled
+                ? addRegion(region.name)
+                : onSubmit?.(region);
+            }
+          }}
```
aTurmo: It's better to extract this to a handleClick method

packages/manager/apps/pci-kubernetes/src/pages/new/steps/LocationStep.component.tsx
```diff
@@ -102,8 +122,8 @@ export function LocationStep({
           code: data.name,
         }),
       );
-      onSubmit(data);
-      setRegion((reg) => ({ ...reg, enabled: true }));
+      onSubmit?.(data);
```
aTurmo: Is there really times where there are no onSubmit ?

packages/manager/apps/pci-kubernetes/src/pages/new/steps/LocationStep.component.spec.tsx
ppprevost: it the common wrapper, i do not use another one.

aTurmo: yes i know that's why it's more a global remark for the project and not for you directly.

What I meant was it's better to have something like :
`export const renderWithWrapper = (ui) => render(ui, { commonWrapper })`
and use tyhis instead of the wrapper

packages/manager/apps/pci-kubernetes/src/pages/new/steps/LocationStep.component.tsx
```diff
@@ -144,18 +168,23 @@ export function LocationStep({
...
             </div>
 
             <div className="grid md:grid-cols-4 gap-4 my-5">
               {tilesData.map(
-                ({ title, pillLabel, description, regionType }) => (
+                ({ title, pillLabel, description, regionType, badge }) => (
                   <KubeDeploymentTile
                     key={regionType}
                     title={title}
+                    badge={badge}
```
lolottetheclash, line 187: badge expects a componentType but here badge could be null, or I am wrong?

ppprevost: it's a componentType that could send html or null

packages/manager/apps/pci-kubernetes/src/pages/new/steps/LocationStep.component.spec.tsx
ppprevost: yes i get it a hoc :)

PR #19104, opened 2025-09-11
fix(ips): update the design of the ip order page

aderghamov:
ref: #MANAGER-19473

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/ips/src/components/CountrySelector/country-selector.component.tsx
+++ b/packages/manager/apps/ips/src/components/CountrySelector/country-selector.component.tsx
@@ -5,6 +5,7 @@ import {
   OdsSelectCustomEvent,
   OdsSelectChangeEventDetail,
   ODS_SPINNER_SIZE,
+  OdsSelectCustomRendererData,
 } from '@ovhcloud/ods-components';
 
 export type CountrySelectorProps = {
@@ -38,6 +39,15 @@ export const CountrySelector: React.FC<CountrySelectorProps> = ({
     );
   }
 
+  const optionRenderingFunction = (data: OdsSelectCustomRendererData) => `
+  <div style="display:flex; gap: 8px; align-items: center;">
+    <span style="width: 22px; height: 16px; background-size: cover; background-image: url('flags/${data.value.toLowerCase()}.svg')"></span>
+    <span>${t(
+      `region-selector-country-name_${data.value.toUpperCase()}`,
+    )}</span>
+  </div>
+  `;
+
   return (
     <OdsSelect
       key={countryCodeList.join('-')}
@@ -46,12 +56,8 @@ export const CountrySelector: React.FC<CountrySelectorProps> = ({
       onOdsBlur={onBlur}
       isReadonly={isReadyOnly}
       customRenderer={{
-        option: (data) => `<div style="display: flex">
-            <span>${t(
-              `region-selector-country-name_${data.value.toUpperCase()}`,
-            )}</span>
-            <span style="width: 32px; height: 24px; margin: auto 15px auto auto; background-image: url('flags/${data.value.toLowerCase()}.svg')"></span>
-          </div>`,
+        item: optionRenderingFunction,
+        option: optionRenderingFunction,
       }}
       {...props}
     >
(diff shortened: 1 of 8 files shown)


packages/manager/apps/ips/src/components/CountrySelector/country-selector.component.tsx
```diff
@@ -38,6 +39,15 @@ export const CountrySelector: React.FC<CountrySelectorProps> = ({
     );
   }
 
+  const optionRenderingFunction = (data: OdsSelectCustomRendererData) => `
+  <div style="display:flex; gap: 8px; align-items: center;">
```
b-fritz-ovh, line 43: Why use inline styles? They are not easy to maintain. If you don't have styles conflicts purpose, I think we should use classes like .flex .align-center ?

packages/manager/apps/ips/src/components/CountrySelector/country-selector.component.tsx
```diff
@@ -38,6 +39,15 @@ export const CountrySelector: React.FC<CountrySelectorProps> = ({
...
+    <span style="width: 22px; height: 16px; background-size: cover; background-image: url('flags/${data.value.toLowerCase()}.svg')"></span>
```
b-fritz-ovh, line 44: In the same idea, why not use .bg-cover, and a class in a style.scss for height and width?

For bg-image, tailwind seems expose a greet feature. Could you try quickly to follow this post ? (I'm not sure it works)

https://stackoverflow.com/questions/70805041/background-image-in-tailwindcss-using-dynamic-url-react-js#answer-70805360

tailwind doc:

https://tailwindcss.com/docs/background-image

packages/manager/apps/ips/src/components/ipVersionOptionCard/IpVersionOptionCard.component.tsx
```diff
@@ -0,0 +1,92 @@
...
+      ) : (
+        <span className="card-children p-4 rounded-b-md">
+          {!isStartingPrice && price === 0 ? (
+            <span className="free-price">{t('free_price')}</span>
+          ) : (
+            <PriceDescription
+              isStartingPrice
+              price={price}
+              suffix={priceSuffix}
+              shouldOverrideStyle={true}
```
b-fritz-ovh, line 85: just `shouldOverrideStyle` ?

packages/manager/apps/ips/src/components/ipVersionOptionCard/ip-version-option-card.scss
```diff
@@ -0,0 +1,24 @@
...
+.option_card {
+  border-color: var(--ods-color-primary-100);
+  .card-children {
+    background-color: var(--ods-color-neutral-050);
+  }
+}
+
+.option_card_selected {
+  border: 2px solid #0050d7;
+  background-color: #f3fdff;
```
b-fritz-ovh: Don't we have some --ods-color var?

packages/manager/apps/ips/src/components/CountrySelector/country-selector.component.tsx
aderghamov: I would love to. 
The problem is that all this dom is rendered through a shadow root. so I cannot make any class work :(

packages/manager/apps/ips/src/components/CountrySelector/country-selector.component.tsx
aderghamov: same answer: I tried but it won't work. :-(

packages/manager/apps/ips/src/components/ipVersionOptionCard/ip-version-option-card.scss
aderghamov: There is no matching ods color for this one.
according to the documentation, it should have been --ods-color-primary-025 but it does no longer exist.

I use the color from the mockup instead

packages/manager/apps/ips/src/components/ipVersionOptionCard/IpVersionOptionCard.component.tsx
aderghamov: it is sufficient, yes.

packages/manager/apps/ips/src/components/ipVersionOptionCard/ip-version-option-card.scss
aderghamov: my bad! I have just found --ods-color-information-500 for the border  😅

PR #19109, opened 2025-09-11
feat(dedicated): removes deprecated  /sws/dedicated/server/ api call

b-fritz-ovh:
ref: #MANAGER-17541

Capture d’écran 2025-09-11 à 13 32 14

### DIFF
--- a/packages/manager/modules/bm-server-components/src/server/server.service.js
+++ b/packages/manager/modules/bm-server-components/src/server/server.service.js
@@ -571,15 +571,30 @@ export default class Server {
   }
 
   getFtpBackupOrderDetail(serviceName, capacity) {
-    return this.OvhHttp.get(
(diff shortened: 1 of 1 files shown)


packages/manager/modules/bm-server-components/src/server/server.service.js
```diff
@@ -571,15 +571,30 @@ export default class Server {
...
-        rootPath: '2api',
-        urlParams: {
-          serviceName,
-          capacity,
+      '/order/dedicated/server/{serviceName}/backupStorage',
+      options,
+    ).then((durations) =>
+      this.OvhHttp.get(
+        '/order/dedicated/server/{serviceName}/backupStorage/{duration}',
+        {
```
tibs245, line 590: Why not put default options here ? I think this can be more readable

tibs245 (approved):

PR #19114, opened 2025-09-11
feat(ips): enabling undefined organisation selection into the organisation section

aderghamov:
ref: #MANAGER-18716

## Description

Ticket Reference: #MANAGER-18716

### DIFF
--- a/packages/manager/apps/ips/src/pages/order/sections/OrganisationSection.component.tsx
+++ b/packages/manager/apps/ips/src/pages/order/sections/OrganisationSection.component.tsx
@@ -55,6 +55,7 @@ export const OrganisationSection: React.FC = () => {
           value={selectedOrganisation}
           placeholder={t('organisation_select_placeholder')}
         >
+          <option value={undefined}>{t('organisation_select_none')}</option>
           {organisations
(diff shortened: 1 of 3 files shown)


packages/manager/apps/ips/src/pages/order/sections/OrganisationSection.component.tsx
```diff
@@ -55,6 +55,7 @@ export const OrganisationSection: React.FC = () => {
           value={selectedOrganisation}
           placeholder={t('organisation_select_placeholder')}
         >
+          <option value={undefined}>{t('organisation_select_none')}</option>
```
chipp972, line 58: `value={undefined}` works ?
I think it should be empty string `value=""` but if it works it's good

aderghamov: It worked when I tested it.
I will test it on more time to be sure.

aderghamov: in fact `value={undefined}` works and `value=""` does not work (the option is not even displayed)

PR #19118, opened 2025-09-11
fix(ips): fix quick filter and ui feedbacks

chipp972:
ref: #MANAGER-19696

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/ips/src/pages/listing/ipListing/components/filters/FilterQuick.tsx
+++ b/packages/manager/apps/ips/src/pages/listing/ipListing/components/filters/FilterQuick.tsx
@@ -1,134 +1,103 @@
-import React, { useContext, useState, useRef, useEffect } from 'react';
+import React, { useContext } from 'react';
 import {
   ODS_BUTTON_SIZE,
   ODS_BUTTON_VARIANT,
   ODS_ICON_NAME,
 } from '@ovhcloud/ods-components';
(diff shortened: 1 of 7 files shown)


packages/manager/apps/ips/src/pages/listing/ipListing/components/filters/FilterQuick.tsx
```diff
@@ -1,134 +1,103 @@
...
-
-    document.addEventListener('mousedown', handleClickOutside);
-    return () => {
-      document.removeEventListener('mousedown', handleClickOutside);
-    };
-  }, []);
 
   return (
-    <div className={`relative ${className}`} ref={menuRef}>
+    <>
```
Steffy29, line 22: Why do you use an empty tag instead of remove it ?

Steffy29 (commented):
Why do you use an empty tag instead of remove it ?

packages/manager/apps/ips/src/pages/listing/ipListing/components/filters/FilterQuick.tsx
chipp972: Because in React the return value should contain only 1 node so I wrapped my 2 components (OdsButton and OdsPopover) in a ReactFragment so it won't add an extra node in the DOM and my React component return value is correct

PR #19121, opened 2025-09-11
refactor(container): align components and data with µ-app structure

oalkabouss:
ref: #MANAGER-18468

## Description

Ticket Reference: #MANAGER-18468


packages/manager/apps/container/src/data/api/preferences.ts
```diff
@@ -10,7 +10,7 @@ export const fetchPreferences = async (
     );
     return data?.value ? JSON.parse(data.value) : null;
   } catch (e) {
-    if (e.response?.status === 404) {
+    if ((e as any).response?.status === 404) {
```
JacquesLarique: `e` should be an `AxiosResponse`

packages/manager/apps/container/src/helpers/suggestions/suggestionsHelper.spec.ts
```diff
@@ -34,7 +34,7 @@ describe('suggestionsHelpers', () => {
       const user: Partial<User> = {
         legalform: 'corporation',
         country: 'FR',
-        companyNationalIdentificationNumber: '12345678901234',
+        companyNationalIdentificationNumber: 12345678901234,
```
JacquesLarique: Build is most likely to be failing because `companyNationalIdentificationNumber` is typed as a string

packages/manager/apps/container/src/data/api/preferences.ts
oalkabouss: I had forgotten to cancel this change, i didn’t mean to keep it, thanks

packages/manager/apps/container/src/container/common/notifications-sidebar/NotificationsSidebar.spec.tsx
```diff
@@ -52,6 +53,7 @@ vi.mock('@ovh-ux/ovh-reket');
 
 describe('NotificationsSidebar', () => {
   beforeAll(() => {
+      vi.spyOn(helpers, 'fromNow').mockResolvedValue('today');
```
JacquesLarique, line 56: Was it necessary to mock this helper ?
If so can you try to remove the timeout as they were used to wait for this helper's response

packages/manager/apps/container/src/components/modals-container/ModalsContainer.component.tsx
JacquesLarique, line 1: Can you rename the folder to `modal-container` and the component to `ModalContainer` as it is made to only ever display one modal at a time

packages/manager/apps/container/src/container/common/notifications-sidebar/NotificationsSidebar.spec.tsx
oalkabouss: Yes, to avoid dynamic date-fns locale imports in tests (flaky). i removed the waitFor timeouts

PR #19126, opened 2025-09-11
fixup! feat(web.exchange): enable logs for web exchange

yjaaouane:
resolves: #MANAGER-19600

## Description

Ticket Reference: #MANAGER-19600

### DIFF
--- a/packages/manager/modules/exchange/src/logs/data-streams/data-streams.template.html
+++ b/packages/manager/modules/exchange/src/logs/data-streams/data-streams.template.html
@@ -1,5 +1,6 @@
 <div data-ovh-alert></div>
 <log-to-customer-list
+    class="exchange-logs"
     data-log-subscription-api-data="$ctrl.logSubscriptionApiData"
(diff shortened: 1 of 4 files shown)


packages/manager/modules/exchange/src/logs/data-streams/data-streams.template.html
```diff
@@ -1,5 +1,6 @@
 <div data-ovh-alert></div>
 <log-to-customer-list
+    class="exchange-logs"
```
darsene, line 3: Is it really necessary as css classes are already imported in the module ?

packages/manager/modules/exchange/src/logs/live-tail/live-tail.template.html
```diff
@@ -1,6 +1,7 @@
 <div data-ovh-alert></div>
 
 <log-to-customer-live-tail
+    class="exchange-logs"
```
darsene, line 4: same here

PR #19142, opened 2025-09-12
feat(pci-instances): redirection to regionalised dashboard

Tsiorifamonjena:
## Description
This PR allow user to be redirected without restriction to the new react page for dashboard. This page is now regionalised

Ticket Reference: #TAPC-5002

### DIFF
--- a/packages/manager/modules/pci/src/projects/project/instances/instance/instance.controller.js
+++ b/packages/manager/modules/pci/src/projects/project/instances/instance/instance.controller.js
@@ -10,6 +10,8 @@ export default class PciInstanceController {
     PciProjectsProjectInstanceService,
     PciProject,
     CHANGELOG,
+    $rootScope,
+    $location,
   ) {
     this.$translate = $translate;
     this.coreConfig = coreConfig;
@@ -19,6 +21,8 @@ export default class PciInstanceController {
     this.PciProject = PciProject;
     this.THREE_AZ_REGION = THREE_AZ_REGION;
     this.CHANGELOG = CHANGELOG;
+    this.$rootScope = $rootScope;
+    this.$location = $location;
   }
 
   $onInit() {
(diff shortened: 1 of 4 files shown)


packages/manager/modules/pci/src/projects/project/instances/instance/instance.routing.js
```diff
@@ -210,30 +210,22 @@ export default /* @ngInject */ ($stateProvider) => {
...
+        const searchParams = new URLSearchParams();
 
         if (message) {
-          promise.then(() =>
-            CucCloudMessage[type](
-              message,
-              'pci.projects.project.instances.instance',
-            ),
-          );
+          searchParams.set(`${type}Messages`, message);
```
lolottetheclash, line 214: I think CucCloudMessage is useless now, or am I wrong?

Tsiorifamonjena: You're right. I have remove all useless injections

aTurmo (dismissed):

packages/manager/modules/pci/src/projects/project/instances/instance/instance.controller.js
```diff
@@ -60,4 +71,23 @@ export default class PciInstanceController {
   refreshMessages() {
     this.messages = this.messageHandler.getMessages();
   }
+
+  redirectToNewInstance() {
+    const { id, region } = this.instance;
+
+    if (
+      !this.hasRedirected &&
```
fredericvilcot: `this.hasRedirected` will always be false no? Because you redirect right after saying  `this.hasRedirected = true;` :)

Tsiorifamonjena: You're right. I put it to avoid redirection loop (this is the case from my localhost) because `redirectToNewInstance` was called both (on init and on every $locationChangeSuccess event). But since react rout will take over it's no more needed

aTurmo (approved):

PR #19163, opened 2025-09-12
feat(pci-block-storage): delete snapshots before retyping

aTurmo:
ref: #TAPC-4569


packages/manager/apps/pci-block-storage/src/api/select/snapshot.ts
```diff
@@ -0,0 +1,5 @@
+import { TVolumeSnapshot } from '@/api/data/snapshot';
+
+export const selectSnapshotForVolume = (volumeId: string) => (
```
lionel95200x, line 3: Why not simple function with two params ?

packages/manager/apps/pci-block-storage/src/pages/retype/RetypeConfirmActionForm.component.tsx
```diff
@@ -0,0 +1,116 @@
...
+                )}
+              </span>
+              <Input
+                {...register('confirmation', {
+                  required: true,
+                  pattern: confirmRegexp,
+                })}
+                className={clsx(
+                  displayFieldError &&
+                    'bg-red-100 border-red-500 text-red-500 focus:text-red-500',
```
lionel95200x: No need to focus text color cause you already have the color without focus

packages/manager/apps/pci-block-storage/src/api/select/snapshot.ts
aTurmo: So I can more easilly use it as a mapper in useVolumeSnapshots
```typescript
export const useVolumeSnapshots = (projectId: string, volumeId: string) => {

  const select = useMemo(
    () => (volume ? selectSnapshotForVolume(volumeId) : undefined),
    [volume],
  );

  return useQuery({
    queryKey: getVolumeSnapshotsQueryKey(projectId, volume?.region),
    queryFn: () => getSnapshotsByRegion(projectId, volume?.region),
    enabled: !!volume,
    select,
  });
};
```

packages/manager/apps/pci-block-storage/public/translations/retype/Messages_fr_FR.json
```diff
@@ -4,8 +4,10 @@
...
+  "pci_projects_project_storages_blocks_retype_confirm_action_label": "Entrer \"{{ confirmWord }}\" pour confirmer votre choix.",
+  "pci_projects_project_storages_blocks_retype_confirm_action_confirm_button": "Confirmer et continuer",
   "pci_projects_project_storages_blocks_retype_detach_volume": "Pour modifier votre type d'offre, vous devez détacher le volume de son instance {{ instance }}",
-  "pci_projects_project_storages_blocks_retype_detach_volume_label": "Entrer \"{{ confirmWord }}\" pour confirmer votre choix.",
-  "pci_projects_project_storages_blocks_retype_detach_volume_confirm_button": "Confirmer et continuer",
-  "pci_projects_project_storages_blocks_retype_detach_volume_error": "Une erreur s'est produite lors du détachement de l'instance. Nous vous invitons à vous rendre sur <0>la page block storage</0> afin de détacher l'instance manuellement"
+  "pci_projects_project_storages_blocks_retype_detach_volume_error": "Une erreur s'est produite lors du détachement de l'instance. Nous vous invitons à vous rendre sur <0>la page block storage</0> afin de détacher l'instance manuellement.",
```
Tsiorifamonjena, line 10: Can you use another than number tag ? I think they are no encoded very well when you request translation. Especially on fast pass

packages/manager/apps/pci-block-storage/src/api/hooks/useSnapshot.ts
```diff
@@ -0,0 +1,52 @@
...
+
+export const getVolumeSnapshotsQueryKey = (
+  projectId: string,
+  volumeId: string,
+) => ['volume-snapshot', projectId, volumeId];
+
+export const useVolumeSnapshots = (projectId: string, volumeId: string) => {
+  const { data: volume } = useVolume(projectId, volumeId);
+
+  const select = useMemo(
```
Tsiorifamonjena, line 16: **suggestions** : How about exit this then the hook can take select as params and if tomorrow you want other use case it will more flexible

packages/manager/apps/pci-block-storage/public/translations/retype/Messages_fr_FR.json
aTurmo: I tried with <link> but it doesn't work

packages/manager/apps/pci-block-storage/src/api/hooks/useSnapshot.ts
aTurmo: No because it's integral to the logic of what we are expecting.
When I do "useVolumeSnapshots" i'm expecting to have the snapshots of the corresponding volum and not every snapshot I have (otherwise it would be just "useSnapshot").
So without the select the method doesn't make sense.

And at some point will we have a regionalized api that will do exactly that for us, and the select will go away.

packages/manager/apps/pci-block-storage/src/api/hooks/useSnapshot.ts
```diff
@@ -0,0 +1,52 @@
...
+    () => (volume ? selectSnapshotForVolume(volumeId) : undefined),
+    [volume],
+  );
+
+  return useQuery({
+    queryKey: getVolumeSnapshotsQueryKey(projectId, volume?.region),
```
aTurmo: Issue (blocking) : pass volume id and not volume region

packages/manager/apps/pci-block-storage/public/translations/retype/Messages_fr_FR.json
aTurmo: image

Tsiorifamonjena: **thought** : Maybe you need a react component convention like <Link />

It's depend on how you do it.  We have the same things with `GuidLinkComponent` on µapp pci-gateway, pci-private-network, pci-load-balancer if you want. But it's up to you. It's non blocking :)

packages/manager/apps/pci-block-storage/src/api/hooks/useSnapshot.ts
Tsiorifamonjena: I understand your point. it's not about to exclude select, but to add a possibility to have another select to filter for example

packages/manager/apps/pci-block-storage/public/translations/retype/Messages_fr_FR.json
aTurmo: I looked it up and for the guide, the all translation key is the link so I don't see how it apply here

packages/manager/apps/pci-block-storage/src/api/hooks/useSnapshot.ts
aTurmo: There is no need to do that right now. It's not aiming for being generic.

PR #19188, opened 2025-09-15
fix(ips): fixing the vmac update display

aderghamov:
ref: #MANAGER-19266

## Description

Ticket Reference: #MANAGER-19266


packages/manager/apps/ips/src/pages/listing/listingContext.tsx
```diff
@@ -34,8 +42,16 @@ export const ListingContextProvider = ({ children }: PropsWithChildren) => {
           setExpiredIps((ips) => ips.concat(expiredIp));
         }
       },
+      getInvalidatedVmacForService: (serviceName: string) =>
+        invalidatedVmacData[serviceName],
```
chipp972:
```suggestion
      invalidatedVmacData,
```

No need to create a function, we can use the state object directly

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
```diff
@@ -29,7 +30,10 @@ export const useGetIpVmacWithIp = ({
   serviceName,
   enabled = true,
 }: UseGetIpVmacWithIpParams) => {
-  const [invalidatedData, setInvalidatedData] = useState(undefined);
+  const {
+    getInvalidatedVmacForService,
+    setInvalidatedVmacForService,
+  } = useContext(ListingContext);
```
chipp972: Actually I think it's a bit weird to use the ListingContext in the data part

Data fetching should be agnostic to which part of the app it's displayed in

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
```diff
@@ -79,7 +87,8 @@ export const useGetIpVmacWithIp = ({
   const formattedResult = {
     isLoading:
       isLoading ||
-      !!results.find((result) => !!result.isLoading || !!invalidatedData),
+      !!getInvalidatedVmacForService(serviceName) ||
```
chipp972: `results.some(query => query.isStale || query.isFetching) ||`

I think that achieve the same result and we don't have to store the invalidate state in a Context that's specific to the listing page, maybe
What do you think ?

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
aderghamov: The problem is not about data fetching but about data refreshing (with tanstack).
In order to properly refresh the vmac display, I found out no other solution than tweaking the `refetchInterval` function of the`useGetIpVmacWithIp` 's `useQuery`.

Here is the logic for setting up the `refetchInterval`when a query is triggered: 
 - if the query is due to data invalidation, we store the invalidated data for a given serviceName and the function returns number value (programming another refresh)
 - if the response of the query matches the stored invalidated data, then it means that no update has been finished in the backend. we assume that the update is still processing
 - if the response does not match the invalidated data, then we set the current invalidated data to `undefined` and `refetchInterval`returns `undefined`(no further refetch is expected).

I used a context because we need a state that is scoped to the whole list instead of the `useGetIpVmacWithIp` hook.
When used into the IpVmac component, the `useGetIpVmacWithIp` reinitialized at each render and its state is reset so I cannot properly manage the invalidated data.

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
aderghamov: I'm afraid but that won't work because `isStale` and `isFetching` won't work with an asynchronous update process.
When invalidating the query, another request is performed. however, if the vmac update process takes more than a few milliseconds (which is always the case), then this request will return the data we previously invalidated (so it is useless).

We need to keep the invalidated data in a state to detect if the add/delete vmac is properly done.

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
chipp972: Ok I understand that there is a task executing in the backend and we have to refetch until it's done.
There is not other way to detect the task is currently executing other than comparing to the old data and storing the fact that we did an action ?
There is no status or a task indicating that we have an action executing currently ?

Because there are 2 problems with this approach :
- first we couple the listing page with the vmac call. If this is the only solution (no task, no status) we should have a separate context spcific to the vmac that we put on the listing page, not the other way around.
- second, if we refresh the page, we lose the state and the status

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
chipp972: ok I see

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
aderghamov: I will try to find out a solutions based on tasks.

chipp972: I checked the API
You can poll the task list filtering on function 'virtualMacAdd' and 'virtualMacDelete' and status 'todo', 'init' or 'doing' : https://eu.api.ovh.com/console/?section=%2Fdedicated%2Fserver&branch=v1#get-/dedicated/server/-serviceName-/task

And keep refetching until there is no task returned then refresh the datagrid from there

```
const params = { functionList: ['virtualMacAdd', 'virtualMacDelete'], status: ['todo', 'init', 'doing'] };
useQuery({
  queryKey: getDedicatedServerTasksQueryKey(params)
  queryFn: getDedicatedServerTasks(params),
  retry: false,
  refetchInterval: query => {
    if (query.state.data?.data?.length > 0) {
      return 5000; // (store in a variable and increase if needed)
    }
    queryClient.invalidateQueries({ // vmac queries for the data grid to refresh });
    return undefined;
  }
});
```

And everytime you trigger a vmac action you invalidate the queryKey to retrigger the task polling

Steffy29 (approved):

packages/manager/apps/ips/src/data/hooks/ip/useGetIpVmacWithIp.ts
aderghamov: Thank you for this clue.
I managed to rework this.

packages/manager/apps/ips/src/utils/dataRefresh.constants.ts
```diff
@@ -1 +1,15 @@
...
+};
+
+export const VMAC_UPDATE_TASKS_QUERY_KEY_PARAMS = {
+  ...UPDATE_TASKS_QUERY_KEY_PARAMS,
+  functionList: [
+    'virtualMacAdd',
+    'addVirtualMac',
+    'removeVirtualMac',
+    'virtualMacDelete',
+  ],
```
chipp972: That's specific to vmac

packages/manager/apps/ips/src/pages/actions/virtualMac/deleteVirtualMac.page.tsx
```diff
@@ -13,9 +13,15 @@ import {
 import { useNavigate, useParams } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 import { ApiError } from '@ovh-ux/manager-core-api';
-import { deleteVirtualMACs, getdedicatedServerVmacQueryKey } from '@/data/api';
+import { deleteVirtualMACs } from '@/data/api';
```
chipp972:
```suggestion
import { deleteVirtualMACs, getDedicatedServerTasksQueryKey } from '@/data/api';
```

packages/manager/apps/ips/src/data/hooks/ip/useGetDedicatedServerTasks.ts
```diff
@@ -0,0 +1,66 @@
...
+    ref.current = value;
+  });
+  return ref.current;
+}
+
+// Only for dedicated server for now
+export const useGetDedicatedServerTasks = ({
+  serviceName,
+  functionList,
+  statuses,
```
chipp972: statusList would be more coherent
you named the above parameter functionList not functions

packages/manager/apps/ips/src/data/hooks/ip/useGetDedicatedServerTasks.ts
```diff
@@ -0,0 +1,66 @@
...
+  enabled = true,
+  queryKeysToInvalidate,
```
chipp972: if it's a single queryKey we should name this param queryKeyToInvalidate

packages/manager/apps/ips/src/data/hooks/ip/useGetDedicatedServerTasks.ts
```diff
@@ -0,0 +1,66 @@
...
+      if (queryData.length > 0) {
+        return INVALIDATED_REFRESH_PERIOD;
+      }
+      return undefined;
+    },
+  });
+
+  const taskIds = dedicatedServerTaskResponse ?? [];
+  const hasOnGoingTask = taskIds.length > 0 || isLoading;
+  const hadOnGoingTask = usePrevious(hasOnGoingTask);
```
chipp972: what is the purpose of this ?
Can't you just invalidate the query in the refetchInterval instead ?

packages/manager/apps/ips/src/data/hooks/ip/useAddVirtualMacToIp.tsx
```diff
@@ -1,7 +1,9 @@
 import { useMutation, useQueryClient } from '@tanstack/react-query';
 import { ApiError, ApiResponse } from '@ovh-ux/manager-core-api';
 import { useNotifications } from '@ovh-ux/manager-react-components';
-import { getdedicatedServerVmacQueryKey, addVirtualMacToIp } from '@/data/api';
+import { addVirtualMacToIp } from '@/data/api';
+import { getDedicatedServerTasksQueryKey } from '../../api';
```
chipp972:
```suggestion
import { addVirtualMacToIp, getDedicatedServerTasksQueryKey } from '@/data/api';
```

packages/manager/apps/ips/src/data/hooks/ip/useAddIpToVirtualMac.tsx
```diff
@@ -1,7 +1,9 @@
 import { useMutation, useQueryClient } from '@tanstack/react-query';
 import { ApiError, ApiResponse } from '@ovh-ux/manager-core-api';
 import { useNotifications } from '@ovh-ux/manager-react-components';
-import { getdedicatedServerVmacQueryKey, addIpToVirtualMac } from '@/data/api';
+import { addIpToVirtualMac } from '@/data/api';
+import { getDedicatedServerTasksQueryKey } from '../../api';
```
chipp972:
```suggestion
import { addIpToVirtualMac, getDedicatedServerTasksQueryKey } from '@/data/api';
```

packages/manager/apps/ips/src/data/api/dedicated-server.ts
```diff
@@ -93,3 +98,56 @@ export const getDedicatedServerAvailableCountries = (
...
+};
+
+export const getDedicatedServerTasksQueryKey = ({
+  serviceName,
+  functionList,
+  statuses,
+}: GetDedicatedServerTasksParams) => [
+  `get/dedicated/server/${encodeURIComponent(
+    serviceName,
+  )}/task?function=${functionList}&status=${statuses}`,
```
chipp972:
```suggestion
  )}/task?function=${functionList.join(',')}&status=${statuses.join(',')}`,
```

packages/manager/apps/ips/src/data/api/dedicated-server.ts
```diff
@@ -93,3 +98,56 @@ export const getDedicatedServerAvailableCountries = (
...
+];
+
+export const getDedicatedServerTask = async (
+  serviceName: string,
+  taskfunction: string,
+  status: string,
+): Promise<IcebergFetchResultV6<{ taskId: number }>> => {
+  return fetchIcebergV6<{ taskId: number }>({
```
chipp972: we don't need iceberg here in my opinion since the goal is just to know how many task there is, not displaying and paginating them

packages/manager/apps/ips/src/data/api/dedicated-server.ts
```diff
@@ -93,3 +98,56 @@ export const getDedicatedServerAvailableCountries = (
...
+      getDedicatedServerTask(serviceName, taskFunction, status),
+    ),
+  );
+
+  return results.reduce(
+    (acc: number[], result) =>
+      acc.concat(result.data.map(({ taskId }) => taskId)),
+    [],
+  );
+};
```
chipp972, line 156: Better use getDedicatedServerTask in combination with `useQueries` in `useGetDedicatedServerTasks` instead of doing a Promise.all here in my opinion

packages/manager/apps/ips/src/data/hooks/ip/useGetDedicatedServerTasks.ts
aderghamov: Doing so triggers too many API requests.
This solution enables a detect a state variation detection for the onGoing tasks and triggers the invalidate only when the state transitions from hasTasks to hasNoTasks

packages/manager/apps/ips/src/utils/dataRefresh.constants.ts
aderghamov: OK, I'll add a specific vmac constant file.

packages/manager/apps/ips/src/data/hooks/ip/useGetDedicatedServerTasks.ts
chipp972: I mean when you stop refetching, just before returning `undefined` you invalidate the query for vmac => it means you trigger 1 new API request if there were some tasks (no need to execute `refetchInterval` otherwise) and there is no task anymore (returning `undefined`)
Why would it trigger too many API requests ?

Additionally, `usePrevious` hook looks like a trick and is looks like an imperative way of doing while react-query is declarative so it doesn't fit too well in term of understanding in my opinion

aderghamov: With the full rework of the solution with useQueries, I managed to make it work in the `refetchInterval`.
Thank you for the comments :)

packages/manager/apps/ips/src/data/hooks/ip/useGetDedicatedServerTasks.ts
```diff
@@ -0,0 +1,110 @@
...
+  functionList: string[];
+  statusList: string[];
+  enabled: boolean;
+  queryKeyToInvalidate: string[];
+};
+
+function isUpdateTaskStatus(status: string): boolean {
+  return status
+    ? UPDATE_TASKS_QUERY_KEY_PARAMS.statusList.includes(status)
+    : true;
```
chipp972:
```suggestion
  return !status
    ?? UPDATE_TASKS_QUERY_KEY_PARAMS.statusList.includes(status);
```

packages/manager/apps/ips/src/data/hooks/ip/useGetDedicatedServerTasks.ts
```diff
@@ -0,0 +1,110 @@
...
+}: UseGetDedicatedServerTasksParams) => {
+  const queryClient = useQueryClient();
+
+  const parameterSets: Array<{ taskFunction: string; status: string }> = [];
+
+  functionList.forEach((taskFunction) => {
+    statusList.forEach((status) => {
+      parameterSets.push({ taskFunction, status });
+    });
+  });
```
chipp972: use useMemo here, we don't want to compute this on every re-render in my opinion
Also avoid mutations in hooks and components in general (`.push`)

```
const parameterSets = React.useMemo(() => functionList.flatMap(taskFunction => statusList.map(status => ({taskFunction, status})), []);
```

You could just include the serviceName also to simplify the function below

```
const parameterSets = React.useMemo(() => 
  functionList.flatMap(taskFunction => 
    statusList.map(status => ({taskFunction, status, serviceName})),
[]);
```

packages/manager/apps/ips/src/data/hooks/ip/useGetDedicatedServerTasks.ts
```diff
@@ -0,0 +1,110 @@
...
+      ),
+    }),
+  });
+
+  const hasOnGoingTask =
+    hasOnGoingTaskResult.data || hasOnGoingTaskResult.isLoading;
+
+  return {
+    hasOnGoingTask,
+  };
```
chipp972: Differentiate isLoading and hasOnGoingTask because if you display a message in case there are on going task, the message will be flashing during the loading state then removed a few frames after.

If it's loading, show loading state.
If it's another state, display it accordingly.

packages/manager/apps/ips/src/data/hooks/ip/useGetDedicatedServerTasks.ts
aderghamov: This syntax won't work.
The `??` works only when the left operand can be null or undefined .

chipp972: Yes you're right : here is the corrected syntax:

`return !status || UPDATE_TASKS_QUERY_KEY_PARAMS.statusList.includes(status);`

PR #19190, opened 2025-09-15
fix(pci-instances): allow unattached volume in the same availability zone

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-4975

### DIFF
--- a/packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/__tests__/useDashboardAction.spec.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/__tests__/useDashboardAction.spec.tsx
@@ -1,0 +1,164 @@
+import { describe, it, vi } from 'vitest';
+import { renderHook, waitFor } from '@testing-library/react';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { useUnattachedVolumes } from '../useDashboardAction';
+import { getInstance } from '@/data/api/instance';
+import { TInstance } from '@/types/instance/entity.type';
+import { getVolumes } from '@/data/api/volume';
+
+const fakeInstance = {
+  id: 'fake-instance-id',
+  name: 'fake-instance',
+  region: {
+    name: 'fake-region',
+    type: 'region',
+    availabilityZone: 'fake-zone-a',
+  },
+  volumes: [
+    {
+      id: 'fake-volume-1',
+      name: 'fake-volume',
+      size: null,
+    },
+  ],
+  task: {
+    isPending: false,
+    status: null,
+  },
+  actions: [{}],
+} as TInstance;
+
+const getVolumesMock = vi.fn();
+
+vi.mock('@/data/api/instance');
+vi.mock('@/data/api/volume');
+vi.mocked(getInstance).mockResolvedValue(fakeInstance);
+vi.mocked(getVolumes).mockImplementation(getVolumesMock);
+
+const render = () =>
+  renderHook(
+    () =>
+      useUnattachedVolumes({
+        projectId: 'fake-projectId',
+        region: 'fake-region',
+        instanceId: 'fake-instance',
+      }),
+    {
+      wrapper: ({ children }) => (
+        <QueryClientProvider client={new QueryClient()}>
+          {children}
+        </QueryClientProvider>
+      ),
+    },
+  );
+
+describe('useUnattachedVolumes', () => {
+  it.each([
+    {
+      description: 'should return empty when volumes are already attached',
+      volumes: [
+        {
+          id: 'fake-volume-1',
+          name: 'fake-volume',
+          availabilityZone: 'fake-zone-a',
+        },
+      ],
+      expected: [],
+    },
+    {
+      description:
+        'should return only unattached volumes with availabilityZone any or null',
+      volumes: [
+        {
+          id: 'fake-volume-1',
+          name: 'fake-volume',
+          availabilityZone: 'fake-zone-a',
+        },
+        {
+          id: 'fake-volume-2',
+          name: 'fake-volume-2',
+          availabilityZone: 'fake-zone-b',
+        },
+        {
+          id: 'fake-volume-3',
+          name: 'fake-volume-3',
+          availabilityZone: 'any',
+        },
+        {
+          id: 'fake-volume-4',
+          name: 'fake-volume-4',
+          availabilityZone: null,
+        },
+      ],
+      expected: [
+        { label: 'fake-volume-3', value: 'fake-volume-3' },
+        { label: 'fake-volume-4', value: 'fake-volume-4' },
+      ],
+    },
+    {
+      description:
+        'should return only unattached volumes with the same instance zone',
+      volumes: [
+        {
+          id: 'fake-volume-1',
+          name: 'fake-volume',
+          availabilityZone: 'fake-zone-a',
+        },
+        {
+          id: 'fake-volume-2',
+          name: 'fake-volume-2',
+          availabilityZone: 'fake-zone-a',
+        },
+        {
+          id: 'fake-volume-2',
+          name: 'fake-volume-2',
+          availabilityZone: 'fake-zone-b',
+        },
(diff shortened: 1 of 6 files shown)


packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/view-models/selectUnattachedResource.ts
```diff
@@ -27,11 +27,15 @@ export const selectUnattachedPrivateNetworks = (
...
   instanceVolumes: TInstanceVolume[],
+  instanceAvailabilityZone: string | null,
 ): TUnattachedResource[] =>
   volumes
     .filter(
-      ({ id }) =>
-        !instanceVolumes.some((instanceVolume) => id === instanceVolume.id),
+      ({ id, availabilityZone }) =>
+        !instanceVolumes.some((instanceVolume) => id === instanceVolume.id) &&
+        (availabilityZone === 'any' ||
```
fredericvilcot, line 36: This use case should be documented, because `string | null` does not prevent this case

Tsiorifamonjena: I'm not sure to understand what do you mean on does not prevent the case ? Isn't the test enough for the documentation ? Maybe it's not more clearer and not cover some case :)

fredericvilcot: Type `string | null` does not handle the explicit case `availabilityZone === 'any'`

So how can we know this use case should be taken in account ?

Tsiorifamonjena: You're right. I update the availabilityZone type with `availabilityZone: string | null | 'any'` I add a comment also for understanding 
```ts
type TVolume = {
  id: string;
  name: string;
  size: number;
  availabilityZone: string | null | 'any'; // any is for Volume Any type which means available for any zone
};
```

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/__tests__/useDashboardAction.spec.tsx
```diff
@@ -0,0 +1,164 @@
...
+import { renderHook, waitFor } from '@testing-library/react';
+import { QueryClient, QueryClientProvider } from '@tanstack/react-query';
+import { useUnattachedVolumes } from '../useDashboardAction';
+import { getInstance } from '@/data/api/instance';
+import { TInstance } from '@/types/instance/entity.type';
+import { getVolumes } from '@/data/api/volume';
+
+const fakeInstance = {
+  id: 'fake-instance-id',
+  name: 'fakse-instance',
```
lolottetheclash: you have a typo here ;)

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/__tests__/useDashboardAction.spec.tsx
```diff
@@ -0,0 +1,164 @@
...
+          id: 'fake-volume-2',
+          name: 'fake-volume-2',
+          availabilityZone: 'fake-zone-b',
+        },
+      ],
+      expected: [{ label: 'fake-volume-2', value: 'fake-volume-2' }],
+    },
+    {
+      description:
+        'should return only unattached volumes with availabilityZone any and the same instance zone',
```
lolottetheclash: or null, no?

fredericvilcot (dismissed):

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/__tests__/useDashboardAction.spec.tsx
```diff
@@ -0,0 +1,164 @@
...
+      wrapper: ({ children }) => (
+        <QueryClientProvider client={new QueryClient()}>
+          {children}
+        </QueryClientProvider>
+      ),
+    },
+  );
+
+describe('useUnattachedVolumes', () => {
+  it.each([
```
aTurmo, line 56: Question (non-blocking): why doing it like that ?

I find this way of testing to be harder to read and quite cluttered.

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/__tests__/useDashboardAction.spec.tsx
```diff
@@ -0,0 +1,164 @@
...
+        },
+      ],
+      expected: [
+        { label: 'fake-volume-3', value: 'fake-volume-3' },
+        { label: 'fake-volume-4', value: 'fake-volume-4' },
+      ],
+    },
+    {
+      description:
+        'should return only unattached volumes with the same instance zone',
```
aTurmo, line 100: question (non-blocking): with the same instace zone as what ?
If it's the same zone as the current instance, it's better then to explecitly use the same value like 
```
availabilityZone: fakeInstance.region.availabilityZone
```

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/__tests__/useDashboardAction.spec.tsx
```diff
@@ -0,0 +1,164 @@
...
+          id: 'fake-volume-2',
+          name: 'fake-volume-2',
+          availabilityZone: 'fake-zone-b',
+        },
+      ],
+      expected: [{ label: 'fake-volume-2', value: 'fake-volume-2' }],
+    },
+    {
+      description:
+        'should return only unattached volumes with availabilityZone any and null in the same instance zone',
```
aTurmo, line 122: same thing as previous comment

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/useDashboardAction.ts
```diff
@@ -56,13 +64,98 @@ export const useUnattachedVolumes = ({
...
+
+export const useAttachVolume = ({
+  projectId,
+  region,
+  instanceId,
+}: TUseUnattachedResource) => {
+  const { addError, addSuccess } = useNotifications();
+  const queryClient = useQueryClient();
+  const navigate = useNavigate();
+  const handleModalClose = () => navigate('..');
```
aTurmo, line 90: issue (blocking): Navigation and notification should not be handle here

useAttachVolume should only handle the data/call part that are domain/technical things.
Notification and navigation depend on the caller and are pure front things.

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/__tests__/useDashboardAction.spec.tsx
Tsiorifamonjena: Honestly, I'm divided. Some people would say that it's more readable and allow to avoid implementation duplication since only the expected data will change.

But I see your point, I will take it account later in other PR

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/__tests__/useDashboardAction.spec.tsx
Tsiorifamonjena: It will be cheating I think. Maybe It comes from the definition of `it.each` which makes it confusing to read. I'll fix it as I said above

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/hooks/useDashboardAction.ts
Tsiorifamonjena: It was already in the front but I want to move the business in a separate presentational hook (not domain)
But I can rework it later and we discussed about it.

PR #19193, opened 2025-09-15
feat(manager-react-components): useDataApi hook

anooparveti:
## Description

Addition of useDataApi for data fetching using hexagonal architecture.
Addition of UseQuery and UseInifiniteQuery adapters.
Addition of v6, v2, and Iceberg ports.

Ticket Reference: #MANAGER-19377    

## Additional Information

BREAKING CHANGE: Removal of useResourcesV6, useIceberg and useResourcesV2 hook


tibs245:
I hope this project can help you I worked on this subject on my personnal time. Not hesitate if you want more details on my reflexion and what i learned. https://github.com/tibs245/server-state-manager-adapter

packages/manager-react-components/src/hooks/data-api/__tests__/Test.utils.tsx
```diff
@@ -0,0 +1,119 @@
...
+  );
+};
+
+export type ResultObj = {
+  name: string;
+  number: string;
+  bool: string;
+};
+
+export const items: ResultObj[] = [
```
helabenkhalfallah: it seems like we mixed the concern between mock data and tests utilities, perhaps we need to extract this to a mock data.

packages/manager-react-components/src/hooks/data-api/adapters/use-infinite-query/__tests__/useInfiniteQuery.spec.tsx
```diff
@@ -0,0 +1,105 @@
...
+          retry: false,
+        },
+      },
+    });
+
+    vi.clearAllMocks();
+    (tanstackUseInfiniteQuery as any).mockReturnValue(mockResult);
+  });
+
+  const wrapper = ({ children }: { children: React.ReactNode }) => (
```
helabenkhalfallah: I think we can move this to TestUtils

packages/manager-react-components/src/hooks/data-api/adapters/use-query/__tests__/useQuery.spec.tsx
```diff
@@ -0,0 +1,144 @@
...
+      },
+    });
+
+    altQueryClient = new QueryClient();
+
+    vi.clearAllMocks();
+    (tanstackUseQuery as any).mockReturnValue(mockResult);
+  });
+
+  const wrapper = ({ children }: { children: React.ReactNode }) => (
```
helabenkhalfallah: duplicated code

packages/manager-react-components/src/hooks/data-api/ports/iceberg/__tests__/useIceberg.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+      }),
+    {
+      wrapper: getWrapper(),
+    },
+  );
+};
+
+describe('useIceberg hook', () => {
+  beforeEach(() => {
+    // vi.mocked(fetchWithIceberg).mockImplementation(() => {
```
helabenkhalfallah: we can remove comment

packages/manager-react-components/src/hooks/data-api/ports/iceberg/__tests__/useIceberg.spec.tsx
```diff
@@ -0,0 +1,124 @@
...
+  });
+
+  it('tests default sorting', () => {
+    renderUseIcebergHook({ defaultSorting: getSorting('number', false) });
+    const callArg = vi.mocked(fetchWithIceberg).mock.calls[0]?.[0];
+    expect((callArg as IcebergFetchParams).sortBy).toEqual('number');
+    expect((callArg as IcebergFetchParams).sortOrder).toEqual('ASC');
+  });
+
+  // it('applies sorting', () => {
```
helabenkhalfallah: we can remove commented code (commented not used, not used not commited)

packages/manager-react-components/src/hooks/data-api/ports/v2/__tests__/useV2.spec.tsx
```diff
@@ -0,0 +1,74 @@
...
+    expect((callArg as FetchV2Params).route).toBe('/dedicated/nasha');
+    expect((callArg as FetchV2Params).pageSize).toBe(10);
+  });
+
+  it('does not make API call when enabled=false', () => {
+    renderUseV2Hook({ enabled: false });
+    expect(fetchV2).not.toHaveBeenCalled();
+  });
+
+  // it('fetches next page', async () => {
```
helabenkhalfallah: we can remove commented code

packages/manager-react-components/src/hooks/data-api/adapters/v6/__tests__/useV6.spec.tsx
```diff
@@ -0,0 +1,274 @@
...
+vi.mock('@tanstack/react-query', async () => {
+  const originalModule = await vi.importActual('@tanstack/react-query');
+  return {
+    ...originalModule,
+    useQuery: vi.fn(),
+  };
+});
+
+function fetchNextPage(fn: () => void, numberOfTimes: number) {
+  for (let i = 0; i < numberOfTimes; i += 1) {
```
helabenkhalfallah, line 28: having loop inside a unit tests is not a good practice.

packages/manager-react-components/src/hooks/data-api/ports/v6/__tests__/useV6.spec.tsx
```diff
@@ -0,0 +1,274 @@
...
+
+  it('fetches next page', () => {
+    const { result } = renderUseV6Hook();
+    act(() => {
+      result.current.fetchNextPage();
+    });
+    expect(result.current.flattenData.length).toEqual(20);
+  });
+
+  // it('fetches data with custom queryFn', () => {
```
helabenkhalfallah: we can remove commented code

packages/manager-react-components/src/hooks/data-api/ports/iceberg/useIceberg.type.ts
```diff
@@ -0,0 +1,22 @@
+import { Dispatch, SetStateAction } from 'react';
+import {
+  IcebergFetchResult,
+  IcebergFetchParams,
+} from '@ovh-ux/manager-core-api';
+import { DatagridColumn, ColumnSort } from '../../../../components';
+import { InfiniteData } from '../../adapters/use-infinite-query';
```
fredericvilcot: In a pure hexagonal architecture, ports should never be coupled with adapters as it's the contract between domain and infra layers, that is fully agnostic from the implementation.

Port should describe its needs, adapter implements it :)

packages/manager-react-components/src/hooks/data-api/ports/v6/filter-and-sort-data/filterAndSort.utils.ts
```diff
@@ -0,0 +1,36 @@
+import { FilterTypeCategories } from '@ovh-ux/manager-core-api';
+
+const getBooleanValue = (value: string) => {
+  return value === 'true';
+};
+
+export const comparatorFn = (
```
helabenkhalfallah: in clean code we dont use in functions or variables names part related to programming language like Fn, Boolean, ... we should use meaningful names related to functional aspect to be sure what we're doing and refactoring in future.

packages/manager-react-components/src/hooks/data-api/ports/v6/filter-and-sort-data/filterAndSort.utils.ts
```diff
@@ -0,0 +1,36 @@
...
+  dataType: FilterTypeCategories,
+  a: string,
```
helabenkhalfallah: a and b are not good and meaningful names

packages/manager-react-components/src/hooks/data-api/ports/v6/useV6.ts
```diff
@@ -0,0 +1,85 @@
+import { useState, useEffect, useMemo, useCallback } from 'react';
+import { v6 } from '@ovh-ux/manager-core-api';
+import { UseDataApiResult } from '../../useDataApi.types';
+import { useDataRetrievalOperations } from '../../utils/data-retrieval-operations/useDataRetrievalOperations';
+import { useQuery } from '../../adapters/use-query';
+import { ResourcesV6Params, FetchV6Result } from './v6.type';
+import { useFilterAndSortData } from './filter-and-sort-data/useFilterAndSortData';
+
+export const useV6 = <TData = unknown, TColumn = unknown>({
```
helabenkhalfallah: why not generic type instead of unknown? unknown may cause big changes in future.

packages/manager-react-components/src/hooks/data-api/ports/v6/v6.type.ts
```diff
@@ -0,0 +1,17 @@
+import { TRefetchInterval } from '../../adapters/use-query';
+import { DatagridColumn, ColumnSort } from '../../../../components';
+
+export type FetchV6Result<TData> = {
+  data: TData[];
+};
+
+export type ResourcesV6Params<TData = unknown, TColumn = unknown> = {
```
helabenkhalfallah: same https://github.com/ovh/manager/pull/19193/files#r2351167972

packages/manager-react-components/src/hooks/data-api/ports/v6/v6.utils.ts
```diff
@@ -0,0 +1,10 @@
+import isDate from 'lodash.isdate';
+import { FilterTypeCategories } from '@ovh-ux/manager-core-api';
+
+export const dataType = (data: unknown) => {
```
helabenkhalfallah: same https://github.com/ovh/manager/pull/19193/files#r2351167972

packages/manager-react-components/src/hooks/data-api/useDataApi.types.ts
```diff
@@ -0,0 +1,55 @@
...
+} from '@ovh-ux/manager-core-api';
+import { UseQueryResult, TRefetchInterval } from './adapters/use-query';
+import {
+  InfiniteData,
+  DefinedUseInfiniteQueryResult,
+} from './adapters/use-infinite-query';
+import { FetchV6Result } from './ports/v6/v6.type';
+import { ColumnSort, FilterWithLabel, DatagridColumn } from '../../components';
+
+export type UseDataApiOptions<TData = unknown, TColumn = unknown> = {
```
helabenkhalfallah: a global feedback about using unknown, if we can privilege more generic or exhaustive typing to have a clear idea about the types.

packages/manager-react-components/src/hooks/data-api/ports/iceberg/useIceberg.ts
```diff
@@ -0,0 +1,119 @@
...
+  useInfiniteQuery,
+  InfiniteData,
+  DefinedUseInfiniteQueryResult,
+} from '../../adapters/use-infinite-query';
+import { DEFAULT_PAGE_SIZE } from '../../useDataApi.constants';
+import { useDataRetrievalOperations } from '../../utils/data-retrieval-operations/useDataRetrievalOperations';
+import { UseIcebergParams, SelectData } from './useIceberg.type';
+import { API_V6_MAX_PAGE_SIZE } from './useIceberg.constants';
+
+export const useIceberg = <TData = unknown, TColumn = unknown>({
```
fredericvilcot: Is it intended that the port, aka domain interface, is the exact mirror of tanstack core implementation? 

In an hexagonal architecture, port should only express the way an adapter can enter the domain layer, example: 

```ts

type IcebergPort = {
  getData: (filters: ... , ...) => TData,
  loadMoreData: (...) => TData,
  abort: (signal: ...) => void,
  ...
}
```

and then in the adapter, we could implement those functions :)

```ts
const tanstackIcebergAdapter = (...) => {
  // here the query logic
}

File: packages/manager-react-components/src/hooks/iam/useOvhIam.spec.tsx [hunk 17]
```diff
@@ -36,20 +37,26 @@ const wrapper = ({ children }: { children: ReactNode }) => (
...
-      v2: {
-        get: vi.fn(),
-      },
-    },
-    fetchIcebergV2: mocks.fetchIcebergV2,
+    useDataApi: mocks.useDataApi,
   };
 });
 
+// vi.mock('@ovh-ux/manager-core-api', async (importOriginal) => {
```
helabenkhalfallah: dead code if possible to delete

packages/manager/core/api/src/iceberg.ts
```diff
@@ -125,6 +19,45 @@ export const getRouteWithParams = (route: string, params: URLSearchParams) => {
...
+  const totalCount = parseInt(headers['x-pagination-elements'] as string, 10);
+
+  return {
+    data,
+    totalCount,
+    status,
+  };
+};
+
+// @deprecated Use fetchWithIceberg
```
helabenkhalfallah, line 60: We need to specify from which version is deprecated and perhaps a small example before and after

packages/manager/core/api/src/utils/headers-builder/headersBuilder.ts
```diff
@@ -0,0 +1,130 @@
+import { transformTagsFiltersToQuery } from '../../filters.js';
+import {
+  FilterTypeCategories,
+  Filter,
+  FilterComparator,
+} from '../../types/filters.type.js';
+
+export function icebergFilter(
```
helabenkhalfallah: function name should start with verb

packages/manager/core/api/src/utils/headers-builder/headersBuilder.ts
```diff
@@ -0,0 +1,130 @@
...
+  comparator: FilterComparator,
+  value: string | string[],
+) {
+  const v = encodeURIComponent(String(value || ''));
```
helabenkhalfallah: v is not a good name

packages/manager-react-components/src/hooks/data-api/ports/iceberg/useIceberg.ts
fredericvilcot: BTW, a port cannot be an implementation of anything so I'm not sure to understand what this code is related to ...

packages/manager-react-components/src/hooks/data-api/ports/iceberg/useIceberg.type.ts
helabenkhalfallah: I agree for this point, thanks for reporting it, we should not see adapter here. External cannot call internal we should move inside always (external to internal), to resolve this we can create a common type instead of calling adapter types inside port types.

packages/manager-react-components/src/hooks/data-api/ports/iceberg/useIceberg.ts
helabenkhalfallah: Yes we can also do like you suggest, port expose only types/contracts and then implement in adapters. In our case we consider that tanstack is the full new driver that we want to attach into our kernel. So we dont now how to communicate with it, then we need an exchange port at first. Then we implement the driver on kernel side. Like what we see with Linux Kernel/Driver/Ports.

packages/manager/core/api/src/iceberg.ts
```diff
@@ -125,6 +19,45 @@ export const getRouteWithParams = (route: string, params: URLSearchParams) => {
...
+  page,
+  pageSize,
+  filters,
+  sortBy,
+  sortOrder,
+  disableCache,
+}: IcebergFetchParams): Promise<IcebergFetchResult<TData>> => {
+  const params = new URLSearchParams();
+  const requestHeaders = buildHeaders()
+    .setPaginationMode()
```
aboungnaseng-ovhcloud, line 34: For API V2, you can request the 10 first items with iceberg, but Iceberg dont return in the response the x-pagination-cursor-next. Did you make a test in on with some endpoints ?

aboungnaseng-ovhcloud:
For the adapters, useQuery and useInfiniteQuery, why not using by importing it directly when we need ? or we can not just export it directly ?

packages/manager-react-components/src/hooks/data-api/ports/iceberg/useIceberg.ts
fredericvilcot: I see the point, but if we decide that _tanstack is the full new driver_, so what's the point having a port? 

I mean, the concept of abstraction using ports in an hexagonal architecture is the core concept, so I see multiple issues doing that (in a pure hexa)

- violation of domain independence : business logic becomes dependent on technical details (tanstack.).
- loss of reversibility : hard to replace an adapter (e.g., switch tanstack → Apollo) without affecting the domain.
- blurred responsibilities : ports start looking like adapters → architecture becomes unclear.
- breaks dependency inversion principle : domain depends on technologies instead of technologies depending on the domain.
- testability : domain tests require mocking external tech or loading unnecessary contexts.

packages/manager/core/api/src/iceberg.ts
anooparveti: Since this will be merged along with MUK v1. I cannot anticipate the version of core-api package.

packages/manager/core/api/src/iceberg.ts
anooparveti: Iceberg has 2 modes,
1. Normal pagination
2. Cursor pagination

So while using iceberg, i am restricting to always use normal pagination irrespective of v2 or v6.

packages/manager-react-components/src/hooks/data-api/adapters/v6/__tests__/useV6.spec.tsx
anooparveti: The hook returns a `fetchNextPage` function. For testing the behaviour of exhaustive fetching until there is no data to be fetched, I will have to call the `fetchNextPage` n times.
So to avoid code duplication, I am using a for loop to call as many times as required.

So, this function will replace
```
result.current.fetchNextPage();
result.current.fetchNextPage();
result.current.fetchNextPage();
result.current.fetchNextPage();
```

with 

```
fetchNextPage(result.current.fetchNextPage , 4);
```

packages/manager-react-components/src/hooks/data-api/ports/iceberg/useIceberg.ts
helabenkhalfallah: I see the confusion, I think since we named as ports create many confusions, but all other aspects are preserved since we have this flow:

1) `useDataApi`: is the adapter for the applications to avoid knowing these details:

```
import { useV6 } from './ports/v6/useV6';
import { useV2 } from './ports/v2/useV2';
import { useIceberg } from './ports/iceberg/useIceberg';
```

So if one day we have another version, we changes only `useDataApi`.

In this part we preserve domain independence, reversibility and other criteria.

`useDataApi` is the adapter for applications.
https://github.com/ovh/manager/blob/e9f327d895426546c8f3bda0ebfdad2301f0f3c4/packages/manager-react-components/src/hooks/data-api/useDataApi.ts

2) Then now `useDataApi` to achieve its works needs to talk with many external peripheral: v6, v2 and iceberg.
To talk with them we need to expose ports from the mother-board `useDataApi`. 

`v6, v2 and iceberg` are semi-externals that depends on full-external dependency `tanstack`. 
https://github.com/ovh/manager/blob/e9f327d895426546c8f3bda0ebfdad2301f0f3c4/packages/manager-react-components/src/hooks/data-api/ports/iceberg/useIceberg.ts#L8
https://github.com/ovh/manager/blob/e9f327d895426546c8f3bda0ebfdad2301f0f3c4/packages/manager-react-components/src/hooks/data-api/ports/v2/useV2.ts#L8
https://github.com/ovh/manager/blob/e9f327d895426546c8f3bda0ebfdad2301f0f3c4/packages/manager-react-components/src/hooks/data-api/ports/v6/useV6.ts#L5

3) Now for the full-external dependency `tanstack` we cannot talk with it directly we don't know how to do this, so we need an adapter pattern.

So we have an adapter for an adapter that uses a port.

Perhaps we need to rename ports as drivers? or choose another name to avoid confusion.

helabenkhalfallah: We can make a small meeting if you want :) :)

packages/manager-react-components/src/hooks/data-api/__mocks__/mock.tsx
```diff
@@ -0,0 +1,87 @@
+import { FilterTypeCategories } from '@ovh-ux/manager-core-api';
+import { DatagridColumn } from '../../../components';
+
+export type ResultObj = {
+  name: string;
+  number: string;
+  bool: string;
+};
+
+export const items: ResultObj[] = [
```
helabenkhalfallah, line 10: excellent for putting in a separated file! :)

helabenkhalfallah:
The final structure should be:

```js
src/
 ├── ports/
 │    ├── useDataApi/
 │    │    ├── useDataApi.ts         // <- the port/facade exported to apps
 │    │    ├── useDataApi.types.ts
 │    │    └── index.ts
 │
 ├── adapters/
 │    ├── api/
 │    │    ├── v2/
 │    │    │    └── useV2.ts
 │    │    ├── v6/
 │    │    │    └── useV6.ts
 │    │    └── iceberg/
 │    │         └── useIceberg.ts
 │    ├── tanstack/
 │    │    ├── useQueryAdapter.ts
 │    │    └── useInfiniteQueryAdapter.ts
 │
 └── index.ts
```

- useDataApi acts as the commander and public port: it exposes a stable, technology-agnostic entry point for consumers.
- useV2, useV6, and useIceberg are adapters implementing that port. Each represents a specific API integration or versioning strategy, hidden behind the public interface.
- useQueryAdapter and useInfiniteQueryAdapter belong to the infra layer, bridging the adapters with TanStack Query (external framework dependency).

This separation preserves Hexagonal Architecture boundaries:
- ports/ → what the app can do (interface & orchestration)
- adapters/ → how it’s done (API-specific implementations)
- infra/ (inside adapters) → with what (libraries/frameworks)
- It also enables independent evolution of API versions, testing, and framework replacement without touching the public useDataApi contract.

If we want to push the separation even further for clarity:

```js
src/
 ├── ports/
 │    ├── useDataApi/
 │    │    ├── useDataApi.ts         // <- the port/facade exported to apps
 │    │    ├── useDataApi.types.ts
 │    │    └── index.ts
 │
 ├── infra/
 │    ├── tanstack/
 │    │    ├── useQueryAdapter.ts
 │    │    └── useInfiniteQueryAdapter.ts
 │
 ├── adapters/
 │    ├── api/
 │    │    ├── v2/
 │    │    │    └── useV2.ts
 │    │    ├── v6/
 │    │    │    └── useV6.ts
 │    │    └── iceberg/
 │    │         └── useIceberg.ts
 │
 └── index.ts
```

This variant makes the infra layer explicit, clarifying that adapters depend on infra (not the other way around). It’s the cleanest and most evolvable version.

- useDataApi = commander/port
- useV2/useV6/useIceberg = adapters
- useQueryAdapter/useInfiniteQueryAdapter = infra

I think this structure strikes the **right balance between architectural purity (clear Hexagonal boundaries) and developer simplicity (easy to navigate, test, and evolve)**.

packages/manager-react-components/src/hooks/data-api/ports/use-data-api/useDataApi.ts
```diff
@@ -0,0 +1,52 @@
+import { useV6 } from '../../adapters/v6/useV6';
+import { useV2 } from '../../adapters/v2/useV2';
+import { useIceberg } from '../../adapters/iceberg/useIceberg';
+import { UseDataApiOptions, UseDataApiResult } from './useDataApi.types';
+
+export const useDataApi = <TData = Record<string, unknown>>({
```
fredericvilcot, line 6: Question: `Port` is still depending from adapter + code implementation?

I thought we decided to say that Port is the contract (only) between domain model and outside. With no interaction with impure side (infra)

helabenkhalfallah: Nop it's not depending at all on implementation, it's a commander or a port that will communicate with adapter to do its works. It's also correct, the flow of communication is always from outside to inside. It's already a contract with domain (apps) and outside (infra + different fech types). For me it's correct and a good balance between simplicity, easy to extends and evolution. We can go like this as a first step and if in future we see we need more enhance we can do it. We need to have a good balance between design patterns, usability, extensibility, evolution, ... If it's complexe it'll not be used or extended and create bug.

PR #19213, opened 2025-09-16
feat(pci-kubernetes): enable etcd

ppprevost:
ref: #TAPC-4958

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-kubernetes/src/components/service/ClusterInformation.component.tsx
+++ b/packages/manager/apps/pci-kubernetes/src/components/service/ClusterInformation.component.tsx
@@ -1,6 +1,10 @@
-import { ODS_THEME_COLOR_INTENT } from '@ovhcloud/ods-common-theming';
 import { useContext, useEffect } from 'react';
-import { useParams } from 'react-router-dom';
+
+import { useTranslation } from 'react-i18next';
+
+import { OdsHTMLAnchorElementTarget } from '@ovhcloud/ods-common-core';
+import { ODS_THEME_COLOR_INTENT } from '@ovhcloud/ods-common-theming';
+
 import {
   ODS_ICON_NAME,
   ODS_ICON_SIZE,
@@ -9,7 +13,7 @@ import {
   ODS_TEXT_SIZE,
   ODS_TILE_VARIANT,
 } from '@ovhcloud/ods-components';
-import { OdsHTMLAnchorElementTarget } from '@ovhcloud/ods-common-core';
+
 import {
   OsdsDivider,
   OsdsText,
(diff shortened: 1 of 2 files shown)


lolottetheclash (dismissed):

packages/manager/apps/pci-kubernetes/src/components/service/ClusterInformation.component.tsx
```diff
@@ -54,16 +53,17 @@ export type ClusterInformationProps = {
...
   const { clearNotifications } = useNotifications();
   const has3AZ = use3AZPlanAvailable();
 
-  const planDocumentationLink = PLAN_DOC_LINKS[ovhSubsidiary] ?? PLAN_DOC_LINKS.DEFAULT;
+  const planDocumentationLink =
+    PLAN_DOC_LINKS[ovhSubsidiary as keyof typeof PLAN_DOC_LINKS] ?? PLAN_DOC_LINKS.DEFAULT;
 
   const { data: regionInformations } = useRegionInformations(projectId, kubeDetail.region);
-  useEffect(() => clearNotifications, []);
+  useEffect(() => clearNotifications, [clearNotifications]);
```
aTurmo: Question (maybe-blocking): Isn't clearNotifications a function ?

Then why is it not called inside the useEffect ?

ppprevost: ClearNotification is a clean up function that will be called automatically on unmount.

aTurmo: Oh yes ok

aTurmo:
I think you need to do a rebase because there is also the static analysis commit in there

aTurmo (approved):

packages/manager/apps/pci-kubernetes/src/components/service/ClusterInformation.component.tsx
```diff
@@ -58,20 +61,21 @@ export default function ClusterInformation({
...
-    PLAN_DOC_LINKS[ovhSubsidiary] ?? PLAN_DOC_LINKS.DEFAULT;
+    PLAN_DOC_LINKS[ovhSubsidiary as keyof typeof PLAN_DOC_LINKS] ??
+    PLAN_DOC_LINKS.DEFAULT;
 
   const { data: regionInformations } = useRegionInformations(
     projectId,
     kubeDetail.region,
   );
-  useEffect(() => clearNotifications, []);
+  useEffect(() => clearNotifications, [clearNotifications]);
```
lionel95200x: Think you forgot to call the method

```suggestion
  useEffect(() => clearNotifications(), [clearNotifications]);
```

aTurmo: Ahah I did the same remark :p
Actually when you do it like this the method will be called when unmounting the component

But since we were both tricked by this, maybe we can replace with
```typescript
useEffect(() => () => clearNotifications(), [clearNotifications]);
```
to make it clearer ?

aTurmo (approved):

PR #19214, opened 2025-09-16
feat(pci-kubernetes): add reset kubeconfig for 3az

ppprevost:
ref: #TAPC-4368

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-kubernetes/src/components/service/ClusterManagement.component.tsx
+++ b/packages/manager/apps/pci-kubernetes/src/components/service/ClusterManagement.component.tsx
@@ -1,3 +1,9 @@
+import { useMemo } from 'react';
+
+import { useHref } from 'react-router-dom';
+
+import { useTranslation } from 'react-i18next';
+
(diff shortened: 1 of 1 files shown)


packages/manager/apps/pci-kubernetes/src/components/service/ClusterManagement.component.tsx
```diff
@@ -89,36 +84,30 @@ export default function ClusterManagement({ kubeDetail }: Readonly<ClusterManage
...
-          <TileButton
-            title={t('kube_service_reset_kubeconfig')}
-            isDisabled={isProcessing(kubeDetail?.status)}
-            href={hrefResetClusterConfig}
-          />
-        )}
-
+        <TileButton
+          title={t('kube_service_reset_kubeconfig')}
+          isDisabled={isProcessing(kubeDetail?.status)}
```
lionel95200x: As you repeating multiple time the same assertion you could declare a const before rendering, and reuse it in all file

```suggestion
const isProcessingKube = isProcessing(kubeDetail?.status)
          isDisabled={isProcessing(kubeDetail?.status)}
```

PR #19236, opened 2025-09-17
fix(pci-instances): Dont show divider after the last item

Tsiorifamonjena:
## Description

In this PR I restructure the DashboardTile putting the divider above to avoid conditioning its display for the last item

Ticket Reference: #TAPC-4967


packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/InstanceGeneralInfoBlock.component.tsx
```diff
@@ -166,7 +164,7 @@ const InstanceGeneralInfoBlock: FC = () => {
         </DashboardTileBlock>
       )}
       {instance?.isDeleteEnabled && (
-        <DashboardTileBlock isLoading={isInstanceLoading} withoutDivider>
```
aTurmo, line 173: Question (non-blocking): Removing "withoutDivider" will make the divider visible no ?

Tsiorifamonjena:
> Issue(bocking): unclear objective
> 
> In the PR description you say that you want to put back the divider, but in the pr title and in the commit message it says that you want it removed.
> 
> Obviously you want to put back the divider so you have to change your commit message to avoid confusion

The subject of this PR is really to remove/hide the last divider of the last item



image


In my description, I just try to explain how I do it: move the divider at the top and then no need to do a condition like `withoutDivider=isLastItem`
So the divider will be always display but at the top of the item and it's matches very well with the design

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/InstanceGeneralInfoBlock.component.tsx
Tsiorifamonjena: Yes, it will always be displayed but at the top of item separating with the above item

aTurmo: Oh I get it now. Good job it's a nice solution

aTurmo: But then there two separator. The one you put above the last element, and the one bellow the previous one since"DashboardTileBlock' contains one

aTurmo: I saw that you but the divider at the top in `DashboardTileBlock`, but wont hte first element of the Layout have a non-wanted divider at the top then ?

Tsiorifamonjena: No because I already remove the divider from `DashboardCardLayout`

PR #19273, opened 2025-09-19
feat(observability): generate the observability µApp

yjaaouane:
resolves: #MAOBS-1

## Description

Ticket Reference: #MAOBS-1


packages/manager/apps/observability/src/index.tsx
```diff
@@ -0,0 +1,51 @@
...
+import './vite-hmr';
+
+const init = async (appName: string) => {
+  // Initialize shell context: auth, env, tracking, APIs
+  const context = await initShellContext(appName);
+
+  // Initialize i18n with default namespaces
+  await initI18n({
+    context,
+    reloadOnLocaleChange: true,
```
MarcMonchablon, line 20: In other micro-app, we had to set `reloadOnLocaleChange` to false, maybe it's preferable to do it here too ?

PR #19295, opened 2025-09-22
feat(account-creation): add new language selector menu

sachinrameshn:
ref: #MANAGER-16322

## Description

- Create new Language Menu component in GCJ module 
- Should save the selected language in local-storage
- Utilise this component inside account creation pages.  

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/account-creation/src/components/header/Header.component.tsx
+++ b/packages/manager/apps/account-creation/src/components/header/Header.component.tsx
@@ -1,12 +1,30 @@
+import { useCallback, useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { saveUserLocale } from '@ovh-ux/manager-config';
+import { LanguageMenu } from '@ovh-ux/manager-gcj-module';
 import ovhCloudLogo from '@/assets/logo-ovhcloud.png';
 
 export default function HeaderComponent() {
+  const { i18n } = useTranslation();
+
+  const onLocalUpdate = useCallback(
+    (newLocale: string) => {
+      i18n.changeLanguage(newLocale);
+      saveUserLocale(newLocale);
+    },
+    [i18n],
+  );
+
   return (
     <div
       className="p-4 flex flex-row items-center justify-between gap-4"
       data-testid="header"
     >
       <img src={ovhCloudLogo} alt="ovh-cloud-logo" className="app-logo ml-3" />
+      <LanguageMenu
+        onLocalUpdate={onLocalUpdate}
+        initialLocal={i18n.language || 'en_GB'}
+      />
     </div>
   );
 }
--- a/packages/manager/modules/gcj/src/components/LanguageMenu/LanguageMenu.tsx
+++ b/packages/manager/modules/gcj/src/components/LanguageMenu/LanguageMenu.tsx
@@ -1,0 +1,71 @@
+import { useCallback, useEffect, useMemo, useState } from 'react';
+import { useMediaQuery } from 'react-responsive';
+import { ODS_BUTTON_VARIANT } from '@ovhcloud/ods-components';
+import { LANGUAGES } from '@ovh-ux/manager-config';
+import { ActionMenu } from '@ovh-ux/manager-react-components';
+import { SMALL_DEVICE_MAX_SIZE } from '../../constants/constants';
+
+interface KeyPairName {
+  name: string;
+  key: string;
+}
+
+export type Props = {
+  onLocalUpdate(locale: string): void;
+  initialLocal?: string;
+};
+
+export const LanguageMenu = ({ onLocalUpdate, initialLocal = '' }: Props) => {
+  const isSmallDevice = useMediaQuery({
+    query: `(max-width: ${SMALL_DEVICE_MAX_SIZE})`,
+  });
+
+  const [currentLocale, setCurrentLocale] = useState<string>(initialLocal);
+  const [availableLanguages, setAvailableLanguages] = useState<KeyPairName[]>(
+    [],
+  );
+
+  useEffect(() => {
+    setAvailableLanguages(
+      LANGUAGES.available.filter(({ key }) => key !== currentLocale),
+    );
+  }, [currentLocale]);
+
+  const onLocaleChange = useCallback(
+    (newLocale: string) => {
+      onLocalUpdate(newLocale);
+      setCurrentLocale(newLocale);
+    },
+    [onLocalUpdate],
+  );
+
+  const currentLanguageLabel = () => {
+    if (!currentLocale) return '';
+    const availableLanguage = LANGUAGES.available.find(
+      (language) => language.key === currentLocale,
+    );
+    if (!availableLanguage?.key) return '';
+    return isSmallDevice
+      ? availableLanguage.key.slice(-2)
+      : availableLanguage.name;
+  };
+
+  const items = useMemo(
+    () =>
+      availableLanguages.map((language, index) => ({
+        label: language.name,
+        onClick: () => onLocaleChange(language.key),
+        id: index,
+      })),
+    [availableLanguages],
+  );
+
+  return (
+    <ActionMenu
+      id="language-menu"
+      items={items}
+      label={currentLanguageLabel()}
+      variant={ODS_BUTTON_VARIANT.ghost}
+    />
+  );
+};
--- a/packages/manager/apps/account-creation/src/i18n.ts
+++ b/packages/manager/apps/account-creation/src/i18n.ts
@@ -1,3 +1,4 @@
+import { LANGUAGES } from '@ovh-ux/manager-config';
 import i18n from 'i18next';
 import I18NextHttpBackend from 'i18next-http-backend';
 import { initReactI18next } from 'react-i18next';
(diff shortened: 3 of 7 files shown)


packages/manager/modules/gcj/src/components/LanguageMenu/LanguageMenu.tsx
```diff
@@ -0,0 +1,69 @@
...
+  userLocale?: string;
+};
+
+export const LanguageMenu = ({ setUserLocale, userLocale = '' }: Props) => {
+  const { i18n } = useTranslation();
+  const isSmallDevice = useMediaQuery({ query: `(max-width: 30em)` });
+
+  const [currentLanguage, setCurrentLanguage] = useState<KeyPairName | null>(
+    null,
+  );
```
JacquesLarique: You can directly initialize `currentLanguage` with the `userLocale`
`const [currentLocale, serCurrentLocale] = useState<string>(userLocale);`
And only updating it inside this component (not from the component that use this component) in order to avoid re-rendering the LanguageMenu component each time you change the locale (also the code using the LanguageMenu component don't need to keep track of the locale using a state)

_Also renaming `userLocale` by `initialLocal` and `currentLanguage` by `currentLocale` could make it clearer_

packages/manager/modules/gcj/src/components/LanguageMenu/LanguageMenu.tsx
```diff
@@ -0,0 +1,69 @@
...
+  };
+
+  const getLanguageLabel = () => {
+    if (currentLanguage) {
+      return isSmallDevice
+        ? currentLanguage.key.slice(-2)
+        : currentLanguage.name;
+    }
+    return '';
+  };
```
JacquesLarique: You should rather use a `useMemo` here, which with the changes I suggest [here](https://github.com/ovh/manager/pull/19295#discussion_r2367189128) would look like:
```typescript
const currentLanguageLabel = useMemo(() => {
    if (!currentLocale) return '';
    const availableLanguage = LANGUAGES.available.find((language) => language.key === currentLocale);
    if (!availableLanguage?.label) return '';
    return isSmallDevice
        ? availableLanguage.label.slice(-2)
        : availableLanguage.label;
}, [currentLocale]);
```

packages/manager/modules/gcj/src/components/LanguageMenu/LanguageMenu.tsx
```diff
@@ -0,0 +1,69 @@
...
+  // Effect to update available languages and set the current language
+  useEffect(() => {
+    const currentLang = LANGUAGES.available.find(
+      ({ key }) => key === userLocale,
+    );
+    setCurrentLanguage(currentLang || null);
+    setAvailableLanguages(
+      LANGUAGES.available.filter(({ key }) => key !== userLocale),
+    );
+  }, [userLocale]);
```
JacquesLarique: I would use a `useMemo` here to fill `availableLanguages`, which with the changes I suggest [here](https://github.com/ovh/manager/pull/19295#discussion_r2367189128) would look like:
```typescript
const availableLanguages = useMemo(
    () => LANGUAGES.available.filter(({ key }) => key !== currentLocale),
    [currentLocale],
);
```

packages/manager/modules/gcj/src/components/LanguageMenu/LanguageMenu.tsx
```diff
@@ -0,0 +1,69 @@
...
+
+  const onLocaleChange = (newLocale: string) => {
+    i18n.changeLanguage(newLocale);
+    setUserLocale(newLocale);
+    saveUserLocale(newLocale);
+  };
```
JacquesLarique: As stated in [this comment](https://github.com/ovh/manager/pull/19295#discussion_r2367189128), I think you should just trigger the `setUserLocale` (callback prop) and `setCurrentLocale` (component state update) here and let the component using `LanguageMenu` make the calls to `i18n.changeLanguage` and `saveUserLocale` (this way if we change i18n librairy we could easily continue to use this component)
Also you should encapsulate this function in a `useCallback`, giving something like:
```typescript
const onLocaleChange = useCallback((newLocale) => {
    setUserLocale(newLocale);
    setCurrentLocal(newLocale)
},[setUserLocale]);
```

packages/manager/modules/gcj/src/components/LanguageMenu/LanguageMenu.tsx
```diff
@@ -0,0 +1,69 @@
...
+        : currentLanguage.name;
+    }
+    return '';
+  };
+
+  const items = availableLanguages.map((language, index) => ({
+    label: language.name,
+    onClick: () => onLocaleChange(language.key),
+    id: index,
+  }));
```
JacquesLarique: A small performance improvement here would be to use a `useMemo`:
```typescript
  const items = useMemo(() =>
    availableLanguages.map((language, index) => ({
        label: language.name,
        onClick: () => onLocaleChange(language.key),
        id: index,
      }), [availableLanguages]);
```

packages/manager/apps/account-creation/src/components/header/Header.component.tsx
JacquesLarique, line 1: A few changes would be necessary in this file after handling the comments on https://github.com/ovh/manager/pull/19295/files#diff-158dbc35497d7cbeee0095c9220668ced185f82b0e7180889f4c51a140c98d51

packages/manager/apps/account-creation/src/components/header/Header.component.tsx
```diff
@@ -1,12 +1,28 @@
+import { useEffect, useState } from 'react';
+import { useTranslation } from 'react-i18next';
+import { saveUserLocale } from '@ovh-ux/manager-config';
+import { LanguageMenu } from '@ovh-ux/manager-gcj-module';
 import ovhCloudLogo from '@/assets/logo-ovhcloud.png';
 
 export default function HeaderComponent() {
+  const { i18n } = useTranslation();
+  const [initialLocal, setInitialLocal] = useState(i18n.language || 'en_GB');
```
JacquesLarique: There is no need to keep track of the locale here, you should just give `i18n.language` to the `LanguageMenu` and define a callback to be executed when the `LanguageMenu` signals for a change:
```typescript
  const onLocaleChange = useCallback(() => {
    i18n.changeLanguage(initialLocal);
    saveUserLocale(initialLocal);
  }, [i18n]);
```

packages/manager/modules/gcj/src/components/LanguageMenu/LanguageMenu.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const isSmallDevice = useMediaQuery({ query: `(max-width: 30em)` });
+
+  const [currentLocale, setCurrentLocale] = useState<string>(initialLocal);
+  const [availableLanguages, setAvailableLanguages] = useState<KeyPairName[]>(
+    [],
+  );
+
+  // Effect to update available languages and set the current language
+  useEffect(() => {
+    setCurrentLocale(currentLocale);
```
JacquesLarique: There is no need to `setCurrentLocale` here since if we enter this `useEffect` it means we just updated it

packages/manager/modules/gcj/src/components/LanguageMenu/LanguageMenu.tsx
```diff
@@ -0,0 +1,70 @@
...
+import { LANGUAGES } from '@ovh-ux/manager-config';
+import { ActionMenu } from '@ovh-ux/manager-react-components';
+
+interface KeyPairName {
+  name: string;
+  key: string;
+}
+
+export type Props = {
+  setInitialLocal(locale: string): void;
```
JacquesLarique: Rename `setInitialLocal` to `onLocalUpdate` for more explicit naming and easier readability

packages/manager/modules/gcj/src/components/LanguageMenu/LanguageMenu.tsx
```diff
@@ -0,0 +1,68 @@
...
+  key: string;
+}
+
+export type Props = {
+  onLocalUpdate(locale: string): void;
+  initialLocal?: string;
+};
+
+export const LanguageMenu = ({ onLocalUpdate, initialLocal = '' }: Props) => {
+  const isSmallDevice = useMediaQuery({ query: `(max-width: 30em)` });
```
aottr: The value `30em` should be in a constants file as e.g. `SMALL_DEVICE_MAX_SIZE`.

packages/manager/modules/gcj/src/components/LanguageMenu/LanguageMenu.tsx
```diff
@@ -0,0 +1,68 @@
...
+    [onLocalUpdate],
+  );
+
+  const currentLanguageLabel = useMemo(() => {
+    if (!currentLocale) return '';
+    const availableLanguage = LANGUAGES.available.find(
+      (language) => language.key === currentLocale,
+    );
+    if (!availableLanguage?.key) return '';
+    return isSmallDevice
```
aottr, line 48: I don't like the use of react-response just for this but the request is very specific and the component from MRC very limiting. I'd have preferred regular media-queries in here (which would be possible by creating the Menu twice and blending one out). Debatable. So I suppose its fine ^^

sachinrameshn: Since MRC wont support label switching, I suppose this is cleaner way with readable code. 
In this case the trade-off is acceptable.

aottr (approved):

PR #19312, opened 2025-09-22
fix(pci-instance): react - fix banner about instances being created

aTurmo:
ref: #TAPC-5090

Small className changes t ofix this problem:
image

The info banner was folling the scroll instead of beong fixed (problem not happening in local).


manoncarbonnel:
*praise*: simple but effective

PR #19313, opened 2025-09-22
fix(pci-instance): react - fix trad and increase polling interval

aTurmo:
ref: #TAPC-5089


packages/manager/apps/pci-instances/src/data/hooks/instance/polling/useInstancesPolling.ts
```diff
@@ -13,7 +13,7 @@ import { instancesQueryKey, isApiErrorResponse } from '@/utils';
 import { TInstance } from '@/types/instance/entity.type';
 import { useOperations } from '@/data/hooks/instance/operation/useOperations';
 
-const POLLING_INTERVAL = 3000;
+const POLLING_INTERVAL = 6_000; // In milliseconds
```
fredericvilcot: This affects global polling in Datagrid (tasks state) which has been set to 3000ms, who decided to change this setting?

packages/manager/apps/pci-instances/src/pages/instances/datagrid/hooks/useDatagridOperationsPolling.ts
```diff
@@ -0,0 +1,60 @@
...
+    if (!operations) return undefined;
+
+    const hasDistantBackupInCopy = operations
+      .filter(isInstanceCreationOperationInProgress)
+      .flatMap((operation) => operation.subOperations)
+      .some(isSubOperationCopyImageInProgress);
+
+    setHasDistantBackupCopyRunning((hadOperationRunning) => {
+      if (hadOperationRunning && !hasDistantBackupInCopy && onComplete) {
+        onComplete();
```
fredericvilcot, line 42: You should never invoke a side effect on a setter action, mainly because: 

- setter is async
- setter has been designed to return a new state, not to produce side effects
- you don't know how many times setter will be called and when

fredericvilcot (changes requested):

packages/manager/apps/pci-instances/src/pages/instances/datagrid/hooks/useDatagridOperationsPolling.ts
aTurmo: fixed

PR #19339, opened 2025-09-23
feat(communication): add tracking to communications listing and detail

aottr:
ref: #MANAGER-18662

## Description

Ticket Reference: #MANAGER-18662

### DIFF
--- a/packages/manager/apps/communication/src/components/notificationContactStatus/NotificationContactStatus.component.tsx
+++ b/packages/manager/apps/communication/src/components/notificationContactStatus/NotificationContactStatus.component.tsx
@@ -21,7 +21,6 @@ export default function NotificationContactStatus({
   notificationId: string;
 }) {
   const { t } = useTranslation('communications');
-
   const highestPriorityContact = useMemo(
     () =>
       contacts.reduce((highest, current) => {
--- a/packages/manager/apps/communication/src/hooks/useTracking/useTracking.tsx
+++ b/packages/manager/apps/communication/src/hooks/useTracking/useTracking.tsx
@@ -1,0 +1,103 @@
+import {
+  ShellContext,
+  getPageProps,
+  getClickProps,
+  TrackingPageParams,
+  TrackingClickParams,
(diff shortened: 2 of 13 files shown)


packages/manager/apps/communication/src/components/notificationContactStatus/NotificationContactStatus.component.tsx
```diff
@@ -53,8 +60,24 @@ export default function NotificationContactStatus({
...
       />
-      <OdsPopover triggerId={`${notificationId}-contact-status`}>
+      <OdsPopover
+        triggerId={`${notificationId}-contact-status`}
+        onOdsShow={() =>
+          trackPage({
+            pageType: PageType.popup,
+            pageName: 'see_warning-subject',
+          })
+        }
```
JacquesLarique: If I'm not wrong we decided not to track this ("Point 2 : That's a good point, so we won't track the "warning" icon.")

aottr: fair, I'll remove both then. I initially thought it was mainly about the banner type.

packages/manager/apps/communication/src/hooks/useTracking/useTracking.tsx
```diff
@@ -0,0 +1,103 @@
...
+    subApp,
+    location,
+    buttonType,
+    actions,
+    actionType,
+  }: CommunicationTrackingClickParams) => {
+    shell.tracking.trackClick(
+      getClickProps({
+        ...TRACKING_CONTEXT,
+        ...pageTracking,
```
JacquesLarique, line 84: Wouldn't this break if `pageTracking` is `undefined` ? (it is possible if if `useMatches` returns an empty array)

aottr: no thats fine, `...undefined` is equal to deconstructing an empty object.

ex:
```js
console.log({...undefined})
> Object { }
```

JacquesLarique (approved):

PR #19371, opened 2025-09-24
feat(iam): add 'advanced search' modal on myPolicies page

MarcMonchablon:
## Description

Ticket Reference: #MAIAM-33


packages/manager/modules/iam/src/components/advancedPolicySearch/advancedPolicySearch.component.js
```diff
@@ -0,0 +1,11 @@
+import controller from './advancedPolicySearch.controller';
+import template from './advancedPolicySearch.template.html';
+
+export default {
+  bindings: {
+    goBack: '<',
+  },
+  controller,
+  template,
+  transclude: true,
```
darsene: Not needed
```suggestion
```

packages/manager/modules/iam/src/components/advancedPolicySearch/advancedPolicySearch.controller.js
```diff
@@ -0,0 +1,218 @@
...
+      actions: { selection: [], isWildcardActive: false },
+      permissionsGroups: [],
+    };
+
+    this.localUsers = { status: 'not-loaded', search: '', items: [] };
+    this.userGroups = { status: 'not-loaded', search: '', items: [] };
+    this.urnSearch = { search: '' };
+
+    this.filterLocalUsersFn = this.filterLocalUsers.bind(this);
+    this.filterUserGroupsFn = this.filterUserGroup.bind(this);
```
darsene, line 39: Is it really necessary ?

packages/manager/modules/iam/src/components/advancedPolicySearch/advancedPolicySearch.controller.js
```diff
@@ -0,0 +1,218 @@
...
+    return this.$q.when(true);
+  }
+
+  trackClick() {
+    this.testTrackClick = true; // TODO
+  }
+
+  trackPage() {
+    this.testTrackPage = true; // TODO
+  }
```
darsene: To remove ?

packages/manager/modules/iam/src/components/advancedPolicySearch/advancedPolicySearch.controller.js
MarcMonchablon: I think so. it's not a $filter in the angularJS sense, but a method to pass to ng-repeat. Since the one I need to pass need to use other classes attributes, we need to bind it to the current class.

I've looked in the project for other instance of a statefull filter function passed as ng-repeat 'filter', and they also do `this.fooFn = this.foo.bind(this)`.

PR #19372, opened 2025-09-24
chore(mrc): add test-id to text-cell component

mathieumousnier:
## Description

Ticket Reference: #19243

We are adding test ids to some components for end-to-end tests.
We need to add many of them on datagrid cell elements.


packages/manager-react-components/src/components/datagrid/text-cell.component.tsx
```diff
@@ -3,14 +3,20 @@ import { OdsText } from '@ovhcloud/ods-components/react';
...
 export function DataGridTextCell({
   className,
   children,
-}: React.PropsWithChildren<{ className?: string }>) {
+  'data-testid': dataTestId,
+}: React.PropsWithChildren<DataGridTextCellProps>) {
   const { t } = useTranslation('datagrid');
   return (
-    <OdsText preset={ODS_TEXT_PRESET.span}>
+    <OdsText preset={ODS_TEXT_PRESET.span} data-testid={dataTestId}>
```
anooparveti, line 19: question: Can it wait for MUKv1, because with ODS v19 you can directly use utils getByText and so on.And this becomes redundant.

mathieumousnier: This is for e2e tests with playwright, not unit test.
Our QA engineer wants to be able to get data on different rows of a datagrid by pointing test-ids like `[ressource-id]-urn` for example

PR #19379, opened 2025-09-25
fix(ips): export ips to csv fixes

chipp972:
ref: #MANAGER-18928

## Description

Ticket Reference: #MANAGER-18928   

- fix losing query parameters when navigating in and out of export csv modal
- fix error displaying on the listing page instead of inside the modal (since we don't close the modal on error)
- fix serviceName parameter when filtering to parked IPs

### DIFF
--- a/packages/manager/apps/ips/src/data/hooks/useExportIpToCsv.ts
+++ b/packages/manager/apps/ips/src/data/hooks/useExportIpToCsv.ts
@@ -1,6 +1,7 @@
 import { useTranslation } from 'react-i18next';
 import { mkConfig, generateCsv, download } from 'export-to-csv';
 import { useMutation } from '@tanstack/react-query';
+import { ApiError } from '@ovh-ux/manager-core-api';
 import { TRANSLATION_NAMESPACES } from '@/utils';
 import { getIpExport, GetIpListParams } from '../api';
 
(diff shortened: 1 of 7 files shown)


packages/manager/apps/ips/src/data/hooks/useExportIpToCsv.ts
```diff
@@ -32,7 +33,10 @@ export const useExportIpToCsv = ({
   const handleExportToCsv = async (): Promise<ExportIpMutationData> => {
     const result = await getIpExport({
       ...apiFilter,
-      serviceName: apiFilter['routedTo.serviceName'],
+      serviceName:
+        apiFilter['routedTo.serviceName'] === null
+          ? '_PARK'
+          : apiFilter['routedTo.serviceName'],
```
aderghamov, line 39:
```suggestion
      serviceName: apiFilter['routedTo.serviceName'] ?? '_PARK',
```

aderghamov (approved):

packages/manager/apps/ips/src/data/hooks/useExportIpToCsv.ts
chipp972: It will break the logic because:
null => filter for IPs routed to parking
undefined => no filter on service (it will filter to parking also with this change)
any other value => filter for service name

PR #19381, opened 2025-09-25
feat(pci-instances): display message when mutli attach is not possible

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-5083

## Additional Information

image

### DIFF
--- a/packages/manager/apps/pci-instances/public/translations/actions/Messages_fr_FR.json
+++ b/packages/manager/apps/pci-instances/public/translations/actions/Messages_fr_FR.json
@@ -74,5 +74,6 @@
   "pci_instances_actions_instance_volume_attach_select": "Selectionner un volume existant",
   "pci_instances_actions_instance_volume_attach_information": "Vous pouvez attacher un volume existant ou créer un nouveau volume à attacher à votre instance. ",
   "pci_instances_actions_instance_volume_attach_create_link": "Créer un volume",
-  "pci_instances_actions_instance_volume_attach_success_message": "Le volume {{ volume }} a été attaché à l'instance {{ instance }}."
+  "pci_instances_actions_instance_volume_attach_success_message": "Le volume {{ volume }} a été attaché à l'instance {{ instance }}.",
+  "pci_instances_actions_instance_volume_attach_detach_warning_message": "Ce volume est déjà attaché à une ou plusieurs instances. Il sera détaché de l'instance sur laquelle il est actuellement attaché."
 }
--- a/packages/manager/apps/pci-instances/src/data/api/mapper/volume.mapper.ts
+++ b/packages/manager/apps/pci-instances/src/data/api/mapper/volume.mapper.ts
@@ -1,0 +1,7 @@
+import { TVolumeDto } from '@/types/volume/api.type';
+import { TVolume } from '@/types/volume/entity.type';
+
+export const mapDtoToVolume = (dto: TVolumeDto): TVolume => ({
+  ...dto,
+  attachedInstances: dto.attachedTo.map(({ id }) => id),
+});
--- a/packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/action/AttachVolume.page.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/action/AttachVolume.page.tsx
@@ -7,6 +7,7 @@ import { useAttachVolume } from '../hooks/useDashboardAction';
 import { useProjectId } from '@/hooks/project/useProjectId';
 import VolumeSelector from '../components/VolumeSelector.component';
 import { useInstanceParams } from '@/pages/instances/action/hooks/useInstanceActionModal';
+import { TUnattachedVolume } from '../view-models/selectUnattachedResource';
 
 const AttachVolume: FC = () => {
   const { t } = useTranslation('actions');
@@ -14,7 +15,7 @@ const AttachVolume: FC = () => {
   const { instanceId, region } = useInstanceParams();
   const navigate = useNavigate();
   const handleModalClose = () => navigate('..');
-  const [volumeId, setVolumeId] = useState<string>('');
+  const [volume, setVolume] = useState<TUnattachedVolume | null>(null);
 
   const { isPending, volumes, attachVolume } = useAttachVolume({
     projectId,
@@ -22,12 +23,18 @@ const AttachVolume: FC = () => {
     instanceId,
   });
 
+  const handleAttachVolume = () => {
+    if (!volume) return;
+
+    attachVolume({ volumeId: volume.value });
+  };
+
   return (
     <Modal
       title={t('pci_instances_actions_instance_volume_attach_title')}
       isPending={isPending}
-      disabled={!volumeId}
-      handleInstanceAction={() => attachVolume({ volumeId })}
+      disabled={!volume}
+      handleInstanceAction={handleAttachVolume}
       onModalClose={handleModalClose}
       variant="primary"
     >
@@ -36,7 +43,11 @@ const AttachVolume: FC = () => {
           <Spinner />
         </div>
       ) : (
-        <VolumeSelector volumes={volumes} onValueChange={setVolumeId} />
+        <VolumeSelector
+          volume={volume}
+          volumes={volumes}
+          onValueChange={setVolume}
+        />
       )}
     </Modal>
   );
--- a/packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/VolumeSelector.component.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/VolumeSelector.component.tsx
@@ -1,4 +1,4 @@
-import { FC } from 'react';
+import { FC, useCallback } from 'react';
 import { useTranslation } from 'react-i18next';
 import {
   FormField,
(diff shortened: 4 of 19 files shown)


packages/manager/apps/pci-instances/public/translations/actions/Messages_fr_FR.json
```diff
@@ -74,5 +74,6 @@
   "pci_instances_actions_instance_volume_attach_select": "Selectionner un volume existant",
   "pci_instances_actions_instance_volume_attach_information": "Vous pouvez attacher un volume existant ou créer un nouveau volume à attacher à votre instance. ",
   "pci_instances_actions_instance_volume_attach_create_link": "Créer un volume",
-  "pci_instances_actions_instance_volume_attach_success_message": "Le volume {{ volume }} a été attaché à l'instance {{ instance }}."
+  "pci_instances_actions_instance_volume_attach_success_message": "Le volume {{ volume }} a été attaché à l'instance {{ instance }}.",
+  "pci_instances_actions_instance_volume_attach_detach_message": "Ce volume est déjà attaché à une ou plusieurs instances. Il sera détaché de l'instance sur laquelle il est actuellement attaché."
```
aTurmo: Nitpick :
```suggestion
  "pci_instances_actions_instance_volume_attach_detach_warning_message": "Ce volume est déjà attaché à une ou plusieurs instances. Il sera détaché de l'instance sur laquelle il est actuellement attaché."
```

packages/manager/apps/pci-instances/src/data/api/mapper/volume.mapper.ts
```diff
@@ -0,0 +1,9 @@
+import { TVolumeDto } from '@/types/volume/api.type';
+import { TVolume } from '@/types/volume/entity.type';
+
+export const mapDtoToVolume = (dto: TVolumeDto): TVolume => ({
+  ...dto,
+  attachedInstances: dto.attachedTo.map(({ id }) => id),
+  availabilityZone:
+    dto.availabilityZone === 'any' ? null : dto.availabilityZone, // any is for Volume Any type which means available for any zone then it is similar to null
```
aTurmo: Question:
I see that the availability zone in the dto can be null. And here you say that any is the same as null, but is it ?
Why it's different from back end side ?

I understant that for your usecase it's the same thing, but can you garanty that it will be the case everytime ?

Furthermore, if you need to comment something like that, it probably means that it's not obvious, meaning that the code can be improve tso that the comment is not needed.

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/VolumeSelector.component.tsx
```diff
@@ -28,6 +28,18 @@ const VolumeSelector: FC<TVolumeSelectorProps> = ({
...
   const createVolumeUrl = `${projectUrl}/storages/blocks`;
+  const [toBeDetach, setToBeDetach] = useState(false);
+
+  const handleVolumeChange = useCallback(
+    ({ value }: { value: string[] }) => {
+      const id = value[0] ?? '';
+      const volume = volumes.find((item) => item.value === id);
+
+      onValueChange(id);
+      setToBeDetach(!!volume?.toBeDetach);
```
aTurmo: Suggestion:
It seems a little wierd to do it like that. I think it would make more sens that you pass the seldcted volume in the props (since the component is a selector it makes sens). And then you can have directly the volume.toBeDetach

packages/manager/apps/pci-instances/src/data/api/mapper/volume.mapper.ts
Tsiorifamonjena: I update the DTO type. Now it's `string | null | 'any'`. 
But yes null and any will always have the same behaviour :
- null is for 1AZ
- any is for 3AZ which mean available for any zone

I dont see another way to document it than putting it in comments. We have already a discussion about it here https://github.com/ovh/manager/pull/19190#discussion_r2349231104 but I'm open if you an idea

The idea of the mapper is to manipulate just two type without depending the region.

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/VolumeSelector.component.tsx
Tsiorifamonjena: I made some changes, I let you see if that seems clear to you

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/VolumeSelector.component.tsx
```diff
@@ -14,21 +14,32 @@ import {
...
 const VolumeSelector: FC<TVolumeSelectorProps> = ({
+  volume,
   volumes,
   onValueChange,
 }) => {
   const { t } = useTranslation('actions');
   const projectUrl = useProjectUrl('public-cloud');
   const createVolumeUrl = `${projectUrl}/storages/blocks`;
 
+  const handleVolumeChange = useCallback(
```
aTurmo, line 35: Very nice :)

packages/manager/apps/pci-instances/src/types/volume/api.type.ts
```diff
@@ -1,3 +1,16 @@
...
+  id: string;
+};
+
+export type TAnyZone = 'any';
+
+export type TVolumeDto = {
+  id: string;
+  name: string;
+  size: number;
+  availabilityZone: string | null | TAnyZone; // TAnyZone means available for any zone then it is similar to null for 1AZ
```
aTurmo: Nitpick : still not a fan of comment ^^

packages/manager/apps/pci-instances/src/data/api/mapper/volume.mapper.ts
aTurmo: I was not clear enough my bad ^^'
I think it's best to keep 'any' and 'null' separated because event if in your case it's the same thing, in the futur it's possible that it will not be the case

Tsiorifamonjena: I understand your point. I let it to be handle by the selector like before so :)

lolottetheclash (dismissed):

packages/manager/apps/pci-instances/src/types/volume/common.type.ts
```diff
@@ -1,6 +1,3 @@
-export type TVolume = {
-  id: string;
-  name: string;
-  size: number;
-  availabilityZone: string | null | 'any'; // any is for Volume Any type which means available for any zone
-};
+export type TStatus = 'available' | 'in-use' | 'error';
+
+export type TAnyZone = 'any';
```
fredericvilcot: Is this alias needed?

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/view-models/selectUnattachedResource.ts
```diff
@@ -22,20 +23,35 @@ export const selectUnattachedPrivateNetworks = (
     )
     .map((network) => ({ label: network.name, value: network.id }));
 
+type TInstance = {
```
fredericvilcot: Be careful to type scope, as `TInstance` is declared several times in µapp

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/view-models/selectUnattachedResource.ts
```diff
@@ -22,20 +23,35 @@ export const selectUnattachedPrivateNetworks = (
...
+  id: string;
+  region: TInstanceRegion;
+};
+
+export type TUnattachedVolume = TUnattachedResource & {
+  canBeDetach: boolean;
```
fredericvilcot:
```suggestion
  canBeDetached: boolean;
```

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/action/AttachVolume.page.tsx
```diff
@@ -7,27 +7,34 @@ import { useAttachVolume } from '../hooks/useDashboardAction';
...
-  const [volumeId, setVolumeId] = useState<string>('');
+  const [volume, setVolume] = useState<TUnattachedVolume | null>(null);
 
   const { isPending, volumes, attachVolume } = useAttachVolume({
     projectId,
     region,
     instanceId,
   });
 
+  const handleAttachVolume = useCallback(() => {
```
fredericvilcot: unnecessary `useCallback` :)

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/VolumeSelector.component.tsx
```diff
@@ -14,21 +14,32 @@ import {
...
+    ({ value }: { value: string[] }) => {
+      const volume = volumes.find((item) => item.value === value[0]);
```
fredericvilcot: Is it resilient to use the `value[0]` through the select component?

packages/manager/apps/pci-instances/src/types/volume/common.type.ts
Tsiorifamonjena: The intention of this is to say that `any` means available for any zone. I wanted avoid comment that caused a lot of talk above. Do you think this is not necessary ?

aTurmo (dismissed):

aTurmo (approved):

packages/manager/apps/pci-instances/src/pages/instances/instance/dashboard/components/VolumeSelector.component.tsx
Tsiorifamonjena: I'm not sure to understand but BTW I replace it by using the items returned by `SelectValueChangeDetail` to be sure that we got the wright volume

lolottetheclash (approved):

manoncarbonnel (approved):

PR #19397, opened 2025-09-25
fix(container): fix node selection in pnr menu for custom path

darsene:
fixes: #MAOBS-23

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/container/src/container/nav-reshuffle/sidebar/navigation-tree/services/bareMetalCloud.ts
+++ b/packages/manager/apps/container/src/container/nav-reshuffle/sidebar/navigation-tree/services/bareMetalCloud.ts
@@ -57,6 +57,8 @@ baremetalUniverse.children = [
         routing: {
           application: 'dedicated-servers',
           hash: '#/',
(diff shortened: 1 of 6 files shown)


packages/manager/apps/container/src/container/nav-reshuffle/sidebar/navigation-tree/services/bareMetalCloud.ts
```diff
@@ -57,6 +57,8 @@ baremetalUniverse.children = [
         routing: {
           application: 'dedicated-servers',
           hash: '#/',
+          pathMatcher: /(\/dedicated\/(configuration|server))|\/dedicated-servers/,
+
```
MarcMonchablon, line 61: Is it OK that this regex don't check for the ending '/' like the other ones ?

MarcMonchablon (approved):

packages/manager/apps/container/src/container/nav-reshuffle/sidebar/navigation-tree/services/bareMetalCloud.ts
darsene: yes, it doesn't change anything here

PR #19405, opened 2025-09-26
feat(*): add translations in common package

MaximeBajeux:
## Description

Ticket Reference: #MANAGER-19820

### DIFF
--- a/packages/manager/modules/common-translations/public/translations/dashboard/Messages_fr_FR.json
+++ b/packages/manager/modules/common-translations/public/translations/dashboard/Messages_fr_FR.json
@@ -23,5 +23,6 @@
   "period": "Période",
   "period_current": "Le mois en cours",
   "period_last": "Le mois dernier",
-  "period_last_months": "Les {{ value }} derniers mois"
+  "period_last_months": "Les {{ value }} derniers mois",
+  "no_result": "Aucun résultat trouvé"
 }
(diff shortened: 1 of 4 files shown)


sachinrameshn (dismissed):

JacquesLarique (approved):

oalkabouss (dismissed):

packages/manager/modules/common-translations/public/translations/dashboard/Messages_fr_FR.json
```diff
@@ -23,5 +23,6 @@
   "period": "Période",
   "period_current": "Le mois en cours",
   "period_last": "Le mois dernier",
-  "period_last_months": "Les {{ value }} derniers mois"
+  "period_last_months": "Les {{ value }} derniers mois",
+  "no_result": "Aucun résultat trouvé"
```
anooparveti: **question:** We have no results in datagrid https://github.com/ovh/manager/blob/master/packages/manager-react-components/src/components/datagrid/translations/Messages_fr_FR.json#L4 but the content is not same. So my question is are both one and the same? If not, maybe we need to enhance the label.

MaximeBajeux: As it's in prevision to use common translations in the catalog app, that keyword doesn't come from a datagrid. I think that can be used in a usage outside datagrid. But, if it's too specific, we can keep on the app side.

anooparveti: Can we make it `no_result_found` instead?

anooparveti (approved):

PR #19411, opened 2025-09-27
refactor(pci-instances): select public ipv4 from view-model for instance action

Tsiorifamonjena:
## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-instances/src/pages/instances/action/view-models/selectInstanceForActionModal.ts
+++ b/packages/manager/apps/pci-instances/src/pages/instances/action/view-models/selectInstanceForActionModal.ts
@@ -1,5 +1,5 @@
 import { TAggregatedInstanceDto } from '@/types/instance/api.type';
-import { TInstance, TInstanceAddress } from '@/types/instance/entity.type';
+import { TInstance } from '@/types/instance/entity.type';
 
 export type TInstanceActionModalViewModel = {
   id: string;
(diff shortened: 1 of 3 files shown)


packages/manager/apps/pci-instances/src/data/hooks/instance/selectors/instances.selector.ts
```diff
@@ -120,21 +120,25 @@ const mapInstanceAddresses = (instance: TAggregatedInstanceDto) =>
     return acc.set(type, [rest]);
   }, new Map<TInstanceAddressType, TAggregatedInstanceAddress[]>());
 
+const isToDisplayAction = ({ name }: { name: TActionName }) => name !== 'vnc';
```
aTurmo:
```suggestion
const isActionToDisplay = ({ name }: { name: TActionName }) => name !== 'vnc';
```

aTurmo: Qeustion:
I see bellow that there is something with 'vncEnabled', is it not related with this filter ?

Tsiorifamonjena: It's related but the below is the selector for dashboard view-model and here we are in listing page selector.
BTW it's handle by this PR https://github.com/ovh/manager/pull/19466/files#diff-79b739a3f4a8b0c6055523c2d0191dfc265eedf81ccc548effb8e93f58544d72R123

fredericvilcot (approved):

PR #19423, opened 2025-09-29
refactor(catalog): migrate static analysis

achrafHafedh:
ref: #MANAGER-19279

## Description

Migrated the catalog app to use the static analysis kit to standardize ESLint, TypeScript, and Prettier configurations. Fixed existing lint issues and applied Prettier formatting.

Ticket Reference: #MANAGER-19279

### DIFF
--- a/packages/manager/apps/catalog/eslint.config.mjs
+++ b/packages/manager/apps/catalog/eslint.config.mjs
@@ -1,0 +1,65 @@
+// Full adoption
+/*import { eslintSharedConfig } from '@ovh-ux/manager-static-analysis-kit';
+
+export default eslintSharedConfig;
+*/
+
+// Progressive adoption
+import { a11yEslintConfig } from '@ovh-ux/manager-static-analysis-kit/eslint/a11y';
+import {
+  complexityJsxTsxConfig,
+  complexityTsJsConfig,
+} from '@ovh-ux/manager-static-analysis-kit/eslint/complexity';
+// import { cssEslintConfig } from '@ovh-ux/manager-static-analysis-kit/eslint/css';
+// import { htmlEslintConfig } from '@ovh-ux/manager-static-analysis-kit/eslint/html';
+// import { importEslintConfig } from '@ovh-ux/manager-static-analysis-kit/eslint/imports';
+import { javascriptEslintConfig } from '@ovh-ux/manager-static-analysis-kit/eslint/javascript';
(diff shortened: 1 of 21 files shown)


packages/manager/apps/catalog/eslint.config.mjs
aottr, line 1: Did you test if the app builds with some rules enabled, maybe even the full adoption ?

packages/manager/apps/catalog/package.json
```diff
@@ -40,6 +43,7 @@
     "tailwindcss": "^3.4.4"
   },
   "devDependencies": {
+    "@ovh-ux/manager-static-analysis-kit": "*",
```
aottr: maybe you could check the lock file for the currently used version in the monorepo so we won't have the `*` here

packages/manager/apps/catalog/eslint.config.mjs
achrafHafedh: Thanks for your feedback!
I ran the catalog app both with and without the container, and it builds without any issues.
I also ran the yarn lint:modern command successfully.
I've updated my PR, feel free to pull the latest changes if you have my branch locally for testing.

packages/manager/apps/catalog/package.json
achrafHafedh: I just updated the version used in the other apps to: "@ovh-ux/manager-static-analysis-kit": "^0.9.0"

packages/manager/apps/catalog/eslint.config.mjs
```diff
@@ -0,0 +1,83 @@
...
+      '@typescript-eslint/no-unsafe-assignment': 'off',
+      '@typescript-eslint/no-explicit-any': 'off',
+      '@typescript-eslint/await-thenable': 'off'
+    },
+  },
+];
+*/
+
+// Progressive and disable full rules
+// import { typescriptEslintConfig } from '@ovh-ux/manager-static-analysis-kit/eslint/typescript';
```
JacquesLarique: You could remove this part

JacquesLarique (approved):

packages/manager/apps/catalog/eslint.config.mjs
achrafHafedh: I understand your comment better now, I just added the rules that don't block the build.

packages/manager/apps/catalog/eslint.config.mjs
achrafHafedh: It's done ! 
I removed the commented part

PR #19426, opened 2025-09-29
feat(communication): add user feedback and minor improvements

aottr:
## Description

Ticket Reference: #MANAGER-19857

### DIFF
--- a/packages/manager/apps/communication/src/pages/contacts/create/Create.page.tsx
+++ b/packages/manager/apps/communication/src/pages/contacts/create/Create.page.tsx
@@ -31,10 +31,9 @@ export default function CreateContactPage() {
     'account:apiovh:notification/contactMean/validate',
   ]);
 
-  const { addSuccess } = useNotifications();
+  const { addSuccess, clearNotifications } = useNotifications();
   const [error, setError] = useState<string | null>(null);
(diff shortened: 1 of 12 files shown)


packages/manager/apps/communication/src/pages/contacts/create/Create.page.tsx
```diff
@@ -57,9 +60,15 @@ export default function CreateContactPage() {
...
     contactMeanId: createdContactMean?.id || '',
     onSuccess: () => {
+      clearNotifications();
       addSuccess(t('add_contact_success_message'));
       navigate(urls.ContactsTab);
     },
+    onError: (err) => {
+      if (err.response?.status === 429) {
+        setError(t('error_rate_limit_message', { ns: 'common' }));
+      }
```
JacquesLarique, line 70: No other error is displayed to the user ?

aottr: Not really, we could display the general error theoretically ^^

JacquesLarique: Seen with Alexis, we should put an error message (for cases like: 400 code is expired / not valid), but we will tackle this in a different ticket

PR #19427, opened 2025-09-29
fix(hub): standardize eslint via static kit

oalkabouss:
ref: #MANAGER-19766

## Description

Ticket Reference: #MANAGER-19766


packages/manager/apps/hub/eslint.config.mjs
JacquesLarique, line 1: Did you try to enable some rules and build ?
If we could enable a few that would be nice, but if a rule trigger errors do not try to resolve them yet (we'll do that at another time) just re-disable the rule

oalkabouss: No blocking errors detected on the hub (lint + build OK).

PR #19438, opened 2025-09-30
feat(container): einvoicing modify the content of the pop up to update

MaximeBajeux:
ref: #MANAGER-19880

## Description

Ticket Reference: #...


packages/manager/apps/container/src/public/translations/suggestion-modal/Messages_fr_FR.json
JacquesLarique, line 1: This is not the requested change: you're changing the buttons label, when it was asked to wrap these button wordings that are inside `suggestion_modal_description_sub`, ie:
```json
"suggestion_modal_description_sub": "Si vous ne constatez aucune erreur dans ces informations, cliquez sur le bouton \"Je valide ces informations\". Sinon merci de cliquer sur le bouton \"Je mets à jour dans mon profil\"."
```
We also suggested to use variable for these wording, so that if we change the button wording there changed in the description as well

JacquesLarique (dismissed):

achrafHafedh (dismissed):

PR #19466, opened 2025-10-01
fix(pci-instances): disable vnc tab instead hiding

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-5134

## Additional Information

image

### DIFF
--- a/packages/manager/apps/pci-instances/src/data/hooks/instance/selectors/instances.selector.ts
+++ b/packages/manager/apps/pci-instances/src/data/hooks/instance/selectors/instances.selector.ts
@@ -120,21 +120,27 @@ const mapInstanceAddresses = (instance: TAggregatedInstanceDto) =>
     return acc.set(type, [rest]);
   }, new Map<TInstanceAddressType, TAggregatedInstanceAddress[]>());
 
+const isActionToDisplay = (name: TActionName) => name !== 'vnc';
+
 const mapInstanceActions = (
   instance: TAggregatedInstanceDto,
   projectUrl: string,
 ): TAggregatedInstanceActions =>
   instance.actions.reduce<TAggregatedInstanceActions>((acc, action) => {
     const { group, name } = action;
-    const newAction = {
-      label: `pci_instances_list_action_${name}`,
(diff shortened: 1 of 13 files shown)


packages/manager/apps/pci-instances/src/data/hooks/instance/selectors/instances.selector.ts
```diff
@@ -120,21 +120,25 @@ const mapInstanceAddresses = (instance: TAggregatedInstanceDto) =>
...
-      label: `pci_instances_list_action_${name}`,
-      link: getActionHrefByName(projectUrl, name, instance),
-    };
-    const foundAction = acc.get(group);
-    if (!foundAction) return acc.set(group, [newAction]);
-    foundAction.push(newAction);
-    return acc;
-  }, new Map() as TAggregatedInstanceActions);
+  instance.actions
+    .filter(isActionToDisplay)
```
lolottetheclash: [suggestion](https://pullpo.io/cc?l=suggestion) : I think you can remove the filter to avoird extra iteration by including your condition into the `reduce`, what do you think?

lolottetheclash (approved):

lolottetheclash (approved):

packages/manager/apps/pci-instances/src/data/hooks/instance/selectors/instances.selector.ts
```diff
@@ -120,21 +120,27 @@ const mapInstanceAddresses = (instance: TAggregatedInstanceDto) =>
     return acc.set(type, [rest]);
   }, new Map<TInstanceAddressType, TAggregatedInstanceAddress[]>());
 
+const isActionToDisplay = (name: TActionName) => name !== 'vnc';
```
fredericvilcot, line 123: We must find later another way to achieve this and thus to prevent from having this business rule into the client logic ;)

Tsiorifamonjena: I agree :)

lolottetheclash (approved):

PR #19479, opened 2025-10-02
fix(pci-instances): react - fix ts error after static kit

aTurmo:


packages/manager/apps/pci-instances/src/utils/index.ts
```diff
@@ -49,3 +49,11 @@ export const replaceToSnakeCase = (str: string): string =>
...
 export const isCustomUrlSection = (str: string): boolean =>
   str.includes('-') || str.includes('/');
+
+export const groupBy = <T, K>(array: T[], getValue: (item: T) => K) => {
+  return array.reduce((acc, item) => {
+    const value = getValue(item);
+    acc.set(value, (acc.get(value) || []).concat(item));
+    return acc;
+  }, new Map<K, T[]>());
+};
```
fredericvilcot, line 59: Great! 

To improve a bit, I'll add some suggestions:

- add documentation and thus because you return a Map, instead of a record ;)

```ts
 * @template T - Type of the elements in the array.
 * @template K - Type of the key used for grouping.
 * @param {T[]} array - The array of elements to group.
 * @returns
 * @example
// etc...
```

- rename the `getValue` function to something more readable as it handle keys (`keyFn`, ..)

- we could imagine (optional) adding some curry, to isolate the iteratee function from the collection :) and in a more ES6 way 😎

```ts
export const groupBy = <T, K>(keyFn: (item: T) => K) => (array: T[]): Map<K, T[]> =>
  array.reduce((acc, item) => {
    const key = keyFn(item);
    acc.set(key, [...(acc.get(key) || []), item]);
    return acc;
  }, new Map<K, T[]>());
```

It's up to you 😌

aTurmo: Thx :)
And ok for the changes

PR #19483, opened 2025-10-02
feat(backup-services): add backup-services listing

pauldkn:
## Description

Ticket Reference: #BKP-496


packages/manager/apps/backup-services/src/data/api/Region.api.ts
```diff
@@ -0,0 +1,8 @@
+import { ApiResponse, v2 } from '@ovh-ux/manager-core-api';
+
+import { ApiRegion } from '@/types/Region.type';
+
+export const GET_REGIONS_ROUTE = '/location';
+
+export const getRegion = async (id: string): Promise<ApiResponse<ApiRegion>> =>
+  v2.get(`${GET_REGIONS_ROUTE}/${id}`);
```
tibs245: As the real endpoint is : `/location/{name}`. I think it's better if we rename `id` by `name`

packages/manager/apps/backup-services/src/data/hooks/useBackupTenants.ts
```diff
@@ -0,0 +1,29 @@
...
+  useQuery({
+    queryKey: getBackupTenantsQueryKey,
+    queryFn: () =>
+      new Promise((resolve) => {
+        console.log('🛜 mocking api call...');
+        setTimeout(() => {
+          resolve(TENANTS_MOCKS);
+        }, 2000);
+      }),
+  });
```
tibs245, line 30: It's just a suggestion on read for next implementation. May be we can implement this pattern with constants to add DEFAULT_MOCK_LATENCE and multiple latence configuration to test easily multiple latence condition. 

What do you think ?

```suggestion
export const useBackupTenantsMocks = (): TUseBackupTenantsResult =>
  useQuery({
    queryKey: getBackupTenantsQueryKey,
    queryFn: () =>
      new Promise((resolve) => {
        console.log('🛜 mocking useBackupTenantsMocks api call...');
        setTimeout(() => {
          resolve(TENANTS_MOCKS);
        }, DEFAULT_MOCK_LATENCE);
      }),
  });
```

packages/manager/apps/backup-services/src/data/hooks/useRegion.ts
```diff
@@ -0,0 +1,14 @@
...
+import { getRegion } from '../api/Region.api';
+
+type TUseRegionResult = UseQueryResult<ApiRegion, Error>;
+
+export const useRegion = (id: string): TUseRegionResult =>
+  useQuery({
+    queryKey: ['region', id],
+    queryFn: () => getRegion(id),
+    select: (res) => res.data,
+  });
```
tibs245:
```suggestion
export const useRegion = (name: string): TUseRegionResult =>
  useQuery({
    queryKey: ['region', name],
    queryFn: () => getRegion(name),
    select: (res) => res.data,
  });
```

PR #19494, opened 2025-10-02
fix(pci-instances): handle messaging after first instance creation

aTurmo:


packages/manager/apps/pci-instances/public/translations/list/Messages_fr_FR.json
```diff
@@ -28,5 +28,6 @@
   "pci_instances_list_action_soft_reboot": "Redémarrer à chaud (soft)",
   "pci_instances_list_action_hard_reboot": "Redémarrer à froid (hard)",
   "pci_instances_list_action_reinstall": "Réinstaller",
-  "pci_instances_operations_running": "L(es) opération(s) sur l(es) instance(s) est en cours"
+  "pci_instances_operations_running": "L(es) opération(s) sur l(es) instance(s) est en cours",
+  "pci_instances_creations_operations_running": "L(es) opération(s) de création d'instance(s) est en cours, merci de rafraichir la page."
```
manoncarbonnel: typo: missing plural in verbs for both lines

packages/manager/apps/pci-instances/src/utils/operation/operations.utils.ts
```diff
@@ -0,0 +1,19 @@
+import { TOperation } from '@/types/operation/entity.type';
+
+export const isInstanceCreationOperationInProgress = (operation: TOperation) =>
+  operation.section === 'instance' &&
```
manoncarbonnel: suggestion: use of constants instead of magic strings. WDYT ?

packages/manager/apps/pci-instances/public/translations/list/Messages_fr_FR.json
aTurmo: It's what's ask from the ux and po

manoncarbonnel (approved):

packages/manager/apps/pci-instances/src/pages/instances/datagrid/hooks/useDatagridOperationsPolling.spec.tsx
```diff
@@ -53,11 +53,25 @@ describe('useDatagridOperationsPolling', () => {
...
             status: 'in-progress',
             subOperations: [
+              { section: 'instance', action: 'create', status: 'in-progress' },
               { section: 'image', action: 'copytoregion', status: 'completed' },
             ],
           },
         ],
       },
+      {
+        description: 'with a in-progress instance creation with copy completed',
```
Tsiorifamonjena: **question** :  is the description reflect really the data ?

If I understand, It should return no operation and not call onComplete **with a in-progress instance creation with completed** but I don't see the copy completed operations.

Tsiorifamonjena (commented):
Just out of curiosity, the instance being created won't be displayed on the listing, will it?

aTurmo:
> Just out of curiosity, the instance being created won't be displayed on the listing, will it?

That's exactly right :)
That's why we are doing this, to detect if an instance is in creation process, since now the call is asynchronous. So that we can inform the customer that it is created even if it does'nt show up on the grid.

packages/manager/apps/pci-instances/src/pages/instances/datagrid/hooks/useDatagridOperationsPolling.spec.tsx
aTurmo: oh yes you're right I've messed up descrioption and data

packages/manager/apps/pci-instances/src/utils/operation/operations.utils.ts
```diff
@@ -0,0 +1,25 @@
+import { TOperation } from '@/types/operation/entity.type';
```
fredericvilcot: You can move these utility functions close to the related hook / component because it only targets the operations logic.

packages/manager/apps/pci-instances/src/utils/operation/operations.utils.ts
```diff
@@ -0,0 +1,25 @@
...
+  operation.status === 'in-progress';
+
+export const isSubOperationCopyImage = (
+  subOperation?: Required<TOperation>['subOperations'][number],
+) =>
+  subOperation &&
+  subOperation.section === 'image' &&
+  subOperation.action === 'copytoregion';
+
+export const isSubOperationCopyImageInProgress = (
```
fredericvilcot: You may want to return only a boolean, not a possible `undefined` value

packages/manager/apps/pci-instances/src/pages/instances/datagrid/hooks/useDatagridOperationsPolling.ts
```diff
@@ -1,70 +1,88 @@
...
+  getNumberOfCreationOperationWithoutImageCopy(operations);
 
 export const useDatagridOperationsPolling = (onComplete?: () => void) => {
   const projectId = useProjectId();
 
   const [
-    numberOfDistantBackupCopyRunning,
-    setNumberOfDistantBackupCopyRunning,
+    numberOfOperationsInProgress,
+    setNumberOfOperationsInProgress,
```
fredericvilcot, line 13: **Nitpick**: we could enhance naming to ensure better understanding.


```suggestion
    ongoingOperationsNb,
    setOngoingOperationsNb,
```

packages/manager/apps/pci-instances/src/pages/instances/datagrid/hooks/useDatagridOperationsPolling.ts
```diff
@@ -1,70 +1,88 @@
...
   ] = useState(0);
 
   const onSuccess = useCallback(
     (operations?: TOperation[]) => {
       if (!operations) return;
```
fredericvilcot: You should also eject `if operations.length === 0` to prevent from calling unnecessary functions

packages/manager/apps/pci-instances/src/utils/operation/operations.utils.ts
aTurmo: yay I had trouble decided where to put this file

packages/manager/apps/pci-instances/src/pages/instances/datagrid/hooks/useDatagridOperationsPolling.ts
aTurmo: I don't know, personally I find what I put clear. And io'm not a fan of putting 'nb' instead of 'number' ^^'

packages/manager/apps/pci-instances/src/pages/instances/datagrid/hooks/useDatagridOperationsPolling.spec.tsx
aTurmo: fixed

packages/manager/apps/pci-instances/src/utils/operation/operations.utils.ts
aTurmo: fixed (i think)

packages/manager/apps/pci-instances/src/utils/operation/operations.utils.ts
aTurmo: fixed

packages/manager/apps/pci-instances/src/pages/instances/datagrid/hooks/useDatagridOperationsPolling.ts
aTurmo: fixed

packages/manager/apps/pci-instances/src/pages/instances/datagrid/hooks/useDatagridOperationsPolling.ts
fredericvilcot: It's up to you, not mandatory 😌

On backend side, we often use the `...nb` convention, so maybe it could be the right time to normalize these naming conventions :)

PR #19502, opened 2025-10-03
refactor(procedures): migrate static analysis

achrafHafedh:
ref: #MANAGER-19277

## Description

Migrated the procedures app to use the static analysis kit to standardize ESLint, TypeScript, and Prettier configurations. Fixed existing lint issues and applied Prettier formatting.

Ticket Reference: #MANAGER-19277

### DIFF
--- a/packages/manager/apps/procedures/package.json
+++ b/packages/manager/apps/procedures/package.json
@@ -42,7 +42,7 @@
     "tailwindcss": "^3.4.4"
   },
   "devDependencies": {
-    "@ovh-ux/manager-static-analysis-kit": "0.9.0",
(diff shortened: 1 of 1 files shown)


packages/manager/apps/procedures/package.json
```diff
@@ -39,11 +42,12 @@
     "tailwindcss": "^3.4.4"
   },
   "devDependencies": {
+    "@ovh-ux/manager-static-analysis-kit": "0.9.0",
```
JacquesLarique: This result in a failure on the step installation for the check "Lint Code Base", you should target version 0.10.0

oalkabouss (approved):

PR #19513, opened 2025-10-03
chore(okms): unit test config migration

mathieumousnier:
## Description

Ticket Reference: #MANAGER-19790


rjamet-ovh (approved):

packages/manager/apps/okms/src/common/pages/OrderOkmsModal/OrderOkmsModal.page.spec.tsx
```diff
@@ -105,19 +105,21 @@ const clickOnConfirmCheckbox = async () => {
     ORDER_OKMS_TC_CONFIRM_CHECKBOX_TEST_ID,
     // eslint-disable-next-line @typescript-eslint/no-explicit-any
   ) as any;
-  await act(() => {
+  act(() => {
```
pauldkn, line 108: shouldn't it be awaited ?

mathieumousnier: Actually it was awaited before, but the IDE showed a warning telling me that it shouldn't.
The way I understand it is that it depends on what's inside the act function, and here the `confirmCheckbox.odsChange.emit` is synchronous, so there is nothing to `await`

PR #19535, opened 2025-10-06
fix(iam): fix pagination issue on condition deletion

MarcMonchablon:
ref: #MAIAM-38

## Description

Ticket Reference: #MAIAM-38

### DIFF
--- a/packages/manager/modules/iam/src/components/conditions/conditionList/conditionList.controller.js
+++ b/packages/manager/modules/iam/src/components/conditions/conditionList/conditionList.controller.js
@@ -4,6 +4,12 @@ export default class IAMConditionListController {
   $onInit() {
     this.condition = null;
     this.closeModal();
+
+    this.pagination = {
+      currentPage: 1,
+      pageSize: 5,
+      offset: null,
+    };
   }
 
   openModal(condition) {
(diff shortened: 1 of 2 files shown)


packages/manager/modules/iam/src/components/conditions/conditionList/conditionList.controller.js
```diff
@@ -4,6 +4,12 @@ export default class IAMConditionListController {
   $onInit() {
     this.condition = null;
     this.closeModal();
+
+    this.pagination = {
+      currentPage: 1,
```
darsene, line 9: As you've created an object to manage the pagination, you can move the init of page size here.
```suggestion
      currentPage: 1,
      pageSize: 5,
```

packages/manager/modules/iam/src/components/conditions/conditionList/conditionList.template.html
```diff
@@ -10,6 +10,8 @@
         data-pagination-mode="arrows"
         data-empty-placeholder="{{:: 'iam_condition_list_empty_placeholder' | translate}}"
         data-page-size="5"
```
darsene:
```suggestion
        data-page-size="{{$ctrl.pagination.pageSize}}"
```

packages/manager/modules/iam/src/components/conditions/conditionList/conditionList.controller.js
MarcMonchablon: Indeed, changed :)

packages/manager/modules/iam/src/components/conditions/conditionList/conditionList.template.html
MarcMonchablon: Fixed :)

PR #19541, opened 2025-10-06
fix(billing): fix behavior on commitment page without discount

MaximeBajeux:
ref: #MANAGER-19948

## Description

Ticket Reference: #...


packages/manager/modules/billing-components/src/components/commitment-pricing-mode/commitment-pricing-mode.html
```diff
@@ -26,9 +26,15 @@
...
-                            >{{$ctrl.upfrontSavings.amountToPay}}</span
+                        <del
+                            class="small"
+                            data-ng-if="$ctrl.upfrontSavings.amountToPay.value !== $ctrl.periodicTotalPrice"
+                            >{{$ctrl.periodicTotalPrice}}</del
+                        >
+                        <span
+                            class="'oui-heading_6 m-0"
+                            data-ng-class="{'oui-color-ae-500': $ctrl.upfrontSavings.amountToPay.value !== $ctrl.periodicTotalPrice}"
+                            >{{$ctrl.upfrontSavings.amountToPay.text}}{{$ctrl.upfrontSavings.value}}</span
```
JacquesLarique: I think you are missing a `.amoutToPay` since there is no `value` property defined on `upfrontSavings`
```suggestion
                            >{{$ctrl.upfrontSavings.amountToPay.text}}{{$ctrl.upfrontSavings.amountToPay.value}}</span
```
_Also, aren't `text` and `value` kind of the same (text just has the currency trailing the price)_

JacquesLarique (approved):

PR #19552, opened 2025-10-06
fix(iam): update scroll when opening dropdown in modals

MarcMonchablon:
ref: #MAIAM-39

## Description

Ticket Reference: #MAIAM-39


yjaaouane (commented):
Since you’re only targeting one element with [0], prefer using an ID or data attribute instead of a class selector to make the code more stable and less dependent on library-specific CSS classes.

packages/manager/modules/iam/src/components/conditions/conditionType/date/dateCondition.controller.js
```diff
@@ -26,4 +31,11 @@ export default class IAMConditionDateController {
       this.updateConditionValues();
     }
   }
+
+  scrollCalendarIntoView() {
+    const calendarElt = this.$element.find('.flatpickr-calendar')[0];
+    if (calendarElt) {
+      calendarElt.scrollIntoView();
+    }
+  }
```
darsene, line 38: As you repeat it in each component, it should be moved into a dedicated service

```suggestion
  scrollCalendarIntoView() {
    this.elementUiService.scrollIntoView('.flatpickr-calendar');
}
```
with that code
```js
export default class ElementUiService {

/* @ngInject */
constructor($element) {
    this.$element = $element;
}

scrollIntoView(elementName) {
   this.$element.find(elementName)[0]?.scrollIntoView();
}
}
myModule.service('elementUiService', ElementUiService)
```

MarcMonchablon: So, I couldn't inject $element into service directly, since it's a component-specific handle to DOM, but I've created a service that helps avoiding the selector duplication.

PR #19576, opened 2025-10-07
feat(dedicated): fixe null value in catch block

b-fritz-ovh:
ref: #MANAGER-18410

## Description

Ticket Reference: #...


packages/manager/modules/netapp/src/dashboard/volumes/dashboard/routing.js
```diff
@@ -96,8 +96,8 @@ export default /* @ngInject */ ($stateProvider) => {
           .get(
             `/storage/netapp/${serviceName}/share/${volumeId}/snapshotReserve`,
           )
-          .then(({ data }) => data?.percent || 5)
-          .catch(() => null),
+          .then(({ data: { percent = 5 } }) => percent)
```
aderghamov: A bit strange!

At least, replace the magic value by a constant.
What's the meaning of catch() => 5 ?
It looks like we dismiss any error that the promise may throw to display a default value.
Hiding the error does not look like a good practice to me.

b-fritz-ovh: I understand. 5% is the default value assigned to the snapshots.

In a short laps of time, when a user create a volume (while it is create), the API throw an error.

Do you have an other idee?

b-fritz-ovh: Done: magic value replaced with DEFAULT_SNAPSHOT_SIZE

Thx

mhelhali-soufien (approved):

PR #19584, opened 2025-10-08
refactor(container): migrate static analysis

oalkabouss:
ref: #MANAGER-19276

## Description

Ticket Reference: #MANAGER-19276


packages/manager/apps/container/src/helpers/dateHelper.ts
```diff
@@ -1,7 +1,7 @@
 import parse from 'date-fns/parse';
 import formatDistanceStrict from 'date-fns/formatDistanceStrict';
 // TODO: upgrade prettier
```
JacquesLarique: Seems like this TODO can be removed

packages/manager/apps/container/src/core/product-nav-reshuffle/ProductNavReshuffleProvider.tsx
```diff
@@ -127,7 +127,7 @@ export const ProductNavReshuffleProvider = ({
     popoverPosition,
     setPopoverPosition
     // FIXME: Can be removed after updating Prettier
-    // eslint-disable-next-line prettier/prettier
+     
   } satisfies ProductNavReshuffleContextType;
```
JacquesLarique: Did you try to remove the `satisfies` as suggested by the FIXME comment ?

packages/manager/apps/container/eslint.config.mjs
```diff
@@ -0,0 +1,49 @@
...
+      '@typescript-eslint/no-unsafe-call': 'off',
+      '@typescript-eslint/no-unused-vars': 'off',
+      '@typescript-eslint/no-redundant-type-constituents': 'off',
+      '@typescript-eslint/unbound-method': 'off',
+      '@typescript-eslint/no-unsafe-enum-comparison': 'off',
+      '@typescript-eslint/no-base-to-string': 'off',
+    },
+  },
+  prettierEslintConfig,
+  { rules: { 'prettier/prettier': 'off' } },
```
JacquesLarique: It seems strange to activate the prettier eslint config and disable the prettier rule

packages/manager/apps/container/src/core/product-nav-reshuffle/ProductNavReshuffleProvider.tsx
aottr: I'd suppose the FIXME comment is about the eslint-disable as the satisfies triggers a typecheck

packages/manager/apps/container/eslint.config.mjs
oalkabouss: Good catch, i forgot to stage that file before committing. i ve just pushed the fix so the config is now consistent. Thanks!

PR #19603, opened 2025-10-08
fix(iam): fix condition modal allowing incomplete conditions

MarcMonchablon:
## Description

Ticket Reference: #MAIAM-39, #MAIAM-40

### DIFF
--- a/packages/manager/modules/iam/src/components/conditions/conditionType/date/dateCondition.controller.js
+++ b/packages/manager/modules/iam/src/components/conditions/conditionType/date/dateCondition.controller.js
@@ -23,7 +23,7 @@ export default class IAMConditionDateController {
       this.timezone?.timeZone,
     )}.${this.criterion?.value}`;
     this.condition.values = {
-      [fullKey]: this.dates.replaceAll(' ', ''),
+      [fullKey]: this.dates?.replaceAll(' ', ''),
     };
     this.condition.hasValue = !!this.condition.values[fullKey];
   }


packages/manager/modules/iam/src/components/conditions/createConditionModal/createConditionModal.template.html
```diff
@@ -5,7 +5,7 @@
                 data-heading="{{:: 'iam_create_condition_configuration' | translate }}"
                 data-primary-action="$ctrl.addCondition()"
                 data-primary-label="{{:: 'iam_create_condition_configure' | translate }}"
-                data-primary-disabled="$ctrl.conditionForm.$invalid"
+                data-primary-disabled="$ctrl.conditionForm.$invalid || !$ctrl.criterion || !$ctrl.condition.hasValue"
```
darsene, line 8: `$ctrl.criterion` should update `$ctrl.conditionForm` controls so I'm not sure you need to check it as well
```suggestion
                data-primary-disabled="$ctrl.conditionForm.$invalid  || !$ctrl.condition.hasValue"
```

MarcMonchablon: The underlying issue for the incomplete conditions is that the form isn't considered invalid even if the field is empty, as long as the corresponding 'oui-select' is opened, hence the test for hasValue. The same behavior can also be replicated by skipping the criterion and first filling the condition value for eg. a product type, then without the `$ctrl.criterion` check the form will also be considered valid once the criterion's selector is opened, even without selecting one.

PR #19607, opened 2025-10-08
feat(backup-vault): setup global mock for use navigation get url

tibs245:
ref: #BKP-525

## Description

Ticket Reference: #...


packages/manager/apps/backup-vaults/vitest.config.js
```diff
@@ -13,6 +13,7 @@ export default mergeConfig(
     test: {
       setupFiles: [
         '@ovh-ux/manager-core-test-utils/src/utils/setup-test.ts',
+        './src/utils/tests/globalsMock.ts',
```
pauldkn, line 16:
```suggestion
        './src/utils/tests/globalMocks.ts',
```

packages/manager/apps/backup-vaults/src/utils/tests/globalsMock.ts
pauldkn, line 1: **typo:** I think filename should be `globalMocks.ts`

PR #19638, opened 2025-10-09
refactor(ovh-payment-method): remove lodash dependency

achrafHafedh:
## Description

Remove lodash dependency and unused classes

Ticket Reference: #MANAGER-18066


packages/components/ovh-payment-method/src/utils.js
aottr, line 1: I know the task was to replace / remove lodash but do we really need a util function to check if `value === null` ?

achrafHafedh: Even though value === null is straightforward, I added isNull in utils for the following reasons:
Consistency: it replaces all existing usages of _.isNull with a single shared helper, making the Lodash removal uniform.
Readability: isNull(value) is more explicit and self-documenting than value === null, especially for developers scanning the code quickly

aottr: I disagree on both points:
Consistency: The usage is limited to 2 occaisons: The isLegacy helper in PaymentMethod and AvailablePaymentMethod. 2 classes which in itself have a big overlap ^^ We don't use this helper anywhere else.

Readability: `===` is a basic operation. Helpers like `isNull` are an abstraction and bloat for minimal gain. Otherwise we would also write helpers like: `isTrue`, `isFalse`, `isUndefined`.

It makes sense for me in a functional programming pattern, but here we already have duplicated code in the class definitions and then write a wrapper around a basic operation.

**Additionally** the use is `!isNull()`, which is imho a lot less readable than `!== null` because of the leading `!` next to an `i`.

Either way, for this package it's non-blocking for me.

PR #19645, opened 2025-10-09
feat(*): add new standard tracking property

JacquesLarique:
## Description

Add new standard tracking property for support language

Ticket Reference: #MANAGER-19491

### DIFF
--- a/packages/components/ovh-shell/src/plugin/tracking/tracking.ts
+++ b/packages/components/ovh-shell/src/plugin/tracking/tracking.ts
@@ -62,7 +62,7 @@ export class TrackingPlugin extends OvhAtInternet {
     };
   }
 
-  configureTracking(region: Region, user: User): void {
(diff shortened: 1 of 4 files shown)


packages/components/ovh-shell/src/plugin/tracking/tracking.ts
```diff
@@ -78,6 +78,7 @@ export class TrackingPlugin extends OvhAtInternet {
       visitorId: user?.customerCode,
       legalform: user?.legalform,
       subsidiary: user?.ovhSubsidiary,
+      language: user?.language,
```
anooparveti, line 83: **question:** is support_language always same as user.language? If not, we should name it as supportLanguage.

JacquesLarique: Yes it is, `user.language` is the language given by /me, aka the support language

PR #19656, opened 2025-10-10
feat(pci-instances): add availability zones

lolottetheclash:
ref: #TAPC-4559

## Description

Ticket Reference: #...


packages/manager/apps/pci-instances/src/pages/instances/create/view-models/__tests__/availabilityZonesViewModel.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+import { selectAvailabilityZones } from '../availabilityZonesViewModel';
+
+const fakeDeps: Deps = {
+  instancesCatalogPort: {
+    selectInstancesCatalog: vi
+      .fn()
+      .mockReturnValue(mockedInstancesCatalogEntity),
+  },
+  messageProviderPort: { getMessage: vi.fn((key: string) => key) },
+};
```
fredericvilcot, line 16: We could imagine having a test helper to prevent from duplicating fake deps for selectors :)

PR #19675, opened 2025-10-10
fix(cookie-policy): update cookie policy urls

cjulien-ovh:
## Description

Update cookie policy banner url to redirect to ovhcloud.com domain instead of ovh.com


anooparveti:
Rebasing with master.

PR #19676, opened 2025-10-10
feat(account-creation): set up tracking

JacquesLarique:
## Description

Set up tracking on account creation app

Ticket Reference: #MANAGER-19491

### DIFF
--- a/packages/manager/apps/account-creation/__mocks__/gcj-module.tsx
+++ b/packages/manager/apps/account-creation/__mocks__/gcj-module.tsx
@@ -1,0 +1,14 @@
+import React from 'react';
+
+export const usePrivacyPolicyLink = () => 'https://fake-link.com';
+
+type DataUsagePolicyProps = {
+  subsidiary: string;
+  region: string;
+};
+
+export const DataUsagePolicy = ({ subsidiary, region }: DataUsagePolicyProps) => (
+  <div data-testid="data-usage-policy">
+    {subsidiary}-{region}
+  </div>
+);
--- a/packages/manager/apps/account-creation/src/context/tracking/tracking.provider.spec.tsx
+++ b/packages/manager/apps/account-creation/src/context/tracking/tracking.provider.spec.tsx
@@ -1,0 +1,240 @@
+import { describe, expect, it, vi, beforeEach } from 'vitest';
+import { render, screen, waitFor } from '@testing-library/react';
+import { TrackingPlugin } from '@ovh-ux/shell';
+import { Region, User } from '@ovh-ux/manager-config';
+import * as shellClient from '@ovh-ux/manager-react-shell-client';
+import { TrackingProvider } from './tracking.provider';
+import { APP_NAME, LEVEL2, SUB_UNIVERSE, UNIVERSE } from '@/tracking.constant';
+import { useTrackingContext } from './useTracking';
+
+const mockConfigureTracking = vi.fn();
+// Mock dependencies
+vi.mock('@ovh-ux/shell', () => ({
+  TrackingPlugin: vi.fn().mockImplementation(() => ({
+    configureTracking: mockConfigureTracking,
+    trackPage: vi.fn(),
+    trackClick: vi.fn(),
+  })),
+}));
+
+vi.mock('@ovh-ux/manager-react-shell-client', () => ({
+  getClickProps: vi.fn((params) => params),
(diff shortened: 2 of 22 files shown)


packages/manager/apps/account-creation/src/context/user/user.provider.tsx
```diff
@@ -50,6 +51,7 @@ export const UserProvider = ({ children = [] }: Props): JSX.Element => {
       setLegalForm(me?.legalform);
       setOvhSubsidiary(me?.ovhSubsidiary);
       setCountry(me?.country as Country);
+      setLanguage(me?.language || undefined);
```
aottr, line 60: the `|| undefined` should be redundant since you're already using optional chaining

packages/manager/apps/account-creation/src/data/hooks/settings/useSettings.ts
```diff
@@ -71,3 +71,20 @@ export const useLanguageSettings = (country?: string, currency?: string) =>
...
   });
+
+export const useSubsidiarySettings = (country?: string, currency?: string, language?: string) => {
+  return useQuery({
+    queryKey: ['settings'],
+    queryFn: () => getSettings(getRegion()),
+    select: (settings: Map<string, BillingCountry[]>) => {
+      if (!(country && currency && language)) {
+        return null;
+      }
```
achrafHafedh: Maybe it would be better to use const enabled  to prevent unnecessary API calls when the parameters are missing, Like :
```
  const enabled = Boolean(country && currency && language);

  return useQuery({
    queryKey: ['settings', country, currency, language],
    queryFn: () => getSettings(getRegion()),
    enabled,
```

packages/manager/apps/account-creation/src/context/user/user.provider.tsx
JacquesLarique: Since `language` is of type `UserLocales | null | undefined`, removing the `|| undefined` result in a TS error, as the `language` state is `UserLocales | undefined` (needed to avoid issue with `null` values in the tracking configuration)

packages/manager/apps/account-creation/src/data/hooks/settings/useSettings.ts
JacquesLarique: Adding `country`, `currency`, and `language` is the `queryKey` will make it so on change it will trigger an API call (which would be unnecessary since the API is only taking the region as a parameter)
Currently only one call to the API is made, and just the `select` functions are called multiple times

achrafHafedh (approved):

packages/manager/apps/account-creation/__mocks__/gcj-module.tsx
```diff
@@ -0,0 +1,9 @@
+import React from 'react';
+
+export const usePrivacyPolicyLink = () => 'https://fake-link.com';
+
+export const DataUsagePolicy = ({ subsidiary, region }: any) => (
+  <div data-testid="data-usage-policy">
+    {subsidiary}-{region}
+  </div>
+);
```
achrafHafedh, line 14:
```suggestion
);

// interface or enum if we have all values
interface DataUsagePolicyProps {
  subsidiary: string;
  region: string;
}

export const DataUsagePolicy: React.FC<DataUsagePolicyProps> = ({
  subsidiary,
  region,
}) => (
     ...
);

```

packages/manager/apps/account-creation/src/context/tracking/tracking.provider.spec.tsx
```diff
@@ -0,0 +1,286 @@
...
+    const TestComponent = () => {
+      contextValue = useTrackingContext();
+      return <div>Test</div>;
+    };
+
+    render(
+      <TrackingProvider region={mockRegion}>
+        <TestComponent />
+      </TrackingProvider>,
+    );
```
achrafHafedh: Maybe you can optimise the extract context like 
```suggestion
   const extractContext = () => {
    let contextValue;
    const TestComponent = () => {
      contextValue = useTrackingContext();
      return <div>Test</div>;
    };
    renderWithTracking(<TestComponent />);
    return contextValue;
  };
```
and you can use : 
```
const ctx = extractContext();
```

achrafHafedh (approved):

MehdiZaghouani (dismissed):

PR #19711, opened 2025-10-13
Ci/pnpm pci instances

fredericvilcot:
This PR makes the PNPM migration for `pci-instances` µapp


package.json
```diff
@@ -130,9 +130,9 @@
     "vite": "^5.4.18",
     "whatwg-fetch": "^3.6.2"
   },
-  "packageManager": "yarn@1.22.22",
   "engines": {
```
helabenkhalfallah, line 134: you should not commit the package json : https://manager.pages.ovhcloud.tools/technical-documentation/manager/technical-migrations/pnpm-incremental-adoption/#files-to-commit-and-push-after-migration-or-rollback

🚫 Do Not Commit

The root package.json must never be committed or pushed.

Any changes there are considered transient or local to the migration process.


I'll ignore by default in future version :)

fredericvilcot: @helabenkhalfallah 
This change has been introduced by the CLI using `yarn pm:add:app --app packages/manager/apps/pci-instances`

helabenkhalfallah: Yes I know dont worry :) :)

fredericvilcot: I removed it :)

packages/manager/apps/pci-instances/src/__mocks__/instance/handlers.ts
```diff
@@ -8,7 +8,7 @@ export type TInstancesServerResponse = DeepReadonly<{
 
 const rootUri = '*/cloud/project/:projectId';
 
-const getResponsePayload = (payload: JsonBodyType): HttpResponse =>
+const getResponsePayload = (payload: JsonBodyType) =>
```
ppprevost, line 11: question: why are you delleting the return type ?

lolottetheclash: @ppprevost  I didn't write this code, but I guess it's because Typescript deduces it on its own.

fredericvilcot: Yes, I had to remove it because of new eslint rules that introduced an error here, so I let TS inferring the type

helabenkhalfallah (approved):

PR #19720, opened 2025-10-14
feat(pci-block-storage): remove retyping encryption

aTurmo:
ref: #TAPC-5125

### DIFF
--- a/packages/manager/apps/pci-block-storage/public/translations/common/Messages_fr_FR.json
+++ b/packages/manager/apps/pci-block-storage/public/translations/common/Messages_fr_FR.json
@@ -19,7 +19,6 @@
   "pci_projects_project_storages_blocks_create_backup_label": "Créer une sauvegarde",
   "pci_projects_project_storages_blocks_delete_label": "Supprimer",
   "pci_projects_project_storages_blocks_add_label": "Créer un volume",
-  "pci_projects_project_storages_blocks_change_encryption": "Chiffrer/déchiffrer le volume",
   "pci_projects_project_storages_blocks_change_type": "Modifier le type du volume",
   "pci_projects_project_storages_blocks_new": "Nouveau",
   "pci_projects_project_storages_blocks_coming_soon": "Bientôt disponible",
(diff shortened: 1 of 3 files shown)


packages/manager/apps/pci-block-storage/src/api/hooks/useCatalogWithPreselection.tsx
```diff
@@ -40,30 +40,26 @@ export const useCatalogWithPreselection = (
...
-    if (!catalogData || !volumeData) return [null, null];
+  const data = useMemo(() => {
+    if (!catalogData || !volumeData) return null;
 
     if (
       is3az(catalogData.regions, volumeData.region) &&
       isClassicMultiAttach(volumeData)
     ) {
-      return [[], null];
+      return [];
```
fredericvilcot: Why not returning `null` too?

package.json
```diff
@@ -5,7 +5,98 @@
   "description": "OVH Control Panel also known as Manager",
   "license": "BSD-3-Clause",
   "workspaces": {
-    "packages": []
+    "packages": [
```
fredericvilcot: This change is related to pnpm migration, should not be here ;)

packages/manager/apps/pci-block-storage/src/api/hooks/useCatalogWithPreselection.tsx
aTurmo: I wanted to distinguish between data not yet available (while query is pending) and data is here but empty (empty = can't retype)

package.json
aTurmo: fixed

packages/manager/apps/pci-block-storage/src/api/hooks/useCatalogWithPreselection.tsx
fredericvilcot: OK so basically, what you want to achieve could be more like that :)
data not yet available (while query is pending --> `undefined`
data is here but empty --> `null`

Manipulating empty arrays, like empty strings, could lead to a mismatch in terms of nullity because you will have to test everywhere some length, instead of manipulating null values that have been created for this purpose ;)

PR #19729, opened 2025-10-14
fix(pci-instances): fix 3AZ milan region for instance creation

aTurmo:
ref: #TAPC-5210

Capture d’écran 2025-10-14 à 11 47 39

### DIFF
--- a/packages/manager/modules/manager-components/src/region/service.js
+++ b/packages/manager/modules/manager-components/src/region/service.js
@@ -72,6 +72,10 @@ export default class ovhManagerRegionService {
     const translatedMacroRegion = this.$translate.instant(
       `manager_components_region_${this.constructor.getMacroRegion(region)}`,
     );
+    if (region === 'EU-SOUTH-LZ-MIL-A') {
+      return `${translatedMacroRegion}-LZ`;
+    }
+
     return translatedMacroRegion || region;
   }
 


packages/manager/modules/manager-components/src/region/service.js
```diff
@@ -72,6 +72,10 @@ export default class ovhManagerRegionService {
     const translatedMacroRegion = this.$translate.instant(
       `manager_components_region_${this.constructor.getMacroRegion(region)}`,
     );
+    if (region === 'EU-SOUTH-LZ-MIL-A') {
+      return `${translatedMacroRegion}-LZ`;
+    }
```
fredericvilcot, line 77: **Question (non-blocking)**: Could we modify the `groupByContinentAndDatacenterLocation` function 
to prevent from having specific predicates? 

I mean, into the function, before grouping, having a look to the regions array to find a duplicate and then modify the way we decide to group (e.g. adding `-LZ` or something else)

aTurmo: I think it will be too complicated to reliably change the grouping for our specific case like that without breaking the expected groupings. It's not ideal but I think it's the safer way to fix our problem.

fredericvilcot: OK, let's go like this so :)

PR #19782, opened 2025-10-15
feat(okms): okms dashboard tiles

rjamet-ovh:
## Description

- rework okms dashboard tile
- update kms okms dashboard page
- add secret-manager okms dashboard page

Ticket Reference: #MANAGER-19793

### DIFF
--- a/packages/manager/apps/okms/src/common/components/okmsDatagrid/OkmsDatagrid.component.spec.tsx
+++ b/packages/manager/apps/okms/src/common/components/okmsDatagrid/OkmsDatagrid.component.spec.tsx
@@ -95,7 +95,6 @@ describe('Okms Datagrid tests suite', () => {
     expect(await screen.findAllByText(columns.servicekeyCount)).toHaveLength(1);
     expect(await screen.findAllByText(columns.region)).toHaveLength(1);
     expect(await screen.findAllByText(columns.status)).toHaveLength(1);
-    expect(screen.getByTestId('header-action')).toBeVisible();
 
     // hidden columns
     expect(screen.queryByText(columns.secretCount)).not.toBeInTheDocument();
@@ -114,6 +113,5 @@ describe('Okms Datagrid tests suite', () => {
     expect(screen.queryByText(columns.kmipCount)).not.toBeInTheDocument();
     expect(screen.queryByText(columns.servicekeyCount)).not.toBeInTheDocument();
     expect(screen.queryByText(columns.region)).not.toBeInTheDocument();
-    expect(screen.queryByTestId('header-action')).not.toBeInTheDocument();
   });
 });
--- a/packages/manager/apps/okms/src/components/layout-helpers/Dashboard/KmsTabs.tsx
+++ b/packages/manager/apps/okms/src/components/layout-helpers/Dashboard/KmsTabs.tsx
@@ -67,6 +67,7 @@ const KmsTabs: React.FC<KmsTabsProps> = ({ tabs }) => {
     >
       {tabs.map((tab: KmsTabProps, index) => (
         <OdsTab
+          data-testid={tab.url}
           key={`ods-tab-bar-item-${tab.url}`}
           id={tab.url}
           isSelected={index === activeTabIndex}
--- a/packages/manager/apps/okms/src/mocks/services/services.mock.ts
+++ b/packages/manager/apps/okms/src/mocks/services/services.mock.ts
@@ -1,4 +1,5 @@
 import { CurrencyCode, ServiceDetails } from '@ovh-ux/manager-react-components';
+import { okmsMock } from '../kms/okms.mock';
 
 export const kmsServicesMock: ServiceDetails = {
   route: {
@@ -68,7 +69,7 @@ export const kmsServicesMock: ServiceDetails = {
     engagementRequest: { pricingMode: '', requestDate: '' },
   },
   resource: {
-    displayName: 'testkms123k',
+    displayName: okmsMock[0].iam.displayName,
     name: '7f3a82ac-a8d8-4c2a-ab0c-f6e86ddf6a7c',
     state: 'active',
     product: {
--- a/packages/manager/apps/okms/src/modules/secret-manager/components/regionSelector/RegionSelector.component.spec.tsx
+++ b/packages/manager/apps/okms/src/modules/secret-manager/components/regionSelector/RegionSelector.component.spec.tsx
@@ -10,7 +10,7 @@ import {
   getOdsButtonByLabel,
   getOdsButtonByIcon,
 } from '@/utils/tests/uiTestHelpers';
-import { initTestI18n } from '@/utils/tests/init.i18n';
+import { initTestI18n, labels } from '@/utils/tests/init.i18n';
 import {
   GeographyGroup,
   RegionOption,
@@ -161,7 +161,7 @@ describe('RegionSelector Component', () => {
       const { container } = await renderRegionSelector();
 
       // Then
-      expect(screen.getByText('region')).toBeInTheDocument();
+      expect(screen.getByText(labels.common.region.region)).toBeInTheDocument();
       await getOdsButtonByLabel({ container, label: mockCurrentRegion.label });
     });
 
--- a/packages/manager/apps/okms/src/pages/dashboard/KmsDashboard.page.spec.tsx
+++ b/packages/manager/apps/okms/src/pages/dashboard/KmsDashboard.page.spec.tsx
@@ -13,7 +13,17 @@ import { KMS_FEATURES } from '@/utils/feature-availability/feature-availability.
 import { SERVICE_KEYS_LABEL } from '@/constants';
 import { KMS_ROUTES_URLS } from '@/routes/routes.constants';
 
-const mockPageUrl = KMS_ROUTES_URLS.kmsDashboard(okmsMock[0].id);
+const mockedOkms = okmsMock[0];
+const mockPageUrl = KMS_ROUTES_URLS.kmsDashboard(mockedOkms.id);
+
+const generalInformationTabTestId = KMS_ROUTES_URLS.kmsDashboard(mockedOkms.id);
+const serviceKeyListTabTestId = KMS_ROUTES_URLS.serviceKeyListing(
+  mockedOkms.id,
+);
+const credentialListTabTestId = KMS_ROUTES_URLS.credentialListing(
+  mockedOkms.id,
+);
+const logTabTestId = KMS_ROUTES_URLS.kmsLogs(mockedOkms.id);
 
 describe('KMS dashboard test suite', () => {
   it('should display an error if the API is KO', async () => {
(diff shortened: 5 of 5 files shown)


packages/manager/apps/okms/src/data/hooks/useUpdateOkmsName.ts
```diff
@@ -56,24 +50,30 @@ export const useUpdateOkmsName = ({
...
         queryKey: okmsQueryKeys.listDatagrid,
       });
-      await queryClient.invalidateQueries({
-        queryKey: getServiceDetailsQueryKey(okmsId),
-      });
-      await queryClient.invalidateQueries({
-        queryKey: okmsQueryKeys.detail(okmsId),
-      });
+
+      // Wait for the renaming to be effective in the backend before refetching the data
```
mathieumousnier: Ouch. I guess you did this because you where forced to, but : 
- what's happening here ? the user sees the name updated after 3s ?
- did we have issues with this before ? I don't remember we did.
- couldn't we fix this with an optimistic update on the detail, so the user sees the name changed, then we might trigger a delayed cache invalidation on the list

Also, no need to await cache invalidations.

Edit: I have just seen that the optimistic update is in the modal component, this is why I coulnd't understand the code. It should be here, in the onMutate callback, using [this pattern](https://tanstack.com/query/v4/docs/framework/react/guides/optimistic-updates)

packages/manager/apps/okms/src/data/hooks/useUpdateOkmsName.ts
```diff
@@ -56,24 +50,30 @@ export const useUpdateOkmsName = ({
...
       clearNotifications();
-      addSuccess(
-        t('key_management_service_service-keys_update_name_success'),
-        true,
-      );
+      addSuccess(t('modify_name_success', { ns: NAMESPACES.ACTIONS }), true);
       onSuccess?.();
     },
     onError: (result: ApiError) => {
+      // Revert the optimistic update if the mutation fails
```
mathieumousnier: This comment mention an optimistic update but I don't see one.
Also, if you code an optimistic update, you can find in the documentation how to save the previous state before the update, pass it to the onError callback, and restore it. All this without doing another API call

packages/manager/apps/okms/src/common/components/okmsUpdateNameModal/OkmsUpdateNameModal.component.tsx
```diff
@@ -0,0 +1,65 @@
...
+          data: {
+            ...prevData.data,
+            iam: {
+              ...prevData.data.iam,
+              displayName: newDisplayName,
+            },
+          },
+        };
+      },
+    );
```
mathieumousnier: Should be in the `useUpdateOkmsName` hook

packages/manager/apps/okms/src/modules/secret-manager/pages/okmsDashboard/updateOkmsNameModal/UpdateOkmsNameModal.page.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { useOutletContext } from 'react-router-dom';
+import OkmsUpdateNameModal from '@/common/components/okmsUpdateNameModal/OkmsUpdateNameModal.component';
+import { OkmsDashboardOutletContext } from '../OkmsDashboard.type';
+
+const UpdateOkmsNameModal = () => {
```
mathieumousnier: Should we name this kind of modal "page component" like this "UpdateOkmsNameModalPage" ?
I did it for the secret Drawers, it differenciates the component from the page.

Also from a naming perspective, both components could be named in a more consistent way like `OkmsUpdateNameModal` and `OkmsUpdateNameModalPage`

packages/manager/apps/okms/src/modules/secret-manager/pages/okmsDashboard/terminateOkmsModal/TerminateOkmsModal.page.tsx
```diff
@@ -0,0 +1,12 @@
+import React from 'react';
+import { useParams } from 'react-router-dom';
+import { LocationPathParams } from '@secret-manager/routes/routes.constants';
+import OkmsTerminateModal from '@/common/components/okmsTerminateModal/OkmsTerminateModal.component';
+
+const TerminateOkmsModal = () => {
```
mathieumousnier: Similar to the OkmsUpdateNameModal where i give more details.

I would name the components `OkmsTerminateModal` and `OkmsTerminateModalPage`

packages/manager/apps/okms/src/common/components/okmsDashboard/billingTile/BillingTile.component.tsx
```diff
@@ -0,0 +1,62 @@
...
+        okmsDomain.id,
+      );
+      break;
+
+    default:
+      break;
+  }
+
+  return resiliateLink;
+};
```
mathieumousnier: Nitpick: I prefer records for this. I find them simpler to read.

```ts 
const useResiliateLink = (okmsDomain: OKMS) => {
  const productType = useProductType();

  const links: Record<ProductType, string> = {
    'key-management-service': KMS_ROUTES_URIS.kmsTerminate,
    'secret-manager': SECRET_MANAGER_ROUTES_URLS.okmsTerminateModal(
      okmsDomain.id,
    ),
  };

  return links[productType];
};
```

It's so short you don't even need the hook anymore

packages/manager/apps/okms/src/common/components/okmsDashboard/generalInformationsTile/items/IdTileItem.component.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+import { ManagerTile, Clipboard } from '@ovh-ux/manager-react-components';
+import { OKMS } from '@/types/okms.type';
+import { ID_LABEL } from '@/constants';
+
+type IdTileItemParams = {
```
mathieumousnier:
```suggestion
type IdTileItemProps = {
```

packages/manager/apps/okms/src/common/components/okmsDashboard/billingTile/BillingTile.component.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+type BillingTileParams = {
```
mathieumousnier:
```suggestion
type BillingTileProps = {
```

packages/manager/apps/okms/src/common/components/okmsDashboard/generalInformationsTile/items/NameTileItem.component.tsx
```diff
@@ -0,0 +1,65 @@
...
+    case 'secret-manager':
+      resiliateLink = SECRET_MANAGER_ROUTES_URLS.okmsUpdateNameModal(okms.id);
+      break;
+
+    default:
+      break;
+  }
+
+  return resiliateLink;
+};
```
mathieumousnier, line 21: Nitpick: a Record would be easier to read

packages/manager/apps/okms/src/common/components/okmsDashboard/generalInformationsTile/items/NameTileItem.component.tsx
```diff
@@ -0,0 +1,65 @@
...
+
+type NameTileItemParams = {
```
mathieumousnier:
```suggestion
type NameTileItemProps = {
```

Looks like all the TileItems have this "Params" name, I stop pointing them here

packages/manager/apps/okms/src/common/components/okmsDashboard/kmipTile/items/KmipEndpointTileItem.spec.tsx
```diff
@@ -0,0 +1,62 @@
...
+
+vi.mock(
+  '@/common/components/downloadOkmsPublicCaLink/DownloadOkmsPublicCaLink',
+  async (original) => ({
+    ...(await original()),
+    DownloadOkmsPublicCaLink: vi.fn(),
+  }),
+);
+
+const renderTileItem = async () => {
```
mathieumousnier: There is a "renderWithI18n" utils if you want.
In fact we never really render without the translations

packages/manager/apps/okms/public/translations/secret-manager/Messages_fr_FR.json
```diff
@@ -30,6 +30,8 @@
   "edit_metadata": "Modifier les paramètres",
   "okms_activation_in_progress": "Veuillez patienter, création en cours.",
   "okms_list": "Liste de domaines OKMS",
+  "okms_dashboard_title": "Détail du domaine OKMS",
+  "okms_manage_label": "gérer mon domaine OKMS",
```
mathieumousnier:
```suggestion
  "okms_manage_label": "Gérer mon domaine OKMS",
```

packages/manager/apps/okms/src/common/hooks/useProductType.ts
```diff
@@ -0,0 +1,25 @@
...
+    productType !== 'key-management-service' &&
+    productType !== 'secret-manager'
+  ) {
+    return null;
+  }
+
+  return productType;
+};
+
+export default useProductType;
```
mathieumousnier, line 26: Nitpick: I would have written this like this:

It's a bit more verbose but : 
- You don't write the type list twice
- You use a proper typegard for type narrowing

```suggestion
import { useLocation } from 'react-router-dom';

const productTypes = ['key-management-service', 'secret-manager'] as const;

export type ProductType = typeof productTypes[number];

const isProductType = (productType: string): productType is ProductType => {
  return productTypes.includes(productType as ProductType);
};

const useProductType = (): ProductType | null => {
  const location = useLocation();

  // Split the pathname by '/' and get the first non-empty segment
  const pathSegments = location.pathname
    .split('/')
    .filter((segment) => segment);
  const productType = pathSegments[0];

  // Ensure the productType is one of the allowed values
  if (!isProductType(productType)) {
    return null;
  }

  return productType;
};

export default useProductType;

```

packages/manager/apps/okms/src/data/hooks/useUpdateOkmsName.ts
```diff
@@ -29,60 +26,66 @@ export type UpdateOkmsNameMutationParams = {
...
+      }, 3000);
+
       clearNotifications();
-      addSuccess(
-        t('key_management_service_service-keys_update_name_success'),
-        true,
-      );
+      addSuccess(t('modify_name_success', { ns: NAMESPACES.ACTIONS }), true);
       onSuccess?.();
     },
```
mathieumousnier, line 84: You can simplify the query invalidations : 
- no need to invalidate the detail cache, you updated it with the name, so it is already ok
- no need to invalidate the list cache after a wait, there is no refetch of the list triggered on this page, the query is only invalidated, and will be refetched when the user goes to a page where it is needed. So you just need to invalidate the cache once, right when the mutation is done

Also, an "optimistic" update is done **before** the query returns, so it would be on the onMutate. Here it is just a normal update of the cache.

```suggestion
    onSuccess: (_, { displayName }) => {
      const previousData = queryClient.getQueryData<{ data: OKMS }>(
        okmsQueryKeys.detail(okms.id),
      );

      // Update the cache
      queryClient.setQueryData(okmsQueryKeys.detail(okms.id), {
        data: {
          ...previousData.data,
          iam: {
            ...previousData.data.iam,
            displayName,
          },
        },
      });

      queryClient.invalidateQueries({
        queryKey: okmsQueryKeys.listDatagrid,
      });

      clearNotifications();
      addSuccess(t('modify_name_success', { ns: NAMESPACES.ACTIONS }), true);
      onSuccess?.();
},
```

rjamet-ovh: I agree that a classical optimistic update should be made in the `onMutate`, but this is not it. We are in the context of optimistically handling a **side effect**.

We need to wait for the change to be propagated from the /services to the /okms databases in the backend, so we must invalidate list and detail queries after a delay.

I have updated the comments to make this clearer.

packages/manager/apps/okms/public/translations/key-management-service/dashboard/Messages_fr_FR.json
```diff
@@ -5,6 +5,14 @@
   "general_informations": "Informations générales",
   "access_certificates": "Certificats d'accès",
   "logs": "Logs",
+  "okms_secret_config": "Paramètres par défault des Secrets",
```
pauldkn: **typo**
```suggestion
  "okms_secret_config": "Paramètres par défaut des Secrets",
```

PR #19813, opened 2025-10-17
feat(*): handle yarn cache restore

hbelkhir:
## Description

Ticket Reference: #MANAGER-19635    

In this screenshot we can see that the cache was restored  and the time of the installation didn't take so much time to be finished
image

And when we check a PR with the old config, we see the enhance:
image


.github/workflows/chech-duplicates-translations.yml
```diff
@@ -56,12 +59,16 @@ jobs:
...
+            ~/.cache/yarn
+            node_modules
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
 
       - name: Install dependencies
         if: steps.gate.outputs.should_continue == 'true'
-        run: yarn install --frozen-lockfile
+        run: yarn install --frozen-lockfile &> /dev/null
```
helabenkhalfallah: why we need to make the command silent? is not good to have the log?

.github/workflows/gh-pages.yaml
```diff
@@ -16,15 +16,15 @@ jobs:
       - name: Install
         run: |
           yarn --version
-          yarn install
+          yarn install &> /dev/null
```
helabenkhalfallah: why we need this change?

.github/workflows/auto-label-merge-conflicts.yaml
```diff
@@ -10,5 +10,5 @@ jobs:
         with:
           CONFLICT_LABEL_NAME: "has conflicts"
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
-          MAX_RETRIES: 5
-          WAIT_MS: 5000
+          MAX_RETRIES: 10
+          WAIT_MS: 10000
```
helabenkhalfallah: why we need this change? this will not cause a long wait ?

.github/workflows/run-bdd-tests.yml
```diff
@@ -13,17 +13,23 @@ jobs:
...
+          path: |
+            ~/.cache/yarn
+            node_modules
+          key: ${{ runner.os }}-yarn-${{ hashFiles('**/yarn.lock') }}
+          restore-keys: |
+            ${{ runner.os }}-yarn-
+            
       - name: Install repository
-        run: yarn install --frozen-lockfile
+        run: yarn install --frozen-lockfile &> /dev/null
```
helabenkhalfallah: why this change?

.github/workflows/check-duplicates-translations.yml
```diff
@@ -56,12 +59,16 @@ jobs:
         if: steps.gate.outputs.should_continue == 'true'
         uses: actions/cache@v3
         with:
-          path: '**/node_modules'
-          key: ${{ runner.os }}-${{ hashFiles('yarn.lock') }}
+          path: |
```
helabenkhalfallah, line 62: what do you think to use https://docs.gitlab.com/ci/yaml/#include 
instead of repeating same blocks?

.github/workflows/auto-label-merge-conflicts.yaml
hbelkhir: Last time we get an error because it didn't get to verify all the checks because these two vars, I already check it with Anoop, just to do one merge on the master

.github/workflows/check-duplicates-translations.yml
hbelkhir: This a gitlab URL, they don't work on github actions

.github/workflows/chech-duplicates-translations.yml
hbelkhir: It's more pretty without logs, but I will take it back :)

.github/workflows/check-duplicates-translations.yml
helabenkhalfallah: strange in docker we can also do include https://docs.docker.com/compose/how-tos/multiple-compose-files/include/

And in github action the doc:
https://docs.github.com/en/actions/tutorials/create-actions/create-a-composite-action#creating-a-composite-action-within-the-same-repository

we can create reusable workflow https://docs.github.com/en/actions/how-tos/reuse-automations/reuse-workflows#creating-a-reusable-workflow

.github/workflows/auto-label-merge-conflicts.yaml
helabenkhalfallah: How you did fix these values? according to which formula? timeout is not a deterministic value, so we'll need to adjust it each time?

.github/workflows/check-duplicates-translations.yml
hbelkhir: I don't understand, we are talking about `path` param of the builtin `actions/cache`, it's a param included on that action, it's not a github action param

.github/workflows/auto-label-merge-conflicts.yaml
hbelkhir: I will remove this, and will check it again if we will get another issue of timeout

helabenkhalfallah (dismissed):

helabenkhalfallah (approved):
I approve, a code enhancement ticket was created. Thanks :) :)

.github/workflows/check-duplicates-translations.yml
anooparveti, line 1: Possible to correct the file name as it has a typo?
`chech` -> `check`

PR #19859, opened 2025-10-20
test(container): add test for beta navigation switch

aottr:
## Description

Ticket Reference: #MANAGER-16569

### DIFF
--- a/packages/manager/apps/container/src/container/common/nav-reshuffle-switch-back/NavReshuffleSwitchBack.spec.tsx
+++ b/packages/manager/apps/container/src/container/common/nav-reshuffle-switch-back/NavReshuffleSwitchBack.spec.tsx
@@ -1,0 +1,133 @@
+import { it, vi, describe, expect, afterEach } from 'vitest';
+import { render, fireEvent, waitFor } from '@testing-library/react';
+import NavReshuffleSwitchBack from './index';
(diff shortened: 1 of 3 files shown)


packages/manager/apps/container/src/container/common/nav-reshuffle-switch-back/NavReshuffleSwitchBack.spec.tsx
```diff
@@ -0,0 +1,116 @@
...
+  },
+}));
+
+vi.mock('@/context', () => ({
+  useShell: () => mockShell.shell,
+}));
+
+vi.mock('@/core/container', () => ({
+  default: () => containerContextStore.current,
+}));
```
JacquesLarique: You should be able to avoid all these mocks by using this [wrapper](https://github.com/ovh/manager/blob/master/packages/manager/apps/container/src/utils/tests/component-wrapper.ts) for which you can find an example [here](https://github.com/ovh/manager/blob/master/packages/manager/apps/container/src/components/modal-container/ModalContainer.component.spec.tsx)
_Note that you will probably need to mock the calls to feature availability and preferences APIs_

JacquesLarique (approved):

MaximeBajeux (approved):

PR #19865, opened 2025-10-20
fix(dedicated.vps): fix issue with commitment if plesk licence

ghyenne:
## Description

Ticket Reference: #PRDCOL-239

### DIFF
--- a/packages/manager/modules/billing-components/src/components/commitment/commitment.controller.js
+++ b/packages/manager/modules/billing-components/src/components/commitment/commitment.controller.js
@@ -126,9 +126,9 @@ export default class {
       commitment.isPeriodic(),
     );
 
-    this.periodicTotalPrice = Number.parseFloat(
-      periodic.totalPrice.value.toFixed(2),
-    );
+    this.periodicTotalPrice = periodic
+      ? Number.parseFloat(periodic.totalPrice.value.toFixed(2))
+      : null;
 
     if (upfront && periodic) {
       // compute discount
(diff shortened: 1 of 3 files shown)


packages/manager/modules/billing-components/src/components/commitment/commitment.controller.js
```diff
@@ -127,7 +127,7 @@ export default class {
     );
 
     this.periodicTotalPrice = Number.parseFloat(
-      periodic.totalPrice.value.toFixed(2),
+      periodic?.totalPrice.value.toFixed(2),
```
helabenkhalfallah: what if totalPrice or value is null or undefined?

packages/manager/modules/models/service/Commitment.class.js
```diff
@@ -16,7 +16,8 @@ export default class Commitment {
   }
 
   addOptionCommitment(optionEngagement) {
-    this.price += optionEngagement.price;
+    this.price +=
+      optionEngagement.pricing.monthlyPriceValue || optionEngagement.price;
```
JacquesLarique: Why this change ?
I'm not sure we want to only consider the monthly price in the total

ghyenne: We need uniformity for display a good price because several option are yearly and other monthly for same main service (VPS is monthly but plesk is yearly).

packages/manager/modules/billing-components/src/components/commitment/commitment.controller.js
ghyenne: It's fixed.
I made:
```js
this.periodicTotalPrice = periodic ? Number.parseFloat(
      periodic.totalPrice.value.toFixed(2),
    ) : null; 
```

PR #19877, opened 2025-10-21
fix(pci-instances): Hide the label when public network does not have name

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-5235

## Additional Information

Before

image

After

image


ppprevost (approved):

lolottetheclash (approved):
👍

PR #19933, opened 2025-10-23
fix(object-storage): fix issue on non available region

lionel95200x:
ref: #TAPC-5206

## Description

- When region is not activated we should display it on region selection

Ticket Reference: #TAPC-5206


packages/manager/apps/pci-object-storage/src/domain/regions/regionFilter.ts
```diff
@@ -91,6 +89,7 @@ export function filterAllowedRegions(
     );
   }
 
+  console.log({ filteredRegions });
```
Tsiorifamonjena: todo: to remove :)

Tsiorifamonjena (approved):

aTurmo (approved):

PR #19944, opened 2025-10-23
fix(billing): fix add new ssh key

achrafHafedh:
ref: #MANAGER-20110

## Description

fix add new ssh key

Ticket Reference: #MANAGER-20110

### DIFF
--- a/packages/manager/modules/billing/src/autoRenew/ssh/user-ssh.service.js
+++ b/packages/manager/modules/billing/src/autoRenew/ssh/user-ssh.service.js
@@ -61,35 +61,48 @@ export default /* @ngInject */ function UserAccountSshService(
   };
 
   self.getCloudSshList = async function getCloudSshList() {
-    const { data } = await iceberg('/cloud/project')
+    const projects = await self.getCloudProjects();
+
+    const sshKeysByProject = await Promise.all(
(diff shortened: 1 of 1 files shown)


packages/manager/modules/billing/src/autoRenew/ssh/user-ssh.service.js
```diff
@@ -92,6 +94,19 @@ export default /* @ngInject */ function UserAccountSshService(
...
+  self.getCloudProjects = function getCloudProjects() {
+    return iceberg('/cloud/project')
+      .query()
+      .expand('CachedObjectList-Pages')
+      .execute()
+      .$promise.then(({ data }) =>
+        map(filter(data, { status: 'ok' }), (project) => ({
+          id: project.project_id,
+          description: project.description,
+        })),
```
JacquesLarique: Please rewrite this part to avoid using lodash:
```suggestion
        data.filter((project) => project.status === 'ok')
          .map((project) => ({
            id: project.project_id,
            description: project.description,
          })),
```

Also, you can now use this function in `getCloudSshList`

achrafHafedh: Good catch! fixed

PR #19949, opened 2025-10-23
fix(okms): level2 tracking id & minor config for download certificate…

mathieumousnier:
Ticket Reference: #MANAGER-18832


packages/manager/apps/okms/src/components/dashboard/downloadKmsPublicCaLink/DownloadKmsPublicCaLink.tsx
```diff
@@ -78,8 +78,8 @@ export const DownloadKmsPublicCaLink = ({
       });
 
       trackClick({
-        location: PageLocation.page,
-        buttonType: ButtonType.button,
+        location: PageLocation.tile,
+        buttonType: ButtonType.link,
         actionType: 'navigation',
```
rjamet-ovh:
```suggestion
        actionType: 'action',
```
Looking at Tracking specs, it should be action here.

PR #19951, opened 2025-10-23
feat(backup-agent): add services pages

tibs245:
## Description

- Move Service Listing page and delete page of tenants on Backup-agent module
- Refactor duplicated code in commons components 

Ticket Reference: #BKP-598


packages/manager/apps/backup-services/tsconfig.json
```diff
@@ -9,6 +9,10 @@
     "outDir": "dist",
     "resolveJsonModule": true
   },
-  "include": ["src"],
+  "include": [
+    "src",
+    "../../modules/backup-agent/src/types/Tenant.type.ts",
+    "../../modules/backup-agent/src/types/VspcTenant.type.ts"
```
mathieumousnier: This looks like a code smell, no path from another module should appear here.
Is it temporary ?

packages/manager/modules/backup-agent/src/data/hooks/vaults/getVault.ts
```diff
@@ -1,6 +1,6 @@
 import { mockVaults } from '@/mocks/vaults/vaults';
 
-// const BACKUP_VAULTS_LIST_ROUTE = '/backup/tenant/vaults';
+// const BACKUP_VAULTS_LIST_ROUTE = '/backup/tenants/vaults';
```
mathieumousnier: Remaining comment here

packages/manager/apps/backup-services/tsconfig.json
```diff
@@ -9,6 +9,10 @@
...
+  ],
```
rjamet-ovh: Could you import the module types without using a relative path in tsconfig?

packages/manager/modules/backup-agent/src/data/hooks/vaults/getVault.ts
tibs245: Yes it's because we don't use for the moment with mock. Endpoint is not already available

rjamet-ovh (approved):

PR #19972, opened 2025-10-24
fix(account-creation): add asterisks and improve ux

MaximeBajeux:
ref: #MANAGER-19805

## Description

Ticket Reference: #MANAGER-19805


packages/manager/apps/account-creation/src/pages/accountDetails/accountDetails.page.tsx
```diff
@@ -77,11 +77,8 @@ function AccountDetailsForm({
     isSMSConsentAvailable,
   } = useUserContext();
 
-  /**
-   * TODO: Import country and language translations from common-translations module once ready
-   */
-  // const { t: tCountry } = useTranslation(NAMESPACES.COUNTRY);
-  // const { t: tLanguage } = useTranslation(NAMESPACES.LANGUAGE);
+  const { t: tCountry } = useTranslation(NAMESPACES.COUNTRY);
```
aottr: I think in this case it would be better to use an alternative way of using `useTranslation` by passing an array of namespaces to the hook and then define the namespace in the `t`-function.

```js
const { t } = useTranslation(['account-details', NAMESPACES.FORM, NAMESPACES.LANGUAGE, NAMESPACES.COUNTRY]);
```

and then 

```js
tCountry(`country_${countryCode}`)
``` 
becomes 
```js
t(`country_${countryCode}`, { ns: NAMESPACES.COUNTRY })
```

aottr: the first namespace (`account-details`) can be accessed without naming the `ns`.
So here it stays the same with `t('asdf');`

ref: https://react.i18next.com/latest/usetranslation-hook#usetranslation-params

packages/manager/apps/account-creation/src/data/api/rules.ts
```diff
@@ -30,6 +33,7 @@ export const getRules = async (params: RulesParam) => {
     throw new Error('Failed to get account creation rules');
   }
   return data.reduce((acc, rule) => {
+    if (rule.fieldName && EXCLUDED_FIELDS.includes(rule.fieldName)) return acc;
```
aottr, line 36: I was questioning the filter on the `getRules` but it ultimatively makes sense in how we use this api and we don't care about the rules of ovhCompany etc.

packages/manager/apps/account-creation/src/pages/accountDetails/accountDetails.page.tsx
```diff
@@ -722,7 +793,7 @@ function AccountDetailsForm({
                       isChecked={Boolean(value)}
                       value={(value as unknown) as string}
                       onClick={() => onChange(!value)}
-                      class="flex-1"
+                      class="flex-[0]"
```
aottr, line 808: You could check if `flex-initial` also works here, but it's just a nitpic

PR #20015, opened 2025-10-28
feat(observability): add region tile selector in tenant creation form

darsene:
resolves: #MAOBS-120

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/observability/src/components/infrastructures/region-badge-type/RegionBadgeType.component.tsx
+++ b/packages/manager/apps/observability/src/components/infrastructures/region-badge-type/RegionBadgeType.component.tsx
@@ -1,4 +1,4 @@
-import React, { useId, useMemo } from 'react';
+import { useId, useMemo } from 'react';
 
 import { useTranslation } from 'react-i18next';
 
--- a/packages/manager/apps/observability/src/components/infrastructures/region-selector/RegionSelector.component.tsx
+++ b/packages/manager/apps/observability/src/components/infrastructures/region-selector/RegionSelector.component.tsx
@@ -63,7 +63,7 @@ export default function RegionSelector() {
                   </OdsTab>
                 ))}
               </OdsTabs>
-              <div className="grid grid-cols-1 gap-4 mt-4 w-1/2">
+              <div className="space-y-4 mt-4">
                 {groupedInfrastructures[activeZone]?.map(({ id, locationDetails }) => (
                   <RadioCard
                     id={id}
--- a/packages/manager/apps/observability/src/__tests__/components/form/information-form/InformationForm.component.spec.tsx
+++ b/packages/manager/apps/observability/src/__tests__/components/form/information-form/InformationForm.component.spec.tsx
@@ -270,14 +270,11 @@ describe('InformationForm', () => {
 
   describe('Component Structure', () => {
     it('should render with correct field types and layout', () => {
-      const { container } = render(
+      render(
         <TestWrapper>
           <InformationForm {...defaultProps} />
         </TestWrapper>,
       );
-
-      const gridContainer = container.querySelector('.grid.grid-cols-1.gap-4.mt-6');
-      expect(gridContainer).toBeInTheDocument();
 
       const titleInput = screen.getByTestId('title-input');
       const descriptionInput = screen.getByTestId('description-input');
--- a/packages/manager/apps/observability/src/__tests__/components/metrics/tenant-configuration-form/TenantConfigurationForm.component.spec.tsx
+++ b/packages/manager/apps/observability/src/__tests__/components/metrics/tenant-configuration-form/TenantConfigurationForm.component.spec.tsx
@@ -648,17 +648,4 @@ describe('TenantConfigurationForm', () => {
       expect(quantityInput).toHaveValue(null);
     });
   });
-
-  describe('Component Structure', () => {
-    it('should render all form fields in correct structure', () => {
-      const { container } = render(
-        <TestWrapper defaultValues={{ infrastructureId: 'infra-123', maxSeries: 100 }}>
-          <TenantConfigurationForm />
-        </TestWrapper>,
-      );
-
-      const gridContainer = container.querySelector('.grid.grid-cols-1.gap-4');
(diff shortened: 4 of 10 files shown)


packages/manager/apps/observability/src/utils/infrastructures.utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  infrastructures: Infrastructure[],
+): Record<string, InfrastructureWithLocation[]> => {
+  const initial: Record<string, InfrastructureWithLocation[]> = { all: [] };
+
+  return infrastructures
+    .filter(hasLocationDetails)
+    .reduce((acc, infrastructure: InfrastructureWithLocation) => {
+      const geographicZone = infrastructure.locationDetails.geographyCode.toLowerCase();
+
+      acc.all!.push(infrastructure);
```
mathieumousnier: Like this you avoid forcing typescript, and if `acc.all` is nullish then the push will be skipped.

```suggestion
      acc.all?.push(infrastructure);
```

packages/manager/apps/observability/src/utils/infrastructures.utils.ts
```diff
@@ -0,0 +1,38 @@
...
+      (acc[geographicZone] ??= []).push(infrastructure);
```
mathieumousnier, line 44: Nitpick: I think I would have wrote this like below, because this syntax pushes me over my limits I guess :)

```suggestion
      acc[geographicZone] = acc[geographicZone] ?? [];
      acc[geographicZone].push(infrastructure);
```

packages/manager/apps/observability/src/components/infrastructures/region-selector/RegionSelector.component.tsx
```diff
@@ -23,35 +28,72 @@ export default function RegionSelector() {
...
+    if (!infrastructures) {
+      return {};
+    }
+    return groupByGeographicZone(infrastructures);
+  }, [infrastructures]);
+
+  const zones = useMemo(() => {
+    const keys = Object.keys(groupedInfrastructures);
+    // Put 'all' first, then other zones
+    return keys.includes('all') ? ['all', ...keys.filter((key) => key !== 'all')] : keys;
```
mathieumousnier: This "all" key is repeated multiple times, in 2 files, and has no type safety check because it is managed as a string.

I think it should at least be declared as a constant in the utility file to cleanup a bit.

Even better, this code should be put in a function in the utility file (`buildZoneList` for example), so this "all" value stays in the utils file.

packages/manager/apps/observability/src/components/infrastructures/region-selector/RegionSelector.component.tsx
```diff
@@ -23,35 +28,72 @@ export default function RegionSelector() {
     types: 'SHARED',
   });
 
+  const groupedInfrastructures = useMemo(() => {
+    if (!infrastructures) {
+      return {};
+    }
```
mathieumousnier: I think I would have managed this case in the groupByGeographicZone utility function.
- It would be covered by the unit tests of the function.
- Code would be cleaner here

packages/manager/apps/observability/src/components/infrastructures/region-selector/RegionSelector.component.tsx
```diff
@@ -23,35 +28,72 @@ export default function RegionSelector() {
...
+    return keys.includes('all') ? ['all', ...keys.filter((key) => key !== 'all')] : keys;
+  }, [groupedInfrastructures]);
+
+  const [activeZone, setActiveZone] = useState<string>('all');
+
+  useMemo(() => {
+    if (zones.length > 0 && !zones.includes(activeZone) && zones[0]) {
+      setActiveZone(zones[0]);
+    }
+  }, [zones, activeZone]);
```
mathieumousnier: I wouldn't use a `useMemo` like this, it is intended to return a value, and not to do state updates.

The most direct fix would be to use a `useEffect`, [or maybe no useEffect at all like this pattern](https://react.dev/learn/you-might-not-need-an-effect#adjusting-some-state-when-a-prop-changes).

But it is still weird.
What is that thing doing anyway ? (I can't get what it does)

packages/manager/apps/observability/src/components/infrastructures/region-badge-type/RegionBadgeType.component.tsx
```diff
@@ -0,0 +1,55 @@
...
+        icon="circle-info"
+        iconAlignment="right"
+        size="sm"
+      />
+      <OdsTooltip triggerId={id} position="right" withArrow>
+        <OdsText preset="caption" className="w-56">
+          <span>{tooltipLabels[type]}</span>
+        </OdsText>
+      </OdsTooltip>
+    </>
```
mathieumousnier, line 54: The `id` shared between the badge and the tooltip should not be shared with any other components.

Ensuring a unique `id` is the only way to ensure the "show on hover" behavior.

packages/manager/apps/observability/src/components/form/radio-card/RadioCard.component.tsx
```diff
@@ -0,0 +1,65 @@
...
+  onChange,
+}: RadioCardProps) => {
+  const isChecked = selected === id;
+
+  return (
+    <label
+      key={id}
+      htmlFor={id}
+      className={clsx(
+        'flex border-solid rounded-md border px-3 py-4',
```
mathieumousnier: To fix the padding quickly: 

```suggestion
        'flex border-solid rounded-md border px-5 py-6',
```

packages/manager/apps/observability/src/utils/infrastructures.utils.ts
darsene: Type created to fix it

packages/manager/apps/observability/src/utils/infrastructures.utils.ts
darsene: It's quite standard https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Nullish_coalescing_assignment

PR #20016, opened 2025-10-28
Fix(pci-workflow): fix region selector ui

aTurmo:
ref: #TAPC-4882, #TAPC-4884, #TAPC-4895

Before:
image

After:
image


fredericvilcot:
@aTurmo question: why do you have 5 commits for such a few changes?
BTW you have to verify your commits :)

aTurmo:
> @aTurmo question: why do you have 5 commits for such a few changes? BTW you have to verify your commits :)

@fredericvilcot I wanted to keep separated the different changes for the review (If think it make it easier to review when there are thematic commits). I'll squash the commits before merging when it's done.

And yes i updated my GPG key but it seems that there is a problem with the email.

aTurmo:
> @aTurmo question: why do you have 5 commits for such a few changes? BTW you have to verify your commits :)

All good now

PR #20068, opened 2025-10-29
feat(public-vcf-aas): add edge gateway tab

pauldkn:
## Description

Ticket Reference: #MANAGER-18386

### DIFF
--- a/packages/manager/apps/hpc-vmware-public-vcf-aas/src/pages/dashboard/datacentre/datacentreDashboard.constants.ts
+++ b/packages/manager/apps/hpc-vmware-public-vcf-aas/src/pages/dashboard/datacentre/datacentreDashboard.constants.ts
@@ -1,2 +1,3 @@
 export const COMPUTE_LABEL = 'Compute';
 export const STORAGE_LABEL = 'Storage';
+export const EDGE_GATEWAY_LABEL = 'Edge Gateway';
--- a/packages/manager/apps/hpc-vmware-public-vcf-aas/src/pages/dashboard/datacentre/edge-gateway/hooks/useEdgeGatewayListingColumns.ts
+++ b/packages/manager/apps/hpc-vmware-public-vcf-aas/src/pages/dashboard/datacentre/edge-gateway/hooks/useEdgeGatewayListingColumns.ts
@@ -1,0 +1,44 @@
+import { VCDEdgeGateway } from '@ovh-ux/manager-module-vcd-api';
+import { DatagridColumn } from '@ovh-ux/manager-react-components';
+import { useTranslation } from 'react-i18next';
+import { NAMESPACES } from '@ovh-ux/manager-common-translations';
+import {
+  EdgeGatewayNameCell,
+  EdgeGatewayConnectivityCell,
+  EdgeGatewayIPBlockCell,
+  EdgeGatewayActionCell,
+} from '../components/EdgeGatewayCells.component';
+
+export const useEdgeGatewayListingColumns = () => {
+  const { t } = useTranslation('datacentres/edge-gateway');
+  const { t: tCommonDashboard } = useTranslation(NAMESPACES.DASHBOARD);
+
+  const columns: Array<DatagridColumn<VCDEdgeGateway>> = [
+    {
+      id: 'name',
+      cell: EdgeGatewayNameCell,
+      label: tCommonDashboard('name'),
+      isSortable: false,
+    },
+    {
+      id: 'connectivity_type',
+      cell: EdgeGatewayConnectivityCell,
+      label: t('edge_connectivity_type'),
+      isSortable: false,
+    },
+    {
+      id: 'ip_block',
+      cell: EdgeGatewayIPBlockCell,
+      label: t('edge_ip_block'),
+      isSortable: false,
+    },
+    {
+      id: 'actions',
+      cell: EdgeGatewayActionCell,
+      label: '',
+      isSortable: false,
+    },
+  ];
+
+  return columns;
+};
--- a/packages/manager/apps/hpc-vmware-public-vcf-aas/src/routes/routes.tsx
+++ b/packages/manager/apps/hpc-vmware-public-vcf-aas/src/routes/routes.tsx
@@ -96,6 +96,12 @@ const DeleteVrackSegmentPage = React.lazy(() =>
   ),
 );
 
+const EdgeGatewayListingPage = React.lazy(() =>
+  import(
+    '@/pages/dashboard/datacentre/edge-gateway/DatacentreEdgeGateway.page'
+  ),
+);
+
 const TerminateOrganizationPage = React.lazy(() =>
   import('@/pages/terminate/TerminateOrganization.page'),
 );
@@ -309,6 +315,11 @@ export default (
               pageType: PageType.popup,
             },
           }}
+        />
+        <Route
+          id={'edge-gateway'}
+          path={urls.edgeGateway}
+          Component={EdgeGatewayListingPage}
         />
       </Route>
       <Route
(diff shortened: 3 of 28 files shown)


packages/manager/apps/hpc-vmware-public-vcf-aas/src/pages/dashboard/datacentre/edge-gateway/hooks/useEdgeGatewayListingColumns.ts
```diff
@@ -0,0 +1,44 @@
...
+import { useTranslation } from 'react-i18next';
+import { NAMESPACES } from '@ovh-ux/manager-common-translations';
+import {
+  EdgeGatewayNameCell,
+  EdgeGatewayConnectivityCell,
+  EdgeGatewayIPBlockCell,
+  EdgeGatewayActionCell,
+} from '../components/EdgeGatewayCells.component';
+
+export const useEdgeGatewayListingColumns = () => {
```
sefiani-ovh, line 12: humm  i think this should not be considered a hook 
actually for me a hook should manage states or context
for me its just a static array where yes yoiu call usetranslaiton hook.
up to you to change the name

packages/manager/apps/hpc-vmware-public-vcf-aas/src/routes/routes.tsx
```diff
@@ -310,6 +316,17 @@ export default (
             },
           }}
         />
+        <Route
+          id={'edge-gateway'}
+          path={urls.edgeGateway}
+          Component={EdgeGatewayListingPage}
+          handle={{
+            tracking: {
```
sefiani-ovh: you have already tracking spec ?
if not no need to handle tracking (we do it when we will have the right ticket)

packages/manager/modules/vcd-api/src/hook/useVcdEdgeGateway.ts
```diff
@@ -0,0 +1,30 @@
...
+
+export const useVcdEdgeGatewaysMocks = (
+  id: string,
+  vdcId: string,
+): UseQueryResult<VCDEdgeGateway[], Error> =>
+  useQuery({
+    queryKey: getVcdEdgeGatewayListQueryKey(id, vdcId),
+    queryFn: () =>
+      new Promise((resolve) => {
+        console.log('🛜 mocking useVcdEdgeGateways api call...');
```
sefiani-ovh, line 24: you want to keep the log?

packages/manager/apps/hpc-vmware-public-vcf-aas/src/pages/dashboard/datacentre/datacentreDashboard.constants.ts
```diff
@@ -1,2 +1,8 @@
+import { VCDDatacentre } from '@ovh-ux/manager-module-vcd-api';
+
 export const COMPUTE_LABEL = 'Compute';
 export const STORAGE_LABEL = 'Storage';
+export const EDGE_GATEWAY_LABEL = 'Edge Gateway';
+
+export const isEdgeCompatibleVDC = (vdc: VCDDatacentre) =>
```
sefiani-ovh: since it's function , i dont know if you can move it to some utilities file helper
but since its used only in one place 
for me just declare 
const OFFER_PROFILE_COMPATIBLE_VDC='NSX'
use the condition directly in the component (but up to you)

packages/manager/apps/hpc-vmware-public-vcf-aas/src/pages/dashboard/datacentre/edge-gateway/hooks/useEdgeGatewayListingColumns.ts
pauldkn: Actually it's defined like this in new µapps from the generator so I decided to use this way of defining listing columns, it allows to split the listing page and make it more readable

packages/manager/modules/vcd-api/src/hook/useVcdEdgeGateway.ts
pauldkn: Yes just for now, as it's only for testing with mocks but it will definitely be removed when prod (this hook using mocks will be deleted)

sefiani-ovh (approved):

packages/manager/modules/vcd-api/src/api/vcd-edge-gateway.ts
```diff
@@ -0,0 +1,9 @@
+import { ApiResponse, v2 } from '@ovh-ux/manager-core-api';
+import { VCDEdgeGateway } from '../types';
+import { getVcdEdgeGatewayListRoute } from '../utils';
+
+export const getVcdEdgeGateways = async (
+  id: string,
+  vdcId: string,
+): Promise<ApiResponse<VCDEdgeGateway[]>> =>
+  v2.get(getVcdEdgeGatewayListRoute(id, vdcId));
```
mathieumousnier: Maybe you used this pattern everywhere so it would be hard to change, but I would really advise to avoid exposing axios internals out of your fetch functions.

Here your returning "data", "status", everything that axios returns, but what you only want is the "data".
This is cleaner, and you won't have to do something like "[data.data](select: (data) => data.data)" on the useQuery hook.

```ts
export const getVcdEdgeGateways = async (id: string, vdcId: string) => {
  const { data } = await v2.get<VCDEdgeGateway[]>(
    getVcdEdgeGatewayListRoute(id, vdcId),
  );
  return data;
};
```

packages/manager/apps/hpc-vmware-public-vcf-aas/src/pages/dashboard/datacentre/edge-gateway/hooks/useEdgeGatewayListingColumns.ts
mathieumousnier: @sefiani-ovh this is a hook because it uses "useTranslation", which is a hook.

packages/manager/apps/hpc-vmware-public-vcf-aas/src/test-utils/test-i18n.ts
```diff
@@ -27,8 +30,10 @@ export const translations = {
   'datacentres/order': datacentresOrder,
   'datacentres/storage': datacentresStorage,
   'datacentres/vrack-segment': datacentresVrackSegment,
+  'datacentres/edge-gateway': datacentresEdgeGateway,
   error,
   terminate,
+  [`${NAMESPACES.DASHBOARD}`]: commonDashboard,
```
mathieumousnier:
```suggestion
  [NAMESPACES.DASHBOARD]: commonDashboard,
```

packages/manager/modules/vcd-api/src/hook/useVcdEdgeGateway.ts
```diff
@@ -0,0 +1,30 @@
...
+import { getVcdEdgeGatewayListQueryKey } from '../utils';
+
+export const useVcdEdgeGateways = (
+  id: string,
+  vdcId: string,
+): UseQueryResult<VCDEdgeGateway[], Error> =>
+  useQuery({
+    queryKey: getVcdEdgeGatewayListQueryKey(id, vdcId),
+    queryFn: () => getVcdEdgeGateways(id, vdcId),
+    select: (data) => data.data,
```
mathieumousnier: As said in another comment, here you are hidding an axios internal property, but this should have been done in the fetch function

packages/manager/modules/vcd-api/src/api/vcd-edge-gateway.ts
pauldkn: I totally agree 👍
I’ve updated this one accordingly and I’m planning a small refactor to apply the same pattern across the other api functions as well

PR #20095, opened 2025-10-30
fix(telecom.sms): delete sms classType on batch sms

ghyenne:
## Description

Ticket Reference: #PRDCOL-247


packages/manager/modules/sms/src/sms/batches/create/translations/Messages_fr_FR.json
```diff
@@ -34,10 +34,6 @@
   "sms_programming": "Programmation de la campagne",
   "sms_programming_date": "Date :",
   "sms_programming_time": "Heure :",
-  "sms_options_type": "Enregistrement du SMS",
-  "sms_options_type_FLASH": "Flash",
-  "sms_options_type_PHONE": "Mémoire téléphone",
-  "sms_options_type_SIM": "Carte SIM",
```
erazafimovh, line 40: Perhaps we should remove it from all translation files

ghyenne: We have to change only the fr_FR file then it's robot who clean the others language file.

erazafimovh (approved):

PR #20128, opened 2025-10-31
feat(common-translations): add area namespace

MaximeBajeux:
ref: #MANAGER-19805

## Description

Ticket Reference: #...


packages/manager/modules/common-translations/public/translations/area/Messages_fr_FR.json
anooparveti, line 1: **question:** Since this is areas for multiple countries, it has a scope to grow large and i think it can be unnecessary to load all areas. Can we split this areas in different sub-namespaces depending on the country and load only what's required?

PR #20159, opened 2025-11-03
fix(pci-block-storage): fix displayed attachable instances

aTurmo:
ref: #TAPC-5257

### DIFF
--- a/packages/manager/apps/pci-block-storage/src/api/hooks/useVolume.tsx
+++ b/packages/manager/apps/pci-block-storage/src/api/hooks/useVolume.tsx
@@ -239,12 +239,14 @@ export const useAttachVolume = ({
         getVolumesQueryKey(projectId),
         (data: { id: string }[]) =>
           data?.map((v) =>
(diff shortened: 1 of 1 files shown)


packages/manager/apps/pci-block-storage/src/api/hooks/useVolume.tsx
```diff
@@ -234,17 +234,19 @@ export const useAttachVolume = ({
   const mutation = useMutation({
     mutationFn: async () => attachVolume(projectId, volumeId, instanceId),
     onError,
-    onSuccess: (volume) => {
+    onSuccess: async (volume) => {
```
ppprevost: question: you're not awaiting something here ? so Why async ?

aTurmo: I added it to test something and forgot to remove it

aTurmo: removed

PR #20169, opened 2025-11-04
feat(backup-agent): add tenants tabs on dashboard page

tibs245:
ref: #BKP-498

## Description

Ticket Reference: #...


packages/manager/modules/backup-agent/public/translations/common/Messages_fr_FR.json
```diff
@@ -1,5 +1,6 @@
 {
   "vaults": "Vaults",
   "services": "Services",
-  "billings": "Facturations"
+  "billings": "Facturations",
```
pauldkn: (it depends on the jira but it seems more logical in the singular form)
```suggestion
  "billing": "Facturation",
```

PR #20177, opened 2025-11-04
feat(account-creation): update design and add popover in choose account creation

achrafHafedh:
ref: #MANAGER-19990

## Description

Ticket Reference: #MANAGER-19990

### DIFF
--- a/packages/manager/apps/account-creation/src/pages/accountType/AccountType.spec.tsx
+++ b/packages/manager/apps/account-creation/src/pages/accountType/AccountType.spec.tsx
@@ -71,12 +71,12 @@ describe('AccountTypePage', () => {
   it('should display radio button for each legal form given by the rules API', async () => {
     render(<AccountTypePage />);
 
-    const corporationElement = screen.getByText('legal_form_corporation');
-    const individualElement = screen.getByText('legal_form_individual');
-    const associationElement = screen.getByText('legal_form_association');
-    expect(corporationElement).toBeInTheDocument();
-    expect(individualElement).toBeInTheDocument();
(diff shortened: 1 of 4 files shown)


packages/manager/apps/account-creation/src/pages/accountType/AccountType.spec.tsx
```diff
@@ -71,12 +71,12 @@ describe('AccountTypePage', () => {
...
-    const associationElement = screen.getByText('legal_form_association');
-    expect(corporationElement).toBeInTheDocument();
-    expect(individualElement).toBeInTheDocument();
-    expect(associationElement).toBeInTheDocument();
+    const corporationElement = screen.getAllByText('legal_form_corporation');
+    const individualElement = screen.getAllByText('legal_form_individual');
+    const associationElement = screen.getAllByText('legal_form_association');
+    expect(corporationElement).toHaveLength(2);
+    expect(individualElement).toHaveLength(2);
+    expect(associationElement).toHaveLength(2);
```
JacquesLarique, line 79: Why do we have two elements for each legal form now ?
Is it because the popover is always in the DOM even if not displayed ?

achrafHafedh: The OdsPopover needs to be mounted from the start, because it's attaches to the triggerId when it mounts. If the component isn’t mounted at the time of the first click, it won’t detect the event correctly.

JacquesLarique (dismissed):

sachinrameshn (approved):

aottr (approved):

PR #20189, opened 2025-11-04
feat(account-creation): add popover in settings page

achrafHafedh:
ref: #MANAGER-20104

## Description

Ticket Reference: #MANAGER-20104


packages/manager/apps/account-creation/src/pages/settings/settings.page.tsx
```diff
@@ -201,9 +224,24 @@ export default function Settings() {
...
               </label>
+              <OdsPopover
+                className="md:w-1/2 p-5"
+                triggerId="site-setting-description"
+                position="top-start"
+                withArrow
+              >
+                <AccountSettingsPopoverContent
+                  description={t('setting_description')}
+                  list={listSettingsDescription}
```
JacquesLarique: Front the ticket description I think there should be no list in this popover

packages/manager/apps/account-creation/src/pages/settings/popover-content/PopoverContent.tsx
```diff
@@ -0,0 +1,25 @@
...
+};
+
+export default function AccountSettingsPopoverContent({
+  description,
+  list,
+}: Props) {
+  return (
+    <div className="flex flex-col">
+      <OdsText preset={ODS_TEXT_PRESET.caption}>{description}</OdsText>
+      <ol className="list-none space-y-2">
```
JacquesLarique: Since we have a case where list should be empty, we could set up a condition to display the `ul`

packages/manager/apps/account-creation/src/pages/settings/settings.page.tsx
achrafHafedh: fixed !  (Racquel isn’t here this afternoon, but I made the fix according to what was requested in the ticket)

packages/manager/apps/account-creation/src/pages/settings/popover-content/PopoverContent.tsx
achrafHafedh: fixed !

sachinrameshn (approved):

packages/manager/apps/account-creation/src/pages/settings/popover-content/PopoverContent.tsx
```diff
@@ -0,0 +1,27 @@
...
+};
+
+export default function AccountSettingsPopoverContent({
+  description,
+  list,
+}: Props) {
+  return (
+    <div className="flex flex-col">
+      <OdsText preset={ODS_TEXT_PRESET.caption}>{description}</OdsText>
+      {list && list.length > 0 && (
```
JacquesLarique:
```suggestion
      {list?.length && (
```

PR #20200, opened 2025-11-05
feat(ldp): add spinner on angular module

yjaaouane:
## Description

Ticket Reference: #MAOBS-107

### DIFF
--- a/packages/manager/modules/log-to-customer/src/tiles/controller.js
+++ b/packages/manager/modules/log-to-customer/src/tiles/controller.js
@@ -100,12 +100,15 @@ export default class LogToCustomerTileCtrl {
   }
 
   getDataStreams() {
-    return this.LogToCustomerService.getDataStreams().then(
-      ({ streams, streamCount }) => {
+    this.loadingLogAccountStreams = true;
+    return this.LogToCustomerService.getDataStreams()
+      .then(({ streams, streamCount }) => {
         this.logAccountStreams = streams;
         this.hasAtLeastOneDataStream = streamCount > 0;
-      },
-    );
+      })
+      .finally(() => {
+        this.loadingLogAccountStreams = false;
+      });
   }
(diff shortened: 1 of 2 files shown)


packages/manager/modules/log-to-customer/src/tiles/template.html
```diff
@@ -133,6 +133,11 @@ <h5 class="text-primary" data-translate="log_tile_section_title"></h5>
                     <span data-translate="log_button_create_account"></span>
                 </a>
 
+                <oui-spinner
+                    size="s"
+                    data-ng-if="!$ctrl.logAccountStreams.length"
+                ></oui-spinner>
```
darsene, line 139: What about whether `logAccountStreams` is empty?
Your spinner will always displayed.
IMO, you should manage it with a loading property rather.

packages/manager/modules/log-to-customer/src/tiles/controller.js
```diff
@@ -100,10 +100,12 @@ export default class LogToCustomerTileCtrl {
...
 
   getDataStreams() {
+    this.loadingLogAccountStreams = true;
     return this.LogToCustomerService.getDataStreams().then(
       ({ streams, streamCount }) => {
         this.logAccountStreams = streams;
         this.hasAtLeastOneDataStream = streamCount > 0;
+        this.loadingLogAccountStreams = false;
       },
     );
```
darsene: To hide spinner in case of error
```suggestion
    this.loadingLogAccountStreams = true;
    return this.LogToCustomerService.getDataStreams().then(
    ({ streams, streamCount }) => {
    this.logAccountStreams = streams;
    this.hasAtLeastOneDataStream = streamCount > 0;
    },
    ).finally( () => {
    this.loadingLogAccountStreams = false;
    });
```

PR #20209, opened 2025-11-05
feat(pci-instances): add list of distributions

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-4629

## Additional Information

image

image


packages/manager/apps/pci-instances/src/pages/instances/create/components/distributionImage/DistributionImageList.component.tsx
```diff
@@ -0,0 +1,115 @@
...
+    <Controller
+      control={control}
+      name="distributionImageName"
+      render={({ field }) => (
+        <RadioGroup
+          className={clsx('grid gap-6 mt-10', {
+            [`grid-cols-${distributions.length % 2 === 0 ? 2 : 3}`]:
+              distributions.length > 1,
+          })}
+          value={selectedDistributionImageName ?? ''}
```
lolottetheclash: I think it should always have a `selectedDistributionImageName`

packages/manager/apps/pci-instances/src/pages/instances/create/components/distributionImage/DistributionImageList.component.tsx
```diff
@@ -0,0 +1,115 @@
...
+              selectable
+              className="justify-center p-5"
+              selected={selectedDistributionImageName === distribution.id}
+            >
+              <PciCard.Header>
+                <Radio value={distribution.id}>
+                  <RadioControl />
+                  <RadioLabel
+                    className={clsx(
+                      `font-bold text-lg text-[--ods-color-heading]`,
```
lolottetheclash: You could move classname management outside of JSX, given the conditions (not blocking)

packages/manager/apps/pci-instances/src/pages/instances/create/components/distributionImage/DistributionImageList.component.tsx
```diff
@@ -0,0 +1,115 @@
...
+    });
+  };
+
+  return (
+    <Controller
+      control={control}
+      name="distributionImageName"
+      render={({ field }) => (
+        <RadioGroup
+          className={clsx('grid gap-6 mt-10', {
```
lolottetheclash: You could move classname management outside of JSX, given the conditions (not blocking)

packages/manager/apps/pci-instances/src/components/distributionImageLabel/DistributionImageLabel.component.tsx
```diff
@@ -0,0 +1,34 @@
...
+  'linux',
+  'n8n',
+  'nvidia',
+  'plesk',
+  'rockylinux',
+  'ubuntu',
+  'windows',
+] as const;
+
+type TDistributionName = typeof DISTRIBUTION[number];
```
lolottetheclash: you could extract all this in common.types, (there is already  `DEPLOYMENT_MODES` etc...)

packages/manager/apps/pci-instances/src/pages/instances/create/components/distributionImage/DistributionImageList.component.tsx
```diff
@@ -0,0 +1,115 @@
...
+                      {
+                        [`${
+                          distributions.length % 2 === 0
+                            ? 'max-w-[300px]'
+                            : 'max-w-[200px]'
+                        }`]: distributions.length > 1,
+                      },
+                    )}
+                  >
+                    <DistributionImageLabel name={distribution.name}>
```
lolottetheclash, line 93: is there a reason to have the name not mandatory?

packages/manager/apps/pci-instances/src/pages/instances/create/components/distributionImage/DistributionImageList.component.tsx
Tsiorifamonjena: Yes there will always a selectedDistributionImageName but since we have `distributionImageNameSchema = z.string().nullable();` typescript throw error. But I cast it to force it to be string then

packages/manager/apps/pci-instances/src/pages/instances/create/components/distributionImage/DistributionImageList.component.tsx
Tsiorifamonjena: name can be undefined in case of a backups

Tsiorifamonjena:
> Nice work, but I have some troubleshootings here: When I click on a card, it only selects the image if I click on the radio button. The selection behavior should be extended to the card. When I change the distribution type, it keeps the previous image selected. Shouldn't it automatically select the first image in the list? When I select “Distribution + apps,” I don't see any images (not a big deal, since they will come from the viewModels).

@lolottetheclash I fix issues. For "Distribution + apps" I add some mock data I don't know exactly all the data to put in there.

lolottetheclash (approved):

PR #20210, opened 2025-11-05
feat(backup-agent): add bucket vault list

tibs245:
ref: #BKP-493

## Description

Ticket Reference: #...


packages/manager/modules/backup-agent/src/pages/vaults/dashboard/buckets/VaultBucket.spec.tsx
```diff
@@ -0,0 +1,21 @@
+import { waitFor, screen } from '@testing-library/react';
+import { renderTest } from '@/test-utils/Test.utils';
+
+const COLUMN_NAME = ["Nom", "Localisation", "Région"]
```
pauldkn: Why not using `labels.name` and so on ?

pauldkn (approved):

packages/manager/modules/backup-agent/src/pages/vaults/dashboard/buckets/VaultBucket.spec.tsx
tibs245: It's true, can be better. I will fix it

PR #20225, opened 2025-11-05
fix(pci-project): clear notifications to avoid stacking and forwardin…

FabienHenon:
…g bw pages

ref: #MANAGER-20213

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-project/src/pages/Layout.tsx
+++ b/packages/manager/apps/pci-project/src/pages/Layout.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useContext, Suspense } from 'react';
+import { useEffect, useContext, Suspense, useState } from 'react';
 import { defineCurrentPage } from '@ovh-ux/request-tagger';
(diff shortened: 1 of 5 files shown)


packages/manager/apps/pci-project/src/pages/Layout.tsx
```diff
@@ -27,6 +31,13 @@ export default function Layout() {
     shell.ux.hidePreloader();
   }, []);
 
+  useEffect(() => {
+    if (originLocation.pathname !== location.pathname) {
+      clearNotifications();
+      setOriginLocation(location);
+    }
+  }, [location.pathname, originLocation.pathname]);
+
```
aTurmo, line 36: Suggestion : I think you will have the same effect if you just have a use effect with clearNotifications(); and location.pathname as a dependency. Since useEffect trigger each time a dependency changes, you don't need to store it.

FabienHenon: Indeed! Thank you

PR #20229, opened 2025-11-05
fix(billing): add error message for renew bulk actions

JacquesLarique:
## Description

Reworked the error management for enabling and disabling autorenew to display the list of errors messages

Ticket Reference: #MANAGER-20226

### DIFF
--- a/packages/manager/modules/billing/src/autoRenew/disable/disable.routing.js
+++ b/packages/manager/modules/billing/src/autoRenew/disable/disable.routing.js
@@ -1,5 +1,6 @@
 import filter from 'lodash/filter';
-import map from 'lodash/map';
+
+import { revertFailedBulkAction } from '../helpers/bulk-action-message.helper';
 
 export default /* @ngInject */ ($stateProvider) => {
   $stateProvider.state('billing.autorenew.disableRedirection', {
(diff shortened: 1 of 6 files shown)


packages/manager/modules/billing/src/autoRenew/disable/disable.routing.js
```diff
@@ -27,13 +26,22 @@ export default /* @ngInject */ ($stateProvider) => {
...
+        ).catch((error) => {
+          for (let i = 0; i < error?.messages?.length ?? 0; i += 1) {
+            const message = error?.messages[i];
+            const service = services.find((s) => s.serviceId === message.id);
+            if (service) {
+              service.setAutomaticRenew();
+            }
+          }
+          return $q.reject(error);
+        }),
```
achrafHafedh, line 42: I think it’s the same code repeating in the catch block of “enable.routing”

packages/manager/modules/billing/src/autoRenew/helpers/bulk-action-message.helper.js
```diff
@@ -0,0 +1,21 @@
...
+  messages = [],
+  revert = () => {},
+) => {
+  for (let i = 0; i < messages.length; i += 1) {
+    const message = messages[i];
+    const service = services.find((s) => s.serviceId === message.id);
+    if (service) {
+      revert(service);
+    }
+  }
```
sachinrameshn: More readable 
```suggestion
  messages.forEach((message) => {
  const service = services?.find((s) => s.serviceId === message.id);
  service && revert(service);
});
```

PR #20230, opened 2025-11-05
feat(pci-block-storage): retyping - disable retyping for encrypted volumes

aTurmo:
ref: #TAPC-5213

The first commit handle the new rules for which volume can retype, with some refactoring to avoid having the rules in multiple place.

Also added a 'cancel' button that close the drawer if the user go to the retype url of a volume that can't be retype (can't go the normal way because the CTA is disabled).
And finally disable scrolling when the drawer is added. I took what was done in object storage (thx @lionel95200x for the solution)


packages/manager/apps/pci-block-storage/public/translations/retype/Messages_fr_FR.json
```diff
@@ -1,5 +1,5 @@
 {
-  "pci_projects_project_storages_blocks_retype_cant_retype": "Les fonctionnalités de modification du type et de chiffrement ne sont pas disponibiles pour les volumes Classic 3AZ",
+  "pci_projects_project_storages_blocks_retype_cant_retype": "Les fonctionnalités de modification du type et de chiffrement ne sont pas disponibiles pour les volumes Classic 3AZ ou encryptés",
```
aTurmo:
```suggestion
  "pci_projects_project_storages_blocks_retype_cant_retype": "Les fonctionnalités de modification du type et de chiffrement ne sont pas disponibles pour les volumes Classic 3AZ ou encryptés",
```

lolottetheclash (approved):
👍🏻

ppprevost (approved):

PR #20251, opened 2025-11-06
build(observability.logs): prepare module for versionning

darsene:
## Description

Ticket Reference: #MAOBS-136

### DIFF
--- a/.github/workflows/run-bdd-tests.yml
+++ b/.github/workflows/run-bdd-tests.yml
@@ -47,6 +47,7 @@ jobs:
           yarn build --filter="./packages/manager/modules/order"
           yarn build --filter="./packages/manager/modules/common-api"
           yarn build --filter="./packages/manager/modules/common-translations"
+          yarn build --filter="./packages/manager/modules/logs-to-customer"
           yarn build --filter="./packages/manager/modules/vcd-api"
           yarn build --filter="./packages/manager/modules/manager-pci-common"
           yarn build --filter="./packages/manager/modules/network-common"
--- a/packages/manager-tools/manager-pm/src/playbook/catalog/pnpm-private-modules.json
+++ b/packages/manager-tools/manager-pm/src/playbook/catalog/pnpm-private-modules.json
@@ -128,10 +128,6 @@
     "pnpm": "packages/manager/modules/iplb"
   },
   {
-    "turbo": "--filter @ovh-ux/logs-to-customer",
-    "pnpm": "packages/manager/modules/logs-to-customer"
-  },
-  {
     "turbo": "--filter @ovh-ux/manager-components",
     "pnpm": "packages/manager/modules/manager-components"
   },
--- a/packages/manager/modules/logs-to-customer/package.json
+++ b/packages/manager/modules/logs-to-customer/package.json
@@ -1,7 +1,6 @@
 {
   "name": "@ovh-ux/logs-to-customer",
   "version": "1.10.0",
-  "private": true,
   "description": "OVHcloud - observability - logs to customer module",
   "license": "BSD-3-Clause",
   "author": "OVH SAS",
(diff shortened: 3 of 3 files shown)


packages/manager/apps/okms/package.json
```diff
@@ -38,6 +38,7 @@
     "@ovhcloud/ods-themes": "^18.6.2",
     "@tanstack/react-query": "^5.51.21",
     "@tanstack/react-table": "^8.20.1",
+    "@tanstack/react-virtual": "^3.10.9",
```
mathieumousnier, line 41: I have seen this in another project and it scares me a little bit :)
This is normally used in datagrids and should be a dep of MRC, but not a direct dep of our app.
Even react-table shouldn't be here...

packages/manager/modules/logs-to-customer/src/data/hooks/useLogService.tsx
```diff
@@ -17,12 +18,12 @@ export const useLogServices = () => {
  * Use log Service
  */
 export const getLogServiceQueryKey = (serviceName: string) => [
-  'getLogService',
+  getLogServicesQueryKey()[0],
```
mathieumousnier, line 21: You use this a few times and this syntax is a bit strange.

I would prefer a destructuration that point to the index 0.
And you might remove the endpoint string.

```ts
export const getLogServiceQueryKey = (serviceName: string) => [
  ...getLogServicesQueryKey(),
  serviceName,
];
```

If you want to go a bit further (but that might not be the best time), the best for querykeys is to make a small factory ([as detailed here](https://tkdodo.eu/blog/effective-react-query-keys#use-query-key-factories)), this would give you this:

```ts
export const logServiceQueryKeys = {
  all: () => ['log-services'],
  detail: (serviceName: string) => [...logServiceQueryKeys.all(), serviceName],
};

export const useLogServices = () => {
  return useQuery({
    queryKey: logServiceQueryKeys.all(),
    queryFn: () => getLogServices(),
  });
};

export const useLogService = (serviceName: string): UseQueryResult<Service> => {
  return useQuery<Service>({
    queryKey: logServiceQueryKeys.detail(serviceName),
    queryFn: () => getLogService(serviceName),
  });
};
```

packages/manager/modules/logs-to-customer/src/data/hooks/useLogSubscriptions.tsx
```diff
@@ -54,29 +51,39 @@ export const usePostLogSubscription = (
...
-  const { t } = useTranslation('logSubscription');
+  const { t } = useTranslation(NAMESPACES.LOG_SUBSCRIPTION);
   const queryClient = useQueryClient();
   const { addError, addSuccess } = useNotifications();
 
-  const queryFunction = apiVersion === 'v2' ? postLogSubscriptionV2 : postLogSubscriptionV6;
-
   return useMutation({
-    mutationKey: getPostLogSubscriptionMutationKey(logSubscriptionUrl, logKind, streamId),
+    mutationKey: getPostLogSubscriptionMutationKey(
```
mathieumousnier, line 59: nitpick: you can remove mutation keys they are useless

packages/manager/modules/logs-to-customer/src/data/hooks/useLogService.tsx
darsene: Could be done in next version #20809

mathieumousnier (approved):

packages/manager/modules/logs-to-customer/src/data/hooks/useLogSubscriptions.tsx
darsene: Will be done in #20809

packages/manager/apps/okms/package.json
darsene: The dependency was already registered, I've just moved it from dep to peerDep

PR #20255, opened 2025-11-06
fix(dedicated): replace distribution with OS

sbraz:
## Description

* Fixes the `des distributions disponible` typo (missing plural).
* Harmonizes the messages to always mention the generic "OS" term instead
  of the Linux-specific "distribution".

This mostly affects the French translations as the English ones already
used OS almost everywhere.

Ticket Reference: MANAGER-20225

## Additional Information

Currently, the name OS is used in some messages:
image
But not everywhere:
image
image


sbraz:
This is being applied in #20401.

PR #20264, opened 2025-11-06
feat(billing): disable cache in services

achrafHafedh:
ref: #MANAGER-20010

## Description

Ticket Reference: #MANAGER-20010

### DIFF
--- a/packages/manager/modules/billing/src/autoRenew/actions/terminate-webcoach/terminate-webcoach.html
+++ b/packages/manager/modules/billing/src/autoRenew/actions/terminate-webcoach/terminate-webcoach.html
@@ -4,7 +4,7 @@
     data-on-error="$ctrl.onError(error)"
     data-service-type="WEBCOACH"
     data-terminate-service="$ctrl.terminateWebCoach()"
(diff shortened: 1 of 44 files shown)


packages/manager/modules/billing/src/autoRenew/actions/terminate-webcoach/terminate-webcoach.html
```diff
@@ -4,7 +4,6 @@
     data-on-error="$ctrl.onError(error)"
     data-service-type="WEBCOACH"
     data-terminate-service="$ctrl.terminateWebCoach()"
-    data-go-back="$ctrl.goBack()"
```
JacquesLarique, line 7: You need to still pass the `go-back` attribute (so that we can close the modal when user click on the X or "cancel" button), you just need to adjust the value you pass it as follow:
`data-go-back="$ctrl.goToAutorenew()"`

achrafHafedh: Fixed !

MehdiZaghouani (approved):

PR #20278, opened 2025-11-06
feat(public-vcf-aas): add edit edge gateway

pauldkn:
## Description

Ticket Reference: #MANAGER-18399

### DIFF
--- a/packages/manager/apps/hpc-vmware-public-vcf-aas/src/pages/dashboard/datacentre/edge-gateway/edit/EditEdgeGateway.page.tsx
+++ b/packages/manager/apps/hpc-vmware-public-vcf-aas/src/pages/dashboard/datacentre/edge-gateway/edit/EditEdgeGateway.page.tsx
@@ -1,4 +1,4 @@
-import { Suspense, useEffect } from 'react';
+import { Suspense } from 'react';
 import { useTranslation } from 'react-i18next';
 import { useNavigate, useParams } from 'react-router-dom';
 import { Controller, useForm } from 'react-hook-form';
@@ -26,7 +26,10 @@ export default function EditEdgeGatewayPage() {
   const { addSuccess, addError } = useMessageContext();
 
   const edgeParams: GetEdgeGatewayParams = { id, vdcId, edgeGatewayId };
-  const { data: edge, isLoading } = useVcdEdgeGatewayMocks(edgeParams);
+  const { data: edge, isLoading, refetch } = useVcdEdgeGatewayMocks({
+    ...edgeParams,
+    staleTime: 5000,
+  });
 
   const {
     mutate: updateEdgeGateway,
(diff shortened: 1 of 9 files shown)


packages/manager/apps/hpc-vmware-public-vcf-aas/src/components/form/SelectField.component.tsx
mathieumousnier, line 1: I found a way to create generic "react-hook-form" fields that I am crazy about, so I can't resist to share : 
https://github.com/ovh/manager/blob/dev/okms-01/packages/manager/apps/okms/src/modules/secret-manager/components/form/SecretDeactivateVersionAfterFormField.component.tsx

- you just have to set name and control props to use the component
- no need to use the Controller component, it's all managed by the useController hook
- you can extends props by adding them to the FormFieldInput type

packages/manager/apps/hpc-vmware-public-vcf-aas/src/pages/dashboard/datacentre/edge-gateway/edit/EditEdgeGateway.page.tsx
```diff
@@ -0,0 +1,111 @@
...
+    updateEdgeGateway({ edgeGatewayNewName: data.edgeGatewayName });
+
+  useEffect(() => {
+    if (!isLoading && edge?.currentState) {
+      reset({
+        edgeGatewayName: edge.currentState.edgeGatewayName,
+        ipBlock: edge.currentState.ipBlock,
+      });
+    }
+  }, [isLoading, edge]);
```
mathieumousnier: This is not good and it is the fault of the guy who designed the drawer: me.
This useEffect could be avoided by putting the form in a child component and loading the data in the parent, but the Drawer does not allow this for now.
A new version of the Drawer will come with MUK, with composable parts, so you will be able to create a child component with a "Drawer.Content" and a "Drawer.Footer" imported from MUK.

[I did something like this here](https://github.com/ovh/manager/blob/dev/okms-01/packages/manager/apps/okms/src/modules/secret-manager/pages/drawers/create-version-drawer/CreateVersionDrawer.page.tsx). I add to recreate the inner parts of the drawer in the uApp, and they will be replaced by MUK components when we will do the migration.

packages/manager/apps/hpc-vmware-public-vcf-aas/src/components/form/SelectField.component.tsx
pauldkn: It's a nice solution indeed, but it seems you have to create a specific component for every field ?
(I see label / helper / tooltip are not variables)

If so, I actually find the use of `Controller` a bit easier

packages/manager/apps/hpc-vmware-public-vcf-aas/src/pages/dashboard/datacentre/edge-gateway/edit/EditEdgeGateway.page.tsx
pauldkn: It's actually one of the recommended way to use the `reset` API according to [the docs](https://react-hook-form.com/docs/useform#defaultValues), but it seems to have some issues with react-query and its caching, as well as using `defaultValues` + `values` to override defaults when data is fetched.

What we found to be the "easiest" solution with @sefiani-ovh, is to use async `defaultValues` with `refetch` and `staleTime` or `enabled: false` from react-query, and it seems to work perfectly. It removes the need for the `reset` API, as well as splitting the component only for that purpose (having `Drawer` composable parts in multiple components)

PR #20283, opened 2025-11-07
feat(cloud-shell): init app

lizardK:
## Description

Create CloudShell app

Ticket Reference: #MANAGER-20239


packages/manager/apps/container/src/container/nav-reshuffle/header/index.tsx
```diff
@@ -40,6 +42,9 @@ function Header({
   const isSmallDevice = useMediaQuery({
     query: `(max-width: ${SMALL_DEVICE_MAX_SIZE})`,
   });
+  const { data: cloudShellAvailability } = useFeatureAvailability(
+    ['cloud-shell'],
+  );
```
MaximeBajeux, line 47: nitpick: we should refacto that to just make one call to feature availability for the container

sachinrameshn (approved):

PR #20291, opened 2025-11-07
feat(pci-instances): add list of ssh keys

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-4623, #TAPC-4626

## Additional Information

image

image

### DIFF
--- a/packages/manager/apps/pci-instances/src/components/banner/Banner.component.tsx
+++ b/packages/manager/apps/pci-instances/src/components/banner/Banner.component.tsx
@@ -1,0 +1,24 @@
+import { FC, PropsWithChildren } from 'react';
+import {
+  Message,
+  MessageIcon,
+  ICON_NAME,
+  MessageBody,
+  MessageProp,
+} from '@ovhcloud/ods-react';
+
+type TBannerProps = PropsWithChildren<MessageProp>;
+
+const Banner: FC<TBannerProps> = ({ children, color, ...props }) => {
+  const icon =
+    color === 'warning' ? ICON_NAME.triangleExclamation : ICON_NAME.circleInfo;
+
+  return (
+    <Message color={color} dismissible={false} {...props}>
+      <MessageIcon name={icon} />
+      <MessageBody>{children}</MessageBody>
+    </Message>
+  );
+};
+
+export default Banner;
--- a/packages/manager/apps/pci-instances/src/pages/instances/create/CreateInstance.page.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/create/CreateInstance.page.tsx
@@ -18,8 +18,8 @@ import {
   quantitySchema,
   distributionImageTypeSchema,
   distributionImageNameSchema,
-  sshNameSchema,
-  sshKeySchema,
+  sshKeyIdSchema,
+  sshPublicKeySchema,
 } from '@/pages/instances/create/CreateInstance.schema';
 
 import { useInstancesCatalog } from '@/data/hooks/catalog/useInstancesCatalog';
@@ -39,8 +39,8 @@ export const instanceCreationSchema = z.object({
   availabilityZone: availabilityZoneSelectionSchema,
   distributionImageType: distributionImageTypeSchema,
   distributionImageName: distributionImageNameSchema,
-  sshName: sshNameSchema.nullable(),
-  sshKey: sshKeySchema.optional(),
+  sshKeyId: sshKeyIdSchema,
+  newSshPublicKey: sshPublicKeySchema.nullable(),
 });
 
 const CreateInstance: FC = () => {
--- a/packages/manager/apps/pci-instances/src/pages/instances/create/components/SshKey.component.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/create/components/SshKey.component.tsx
@@ -1,14 +1,5 @@
-import { useEffect, useState } from 'react';
-import {
-  Button,
-  Divider,
-  Icon,
-  ICON_NAME,
-  Message,
-  MessageBody,
-  MessageIcon,
-  Text,
-} from '@ovhcloud/ods-react';
+import { useEffect, useMemo, useState, useCallback } from 'react';
+import { Button, Divider, Icon, Text } from '@ovhcloud/ods-react';
 import { useTranslation } from 'react-i18next';
 import { SshKeyHelper } from './sshKey/SshKeyHelper.component';
 import { useSshKeys } from '@/data/hooks/ssh/useSshKeys';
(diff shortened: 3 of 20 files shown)


lolottetheclash (changes requested):
I think you should improve the SSH key regex: it should not valid the fact that it can contain some blank spaces into the ssh key, and also you should trim it before testing it

packages/manager/apps/pci-instances/src/components/banner/Banner.component.tsx
```diff
@@ -0,0 +1,26 @@
...
+type TBannerProps = MessageProp & {
+  message: string;
+};
+
+const Banner: FC<TBannerProps> = ({ message, color, ...props }) => {
+  const icon =
+    color === 'warning' ? ICON_NAME.triangleExclamation : ICON_NAME.circleInfo;
+
+  return (
+    <Message color={color} {...props}>
```
aTurmo: Suggestion: By default the ODS19 'Message' is dismissible (as we can see the cross on the left on your screen captures), but ODS19 doesn't handle clicking on the cross. Si I suggest you either handle clicking on the cross via 'onRemove' or remove the corss via 'dismissible={false}'

packages/manager/apps/pci-instances/src/components/banner/Banner.component.tsx
```diff
@@ -0,0 +1,26 @@
...
+      <MessageIcon name={icon} />
+      <MessageBody>{message}</MessageBody>
```
aTurmo: suggestion : children instead of just a string message ?

packages/manager/apps/pci-instances/src/pages/instances/create/components/SshKey.component.tsx
```diff
@@ -28,27 +32,75 @@ const SshKey = ({ microRegion }: TSshKeyProps) => {
...
+
   const handleAddSshKey: SubmitHandler<TAddSshKeyForm> = ({
     sshName,
     sshKey,
   }) => {
     setValue('sshName', sshName);
     setValue('sshKey', sshKey);
-    setOpenSshKeyForm(false);
+
+    setSshKeyItems((prevSshKeys) => [
```
aTurmo: Question: what will happen if I put a ssh key with an existing label ?

packages/manager/apps/pci-instances/src/pages/instances/create/components/SshKey.component.tsx
```diff
@@ -28,27 +32,75 @@ const SshKey = ({ microRegion }: TSshKeyProps) => {
...
+      ...prevSshKeys,
+    ]);
+
+    handleCloseSshKeyForm();
+    setIsSshKeyAdded(true);
+  };
+
+  const updateSshKeyFields = useCallback(
+    (sshKeys: TSshKeyData[]) => {
+      const selectedSshKey = sshKeys[0];
```
aTurmo: question: are you sure you can always take the sshKeys[0] ?

packages/manager/apps/pci-instances/src/pages/instances/create/components/SshKey.component.tsx
```diff
@@ -28,27 +32,75 @@ const SshKey = ({ microRegion }: TSshKeyProps) => {
...
+    handleCloseSshKeyAddedMessage();
   };
 
+  useEffect(() => {
+    if (!isLoading) setSshKeyItems(sshKeys);
+  }, [isLoading, sshKeys]);
+
   useEffect(() => {
     if (!isLoading && sshKeys.length === 0) setOpenSshKeyForm(true);
   }, [isLoading, sshKeys]);
```
aTurmo, line 102: You can merge those two use effect (same dependencies)

packages/manager/apps/pci-instances/src/components/banner/Banner.component.tsx
Tsiorifamonjena: good point, I let the dismissible to be false by default

lolottetheclash:
@Tsiorifamonjena I was thinking that ssh key should not contain any blank space but I think it's wrong, it can contain between each part. To trim it is a good practise, but maybe check the right ssh regex as it's already present in the manager :)

packages/manager/apps/pci-instances/src/pages/instances/create/components/SshKey.component.tsx
Tsiorifamonjena: That's why I control on #L79. This function is used to update fields when user select existing ssh key and when setting the default value if there are already some ssh keys.

aTurmo: Yes ok, but I mean what about the other sshKeys ? (for index > 0)

packages/manager/apps/pci-instances/src/pages/instances/create/components/SshKey.component.tsx
Tsiorifamonjena: Now it can't be possible anymore, the form will throw error.

packages/manager/apps/pci-instances/src/pages/instances/create/components/SshKey.component.tsx
Tsiorifamonjena: Ah, it will always be the index 0 whether it's to set the default value or on select because select return an array with one entry. That's also why I put the new added ssh key at the first position

Tsiorifamonjena:
> @Tsiorifamonjena I was thinking that ssh key should not contain any blank space but I think it's wrong, it can contain between each part. To trim it is a good practise, but maybe check the right ssh regex as it's already present in the manager :)

You were right @lolottetheclash :)

here is the ssh key format `<type> <base64-encoded-key>` and `base64-encoded-key` should not contain any blank space. I searched the entire manager but there is no regex. Validation is done on the backend side for existing form.

I improve the regex and I let you see

lolottetheclash (commented):
When I click on "create a new ssh key" for the first time, I have the impression that page reloads, do you have the same behavior?

Tsiorifamonjena:
> When I click on "create a new ssh key" for the first time, I have the impression that page reloads, do you have the same behavior?

@lolottetheclash the translation was probably loaded in my local, that's why I didn't have the same behavior. 
The problem was from common translation not yet load on mounting the AddSshKey component for the first time (after click create a new ssh key) then react-i18next fetch and there is a fallback.

I fix it by adding the namespace in the `init` I let you see

packages/manager/apps/pci-instances/src/pages/instances/create/components/SshKey.component.tsx
aTurmo: Ok i'm starting to understand but it seems complicated.

packages/manager/apps/pci-instances/src/pages/instances/create/CreateInstance.page.tsx
```diff
@@ -39,8 +39,8 @@ export const instanceCreationSchema = z.object({
   availabilityZone: availabilityZoneSelectionSchema,
   distributionImageType: distributionImageTypeSchema,
   distributionImageName: distributionImageNameSchema,
-  sshName: sshNameSchema.nullable(),
-  sshKey: sshKeySchema.optional(),
+  sshKeyId: sshKeyIdSchema,
+  newSshPublicKey: sshPublicKeySchema.nullable(),
```
lolottetheclash: could you integrate nullable into `sshPublicKeySchema`? as you did for `sshKeyIdSchema`

lolottetheclash (approved):

PR #20300, opened 2025-11-07
feat(pci-volume-snapshot): make volume name mandatory when creating a new one from snapshot

aTurmo:
ref: #TAPC-3698

### DIFF
--- a/packages/manager/apps/pci-volume-snapshot/src/pages/create-volume/VolumeEdit.component.spec.tsx
+++ b/packages/manager/apps/pci-volume-snapshot/src/pages/create-volume/VolumeEdit.component.spec.tsx
@@ -1,5 +1,6 @@
-import { describe, it, expect, vi, beforeEach } from 'vitest';
-import { render, fireEvent, act } from '@testing-library/react';
+import { beforeEach, describe, expect, it, vi } from 'vitest';
+import userEvent from '@testing-library/user-event';
+import { act, render } from '@testing-library/react';
 import {
   TRegion,
   TRegionQuota,
@@ -9,6 +10,7 @@ import {
 } from '@/api/api.types';
 import VolumeEdit from '@/pages/create-volume/VolumeEdit.component';
 import { createWrapper } from '@/wrapperRenders';
+import { VOLUME_NAME_MAX_LENGTH } from '@/constants';
 
 const MOCK_VOLUME: TVolume = {
   id: 'fake-volume-id',
@@ -105,7 +107,7 @@ describe('VolumeEdit', () => {
 
   describe('Name input', () => {
     it('uses suggested name as default value if specified', async () => {
-      const { container } = render(
+      const { getByRole } = render(
         <VolumeEdit
           projectId="fake-project-id"
           volume={MOCK_VOLUME}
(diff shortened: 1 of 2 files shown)


Masta2000 (dismissed):

packages/manager/apps/pci-volume-snapshot/src/pages/create-volume/VolumeEdit.component.tsx
```diff
@@ -87,11 +87,15 @@ export default function VolumeEdit({
   );
 
   const errorState = {
+    nameIsAbsent:
```
fredericvilcot: I would have normalized errors, like we do on backend, something like `NO_NAME_EXCEPTION`, `NAME_TOO_LONG_EXCEPTION`, etc..

fredericvilcot (approved):

packages/manager/apps/pci-volume-snapshot/src/pages/create-volume/VolumeEdit.component.spec.tsx
```diff
@@ -176,24 +209,28 @@ describe('VolumeEdit', () => {
...
-      const submitButton = queryByText('submit-label');
-      expect(submitButton).toBeDisabled();
-      const formElt = container.querySelector('form');
-      act(() => fireEvent.submit(formElt as Element));
-      expect(onSubmit).not.toHaveBeenCalled();
+      // I don't know why, but the button is correctly of type submit when running the app, but is of type 'button' when testing
+      // This is a band-aid that should be removed when migrating to ODS19
+      submitButton.setAttribute('type', 'submit');
+
+      await act(async () => {
```
ppprevost, line 226: you do not need act

aTurmo: If i dont put it i get warning everywhere in test. It's the only way to not have tests logs polluted.

packages/manager/apps/pci-volume-snapshot/src/pages/create-volume/VolumeEdit.component.tsx
aTurmo: I don't really understand how it applies here. You want an enum for each field ?

fredericvilcot: `nameIsAbsent` is not normalized as en error, regarding naming. I'm talking about exception concepts, I'll wish to have explicit error handlings, out of the component ;)

packages/manager/apps/pci-volume-snapshot/src/pages/create-volume/VolumeEdit.component.tsx
```diff
@@ -87,11 +87,15 @@ export default function VolumeEdit({
...
+      volumeName === null ||
+      volumeName === undefined ||
```
lolottetheclash: may be you can add null & undefined tests as well

aTurmo: It's a good popint but actually i think volumeName can't be null or undefined. I added those check from habits but I think it's unnecessary. In any case it's not possible to the this directly because you can't put null or undefined into a controlled input.
I'll removed thoses checks

PR #20303, opened 2025-11-07
feat(*): make the shell initializable without the configuration

JacquesLarique:
## Description

- Make the shell initializable without the configuration
- Adding a method to finish the shell initialization (by giving it the environment)
- Adding a method to update the shell environment plugin and tracking configuration (to avoid reloading the page when, after a maintenance error, the site is back up and we receive a proper answer from /configuration)

Ticket Reference: #MANAGER-19571

### DIFF
--- a/packages/manager/apps/container/src/components/sso-auth-modal/SSOAuthModal.component.tsx
+++ b/packages/manager/apps/container/src/components/sso-auth-modal/SSOAuthModal.component.tsx
@@ -54,8 +54,10 @@ const SSOAuthModal = (): JSX.Element => {
   const userId = uxPlugin.getUserIdCookie();
 
   useEffect(() => {
+    if (!uxPlugin) return;
+
     const updateMode = () => {
-      setMode(uxPlugin?.getSSOAuthModalMode(userId));
+      setMode(uxPlugin.getSSOAuthModalMode(userId));
     };
     document.addEventListener('visibilitychange', updateMode);
 
--- a/packages/manager/apps/container/src/components/user-name/UserName.component.tsx
+++ b/packages/manager/apps/container/src/components/user-name/UserName.component.tsx
@@ -1,10 +1,31 @@
+import { useMemo } from 'react';
+
+import { ODS_THEME_COLOR_INTENT } from '@ovhcloud/ods-common-theming';
+import { ODS_TEXT_LEVEL, ODS_TEXT_SIZE } from '@ovhcloud/ods-components';
+import { OsdsText } from '@ovhcloud/ods-components/react';
+
 import { LEGAL_FORMS } from '@/container/common/constants';
 import useUser from "@/hooks/user/useUser";
 
-export const UserName = () => {
+type Props = {
+  size: ODS_TEXT_SIZE,
+};
+
+export const UserName = ({ size }: Props) => {
   const user = useUser();
+  const name = useMemo(() =>
+    user.legalform === LEGAL_FORMS.CORPORATION
+      ? user.organisation
+      : `${user.firstname} ${user.name}`,
+    [user]);
   
-  return user.legalform === LEGAL_FORMS.CORPORATION
-    ? user.organisation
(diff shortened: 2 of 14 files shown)


packages/manager/apps/container/src/components/user-name/UserName.component.tsx
```diff
@@ -0,0 +1,10 @@
+import { LEGAL_FORMS } from '@/container/common/constants';
+import useUser from "@/hooks/user/useUser";
+
+export const UserName = () => {
```
MehdiZaghouani: This isn’t really a React component since it returns a string instead of JSX. It behaves more like a helper function.

packages/manager/apps/container/src/container/common/language/index.tsx
```diff
@@ -9,17 +9,30 @@ import LanguageList from './List';
...
+import useContainer from '@/core/container';
+import { ODS_SKELETON_SIZE } from '@ovhcloud/ods-components';
 
 export type Props = {
   onChange(show: boolean): void;
-  setUserLocale(locale: string): void;
-  userLocale?: string;
+};
+
+export const LanguageMenuContainer = ({ onChange }: Props) => {
```
MehdiZaghouani: It’s generally better to keep one exported component per file. Exporting multiple components from the same file can make the code harder to navigate and maintain

packages/manager/apps/container/src/container/nav-reshuffle/header/user-account-menu/index.tsx
```diff
@@ -76,6 +55,9 @@ export const UserAccountMenu = ({ onToggle }: Props): JSX.Element => {
       <UserAccountMenuButton
         show={isAccountSidebarOpened}
         onClick={(nextShow) => {
+          if (!isReady) {
```
MehdiZaghouani: i suggest to moving this logic into a function

packages/manager/apps/container/src/components/sso-auth-modal/SSOAuthModal.component.tsx
```diff
@@ -49,15 +49,18 @@ const SSOAuthModal = (): JSX.Element => {
   const authPlugin = shell.getPlugin('auth');
   const uxPlugin = shell.getPlugin('ux');
   const { environment } = useApplication();
-  const { user } = environment;
+  const { user } = environment ?? {};
   const [connectedUser, setConnectedUser] = useState<User>({} as User);
   const userId = uxPlugin.getUserIdCookie();
 
   useEffect(() => {
```
aottr, line 56: you should exit the useEffect early if `uxPlugin` is undefined to avoid registering the listener with an empty handler / assigning `undefined` to the `mode` state

packages/manager/apps/container/src/container/nav-reshuffle/header/index.tsx
```diff
@@ -86,20 +85,18 @@ function Header({
                 </div>
               )}
               <div className={`oui-navbar-list__item ${style.navbarListItem}`}>
-                <LanguageMenu
-                  setUserLocale={setUserLocale}
-                  userLocale={userLocale}
+                <LanguageMenuContainer
                   onChange={(show: boolean) => {
```
MehdiZaghouani, line 89: we don't need to add show type

packages/manager/apps/container/src/container/common/language/index.tsx
aottr: Yes, especially since you export it, it should be in two different files

packages/manager/apps/container/src/container/nav-reshuffle/header/index.tsx
JacquesLarique: With this the `yarn build` there's no issue to do it, however the `yarn build:strict` is failing with the error `Parameter 'show' implicitly has an 'any' type.`
So I think we should keep it this way

PR #20305, opened 2025-11-10
build(observability): pnpm migration of module observability-to-customer

yjaaouane:
## Description

Ticket Reference: #MAOBS-129


darsene (commented):
@yjaaouane Vitest config file should be also updated normally

MarcMonchablon (approved):

PR #20314, opened 2025-11-10
feat(account-creation): manage redirection from settings page

JacquesLarique:
## Description

Set up redirection url retrieval from the url
Urls sanitization
Redirection post form validation

Ticket Reference: #MANAGER-17394

### DIFF
--- a/packages/manager/apps/account-creation/src/data/hooks/settings/useSettings.ts
+++ b/packages/manager/apps/account-creation/src/data/hooks/settings/useSettings.ts
@@ -71,3 +71,30 @@ export const useLanguageSettings = (country?: string, currency?: string) =>
       );
     },
   });
+
+export const useSubsidiarySettings = (
+  country?: string,
+  currency?: string,
+  language?: string,
+) => {
+  return useQuery({
+    queryKey: ['settings'],
+    queryFn: () => getSettings(getRegion()),
+    select: (settings: Map<string, BillingCountry[]>) => {
+      if (!(country && currency && language)) {
+        return null;
+      }
+      const foundCountry = settings
+        .get(country)
+        ?.find(
+          (billingCountry) =>
+            billingCountry.currency.code === currency &&
+            billingCountry.ietfLanguageTag === language,
+        );
+      if (!foundCountry) {
+        return null;
+      }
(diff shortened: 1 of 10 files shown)


packages/manager/apps/account-creation/src/pages/settings/settings.page.tsx
```diff
@@ -97,8 +110,9 @@ export default function Settings() {
   const submitSettings: SubmitHandler<SettingsFormData> = useCallback(
     ({ country, currency, language }: SettingsFormData) => {
       console.log({ country, currency, language });
+      window.location.href = `${redirectToSignUpUrl}?ovhSubsidiary=${ovhSubsidiary}`;
```
aottr: I think we can clean up the console.log.

Also: If signupURL is invalid, we will still redirect to null?ovhSubsidary=FR

JacquesLarique: I'll keep the console.log, to avoid a lint issue with the parameters being unused (and avoid conflict on the tracking implementation branch)

packages/manager/apps/account-creation/src/data/hooks/settings/useSettings.ts
```diff
@@ -71,3 +71,30 @@ export const useLanguageSettings = (country?: string, currency?: string) =>
...
     },
   });
+
+export const useSubsidiarySettings = (
+  country?: string,
+  currency?: string,
+  language?: string,
+) => {
+  return useQuery({
+    queryKey: ['settings'],
```
sachinrameshn, line 81: Can we move the query key inside data/api/setting folder as getSettingsQueryKey 
export const getSettingsQueryKey = (region) => ['/account/settings']; as of now.
And use this once API is ready 
export const getSettingsQueryKey = (region) => ['/account/settings?ovhIS=${region}'];
```suggestion
    queryKey: getSettingsQueryKey,
```

JacquesLarique: It is planned to be done in MANAGER-19878

aottr (approved):

PR #20315, opened 2025-11-10
fix(okms): fix dashboard tile text overflow

rjamet-ovh:
ref: #MANAGER-16519

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/okms/src/modules/key-management-service/pages/credential/general-informations/generalInformations.page.tsx
+++ b/packages/manager/apps/okms/src/modules/key-management-service/pages/credential/general-informations/generalInformations.page.tsx
@@ -56,7 +56,11 @@ const CredentialGeneralInformations = () => {
       id: 'name',
       label: t('key_management_service_credential_dashboard_name'),
       value: (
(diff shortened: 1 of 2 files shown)


mathieumousnier (approved):

packages/manager/apps/okms/src/modules/key-management-service/pages/credential/general-informations/generalInformations.page.tsx
```diff
@@ -70,7 +74,11 @@ const CredentialGeneralInformations = () => {
       id: 'description',
       label: t('key_management_service_credential_dashboard_description'),
       value: (
-        <OdsText preset={ODS_TEXT_PRESET.span}>
+        <OdsText
+          preset={ODS_TEXT_PRESET.span}
+          // Temporary fix: wrap text wihtout whitespace
```
tibs245:
```suggestion
          // Temporary fix: wrap text without whitespace
```

😛

rjamet-ovh: done

PR #20317, opened 2025-11-10
fix(object-storage): avoid submit form bubbling

lionel95200x:
ref: #DATATR-2651

## Description

- Avoid trigger event of object creation

Ticket Reference: #DATATR-2651

### DIFF
--- a/packages/manager/apps/pci-object-storage/src/pages/object-storage/users/create/AddUserForm.component.tsx
+++ b/packages/manager/apps/pci-object-storage/src/pages/object-storage/users/create/AddUserForm.component.tsx
@@ -22,6 +22,7 @@ import { useGetUser } from '@/data/hooks/user/useGetUser.hook';
 import UserInformation from '@/components/user-information/userInformation.component';
 import { POLLING } from '@/configuration/polling.constants';
 import { FormField } from '@/components/form-field/FormField.component';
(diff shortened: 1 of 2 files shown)


packages/manager/apps/pci-object-storage/src/pages/object-storage/users/create/AddUserForm.component.tsx
```diff
@@ -75,15 +75,19 @@ const AddUserForm = ({ onClose }: AddUserFormProps) => {
...
+    form.handleSubmit((formValues) => {
+      addUser({
+        projectId,
+        newUser: {
+          description: formValues.userName,
+          role: user.RoleEnum.objectstore_operator,
+        },
+      });
+    })();
+  };
```
jperchoc: handleSubmit actually exposes the event: https://react-hook-form.com/docs/useform/handlesubmit

So I think it would be cleaner to stop the propagation like this:

```suggestion
  const onSubmit = form.handleSubmit((formValues, e) => {
  e?.preventDefault();
  e?.stopPropagation();

  addUser({
    projectId,
    newUser: {
      description: formValues.userName,
      role: user.RoleEnum.objectstore_operator,
    },
  });
});
```

jperchoc (approved):

PR #20385, opened 2025-11-13
feat(account-migration): clean up account migration modules

MehdiZaghouani:
ref: #MANAGER-14290

The purpose of this PR is to clean up the `account-migration` module, which has not been used since 2020.
I split the work into three parts:

1.  Remove all `account-migration` imports across all apps and modules
2.  Remove the `account-migration` services
3.  Delete the module itself


packages/manager/apps/dedicated/client/app/app.module.js
```diff
@@ -208,8 +207,6 @@ export default async (containerEl, shellClient) => {
         ngOvhExportCsv,
         ngPaginationFront,
         ngQAllSettled,
-        'ovh-api-services',
```
JacquesLarique: Not to be removed

JacquesLarique (approved):

PR #20386, opened 2025-11-13
feat(okms): secret-manager guides

rjamet-ovh:
## Description

Ticket Reference: #...


packages/manager/apps/okms/src/modules/secret-manager/components/guides/guide-kv2-api/useGuideItemKv2Api.spec.ts
mathieumousnier, line 1: I have a lot to say about the file names :)

**secret-manager-guides**
Can we call the `secret-manager-guides` folder to `guides` ?
Same as on the `key-management-service` side.

**guide-KV2-api**
Acronyms is a topic, and numbers is another one, but anyways, I would avoid any uppercase in theses kebab-case folder: `guide-kv2-api`

**useGuideItemKV2Api, guideKV2**
The toughest one.
Still, even with acronyms and numbers, snakeCase will make : `useGuideItemKv2Api` and `guideKv2`.
The same way we should have `getOkms`, `createOkms`.

Variables and functions should then be renamed the same way !

rjamet-ovh: done

mathieumousnier (approved):

PR #20398, opened 2025-11-14
refactor(observability): migrate observability-to-customer module to muk

yjaaouane:
fixes: #MAOBS-139

## Description

Ticket Reference: #MAOBS-139


packages/manager/modules/observability-to-customer/src/components/timeControls/TimeControls.component.tsx
```diff
@@ -72,8 +72,11 @@ export const TimeControls: React.FC<Readonly<TimeControlsProps>> = ({
                       ${value === 'custom' ? 'bg-[var(--ods-color-primary-200)] text-initial' : ''}
                     `}
                 >
-                  <span>{t('option_custom')}</span>
-                  <Icon name={ICON_NAME.calendar} />
+                  <>
+                    <span>{t('option_custom')}</span>
+                    <Icon name={ICON_NAME.calendar} />
+                  </>
```
MarcMonchablon: Probably a small mistake here, cause I don't think the ReactFragment is needed here :)

packages/manager/modules/observability-to-customer/src/pages/metric/DashboardWidgetModal.page.tsx
```diff
@@ -2,21 +2,24 @@ import { useNavigate, useParams } from 'react-router-dom';
 
 import { useTranslation } from 'react-i18next';
 
+import {
+  ModalBody,
+  ModalContent,
+  ModalOpenChangeDetail,
+} from '@ovhcloud/ods-react';
+
```
darsene: Should be deleted as you use muk

packages/manager/modules/observability-to-customer/src/components/timeControls/TimeControls.component.tsx
yjaaouane: The button component of muk not accept multiple children, i am agree not nice :'(
Type 'Element[]' is not assignable to type 'string | Element'

also no icon property to define on button

PR #20426, opened 2025-11-14
feat(ips): adding a survey link ot the iplisting page

aderghamov:
ref: #MANAGER-20275

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/ips/public/translations/listing/Messages_fr_FR.json
+++ b/packages/manager/apps/ips/public/translations/listing/Messages_fr_FR.json
@@ -113,5 +113,7 @@
   "all_items_failover": "Toutes les Additional IP",
   "slice_in_progress": "Segmentation en cours",
   "aggregate_in_progress": "Agrégation en cours",
-  "creation_in_progress": "Création en cours"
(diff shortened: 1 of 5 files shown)


packages/manager/apps/ips/src/utils/survey.constants.ts
```diff
@@ -0,0 +1 @@
+export const SURVEY_BASE_URL = 'https://survey.ovh.com/index.php/933956';
```
chipp972: Just put it directly in the SurveyLink.tsx file
There won't be any other constants related to this and other components won't need to access it

packages/manager/apps/ips/public/translations/survey/Messages_fr_FR.json
```diff
@@ -0,0 +1,4 @@
+{
+  "survey_link": "Vous aimez?",
+  "survey_link_tooltip": "Vous aimez cette page ? Faites nous part de vos remarques !"
+}
```
chipp972: Since it appears on the listing page, we can add this to the listing translation file I think

Steffy29 (approved):

PR #20482, opened 2025-11-18
feat(pci-instances): add distribution version selector

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-4630

## Additional Information

image


packages/manager/apps/pci-instances/src/__mocks__/instance/constants.ts
```diff
@@ -1117,70 +1117,247 @@ export const mockedDistributionImageList = [
     type: 'unix',
     label: 'Linux',
     name: 'linux',
+    pricing: {
+      hour: '0.05 €',
+    },
+    versions: [
+      {
+        label: 'Rocky Linux 10',
+        value: 'Rocky Linux 10',
```
aTurmo: Suggestion: Maybe distinguish between each 'label' and 'value' to be sure that the correct one is used each time. Even if in reality they are the same (and if they are always the same why do we need both ?)

packages/manager/apps/pci-instances/src/pages/instances/create/components/CreationCart.component.tsx
```diff
@@ -89,9 +105,21 @@ export const CreationCart = () => {
               'creation:pci_instance_creation_cart_distribution_image_title',
             ),
             description: (
-              <Text preset="heading-6" className="text-[--ods-color-heading]">
-                {distributionImageName} - version x
-              </Text>
+              <>
+                <Text preset="heading-6" className="text-[--ods-color-heading]">
+                  {[distributionImageName, distributionImageVersion]
+                    .filter((distribution) => distribution)
```
aTurmo:
```suggestion
                    .filter(Boolean)
```

packages/manager/apps/pci-instances/src/pages/instances/create/hooks/useForm.ts
```diff
@@ -64,10 +64,19 @@ export const useForm = (projectId: string) => {
...
+
+  const distributionImageNameDefaultValue = distributionImage?.id ?? null;
+
+  const flavorImagePriceDefaultValue = distributionImage?.pricing?.hour ?? null;
+
+  const distributionImageVersionDefaultValue =
+    distributionImage?.versions?.[0]?.value ?? null;
+
+  const distributionLicencePrice =
+    distributionImage?.pricing?.licence?.total ?? null;
```
aTurmo: Suggestion:
I see this in at least three different files (DistributionImageList.component.tsx, DistributionImageType.component.tsx and this one here).
So I think you could introduce some king of method like 'getDistributionLicensePrice = (distributionImage) => ...'

packages/manager/apps/pci-instances/src/__mocks__/instance/constants.ts
Tsiorifamonjena: yeah it was just a mock for the Select component where both label and value are mandatory. I update and surely readjust when we connect with data

packages/manager/apps/pci-instances/src/pages/instances/create/hooks/useForm.ts
Tsiorifamonjena: you were right, I refactor

PR #20513, opened 2025-11-19
build(pci-block-storage): migrate to pnpm

Taelar:
ref: #TAPC-5334

## Description

Ticket Reference: #...


packages/manager/apps/pci-block-storage/vitest.config.js
```diff
@@ -33,6 +34,7 @@ export default defineConfig({
     },
   },
   resolve: {
+    dedupe: [...defaultDedupedDependencies],
```
aTurmo, line 37: Question: was this added automatically ?

Taelar: I think it was yes. I also double-checked other pnpm apps and they all have this conf so I assumed it was required

PR #20516, opened 2025-11-19
fix(ips): fixing the anti-hack ip unblock API call

aderghamov:
ref: #MANAGER-18914

## Description

Ticket Reference: #...


packages/manager/apps/ips/src/pages/actions/antiHack/unblockAntiHack.page.tsx
chipp972, line 115:
```suggestion
        queryKey: getIpAntihackQueryKey({ ip: parentIp }),
```

Like the other ? Or is it working fine ?

aderghamov: I think that this should be the intended behavior.
The query for getting the antihack requires the ipblock (ipParent)

PR #20518, opened 2025-11-19
fix(muk): correct hook useTranslatedMicroRegion

aboungnaseng-ovhcloud:
## Description

- Delete duplication keys in commons translations for FR region
- Correct translations in useTranslatedMicroRegion hook
- Update story in manager wiki

Ticket Reference: #MANAGER-20207

### DIFF
--- a/packages/manager-wiki/.storybook/i18n.ts
+++ b/packages/manager-wiki/.storybook/i18n.ts
@@ -1,14 +1,31 @@
+/// <reference types="vite/client" />
 import i18n from 'i18next';
 import { initReactI18next } from 'react-i18next';
+import I18NextHttpBackend from 'i18next-http-backend';
 
-i18n.use(initReactI18next).init({
+i18n
+.use(initReactI18next)
+.use(I18NextHttpBackend)
+.use({
+  type: 'postProcessor',
+  name: 'normalize',
+  process: (value: string) => (value ? value.replace(/&amp;/g, '&') : value),
+})
+.init({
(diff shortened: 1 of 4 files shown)


packages/manager-wiki/.storybook/i18n.ts
```diff
@@ -1,14 +1,38 @@
...
+  i18n
+    .addResources('fr_FR', NAMESPACES.REGION, regionFrFR)
+    .addResources('en_GB', NAMESPACES.REGION, regionEnGB)
+    .addResources('de_DE', NAMESPACES.REGION, regionDeDE)
+    .addResources('es_ES', NAMESPACES.REGION, regionEsES)
+    .addResources('it_IT', NAMESPACES.REGION, regionItIT)
+    .addResources('pt_PT', NAMESPACES.REGION, regionPtPT)
+    .addResources('pl_PL', NAMESPACES.REGION, regionPlPL)
+    .addResources('fr_CA', NAMESPACES.REGION, regionFrCA);
+}
```
anooparveti: **question:** Do we need to import and add all the translations that we use in MUK?

aboungnaseng-ovhcloud: That how I resolve it in the storybook, But I have test it in generated app, we dont need to import it in this way

aboungnaseng-ovhcloud: deleted 👍  I added another confiugration in the public folder

PR #20526, opened 2025-11-19
feat(pci-block-storage): fix region next button being visible when it should not

aTurmo:
ref: #TAPC-5343


packages/manager/apps/pci-block-storage/src/pages/new/components/LocationStep.component.tsx
```diff
@@ -43,7 +43,12 @@ export function LocationStep({
   ] = useState<TDeployment | null>(null);
   const [selectedLocalisation, setSelectedLocalisation] = useState<
     TLocalisation
-  >(undefined);
+  >(null);
+
+  useEffect(() => {
+    if (!step.isLocked) setSelectedLocalisation(null);
```
lolottetheclash, line 49: you can add also  a check on selectedLocalisation, if it's already null

lolottetheclash (approved):

PR #20529, opened 2025-11-20
feat(vcfaas): disable delete network acl when default one is remaining

sefiani-ovh:
ref: #MANAGER-20293


packages/manager/apps/hpc-vmware-public-vcf-aas/src/pages/listing/networkAcl/NetworkAclListContent.component.tsx
```diff
@@ -46,6 +46,9 @@ export default function NetworkAclListContent() {
   const { addInfo, clearMessage } = useMessageContext();
   const msgUidActiveTaskRef = useRef<number | null>(null);
 
+  const isDefaultAcl = useMemo(() => {
+    return networks.length === 1 && networks[0].network === DEFAULT_ACL_NETWORK;
+  }, [networks]);
```
pauldkn: I don't think you need `useMemo` for this, it's not a heavy operation

PR #20532, opened 2025-11-20
feat(object-storage): improve dashboard object

lionel95200x:
ref: #DATATR-2681

## Description

Ticket Reference: #...


packages/manager/apps/pci-object-storage/public/translations/regions/Messages_fr_FR.json
```diff
@@ -16,6 +16,8 @@
   "region_TOR_micro": "Toronto ({{ micro }})",
   "region_EU-WEST-PAR": "Paris",
   "region_EU-WEST-PAR_micro": "Paris ({{ micro }})",
+  "region_EU-SOUTH-MIL": "Milan",
+  "region_EU-SOUTH-MIL_micro": "Milan ({{ micro }})",
```
jperchoc, line 24: I think you need to also add those:
```
region_MIL
region_MIL_micro
region_MIL_micro
region_continent_MIL
region_continent_EU-SOUTH-MIL
```

Actually, I'm not 100% sure that EU-SOUTH-MIL is really used, as we are using the `getMacroRegion` in `packages/manager/apps/pci-object-storage/src/hooks/useTranslatedMicroRegions.tsx` which transform the EU-SOUTH-MIL region to MIL

packages/manager/apps/pci-object-storage/src/components/route-modal/RouteModal.tsx
```diff
@@ -20,7 +20,13 @@ const RouteModal = ({
 }: RouteModalProps) => {
   const navigate = useNavigate();
   const onOpenChange = (open: boolean) => {
-    if (!open) navigate(backUrl);
+    if (!open) {
+      if (typeof backUrl === 'number') {
+        navigate(backUrl);
+      } else {
+        navigate(backUrl);
+      }
```
AkdM: Both conditions call the same function, so the check isn't needed:

```suggestion
      navigate(backUrl);
```

packages/manager/apps/pci-object-storage/public/translations/regions/Messages_fr_FR.json
jperchoc: Image

packages/manager/apps/pci-object-storage/src/components/route-modal/RouteModal.tsx
```diff
@@ -9,7 +9,7 @@ import {
 } from '@datatr-ux/uxlib';
 
 interface RouteModalProps {
-  backUrl?: string;
+  backUrl?: string | number;
```
jperchoc: Do we have any benefit allowing a number here ? I see you are using `-1` as a backUrl, but `..` would do the same I guess ?

packages/manager/apps/pci-object-storage/src/pages/object-storage/storage/s3Id/S3.layout.tsx
```diff
@@ -40,8 +40,17 @@ export function breadcrumb() {
...
 const HIDE_TABS_ROUTE_PATTERNS = ['s3.object.'];
 
+const shouldHideS3Header = (
+  matches: ReturnType<typeof useMatches>,
+): boolean => {
+  return matches.some((match) =>
+    HIDE_HEADER_ROUTE_PATTERNS.some((pattern) => match.id?.startsWith(pattern)),
+  );
+};
+
```
jperchoc, line 53: This will be the exact same result as shouldHides3Tabs, as the pattern is the same. Any reason why we want to duplicate here ?

packages/manager/apps/pci-object-storage/src/pages/object-storage/storage/s3Id/objects/_components/SearchBar.component.tsx
```diff
@@ -0,0 +1,119 @@
...
+    setSearchOpen(value.length > 0);
+  };
+
+  const showResults =
+    searchOpen && deferredSearchQuery && deferredSearchQuery.length > 0;
+
+  return (
+    <Popover open={showResults} onOpenChange={setSearchOpen}>
+      <Command className="rounded-lg border-none shadow-none">
+        <div className="flex">
```
jperchoc: This seems to fix the weird border issue, while we find a better solution
```suggestion
        <div className="flex pl-4">
```

packages/manager/apps/pci-object-storage/src/pages/object-storage/storage/s3Id/objects/delete/DeleteObject.modal.tsx
```diff
@@ -41,7 +48,7 @@ const DeleteSwiftObject = () => {
           name: objectKey,
         }),
       });
-      navigate('../');
+      navigate(-1);
```
jperchoc: Not sure to understand this change

packages/manager/apps/pci-object-storage/src/pages/object-storage/storage/s3Id/objects/delete/DeleteObject.modal.tsx
```diff
@@ -55,7 +62,7 @@ const DeleteSwiftObject = () => {
   };
 
   return (
-    <RouteModal isLoading={!projectId && !objectKey}>
+    <RouteModal backUrl={-1} isLoading={!projectId && !objectKey}>
```
jperchoc: Same here, why not keeping the default backurl ?

packages/manager/apps/pci-object-storage/src/pages/object-storage/storage/s3Id/objects/object/_components/ObjectDownloadButton.component.tsx
```diff
@@ -0,0 +1,66 @@
...
+  });
+
+  return (
+    <TooltipProvider>
+      <Tooltip>
+        <TooltipTrigger asChild>
+          <span
+            className="inline-block"
+            tabIndex={isDownloadActionDisabled ? 0 : -1}
+          >
```
jperchoc: Do we need a span here ? There is only a button in the content

jperchoc (approved):

PR #20533, opened 2025-11-20
feat(backup-agent): move baremetal logics in backup-agent module

tibs245:
ref: #BKP-652

## Description

Ticket Reference: #...


packages/manager/apps/bmc-backup-agent-baremetal/package.json
```diff
@@ -32,26 +35,30 @@
     "@ovhcloud/ods-components": "~18.6.2",
     "@ovhcloud/ods-themes": "~18.6.2",
     "@tanstack/react-query": "5.51.21",
+    "@tanstack/react-virtual": "^3.10.9",
```
mathieumousnier, line 38: I might be missing something with modules and dependencies, but it looks weird to see this here. Do you use it in the project outside of the datagrid component ?

packages/manager/modules/backup-agent/src/components/Fields/RhfField.component.tsx
mathieumousnier, line 1: There is a rule in the eslint static kit that specifies to avoid `index` files.
This would be better not to add more of them.

Personnaly, it takes me 5 minutes each time to find the main component in a folder where it is called `index`

packages/manager/modules/backup-agent/src/data/api/index.ts
mathieumousnier, line 1: Barrel files looks cool, [but they induces many issues](https://www.google.com/search?q=stop%20using%20barrel%20files).
- circular dependencies
- loading to much context in tests and making them very long

I would advise not using them so much.

packages/manager/apps/bmc-backup-agent-baremetal/src/pages/onboarding/Onboarding.spec.tsx
```diff
@@ -7,13 +7,17 @@ import { BAREMETAL_MOCK } from '@/mocks/baremetals/baremetals.mocks';
...
 
-vi.mock('react-router-dom', () => ({
-  Link: ({ to, children, ...props }: { to: string; children: React.ReactNode }) => (
-    <a href={to} {...props}>
-      {children}
-    </a>
-  ),
-}));
+vi.mock('react-router-dom', async () => {
+  const actual = await vi.importActual('react-router-dom');
```
vovh, line 11: Out of curiosity, is importActual used to avoid multiple instances ?

vovh (approved):

packages/manager/modules/backup-agent/src/components/Fields/RhfField.component.tsx
```diff
@@ -0,0 +1,123 @@
...
+  }
+>) => {
+  const { helperMessage, controller } = useRhfFieldContext();
+
+  if (!controller?.fieldState.error) {
+    return <RhfHelper {...rest}>{customHelperMessage ?? helperMessage}</RhfHelper>;
+  }
+
+  return <></>;
+};
```
mathieumousnier, line 100: - Common pattern is to return early if there is an error, and let the nominal case in the component body.
- I would really avoid all the complexity with labels coming from the global state, and overriden with the props...

packages/manager/apps/bmc-backup-agent-baremetal/src/pages/onboarding/Onboarding.spec.tsx
tibs245: No it's to mock Link but keep `useNavigate` . 

To avoid multiple instance it's a reorganization of module in package.json and add react-hook-form in `dedupe` (See: `packages/manager/apps/bmc-backup-agent-baremetal/vitest.config.js`)

packages/manager/modules/backup-agent/src/components/Fields/RhfField.component.tsx
```diff
@@ -0,0 +1,123 @@
...
+  ...rest
+}: Readonly<RhfFieldProps<TFieldValues, TName>>) => {
+  const id = useId();
+  const controller = useController({ ...controllerParams, control });
+
+  const contextValues = {
+    id,
+    helperMessage,
+    controller,
+  };
```
mathieumousnier, line 38: I really think this global state is overkill. It adds complexity for really small gain.
These components would look sharper and would be easier to use without it.

- `controller` is reachable from the useController hook. So this is a copy in a context of a data that is already in an internal context of the rhf lib, then it would be easier for coworkers to use the rhf hook and not having to learn a custom implementation
- `helperMessage`can be set on the concerned component
- `id` also

packages/manager/modules/backup-agent/src/components/Fields/RhfField.component.tsx
```diff
@@ -0,0 +1,123 @@
...
+};
+
+export const RhfField = <TFieldValues extends FieldValues, TName extends Path<TFieldValues>>({
+  controllerParams,
+  className,
+  helperMessage,
+  isHiddenError,
+  control,
+  ...rest
+}: Readonly<RhfFieldProps<TFieldValues, TName>>) => {
```
mathieumousnier, line 30: I see that you call you component like this:

```tsx
      <RhfField
        control={control}
        data-testid="rhfField"
        controllerParams={register('inputTest')}
        helperMessage={HELPER_MESSAGE}
      >
```

But you don't need to (and shouldn't) use `register` and `control` together, they are both intended to register components, but in different ways.
I you want to be able to call `usController`, you just need `control` and `name`

packages/manager/modules/backup-agent/src/pages/services/dashboard/agent/add-configuration/AddConfiguration.page.tsx
```diff
@@ -0,0 +1,133 @@
...
+      className="flex flex-col gap-4"
+      noValidate
+      onSubmit={(e) => void handleSubmit(onSubmit)(e)}
+    >
+      <Drawer
+        isOpen
+        heading={t(`${BACKUP_AGENT_NAMESPACES.AGENT}:add_configuration`)}
+        onDismiss={goBack}
+        primaryButtonLabel={t(`${NAMESPACES.ACTIONS}:add`)}
+        onPrimaryButtonClick={() => formRef.current?.requestSubmit()}
```
mathieumousnier, line 76: Why don't you just call `handleSubmit(onSubmit)` ?
You wouldn't need `formRef`

packages/manager/modules/backup-agent/src/pages/vaults/listing/Listing.spec.tsx
```diff
@@ -25,7 +25,7 @@ describe('[INTEGRATION] - Listing page', () => {
           container.querySelector(`ods-link[label="${mockVaults[0]!.currentState.name}"]`),
         ).toBeVisible();
       },
-      { timeout: 5000 },
+      { timeout: 10_000 },
```
mathieumousnier, line 28: Waiting 10 sec in a test is very long. It looks like there is an issue somewhere.

packages/manager/apps/bmc-backup-agent-baremetal/package.json
tibs245: I has an error with MRC and this dependancies

packages/manager/modules/backup-agent/src/components/Fields/RhfField.component.tsx
tibs245: I will discuss with Hela. But in this case as it's a module I need sometime import logic of this module on app. It's better I think instead duplicate logic or create multiple modules.

packages/manager/modules/backup-agent/src/pages/vaults/listing/Listing.spec.tsx
tibs245: My procs with Webex need more 5 seconds :/

packages/manager/modules/backup-agent/src/pages/services/dashboard/agent/add-configuration/AddConfiguration.page.tsx
tibs245: To avoid duplicated logic. With that I can directly have one point for verity

PR #20539, opened 2025-11-20
chore(okms): activate typescript strict=true

mathieumousnier:
## Description

Ticket Reference: #MANAGER-19870

### DIFF
--- a/packages/manager/apps/okms/src/common/components/okms-dashboard/billing-tile/BillingTile.spec.tsx
+++ b/packages/manager/apps/okms/src/common/components/okms-dashboard/billing-tile/BillingTile.spec.tsx
@@ -1,4 +1,4 @@
-import { okmsMock } from '@key-management-service/mocks/kms/okms.mock';
+import { okmsRoubaix1Mock } from '@key-management-service/mocks/kms/okms.mock';
 import { KMS_ROUTES_URIS } from '@key-management-service/routes/routes.constants';
 import { SECRET_MANAGER_ROUTES_URLS } from '@secret-manager/routes/routes.constants';
 import { act, screen } from '@testing-library/react';
@@ -11,7 +11,7 @@ import { renderWithI18n } from '@/common/utils/tests/testUtils';
 import { BillingTile } from './BillingTile.component';
 import { BILLING_TILE_TEST_IDS } from './BillingTile.constants';
 
-const okmsMocked = okmsMock[0];
+const okmsMocked = okmsRoubaix1Mock;
 
 vi.mock('@/common/hooks/useProductType');
 
--- a/packages/manager/apps/okms/src/common/components/okms-dashboard/secret-config-tile/items/CasTileItem.spec.tsx
+++ b/packages/manager/apps/okms/src/common/components/okms-dashboard/secret-config-tile/items/CasTileItem.spec.tsx
@@ -6,16 +6,14 @@ import { labels } from '@/common/utils/tests/init.i18n';
 import { renderWithI18n } from '@/common/utils/tests/testUtils';
 
 import { SECRET_CONFIG_TILE_TEST_IDS } from '../SecretConfigTile.constants';
-import { CasTileItem } from './CasTileItem.component';
+import { CasTileItem, CasTileItemProps } from './CasTileItem.component';
 
-const renderTileItem = async ({
-  isPending = false,
-  secretConfig,
-}: {
-  isPending?: boolean;
-  secretConfig?: SecretConfig;
-}) => {
-  return renderWithI18n(<CasTileItem secretConfig={secretConfig} isPending={isPending} />);
+const buildProps = (props: Partial<CasTileItemProps> = {}): CasTileItemProps => {
+  return props as CasTileItemProps;
+};
+
+const renderTileItem = async (secretConfigOkmsQuery: CasTileItemProps) => {
+  return renderWithI18n(<CasTileItem {...secretConfigOkmsQuery} />);
 };
 
 describe('OKMS - secret config - cas item Tile Item test suite', () => {
@@ -38,7 +36,7 @@ describe('OKMS - secret config - cas item Tile Item test suite', () => {
     // GIVEN
 
     // WHEN
-    await renderTileItem({ secretConfig, isPending: false });
+    await renderTileItem(buildProps({ data: secretConfig, isPending: false, isError: false }));
 
     // THEN
     expect(screen.getByText(labels.secretManager.cas_with_description)).toBeVisible();
@@ -50,7 +48,7 @@ describe('OKMS - secret config - cas item Tile Item test suite', () => {
     // GIVEN
 
     // WHEN
-    await renderTileItem({ isPending: true });
+    await renderTileItem(buildProps({ isPending: true, isError: false, data: undefined }));
 
     // THEN
     expect(screen.getByTestId(SECRET_CONFIG_TILE_TEST_IDS.skeleton)).toBeInTheDocument();
--- a/packages/manager/apps/okms/src/common/pages/order-okms-modal/OrderOkmsModal.page.spec.tsx
+++ b/packages/manager/apps/okms/src/common/pages/order-okms-modal/OrderOkmsModal.page.spec.tsx
@@ -126,13 +126,13 @@ const clickOnConfirmButton = async (user: UserEvent) => {
   // Fails intermittently without this - button click does not always works
   await wait(300);
 
-  let confirmButton: HTMLElement;
+  let confirmButton: HTMLElement | undefined = undefined;
   await waitFor(() => {
     confirmButton = screen.getByTestId(ORDER_OKMS_TC_CONFIRM_BUTTON_TEST_ID);
     expect(confirmButton).toHaveAttribute('is-disabled', 'false');
   });
   await act(async () => {
-    await user.click(confirmButton);
+    await user.click(confirmButton!);
   });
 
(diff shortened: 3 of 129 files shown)


packages/manager/apps/okms/src/common/components/okms-dashboard/billing-tile/BillingTile.spec.tsx
```diff
@@ -11,7 +11,7 @@ import { renderWithI18n } from '@/common/utils/tests/testUtils';
 import { BillingTile } from './BillingTile.component';
 import { BILLING_TILE_TEST_IDS } from './BillingTile.constants';
 
-const okmsMocked = okmsMock[0];
+const okmsMocked = okmsRoubaix1Mock;
```
pauldkn, line 14: **nitpick:** Maybe you could use a function like your `invariant` to be able to keep `okmsMock[0]`, something like `definedValue(okmsMock[0])` and it might avoid having to create a individual mock

packages/manager/apps/okms/src/common/components/okms-dashboard/secret-config-tile/items/CasTileItem.spec.tsx
```diff
@@ -6,16 +6,14 @@ import { labels } from '@/common/utils/tests/init.i18n';
...
-}: {
-  isPending?: boolean;
-  secretConfig?: SecretConfig;
-}) => {
-  return renderWithI18n(<CasTileItem secretConfig={secretConfig} isPending={isPending} />);
+import { CasTileItem, CasTileItemProps } from './CasTileItem.component';
+
+const buildProps = (props: Partial<CasTileItemProps> = {}): CasTileItemProps => {
+  return props as CasTileItemProps;
+};
```
pauldkn, line 13: Are these functions "tricks" only for TS ?
(you're not adding properties making a complete object)

packages/manager/apps/okms/src/common/utils/tests/uiTestHelpers.ts
```diff
@@ -60,6 +60,9 @@ const getOdsButton = async ({
     },
     { ...WAIT_FOR_DEFAULT_OPTIONS, ...options },
   );
+  if (!button) {
+    throw new Error(`Button ${label ?? altLabel ?? iconName ?? 'not found'} not found`);
```
pauldkn: To avoid "Button not found not found"
```suggestion
    throw new Error(`Button ${label ?? altLabel ?? iconName ?? ''} not found`);
```

packages/manager/apps/okms/src/common/utils/tools/filterNullish.ts
pauldkn, line 1: **nitpick:** `Nullish` word is a bit confusing, I wouldn't expect it to filter `false`

You could use `filterFalsy` as you're restricting the type of T, it wouldn't accept `''` or `0` anyway so there would be less confusion imo

packages/manager/apps/okms/src/common/components/okms-dashboard/secret-config-tile/items/CasTileItem.spec.tsx
mathieumousnier: This function is only intended for tests, I would NEVER use this in the project code.
So yes, it lies to typescript end says the props have all the required properties.

I had to do this because I typed the component props from the UseQueryResult type, to be able to pass the whole useQuery result object to the component.
I am not really fan to couple project components with react-query like this, but it is a good way to keep the relation between `data` `isPending` and `isError`. If I declare the props separately I will lose the relation

packages/manager/apps/okms/src/common/components/okms-dashboard/billing-tile/BillingTile.spec.tsx
pauldkn: After giving it some thought and discussing together about it, your approach is better 👍

(ex: allows selecting a specific mock, order is not important)

mathieumousnier: Good suggestion but I prefer to use dedicated mocks for resource units than importing the full list and pointing to one index.

packages/manager/apps/okms/src/common/utils/tools/filterNullish.ts
mathieumousnier: This is right, I fixed it, and even added a test file for the filterFalsy function

packages/manager/apps/okms/src/common/utils/tools/filterFalsy.spec.ts
pauldkn, line 1: **nitpick:** I think you could reduce the tests to 4 main cases:
1. mixed: 1 truthy value + all falsy values → returns array with only the truthy value
2. only falsy values → returns empty array
3. only truthy values → returns the same array
4. empty array → returns an empty array

pauldkn (approved):

packages/manager/apps/okms/src/common/utils/tools/filterFalsy.spec.ts
mathieumousnier: done ✅

packages/manager/apps/okms/src/common/pages/order-okms-modal/OrderOkmsModal.page.spec.tsx
```diff
@@ -126,12 +126,16 @@ const clickOnConfirmButton = async (user: UserEvent) => {
...
   await waitFor(() => {
     confirmButton = screen.getByTestId(ORDER_OKMS_TC_CONFIRM_BUTTON_TEST_ID);
     expect(confirmButton).toHaveAttribute('is-disabled', 'false');
   });
+
   await act(async () => {
+    if (!confirmButton) {
+      throw new Error('Confirm button not found');
+    }
     await user.click(confirmButton);
```
tibs245: Just to suggest 2 alternative for the 3 is good for me : 

## Option 2: 
```suggestion
  let confirmButton: HTMLElement | undefined = undefined;
await waitFor(() => {
confirmButton = screen.getByTestId(ORDER_OKMS_TC_CONFIRM_BUTTON_TEST_ID);
expect(confirmButton).toHaveAttribute('is-disabled', 'false');
});

await act(async () => {
await user.click(confirmButton!);
```

## Option 3



```suggestion
// On top of file : 
const getConfirmButton = () => screen.getByTestId(ORDER_OKMS_TC_CONFIRM_BUTTON_TEST_ID)

const getEnabledConfirmButton = () => {
await waitFor(() => expect(getConfirmButton()).toHaveAttribute('is-disabled', 'false');
);

return getConfirmButton();
}

///
  let confirmButton: HTMLElement | undefined = undefined;

await act(async () => {
await user.click(getEnabledConfirmButton());
```

packages/manager/apps/okms/src/modules/key-management-service/hooks/service-key/useServiceKeyActions.spec.tsx
```diff
@@ -11,6 +11,12 @@ import { ODS_BUTTON_COLOR } from '@ovhcloud/ods-components';
 
 import useServiceKeyActionsList from './useServiceKeyActionsList';
 
+const okmsItemMock = okmsRoubaix1Mock;
+
+if (!okmsItemMock) {
+  throw new Error('OKMS item mock not found');
+}
+
```
rjamet-ovh:
```suggestion
const okmsItemMock = okmsRoubaix1Mock;

```
since `okmsItemMock` is a mock, it can't be undefined

packages/manager/apps/okms/src/modules/key-management-service/pages/create/CreateKms.spec.tsx
```diff
@@ -29,10 +29,14 @@ vi.mock('react-router-dom', async (importOriginal) => {
   };
 });
 
-const regionListMock = catalogMock.plans[0].configurations[0].values;
-const firstRegionMock = locationsMock.find((location) => location.name === regionListMock[0]);
+const regionListMock = catalogMock?.plans[0]?.configurations[0]?.values;
```
rjamet-ovh:
```suggestion
const regionListMock = catalogMock.plans[0]?.configurations[0]?.values;
```
this mock can't be undefined

packages/manager/apps/okms/src/common/pages/order-okms-modal/OrderOkmsModal.page.spec.tsx
mathieumousnier: I took option 2

packages/manager/apps/okms/src/modules/secret-manager/hooks/useCurrentRegion.ts
```diff
@@ -10,7 +9,7 @@ import { LocationPathParams } from '@secret-manager/routes/routes.constants';
  * If neither is found, it returns undefined.
  */
 export const useCurrentRegion = (okmsList: OKMS[]) => {
-  const { okmsId, region } = useParams<LocationPathParams>();
+  const { okmsId, region } = useParams();
```
rjamet-ovh, line 12: should we use `useRequiredParams` here as well?

mathieumousnier: No because according to the place we use it, there will be a `okmsId` or a `region` parameter.
And the `undefined` cases are managed in the hook: 

image

PR #20559, opened 2025-11-21
feat(pci-block-storage): few retyping ui fixes

aTurmo:
ref: #TAPC-5344

### DIFF
--- a/packages/manager/apps/pci-block-storage/src/api/hooks/useVolume.tsx
+++ b/packages/manager/apps/pci-block-storage/src/api/hooks/useVolume.tsx
@@ -5,7 +5,7 @@ import {
   PaginationState,
   useCatalogPrice,
 } from '@ovh-ux/manager-react-components';
-import { useMutation, useQueries, useQuery } from '@tanstack/react-query';
+import { useMutation, useQueries } from '@tanstack/react-query';
 import { useTranslation } from 'react-i18next';
 import { pipe } from 'lodash/fp';
 import { NAMESPACES } from '@ovh-ux/manager-common-translations';
--- a/packages/manager/apps/pci-block-storage/src/components/icon/Icon.tsx
+++ b/packages/manager/apps/pci-block-storage/src/components/icon/Icon.tsx
@@ -3,6 +3,6 @@ import { ComponentProps, PropsWithoutRef } from 'react';
 
 type Props = PropsWithoutRef<ComponentProps<'span'> & { name: IconName }>;
 
-export const Icon = ({ name, className, ...spanProps }: Props) => (
+export const Icon = ({ name, ...spanProps }: Props) => (
   <OdsIcon name={name} {...spanProps} />
 );
--- a/packages/manager/apps/pci-block-storage/src/hooks/useSearchFormParams.ts
+++ b/packages/manager/apps/pci-block-storage/src/hooks/useSearchFormParams.ts
@@ -40,6 +40,7 @@ export function useSearchFormParams<
         },
       );
     } catch (e) {
+      // eslint-disable-next-line no-console
       console.error('Error while parsing search params into form', e);
     }
   }, [searchParams]);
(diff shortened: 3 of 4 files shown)


packages/manager/apps/pci-block-storage/src/components/VolumeModelTilesInput.component.tsx
```diff
@@ -71,31 +75,33 @@ export const VolumeModelTilesInput = ({
...
-        price: m.hourlyPrice,
+      putPreselectedModelFirst(volumeModels).map((model) => ({
+        ...model,
+        label: model.displayName,
+        description: getDescription(model),
+        badges: hideBadges
+          ? []
+          : [
+              model.encrypted
+                ? {
```
ppprevost: [suggestion](https://pullpo.io/cc?l=suggestion) 
```
  {
    label: t(
      `common:pci_projects_project_storages_blocks_encryption_${
        model.encrypted ? 'available' : 'unavailable'
      }`
    ),
    backgroundColor: model.encrypted ? '#D2F2C2' : '#FFCCD9',
    textColor: model.encrypted ? '#113300' : '#4D000D',
    icon: 'lock' as const,
  },
```

aTurmo: why not

aTurmo: done

packages/manager/apps/pci-block-storage/src/api/select/catalog.ts
```diff
@@ -404,3 +404,12 @@ export const mapRetypingVolumeCatalog = (
         mapPreselection(type),
       ),
     );
+
+export const putPreselectedModelFirst = <T = TVolumeModel | TVolumeRetypeModel>(
```
fredericvilcot: `put` refers to a CRUD operation, I'll suggest to make the sort understandable


```suggestion
export const sortByPreselectedModel = <T = TVolumeModel | TVolumeRetypeModel>(
```

packages/manager/apps/pci-block-storage/src/api/select/catalog.ts
```diff
@@ -404,3 +404,12 @@ export const mapRetypingVolumeCatalog = (
...
+  volumeModels: T[],
+): T[] =>
+  [...volumeModels].sort((a, b) => {
+    if (a['isPreselected' as keyof TVolumeRetypeModel] === true) return -1;
```
fredericvilcot: You should not cast `T[a]` or `T[b]` `as keyof TVolumeRetypeModel` because you are making a wrong usage of your generic type that can be either `TVolumeRetypeModel` or `TVolumeModel` that has no preselected property.

`a['isPreselected' as keyof TVolumeRetypeModel]` will return undefined if property does not exist that will mislead your sort logic ;)

packages/manager/apps/pci-block-storage/src/components/VolumeModelTilesInput.component.tsx
```diff
@@ -111,9 +110,10 @@ export const VolumeModelTilesInput = ({
         horizontal && '[&_.config-card\\_\\_badges]:w-full whitespace-pre-line',
       )}
     >
+      <Text preset={TEXT_PRESET.heading5}>{label}</Text>
       <TilesInput
         name="volume-type"
-        label={label}
+        label=""
```
fredericvilcot: You could use `undefined` instead to say that you are not using a label for `TilesInput` :)

packages/manager/apps/pci-block-storage/src/api/select/catalog.ts
aTurmo: The cast as keyof TVolumeRetypeModel is on 'isPreselected' and not T[a]. And having `a['isPreselected' as keyof TVolumeRetypeModel]` returning undefined is what I want for TVolumeModel. That way I return 0 and therefore the array stay the same.

I've thoroughly tested the method to be sure it works as intended.

fredericvilcot: Yes, but if tomorrow you'll have another sort logic for `TVolumeModel `, it will lead to unexpected behavior manipulating 0 values for another sort ;)

PR #20575, opened 2025-11-24
feat(billing-components): fix design status column in table services

achrafHafedh:
ref: #MANAGER-20011

## Description

Ticket Reference: #MANAGER-20011


packages/manager/modules/billing-components/src/components/service-status/service-status.html
```diff
@@ -1,4 +1,4 @@
-<p>
+<p class="mb-0">
```
JacquesLarique: You can also just remove the `<p>`, I dont think it make sense anymore

aottr (approved):

PR #20582, opened 2025-11-24
feat(billing): add punycode in datagrid popin and csv

achrafHafedh:
ref: #MANAGER-20025

## Description

Ticket Reference: #MANAGER-20025

### DIFF
--- a/packages/manager/modules/billing/src/autoRenew/services/services.html
+++ b/packages/manager/modules/billing/src/autoRenew/services/services.html
@@ -122,7 +122,7 @@
 
         <!-- Columns -->
         <oui-datagrid-column
-            data-property="serviceId"
+            data-property="serviceIdUnicode"
             data-title="::'billing_autorenew_service_name' | translate"
             data-type="string"
             data-searchable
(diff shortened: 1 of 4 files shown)


packages/manager/modules/billing/src/autoRenew/services/services.html
```diff
@@ -142,15 +142,15 @@
...
             <span
                 data-ng-if="!$row.url || $row.isByoipService()"
-                data-ng-bind="::$row.serviceId | limitTo:80"
+                data-ng-bind="::$row.serviceIdUnicode | limitTo:80"
             ></span>
             <small
                 class="d-block"
                 data-ng-if="$row.alternativeId !== $row.serviceId"
-                data-ng-bind="$row.alternativeId"
+                data-ng-bind="$row.serviceIdUnicode"
```
JacquesLarique: I don't see the request to encode the `alternativeId` in the ticket, but if it is the case you should parse the property directly in the class (would be OK since no call to API are using this property) and still display it there
_you could add some one-time binding (`::`) to enhance perf_

achrafHafedh: Yes, you’re right, I don’t need this change. I rolled back this modification

JacquesLarique (approved):

MehdiZaghouani (approved):

MaximeBajeux (approved):

PR #20601, opened 2025-11-25
fix(muk): datagrid header table fixed sticky

aboungnaseng-ovhcloud:
ref: #MANAGER-20389

## Description

- Fix header sticky, table fixed css
Screenshot 2025-11-25 at 17 12 07

- Fix datagrid top bar sm size alignment
Screenshot 2025-11-25 at 17 10 43

- Update translate Y for stick topbar for safari
Screenshot 2025-11-25 at 16 44 01

Screenshot 2025-11-25 at 16 43 52

- Test in Edge Browser
Screenshot 2025-11-25 at 16 46 03

### DIFF
--- a/packages/manager-ui-kit/src/components/datagrid/table/table-body/TableBody.component.tsx
+++ b/packages/manager-ui-kit/src/components/datagrid/table/table-body/TableBody.component.tsx
@@ -1,4 +1,4 @@
-import { Fragment, useCallback, useEffect, useMemo, useState } from 'react';
+import { Fragment, useCallback, useEffect, useMemo } from 'react';
 
 import { flexRender } from '@tanstack/react-table';
 import { useVirtualizer } from '@tanstack/react-virtual';
@@ -24,9 +24,9 @@ export const TableBody = <T,>({
   tableContainerRef,
   contentAlignLeft = true,
 }: TableBodyProps<T>) => {
-  const [browserName, setBrowserName] = useState<string>('unknown');
   const { rows } = rowModel;
   const previousRowsLength = usePrevious(rows?.length);
+  const browserName = useMemo(() => getBrowserName(), []);
   const rowVirtualizer = useVirtualizer({
     count: rows.length,
     estimateSize: useCallback(() => maxRowHeight, [maxRowHeight]),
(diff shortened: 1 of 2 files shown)


packages/manager-ui-kit/src/components/datagrid/table/table-body/TableBody.component.tsx
```diff
@@ -91,7 +97,10 @@ export const TableBody = <T,>({
               style={{
                 left: leftPosition,
                 height: `${maxRowHeight}px`,
-                transform: `translateY(${translateY}px)`,
+                transform:
+                  browserName === 'Safari' && !hideHeader
+                    ? `translateY(${translateY + maxRowHeight}px)`
+                    : `translateY(${translateY}px)`,
```
anooparveti, line 99: **question:** Why should we have browser specific implementation? Any references that propose this solution?

aboungnaseng-ovhcloud: Screenshot 2025-11-26 at 09 11 27

Cause in safari the sticky position with tailwind dont work well, our implementation with a Thead in sticky and scroll in the Tbody dont work specially in safari.

If we want to make it work on safari withtout adding this, we need to change our thead by a div and the tbody div too, but semanticly its not good

stif59100 (approved):

packages/manager-ui-kit/src/components/datagrid/table/table-body/TableBody.component.tsx
anooparveti: Agreed to proceed with the fix and revisit this topic with a new ticket.

helabenkhalfallah: As discussed we create a ticket for a more clean solution.

packages/manager-ui-kit/src/components/datagrid/utils/Datagrid.utils.ts
```diff
@@ -0,0 +1,10 @@
+export const getBrowserName = () => {
```
helabenkhalfallah, line 1: this can be a generic utility

PR #20612, opened 2025-11-25
fix(ips): fixing the redirection with serviceName and toggleFirewall …

aderghamov:
…action

ref: #MANAGER-19926

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/dedicated/client/app/ip/ip.routing.js
+++ b/packages/manager/apps/dedicated/client/app/ip/ip.routing.js
@@ -46,6 +46,10 @@ export const urlDynamicParts = {
   service: ':service',
 };
 
+const targetQueryParams = {
+  serviceName: 'routedTo.serviceName',
+};
+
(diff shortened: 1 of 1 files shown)


packages/manager/apps/dedicated/client/app/ip/ip.routing.js
```diff
@@ -91,7 +98,9 @@ export function mapAngularStateToReactUrl(stateName, params) {
   };
 
   // Fallback – just send the user to the React root
-  return stateMap[stateName] || baseReactPath;
+  return stateName === 'app.ip.dashboard' && params.action === 'toggleFirewall'
```
chipp972, line 101: for the firewall it redirects to the configure edge network firewall page ?

aderghamov: I may be mistaken but I think it should be redirected to this page.
Do you know a better target for this redirection?

PR #20617, opened 2025-11-25
feat(hub): add fix and verify html tests

achrafHafedh:
ref: #MANAGER-20181

## Description

Ticket Reference: #MANAGER-20181

### DIFF
--- a/packages/manager/apps/hub/src/billing/components/billing-status/BillingStatus.spec.tsx
+++ b/packages/manager/apps/hub/src/billing/components/billing-status/BillingStatus.spec.tsx
@@ -1,0 +1,45 @@
+import { render } from '@testing-library/react';
+import { describe, expect, it } from 'vitest';
+
+import { FourServices } from '@/__mocks__/billingServices';
+
+import BillingStatus from './BillingStatus.component';
+
+describe('BillingStatus W3C Validation', () => {
+  const [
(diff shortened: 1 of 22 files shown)


MehdiZaghouani (approved):

packages/manager/apps/hub/src/pages/dashboard/PaymentStatus.component.tsx
JacquesLarique, line 1: This should not be fixed this way, as we are saying all cells are headers which is not good for accessibility
We should consider reworking this part to not use a table (since it has no header)
I let you free to either fix it in this PR or create a dedicated ticket (in which we could include the support ticket table as well)

packages/manager/apps/hub/src/billing/components/billing-status/BillingStatus.spec.tsx
JacquesLarique, line 1: The name of this file should be `BillingStatus.spec.tsx`

MehdiZaghouani (dismissed):

packages/manager/apps/hub/src/billing/components/billing-status/BillingStatus.spec.tsx
achrafHafedh: fixed

PR #20646, opened 2025-11-26
feat(pci-worklow): fix instances selection filters and sorting

aTurmo:
ref: #TAPC-5370


packages/manager/apps/pci-workflow/src/api/hooks/instance/selector/instances.selector.ts
```diff
@@ -146,16 +140,37 @@ export const instancesSelector = (
     status: mapInstanceStatus(instance),
     region: regionMapper(instance),
     autoBackup: autoBackupMapper(instance),
+    searchField: `${instance.name} ${regionMapper(instance).label} ${
+      mapInstanceStatus(instance).name
+    } ${mapInstanceFlavor(flavors)(instance)}`,
```
lolottetheclash: I guess it's correct but not easily readable...

aTurmo: improved (i hope ^^)

PR #20649, opened 2025-11-26
feat(github): update muk templates and label

aboungnaseng-ovhcloud:
## Description

- update muk templates and label

Ticket Reference: #MANAGER-19477


.github/labeler.yml
```diff
@@ -65,6 +65,8 @@ enabler:
   - packages/manager/apps/iam/**
 manager-components:
   - packages/manager-components/src/**
```
anooparveti: We can remove this.

anooparveti (approved):

PR #20650, opened 2025-11-26
fix(account-creation): fix crash in settings page

achrafHafedh:
ref: #MANAGER-20305

## Description

Ticket Reference: #MANAGER-20305


JacquesLarique (changes requested):
Can you remove the `isPreloaderHide` prop on the `ErrorBoundary` component in packages/manager/apps/account-creation/src/routes/routes.tsx please

MehdiZaghouani (approved):

PR #20663, opened 2025-11-27
feat(pci-instances): link image types with api

lolottetheclash:
ref: #TAPC-5220

FYI some data will be moved into view models in another PR

## Description

Ticket Reference: #...


packages/manager/apps/pci-instances/src/utils/index.ts
```diff
@@ -63,24 +63,22 @@ export const getLocalZoneTranslationKey = (regionName: string) =>
 
 export const isEven = (nb: number) => nb % 2 === 0;
 
-export const getRegionalizedFlavorId = (
-  flavorName: string,
-  regionName: string,
-) => `${flavorName}_${regionName}`;
+export const getRegionalizedItemId = (itemName: string, regionName: string) =>
```
Tsiorifamonjena: Isn't a bit more generalised "item" ? Maybe I don't really understand the purpose. I understand it will be to get the Id but item of what ?

packages/manager/apps/pci-instances/src/adapters/i18n/left/i18nMessageProvider.ts
Tsiorifamonjena, line 1: I guess it will no more necessary ?

lolottetheclash: exactly, it's handled in next PR :)

packages/manager/apps/pci-instances/src/utils/index.ts
lolottetheclash: Actually, I had generalized it because I was using it for other things, but in the end I didn't use it, so I'm renaming it. Thank you!

PR #20671, opened 2025-11-27
feat(pci-kubernetes): update kubeconfig actions

Taelar:
## Description

Ticket Reference: #TAPC-5144


packages/manager/apps/pci-kubernetes/src/components/service/ClusterAccessAndSecurity.component.tsx
```diff
@@ -125,8 +81,8 @@ export default function ClusterAccessAndSecurity({
   };
 
   return (
-    <OsdsTile className="flex-col w-full shadow-lg" inline rounded variant={ODS_TILE_VARIANT.ghost}>
-      <div className="flex flex-col w-full">
+    <OsdsTile className="w-full flex-col shadow-lg" inline rounded variant={ODS_TILE_VARIANT.ghost}>
+      <div className="flex w-full flex-col">
```
aTurmo, line 85: Question: Why did you changed the classes order ? Of course not important but why ? ^^
And btw I think it's better to have flex related classes together like it was before with flew and flex-col on line 85

packages/manager/apps/pci-kubernetes/src/components/service/ClusterAccessAndSecurity.component.tsx
```diff
@@ -315,7 +249,7 @@ export default function ClusterAccessAndSecurity({
                                   key !== 'clientId',
                               )
                               .map(([key, value]) => (
-                                <div key={key} className="mb-4 mt-4 flex flex-col max-w-[400px]">
+                                <div key={key} className="my-4 flex max-w-[400px] flex-col">
```
aTurmo, line 252: same thing here with flex classes

packages/manager/apps/pci-kubernetes/src/components/service/ClusterConfigFileActions.component.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+  vi.spyOn(KubeAPI, 'postKubeConfig').mockReturnValue(new Promise(() => {}));
+
+  const { getByTestId, queryByTestId } = render(
+    <ClusterConfigFileActions projectId="" kubeDetail={{ status: 'READY' } as TKube} />,
+    { wrapper },
+  );
+
+  await userEvents.click(getByTestId('ClusterConfigFileActions-DownloadKubeConfig'));
+
+  expect(getByTestId('clusterConfigFileActions-spinnerDownloadKubeConfig')).toBeInTheDocument();
```
aTurmo:
```suggestion
  expect(getByTestId('clusterConfigFileActions-spinnerDownloadKubeConfig')).toBeVisible();
```

packages/manager/apps/pci-kubernetes/src/components/service/ClusterConfigFileActions.component.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+  vi.spyOn(KubeAPI, 'postKubeConfig').mockReturnValue(new Promise(() => {}));
+
+  const { getByTestId, queryByTestId } = render(
+    <ClusterConfigFileActions projectId="" kubeDetail={{ status: 'READY' } as TKube} />,
+    { wrapper },
+  );
+
+  await userEvents.click(getByTestId('ClusterConfigFileActions-CopyKubeConfig'));
+
+  expect(getByTestId('clusterConfigFileActions-spinnerCopyKubeConfig')).toBeInTheDocument();
```
aTurmo:
```suggestion
  expect(getByTestId('clusterConfigFileActions-spinnerCopyKubeConfig')).toBeVisible();
```

packages/manager/apps/pci-kubernetes/src/components/service/ClusterConfigFileActions.component.spec.tsx
```diff
@@ -0,0 +1,72 @@
...
+import { wrapper } from '@/wrapperRenders';
+
+import { ClusterConfigFileActions } from './ClusterConfigFileActions.component';
+
+it('disables download kubeconfig button when status is installing', () => {
+  const { getByTestId } = render(
+    <ClusterConfigFileActions projectId="" kubeDetail={{ status: 'INSTALLING' } as TKube} />,
+    { wrapper },
+  );
+  expect(getByTestId('ClusterConfigFileActions-DownloadKubeConfig')).toBeDisabled();
```
aTurmo: Remark (blocking):
Short: avoid using data-testid for tests
Long: we already have a discussion about that and we promote to not use data-testid in component for testing. It's better to use getByText or getByRole because it represent true things the user can interact with, whereas testId is just a hacky way of getting an element.
There are cases where it can be necessary to use data-testid, an exemple is with spinner where there is nothing tangible to rely on.
But in your cases you get button so it is accessible by role or text.

Of course this comment is for the whole file not just this test.

packages/manager/apps/pci-kubernetes/src/helpers/matchers/matchers.ts
```diff
@@ -12,3 +12,20 @@ export const isNodePoolNameValid = (input: string) =>
   NODE_POOL_NAME_CONSTRAINTS.PATTERN.test(input);
 
 export const isClusterNameValid = (input: string) => CLUSTER_NAME_CONSTRAINTS.PATTERN.test(input);
+
+const CLUSTER_API_URL_SCHEME =
```
aTurmo: Optionnal
```suggestion
const CLUSTER_API_URL_REGEX =
```

aTurmo (changes requested):

packages/manager/apps/pci-kubernetes/src/components/service/ClusterConfigFileActions.component.tsx
```diff
@@ -0,0 +1,113 @@
...
+        {isKubeConfigPending && ongoingKubeConfigAction === 'copy' && (
+          <Spinner size="sm" data-testid="clusterConfigFileActions-spinnerCopyKubeConfig" />
+        )}
+      </div>
+
+      <div className="flex items-center">
+        <Button
+          size="sm"
+          variant="ghost"
+          data-testid="ClusterConfigFileActions-DownloadKubeConfig"
```
ppprevost: [question](https://pullpo.io/cc?l=question) 
: why do you not use kebab case for data-test id ?

packages/manager/apps/pci-kubernetes/src/helpers/matchers/matchers.ts
```diff
@@ -12,3 +12,20 @@ export const isNodePoolNameValid = (input: string) =>
...
+  /^(?<shortId>[\da-z]+)\.(?<subRegion>[\da-z]+)\.(?<region>[\da-z]+)\.k8s\.ovh\.net$/;
+
+export const getClusterUrlFragments = (url: string) => {
+  const match = url.match(CLUSTER_API_URL_SCHEME);
+
+  if (!match) {
+    return null;
+  }
+
+  return {
```
ppprevost, line 31: [suggestion](https://pullpo.io/cc?l=suggestion) 
i would prefer to handle nullity : 
```
const { shortId, subRegion, region } = match.groups;
 if (!shortId || !subRegion || !region) {
    return null;
  }
```

packages/manager/apps/pci-kubernetes/src/components/service/ClusterAccessAndSecurity.component.tsx
Taelar: It was done by the eslint/tailwind plugin, even though I agree that flex related classes should be grouped. Since it displays a warning otherwise, I think it's better that we let the plugin do the sorting automatically

packages/manager/apps/pci-kubernetes/src/components/service/ClusterConfigFileActions.component.tsx
Taelar: I must admit I just used the existing pattern I found on the component without checking if it was consistent with the rest of the app. Do we have any guidelines on how they should be structured ?

aTurmo: The guidelines is to not use data-testid in the first place when possible ^^
Then it's best if you follow the same pattern as the rest of the uapp you're working on. I don't think there is a general guideline for this.

packages/manager/apps/pci-kubernetes/src/components/service/ClusterConfigFileActions.component.tsx
```diff
@@ -0,0 +1,113 @@
...
+    });
+  };
+
+  return (
+    <div className="flex flex-col items-start gap-3">
+      <div className="flex items-center">
+        <Button
+          size="sm"
+          variant="ghost"
+          data-testid="copy-kubeconfig"
```
aTurmo: Great job removing data-testid from tests but you can also remove it from prod code ^^
I noticed here and line 101

aTurmo (approved):

aTurmo (approved):

packages/manager/apps/pci-kubernetes/src/components/service/ClusterConfigFileActions.component.tsx
```diff
@@ -0,0 +1,134 @@
...
+  const copyConfigFile = () => {
+    setOngoingKubeConfigAction('copy');
+    postKubeConfig(undefined, {
+      onSuccess: async (configFile) => {
+        try {
+          await navigator.clipboard.writeText(configFile.content);
+
+          toast(t('kube_service_copy_kubeconfig_success'), {
+            color: 'success',
+            duration: 5000,
```
ppprevost: we could use a const for the duration

ppprevost (approved):

PR #20673, opened 2025-11-27
fix(pci-worklow): fix listing worklow execution date and status

aTurmo:
ref: #TAPC-5376

### DIFF
--- a/packages/manager/apps/pci-workflow/public/translations/listing/Messages_en_GB.json
+++ b/packages/manager/apps/pci-workflow/public/translations/listing/Messages_en_GB.json
@@ -15,6 +15,7 @@
   "pci_workflow_execution_status_SUCCESS": "Success",
   "pci_workflow_execution_status_ERROR": "Error",
   "pci_workflow_execution_status_PAUSED": "Pause",
+  "pci_workflow_execution_status_CANCELED": "Cancelled",
   "pci_workflow_executions_title": "'{{ workflowName }}' workflow execution",
   "pci_workflow_executions_empty": "No executions found",
   "pci_workflow_executions_ececution_date": "Date (UTC)",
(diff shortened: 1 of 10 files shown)


packages/manager/apps/pci-workflow/public/translations/listing/Messages_en_GB.json
```diff
@@ -15,6 +15,7 @@
   "pci_workflow_execution_status_SUCCESS": "Success",
   "pci_workflow_execution_status_ERROR": "Error",
   "pci_workflow_execution_status_PAUSED": "Pause",
+  "pci_workflow_execution_status_CANCELED": "Cancelled",
```
Tsiorifamonjena, line 18: question: is this correct ? it's not Canceled ? (I'm not good at English anyway)

aTurmo: It's a good question because I asked myself the same thing, but "Canceled" is the american way and "cancelled" is the british way. And when looking for trads elsewhere in manager I always found the british way "cancelled" to translate "annulé"

packages/manager/apps/pci-workflow/src/api/hooks/workflows.tsx
```diff
@@ -33,13 +32,18 @@ import { useMe } from './user';
...
   lastExecutionStatus: TExecutionState;
 };
 
+export type TWorkflow = TRemoteWorkflow &
+  TWorkflowRegion &
+  TWorkflowLastExecution & {
+    instanceName: string;
+  };
+
 const getWorkflowQueryOptions = (projectId: string, regionName: string) =>
```
ppprevost: is that a hook ? Because you file is called with hook ?

lolottetheclash (approved):
well done & tested 👌
You should just rename your commits with fix instead of feat

PR #20674, opened 2025-11-27
fix(okms): fix breadcrumb items

rjamet-ovh:
ref: #MANAGER-20414

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/okms/src/modules/key-management-service/components/breadcrumb/Breadcrumb.tsx
+++ b/packages/manager/apps/okms/src/modules/key-management-service/components/breadcrumb/Breadcrumb.tsx
@@ -1,4 +1,4 @@
-import { useNavigate } from 'react-router-dom';
+import { useHref } from 'react-router-dom';
 
 import {
   BreadcrumbItem,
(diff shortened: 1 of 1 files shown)


packages/manager/apps/okms/src/modules/key-management-service/components/breadcrumb/Breadcrumb.tsx
```diff
@@ -12,25 +12,20 @@ export type BreadcrumbProps = {
...
-          }}
-          href={''}
-        />
+        <Item key={id} id={id} label={label} navigateTo={navigateTo} />
       ))}
     </OdsBreadcrumb>
   );
 }
 
+const Item = ({ id, label, navigateTo }: BreadcrumbItem) => {
```
tibs245, line 26: The linter not catch with rule "Only one component by file" ?

rjamet-ovh: disabled rule for now, we're migrating step by step :)

PR #20722, opened 2025-12-01
fix(pci-os): unit measure are not link to locale

abullet33:
ref: #DATATR-2660

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-object-storage/src/components/file-input/FileInput.component.tsx
+++ b/packages/manager/apps/pci-object-storage/src/components/file-input/FileInput.component.tsx
@@ -3,8 +3,8 @@ import { Upload, X } from 'lucide-react';
 import { Button } from '@datatr-ux/uxlib';
 import clsx from 'clsx';
 import { useTranslation } from 'react-i18next';
-import { octetConverter } from '@/lib/bytesHelper';
 import FileIcon from '../file-icon/FileIcon.component';
+import { useLocaleBytesConverter } from '@/hooks/useLocaleByteConverter.hook';
 
 interface FileInputProps
   extends Omit<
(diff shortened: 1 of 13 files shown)


packages/manager/apps/pci-object-storage/src/components/file-input/FileInput.component.tsx
```diff
@@ -3,8 +3,9 @@ import { Upload, X } from 'lucide-react';
 import { Button } from '@datatr-ux/uxlib';
 import clsx from 'clsx';
 import { useTranslation } from 'react-i18next';
-import { octetConverter } from '@/lib/bytesHelper';
+import { localeBytesConverter, octetConverter } from '@/lib/bytesHelper';
```
jperchoc: Is `octetConverter` still used ?

packages/manager/apps/pci-object-storage/src/pages/object-storage/storage/_components/StorageListColumns.component.tsx
```diff
@@ -18,9 +18,10 @@ import storages, {
   FormattedStorage,
   ObjectStorageTypeEnum,
 } from '@/types/Storages';
-import { octetConverter } from '@/lib/bytesHelper';
+import { localeBytesConverter } from '@/lib/bytesHelper';
 import { RegionTypeBadge } from '@/components/region-type-badge/RegionTypeBadge.component';
 import RegionWithFlag from '@/components/region-with-flag/RegionWithFlag.component';
+import { useLocale } from '@/hooks/useLocale';
```
jperchoc: Maybe it would be easier to create a hook `useLocalBytesConverter` that would get the local under the hood. This would reduce the complexity by avoiding using useLocale everywhere:

```
import { useBytes } from '@/hooks/useBytes';
const bytesConverter = useBytes();

...

bytesConverter(row.original?.objectsSize, false, 2)
```

abullet33: Yes, I changed to use a Hook that manage everything.

Emiri-i (approved):

PR #20731, opened 2025-12-01
fix(veeam): update service and product link in onboarding page

sefiani-ovh:
ref: #MANAGER-20367


packages/manager/apps/veeam-backup/src/veeam-backup.config.ts
```diff
@@ -1,11 +1,13 @@
 export const appName = 'veeam-backup';
 
-export const productFullName = 'Managed Veeam for VMware Cloud Director';
+export const productFullName = 'Managed Veeam for Public VCF as-a-Service';
 export const VMWARE_CLOUD_DIRECTOR_PRODUCT_NAME = 'Public VCF as-a-Service';
 export const productName = 'Managed Veeam for VCD';
 
 export const vcdOrganizationAppName = 'hpc-vmware-public-vcf-aas';
 
+export const serviceName = 'Public VMware Cloud Foundation as-a-Service';
```
pauldkn, line 9: It might be better to specify it's the VCF serviceName, to remove any confusion (as it's in veeam-backup µapp)
```suggestion
export const vcfServiceName = 'Public VMware Cloud Foundation as-a-Service';
```

pauldkn (changes requested):
It should not target `project/vcf-edge-gateway` branch.
It's related to backup so it could be an independant PR

PR #20750, opened 2025-12-02
fix(tracking): add additional data for page tracking

seven-amid:
## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/core/shell-client/src/hooks/useOvhTracking.ts
+++ b/packages/manager/core/shell-client/src/hooks/useOvhTracking.ts
@@ -8,14 +8,17 @@ import { ShellContext, TrackingContextParams } from '../ShellContext';
 
 export const getLevel2TrackingParam = (universe: string) => {
   const result = Object.keys(AT_INTERNET_LEVEL2).filter((element) =>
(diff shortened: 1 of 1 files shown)


packages/manager/core/shell-client/src/hooks/useOvhTracking.ts
```diff
@@ -49,9 +52,14 @@ export enum ButtonType {
   tab = 'go-to-tab',
 }
 
+type AdditionalData = {
+  pci_project_mode?: 'full' | 'discovery' | '';
```
anooparveti: **nitpick:** If its additional data in a common package, we should not pollute it with PCI references. Maybe define it as `Record<string, string>`

helabenkhalfallah (approved):

packages/manager/core/shell-client/src/hooks/useOvhTracking.ts
seven-amid: Done

aboungnaseng-ovhcloud (approved):

anooparveti (approved):

PR #20756, opened 2025-12-02
Feat/backup agent disabled actions

tibs245:
## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/modules/backup-agent/src/pages/services/listing/delete/DeleteTenant.page.tsx
+++ b/packages/manager/modules/backup-agent/src/pages/services/listing/delete/DeleteTenant.page.tsx
@@ -5,21 +5,28 @@ import { useNavigate, useSearchParams } from 'react-router-dom';
 import { useTranslation } from 'react-i18next';
 
 import { ODS_MODAL_COLOR } from '@ovhcloud/ods-components';
-import { OdsText } from '@ovhcloud/ods-components/react';
+import { OdsMessage, OdsText } from '@ovhcloud/ods-components/react';
 
 import { NAMESPACES } from '@ovh-ux/manager-common-translations';
+import { useFeatureAvailability } from '@ovh-ux/manager-module-common-api';
 import { Modal, useNotifications } from '@ovh-ux/manager-react-components';
 
 import { BACKUP_AGENT_NAMESPACES } from '@/BackupAgent.translations';
 import { useDeleteVSPCTenant } from '@/data/hooks/tenants/useDeleteTenant';
 import { useVSPCTenantsMocks } from '@/data/hooks/tenants/useVspcTenants';
+import { useGetFeaturesAvailabilityNames } from '@/hooks/useGetFeatureAvailability';
 
 export default function DeleteTenantPage() {
+  const [searchParams] = useSearchParams();
   const { t } = useTranslation([BACKUP_AGENT_NAMESPACES.SERVICE_LISTING, NAMESPACES.ACTIONS]);
   const navigate = useNavigate();
   const closeModal = () => navigate('..');
   const { addSuccess, addError } = useNotifications();
-  const [searchParams] = useSearchParams();
+  const { DELETE_TENANT: deleteAgentFeatureName } = useGetFeaturesAvailabilityNames([
+    'DELETE_TENANT',
+  ]);
+  const { data: featureAvailabilities } = useFeatureAvailability([deleteAgentFeatureName]);
+  const isDeleteTenantFeatureAvailable = featureAvailabilities?.[deleteAgentFeatureName] ?? false;
 
   const { data: tenants } = useVSPCTenantsMocks();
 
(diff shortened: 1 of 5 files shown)


package.json
pauldkn, line 1: I don't think this should be committed

packages/manager/modules/backup-agent/src/hooks/useGetFeatureAvailability.ts
```diff
@@ -0,0 +1,14 @@
+import { useContext } from 'react';
+
+import { BackupAgentContext } from '@/BackupAgent.context';
+import { FEATURE_AVAILABILITY } from '@/module.constants';
+
+export type UseGetFeaturesAvailabilitiesNameParams = (keyof typeof FEATURE_AVAILABILITY)[];
+
+export const useGetFeaturesAvailabilitiesName = (
```
pauldkn: I think it would make more sense
```suggestion
export const useGetFeaturesAvailabilityNames = (
```

packages/manager/modules/backup-agent/src/hooks/useGetFeatureAvailability.ts
```diff
@@ -0,0 +1,14 @@
+import { useContext } from 'react';
+
+import { BackupAgentContext } from '@/BackupAgent.context';
+import { FEATURE_AVAILABILITY } from '@/module.constants';
+
+export type UseGetFeaturesAvailabilitiesNameParams = (keyof typeof FEATURE_AVAILABILITY)[];
```
pauldkn: I think it would make more sense
```suggestion
export type UseGetFeaturesAvailabilityNamesParams = (keyof typeof FEATURE_AVAILABILITY)[];
```

packages/manager/modules/backup-agent/src/pages/services/dashboard/Dashboard.page.tsx
```diff
@@ -31,33 +31,35 @@ export default function DashboardPage() {
...
+              >
+                <OdsTab isSelected={tab.isActive}>{t(tab.title)}</OdsTab>
+              </NavLink>
+            ))}
+          </OdsTabs>
+        }
+      >
+        <Suspense fallback={null}>
+          <Outlet />
+        </Suspense>
```
pauldkn, line 61: Why do you need to wrap the `outlet` with a `suspense`?

packages/manager/modules/backup-agent/src/test-utils/testWrapperProviders.tsx
```diff
@@ -46,3 +48,14 @@ export const addShellContextProvider = async (providers: TestProvider[]) => {
     <ShellContext.Provider value={context}>{children}</ShellContext.Provider>
   ));
 };
+
+export const addAppProvider = (
```
pauldkn: Maybe `addContextProvider` or `addAppContextProvider` would be clearer

packages/manager/modules/backup-agent/src/pages/services/dashboard/Dashboard.page.tsx
tibs245: Problem of margin

packages/manager/modules/backup-agent/src/BackupAgent.context.tsx
```diff
@@ -8,6 +8,6 @@ export interface BackupAgentProviderProps {
 }
 
 export const BackupAgentContext = createContext<BackupAgentProviderProps>({
-  appName: 'Backup Agent',
+  appName: 'backup-agent',
```
mathieumousnier: This context could have just been some initialization variables.

Context is for "reactive" state: 
1. things that changes
2. and should trigger re-renders when they change

This is more global variables for your module, not global state.

packages/manager/modules/backup-agent/src/hooks/useGetFeatureAvailability.ts
```diff
@@ -0,0 +1,12 @@
+import { useContext } from 'react';
+
+import { BackupAgentContext } from '@/BackupAgent.context';
+import { FEATURE_AVAILABILITY } from '@/module.constants';
+
+export type UseGetFeaturesAvailabilityNamesParams = (keyof typeof FEATURE_AVAILABILITY)[];
```
mathieumousnier:
```suggestion
export type Feature = (keyof typeof FEATURE_AVAILABILITY)[];
```

packages/manager/modules/backup-agent/src/pages/services/listing/delete/DeleteTenant.page.tsx
```diff
@@ -5,21 +5,29 @@ import { useNavigate, useSearchParams } from 'react-router-dom';
...
+  const [searchParams] = useSearchParams();
   const { t } = useTranslation([BACKUP_AGENT_NAMESPACES.SERVICE_LISTING, NAMESPACES.ACTIONS]);
   const navigate = useNavigate();
   const closeModal = () => navigate('..');
   const { addSuccess, addError } = useNotifications();
-  const [searchParams] = useSearchParams();
+  const [deleteAgentFeatureName] = useGetFeaturesAvailabilityNames(['DELETE_TENANT']);
+  const { data: featureAvailabilities } = useFeatureAvailability([
+    deleteAgentFeatureName as string,
+  ]);
```
mathieumousnier: This "as string" is a bit concerning :)

When you do : 
```ts
const [deleteAgentFeatureName] = useGetFeaturesAvailabilityNames(['DELETE_TENANT']);
```

The resulting `deleteAgentFeatureName]` is `string | undefined`, so the type is not strict.

This `useGetFeaturesAvailabilityNames` function is very hard to type with its model of taking an array as input, and giving an array as output.
If you treat each names independantly it's a lot easier, this spits a "string" each time.
```ts
const deleteVaultFeatureName = useGetFeatureAvailabilityName('DELETE_VAULT');
```

There might be a way to type useGetFeatureAvailabilityName with arrays but i don't see it right now

tibs245: Me too. I noted this point for later

mathieumousnier: You mean you will fix the tech debt later ?

tibs245: Yes. But I fixed with a use of object key/value. It's fast solution to avoid this problem

PR #20764, opened 2025-12-02
fix(container): update live-chat URLs to load on ovhcloud.com

anooparveti:
## Description

Ticket Reference: #...


anooparveti:
Bypassing for hotfix.

PR #20784, opened 2025-12-03
Feat/add download agent

tibs245:
## Description

Ticket Reference: #BKP-663

### DIFF
--- a/packages/manager/apps/bmc-backup-agent-baremetal/src/components/onboarding/onboardingLayout/OnboardingLayout.component.tsx
+++ b/packages/manager/apps/bmc-backup-agent-baremetal/src/components/onboarding/onboardingLayout/OnboardingLayout.component.tsx
@@ -45,7 +45,7 @@ const OnboardingLayoutButton: React.FC<OnboardingLayoutButtonProps> = ({
     return null;
   }
   return (
-    <div className="flex sm:pt-8 xs:pt-2.5 flex-row items-center space-x-4 justify-center">
+    <div className="flex flex-row items-center justify-center space-x-4 xs:pt-2.5 sm:pt-8">
       <Link id={tooltipId} to={!isOrderDisabled && !isOrderLoading ? orderHref : ''}>
         <OdsButton
           size={ODS_BUTTON_SIZE.md}
@@ -90,14 +90,13 @@ export const OnboardingLayout: React.FC<OnboardingLayoutProps> = ({
 }) => {
   const { ...imgProps } = img;
   return (
-    <div className="flex flex-col mx-auto sm:px-10">
+    <div className="mx-auto flex flex-col sm:px-10">
       {!hideHeadingSection && (
         <section className="flex flex-col items-center">
           {(img?.src || placeholderImg) && (
-            <div className="flex justify-center pt-8 max-h-28 w-full">
+            <div className="flex max-h-28 w-full justify-center pt-8">
               <img
                 {...imgProps}
-                className="max-h-[150px]"
                 src={img?.src ?? placeholderImg}
                 alt=""
                 width={img?.width}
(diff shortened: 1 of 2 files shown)


packages/manager/modules/backup-agent/src/module.constants.ts
```diff
@@ -7,3 +7,9 @@ export const FEATURE_AVAILABILITY = Object.freeze({
   DELETE_TENANT: 'deleteTenant',
   DELETE_VAULT: 'deleteVault',
 });
+
+export const OS = {
+  LINUX: 'Linux',
+  WINDOWS: 'Windows',
+  MAC: 'Mac',
+};
```
mathieumousnier, line 15: This key:values pattern for constants always look a bit like JAVA to me.

If you change this for :
```ts
export const OS_VALUES = ['Linux', 'Windows', 'Mac'] as const;
export type OS = (typeof OS_VALUES)[number];
```

This will simplify and beautify all your code.

1. You will have an `OS` type way cleaner than using `keyof typeof OS` everywhere.

2. You can avoid this `as` in AgentDownload.page.tsx
```tsx
// AgentDownload.page.tsx line 60
{Object.keys(OS).map((osKey) => (
  <option key={osKey} value={osKey}>
    {OS[osKey as keyof typeof OS]}
  </option>
))}
// Replaced by this
{OS_VALUES.map((osKey) => (
  <option key={osKey} value={osKey}>
    {osKey}
  </option>
))}
```
3. You get typescript validation on direct values here : 
```ts
// getDownloadLinkAgent.ts
select: (data) => {
  switch (os) {
    case 'LINUX':
      return data.linuxUrl;
    case 'WINDOWS':
      return data.windowsUrl;
    case 'MAC':
      return data.macUrl;
    default:
      return data.linuxUrl;
  }
},
// Replaced by this
select: (data) => {
  switch (os) {
    case 'Linux':
      return data.linuxUrl;
    case 'Windows':
      return data.windowsUrl;
    case 'Mac':
      return data.macUrl;
    default:
      return data.linuxUrl;
  }
},
```

packages/manager/modules/backup-agent/src/pages/services/dashboard/agent/download/__tests__/AgentDownload.test.tsx
```diff
@@ -0,0 +1,168 @@
...
+          children: React.ReactNode;
+          label: string;
+          isLoading: boolean;
+          isDisabled: boolean;
+        }) => (
+          <button data-is-loading={isLoading} data-is-disabled={isDisabled} {...props}>
+            {label}
+          </button>
+        ),
+      ),
```
mathieumousnier, line 40: This is more a general remark and not something to fix here, but don't you feel like you repeat those mocks a lot ?
There is ways to centralise this, it would be less code to write everywhere, but above all less code to maintain

tibs245: Yes it's noted on my side to improve this. But effectively I prefer take the time to find the better method to do that

packages/manager/modules/backup-agent/src/module.constants.ts
mathieumousnier: Oh, I just had a case where I need to create some mapping object like the OS one to get centralized values for tests.
The issue in tests is that what you put in expect().toBe('bla-bla') is not typed...

But in the code, using a union type with the values is most of the time the most straight-forward solution

tibs245: It's already a union type via keyof typeof. I create a type alias for lisibility. But I prefer keep label displayed separatly

mathieumousnier: This issue here is that your union type is on the keys, not the values.
So you still have to use this "as" casting here, and you need to map on keys, to finally get the values:
```ts
  {Object.keys(OS_LABELS).map((osKey) => (
    <option key={osKey} value={osKey}>
      {OS_LABELS[osKey as OS]}
    </option>
  ))}
```
This is going left, then right instead of going straight to the point.

But okay

mathieumousnier (approved):

PR #20792, opened 2025-12-03
fix(account-creation): add query params persistence

JacquesLarique:
## Description

Add `onsuccess` query param persistence when redirection user from the old version to the new version

Ticket Reference: #MANAGER-20472

### DIFF
--- a/packages/manager/apps/sign-up/src/routing.js
+++ b/packages/manager/apps/sign-up/src/routing.js
@@ -16,9 +16,10 @@ export const state = {
   url: '/?lang&ovhSubsidiary&ovhCompany&callback&onsuccess',
   component: signupFormComponent.name,
   resolve: {
-    getRedirectLocation: /* @ngInject */ ($location) => (nic) => {
-      const { callback, onsuccess } = $location.search();
-
+    queryParams: /* @ngInject */ ($location) => $location.search(),
+    callback: /* @ngInject */ (queryParams) => queryParams.callback,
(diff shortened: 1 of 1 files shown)


packages/manager/apps/sign-up/src/routing.js
```diff
@@ -65,8 +66,12 @@ export const state = {
         .getSsoAuthPendingPromise()
         .then(() => ssoAuthentication.user),
 
-    cancelStep: /* @ngInject */ ($location, ssoAuthentication) => () => {
-      ssoAuthentication.logout($location.search().onsuccess);
+    cancelStep: /* @ngInject */ (
+      $location,
```
sachinrameshn:
```suggestion
    cancelStep: /* @ngInject */ (
```

sachinrameshn (commented):
https://github.com/ovh/manager/pull/20792/files#diff-66e19f464f40fd0292ba24f525c4b6aa4f3ec5442e2817692703a19b58a19607L229
Can we modify here as below by injecting queryParams directly:

 subsidiary: /* @ngInject */ (queryParams, coreConfig) =>
      queryParams.ovhSubsidiary || coreConfig.getUser().ovhSubsidiary,

packages/manager/apps/sign-up/src/routing.js
JacquesLarique: Fixed

PR #20804, opened 2025-12-03
feat(pci-block-storage): delete snapshots before retyping

aTurmo:
ref: #TAPC-4569

### DIFF
--- a/packages/manager/apps/pci-block-storage/src/api/hooks/operation/useOperationPolling.ts
+++ b/packages/manager/apps/pci-block-storage/src/api/hooks/operation/useOperationPolling.ts
@@ -1,0 +1,53 @@
+import { Query, useQuery, UseQueryOptions } from '@tanstack/react-query';
+import { useEffect } from 'react';
+import { getOperation, TOperation } from '@/api/data/operation';
+
+const getOperationQueryKey = (projectId: string, operationId: string) => [
+  'operation',
+  projectId,
+  operationId,
+];
+
+export type TUseOperationPollingQueryOptions = Omit<
+  UseQueryOptions<TOperation>,
+  'queryFn' | 'queryKey'
+>;
+
+const DEFAULT_INTERVAL_IN_MS = 1000;
+
+const defaultQueryOptions: TUseOperationPollingQueryOptions = {
+  refetchIntervalInBackground: true,
+  gcTime: 0,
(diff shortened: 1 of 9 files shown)


packages/manager/apps/pci-block-storage/src/api/hooks/useSnapshot.ts
```diff
@@ -29,24 +38,93 @@ export const useVolumeSnapshots = (projectId: string, volumeId: string) => {
...
   volumeId,
-  snapshotsIds,
 }: DeleteVolumeSnapshotsProps) => {
+  const { data: volume } = useVolume(projectId, volumeId);
+
+  const [deleteOperation, setDeleteOperation] = useState<TOperation>(null);
+  const [hasReachMaxCalls, setHasReachMaxCalls] = useState(false);
+  const [hadAnError, setHadAnError] = useState(false);
+
   const mutation = useMutation({
```
Taelar: thought: This const could use a more specific name, like "snapshotDeleteMutation" so it's easier to understand what it does

packages/manager/apps/pci-block-storage/src/api/hooks/operation/useOperationPolling.ts
```diff
@@ -0,0 +1,53 @@
...
+    ...defaultQueryOptions,
+    ...queryOptions,
+  });
+
+  useEffect(() => {
+    if (isLoading || !operationId) return undefined;
+    if (error) {
+      return onError?.();
+    }
+    return onSuccess?.(data);
```
ppprevost: [question(blocking)](https://pullpo.io/cc?l=question&d=blocking) 
why mixing useQuery and useMutation logic ? 
it creates ambiguity : 
When to call onSuccess ? what happen if data changes ? 
Why retuning undefined if it's loading state?
You forget operationId inside the useEffect, plus onSuccess and onError.
In my opinion you loose predictability when you using callback instead of classic useQuery return.

packages/manager/apps/pci-block-storage/src/api/hooks/operation/useOperationPolling.ts
```diff
@@ -0,0 +1,53 @@
...
+  }, [data, error, isLoading]);
```
aTurmo: ajouter le operationId

PR #20805, opened 2025-12-03
feat(telecom): replace action menu in voip page  with two buttons

amine-daoussi:
ref: #PRDCOL-277

## Description

In Telecom > voip & fax, replace the action menu with 2 buttons (order and manage my repayments)

image

### DIFF
--- a/packages/manager/apps/telecom/src/app/telecom/telephony/telephony.html
+++ b/packages/manager/apps/telecom/src/app/telecom/telephony/telephony.html
@@ -33,20 +33,20 @@
         data-columns="$ctrl.columnsConfig"
     >
         <oui-datagrid-topbar>
-            <oui-action-menu
-                text="{{:: 'telephony_table_header_action' | translate }}"
+            <oui-button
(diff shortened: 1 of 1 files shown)


packages/manager/apps/telecom/src/app/telecom/telephony/telephony.html
```diff
@@ -33,20 +33,18 @@
...
+                class="btn btn-default"
+                data-ng-click="$ctrl.gotoOrder()"
+                data-translate="telephony_order">
+            </button>
+            <button
+                type="button"
+                class="btn btn-default"
+                data-ng-click="$ctrl.goToTelephonyRepayments()"
+                data-translate="telephony_table_action_item_telephony_repayments">
+            </button>
```
ghyenne: I think it should to have one primary button (order) and one secondary button (repayments).
You can also use the ui-kit https://ovh.github.io/ovh-ui-kit/?path=/docs/design-system-components-buttons-webcomponent-secondary--default

PR #20845, opened 2025-12-04
build(pci-kubernetes): migrate to pnpm

Taelar:
## Description

Ticket Reference: #TAPC-5407


packages/manager/apps/pci-kubernetes/package.json
```diff
@@ -32,21 +32,21 @@
     "@ovhcloud/ods-components": "17.2.2",
     "@ovhcloud/ods-react": "^19.0.2",
     "@ovhcloud/ods-theme-blue-jeans": "17.2.2",
-    "@ovhcloud/ods-themes": "^19.0.2",
-    "@tanstack/react-query": "^5.51.21",
-    "@tanstack/react-table": "^8.17.3",
+    "@ovhcloud/ods-themes": "18.6.2",
```
fredericvilcot, line 35: Why ODS themes has been downgraded to legacy 18 version? 

@helabenkhalfallah I think we should do the same thing for other migrations: update versions to what is really used in the monolith?

packages/manager/apps/pci-kubernetes/vitest.config.js
```diff
@@ -22,16 +23,11 @@ export default mergeConfig(
           ...defaultExcludedFiles,
           // App-specific exclusions (not in shared config):
           'vite-*.ts',
```
fredericvilcot, line 25: Are these exclusions all inherited from `defaultExcludedFiles` ?

helabenkhalfallah:
The PR is not a good migration for pnpm, it changes too many files that's supposed to not to do, we should change only these:
https://github.com/ovh/manager/pull/20812 

To keep stability and predictability.

Taelar:
Migration done in #20878

PR #20848, opened 2025-12-04
fix(okms): reorder credentials validity options

mathieumousnier:
Ticket Reference: #MANAGER-16442


packages/manager/apps/okms/src/common/utils/ods/odsSelect.ts
```diff
@@ -0,0 +1,13 @@
+// This 2 functions are used to set string values to the ODS 18 Select component
```
tibs245, line 1: No little TU ?

PR #20854, opened 2025-12-04
fix(muk): correct table container ref in datagrid

aboungnaseng-ovhcloud:
## Description

- correct table container ref in datagrid
- fix issues of build

Ticket Reference: #...

### DIFF
--- a/packages/manager-ui-kit/src/components/error/Error.component.tsx
+++ b/packages/manager-ui-kit/src/components/error/Error.component.tsx
@@ -13,7 +13,6 @@ import {
   Text,
 } from '@ovhcloud/ods-react';
 
-import { Environment } from '@ovh-ux/manager-config';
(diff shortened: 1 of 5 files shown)


helabenkhalfallah (dismissed):

packages/manager-ui-kit/src/components/error/Error.component.tsx
```diff
@@ -40,7 +41,7 @@ export const Error = ({
 
   useEffect(() => {
     const env = shell?.environment?.getEnvironment();
-    env?.then((response) => {
+    env?.then((response: Environment) => {
```
anooparveti: **nitpick:** It need not be `Environment`. Error page can be shown on any API failure.

aboungnaseng-ovhcloud: 👍

helabenkhalfallah (approved):

PR #20859, opened 2025-12-05
fix(communication): resolve issue with the link

sachinrameshn:
ref: #MANAGER-20465

## Description

Ticket Reference: #...


sachinrameshn:
> Why adding the app to the eslintignore ?

Since communication app is not adapted to static kit rules, eslint is failing during commit. In order to by pass this I added the app to the eslintignore. But this as to be removed during static kit adaption.

anooparveti (approved):

PR #20904, opened 2025-12-08
feat(account-creation): change logo

achrafHafedh:
ref: #MANAGER-20304

## Description

Ticket Reference: #MANAGER-20304


packages/manager/apps/account-creation/src/components/header/Header.component.tsx
```diff
@@ -9,7 +9,7 @@ import {
   usePageTracking,
 } from '@ovh-ux/manager-react-shell-client';
 
-import ovhCloudLogo from '@/assets/logo-ovhcloud.png';
+import OvhCloudLogo from '@/assets/logo-ovhcloud.svg?url';
```
sachinrameshn, line 12: Why we are adding "url" at the end ?

achrafHafedh: The ?url at the end is a Vite hint that tells the bundler: Import this file as a URL string instead of as an SVG component/module (This forces Vite to return a public URL to the file)
We have the same thing in several places, such as in account-creation layout.tsx, in the container HostedPrivateCloudSidebar.tsx, and in other parts of the code as well

JacquesLarique (approved):

PR #20943, opened 2025-12-09
fix(okms): broken download service key from list

mathieumousnier:
ref: #MANAGER-20360

### DIFF
--- a/packages/manager/apps/okms/src/modules/key-management-service/components/listing/ListingCells.tsx
+++ b/packages/manager/apps/okms/src/modules/key-management-service/components/listing/ListingCells.tsx
@@ -1,6 +1,6 @@
 import { useNavigate } from 'react-router-dom';
 
-import useServiceKeyActionsList from '@key-management-service/hooks/service-key/useServiceKeyActionsList';
+import { useServiceKeyActionsList } from '@key-management-service/hooks/service-key/useServiceKeyActionsList';
 import { useServiceKeyTypeTranslations } from '@key-management-service/hooks/service-key/useServiceKeyTypeTranslations';
 import { useFormattedDate } from '@key-management-service/hooks/useFormattedDate';
 import { OKMS } from '@key-management-service/types/okms.type';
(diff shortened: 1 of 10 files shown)


packages/manager/apps/okms/src/modules/key-management-service/hooks/service-key/useServiceKeyDownload.tsx
rjamet-ovh, line 1: this hook file should be named `useServiceKeyDownload.tsx`

mathieumousnier: done

packages/manager/apps/okms/src/modules/key-management-service/components/listing/ListingCells.tsx
```diff
@@ -67,14 +67,22 @@ export const DatagridStatus = (props: OkmsServiceKey) => {
 };
 
 export const DatagridServiceKeyActionMenu = (serviceKey: OkmsServiceKey, okms: OKMS) => {
-  const actionList = useServiceKeyActionsList(okms, serviceKey, true);
+  const actions = useServiceKeyActionsList(okms, serviceKey, 'list');
+
+  // Add id and remove icon
+  const actionsWithId = actions.map((action, index) => ({
+    ...action,
+    id: index + 1,
```
tibs245: Why `+ 1` ? If it's to use as key it's not possible take more stable attribut as reference ?

mathieumousnier: It would be nice.
But this is a key for the ActionMenuItem component from MRC, which only accepts numbers 🤪

```ts
export interface ActionMenuItem {
    id: number;
    rel?: string;
    href?: string;
    download?: string;
    target?: string;
    onClick?: () => void;
```

mathieumousnier: +1 might not be useful, I could blame IA for writing this but I didn't corrected it :)

mathieumousnier: I removed the +1

PR #20951, opened 2025-12-09
feat(pci-block-storage): handle force refetching linked instances and snapshots

aTurmo:
ref: #TAPC-5456

### DIFF
--- a/packages/manager/apps/pci-block-storage/src/api/hooks/helpers.ts
+++ b/packages/manager/apps/pci-block-storage/src/api/hooks/helpers.ts
@@ -9,13 +9,8 @@ export type TForceReloadUseQueryParams = Pick<
   'refetchOnMount' | 'refetchOnWindowFocus' | 'refetchOnReconnect'
 >;
 
-export const getForceReloadUseQueryParams = (
-  forceReload?: boolean,
-): TForceReloadUseQueryParams =>
-  forceReload
-    ? {
-        refetchOnMount: 'always',
-        refetchOnWindowFocus: 'always',
-        refetchOnReconnect: 'always',
-      }
-    : {};
+export const forceReloadUseQueryOptions: TForceReloadUseQueryParams = {
+  refetchOnMount: 'always',
+  refetchOnWindowFocus: 'always',
+  refetchOnReconnect: 'always',
+};
--- a/packages/manager/apps/pci-block-storage/src/api/hooks/useInstance.ts
+++ b/packages/manager/apps/pci-block-storage/src/api/hooks/useInstance.ts
@@ -7,7 +7,7 @@ import {
   selectAttachedInstances,
 } from '@/api/select/instances';
 import {
-  getForceReloadUseQueryParams,
+  forceReloadUseQueryOptions,
   TApiHookOptions,
 } from '@/api/hooks/helpers';
 
@@ -48,13 +48,11 @@ export const useAttachedInstances = (
     [volume],
   );
 
-  const forceReloadOptions = getForceReloadUseQueryParams(options.forceReload);
-
   return useQuery({
     queryKey: getInstancesQueryKey(projectId, volume?.region),
     queryFn: () => getInstancesByRegion(projectId, volume.region),
     enabled: !!volume,
     select,
-    ...forceReloadOptions,
+    ...(options.forceReload && forceReloadUseQueryOptions),
   });
 };
--- a/packages/manager/apps/pci-block-storage/src/api/hooks/useSnapshot.ts
+++ b/packages/manager/apps/pci-block-storage/src/api/hooks/useSnapshot.ts
@@ -16,7 +16,7 @@ import {
   isOperationInProgress,
 } from '@/api/select/operation';
 import {
-  getForceReloadUseQueryParams,
+  forceReloadUseQueryOptions,
   TApiHookOptions,
 } from '@/api/hooks/helpers';
 
@@ -39,14 +39,12 @@ export const useVolumeSnapshots = (
     [volume],
   );
 
-  const forceReloadOptions = getForceReloadUseQueryParams(options.forceReload);
-
   return useQuery<TVolumeSnapshot[]>({
     queryKey: getVolumeSnapshotsQueryKey(projectId, volumeId),
     queryFn: () => getSnapshotsByRegion(projectId, volume?.region),
     enabled: !!volume,
     select,
-    ...forceReloadOptions,
+    ...(options.forceReload && forceReloadUseQueryOptions),
   });
 };
 
--- a/packages/manager/apps/pci-block-storage/src/pages/retype/Retype.page.tsx
+++ b/packages/manager/apps/pci-block-storage/src/pages/retype/Retype.page.tsx
@@ -38,13 +38,13 @@ const RetypePage = () => {
     data: instances,
     isPending: isInstancesPending,
     isFetching: isInstancesFetching,
-  } = useAttachedInstances(projectId, volumeId, true);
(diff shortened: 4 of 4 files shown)


packages/manager/apps/pci-block-storage/src/api/hooks/useInstance.ts
```diff
@@ -32,18 +32,34 @@ export const useAttachableInstances = (projectId: string, volumeId: string) => {
   });
 };
 
-export const useAttachedInstances = (projectId: string, volumeId: string) => {
+export const useAttachedInstances = (
+  projectId: string,
+  volumeId: string,
+  forceReload = false,
```
fredericvilcot: We could prevent from adding too much arguments, so I'll suggest to keep the `projectId` and the `volumeId` as main arguments, and to add a configuration / ctx object as last parameter to handle extra context args

packages/manager/apps/pci-block-storage/src/api/hooks/useSnapshot.ts
```diff
@@ -19,19 +23,35 @@ export const getVolumeSnapshotsQueryKey = (
...
+) => {
   const { data: volume } = useVolume(projectId, volumeId);
 
   const select = useMemo(
     () => (volume ? selectSnapshotForVolume(volumeId) : undefined),
     [volume],
   );
 
-  return useQuery({
+  const forceReloadOption: Pick<
```
fredericvilcot: Please don't repeat yourself, and make a single function to return right options :)

packages/manager/apps/pci-block-storage/src/pages/retype/Retype.page.tsx
```diff
@@ -37,15 +37,21 @@ const RetypePage = () => {
...
-  );
+  const {
+    data: snapshots,
+    isPending: isSnapshotsPending,
+    isFetching: isSnapshotsFetching,
+  } = useVolumeSnapshots(projectId, volumeId, true);
 
   const isPending =
-    isCatalogPending || isInstancesPending || isSnapshotsPending;
+    isCatalogPending ||
```
lolottetheclash, line 50: I can see you're using isFetching & isPending for snapshots & instances, no need to use isFetching for catalog?

packages/manager/apps/pci-block-storage/src/api/hooks/useCatalogWithPreselection.tsx
```diff
@@ -35,7 +35,11 @@ export const useCatalogWithPreselection = (
     { data: catalogData, ...restCatalogQuery },
   ] = useQueries({
     queries: [
-      getVolumeQuery(projectId, volumeId),
+      {
+        ...getVolumeQuery(projectId, volumeId),
+        refetchOnWindowFocus: true,
+        refetchOnMount: true,
+      },
```
lolottetheclash: no need for `refetchOnReconnect` here?

packages/manager/apps/pci-block-storage/src/api/hooks/useInstance.ts
lolottetheclash: This also avoids using a Boolean as a direct argument, making the function call more readable with `callFn({forceReload:true})` rather than `callFn(true)`.

lolottetheclash (commented):
just a question :)

packages/manager/apps/pci-block-storage/src/pages/retype/Retype.page.tsx
aTurmo: Good question but no because we are not re-fetching catalog.

packages/manager/apps/pci-block-storage/src/api/hooks/useCatalogWithPreselection.tsx
aTurmo: We want to be sure to have the freshest data to I think it's better to add this option

packages/manager/apps/pci-block-storage/src/api/hooks/useInstance.ts
aTurmo: done

packages/manager/apps/pci-block-storage/src/api/hooks/useSnapshot.ts
aTurmo: done

packages/manager/apps/pci-block-storage/src/api/hooks/useInstance.ts
aTurmo: done

packages/manager/apps/pci-block-storage/src/api/hooks/useCatalogWithPreselection.tsx
aTurmo: Ok i got confused, we actually don't need to refresh here

packages/manager/apps/pci-block-storage/src/api/hooks/helpers.ts
```diff
@@ -0,0 +1,21 @@
...
+};
+
+export type TForceReloadUseQueryParams = Pick<
+  UseQueryOptions,
+  'refetchOnMount' | 'refetchOnWindowFocus' | 'refetchOnReconnect'
+>;
+
+export const getForceReloadUseQueryParams = (
+  forceReload?: boolean,
+): TForceReloadUseQueryParams =>
```
fredericvilcot: Reagrding `getForceReloadUseQueryParams` naming, it suggests that forceReload args are always available, but if argument is `false` or `undefined` , it's not the case..

So I'll suggest to make the function predictable and always return a `TForceReloadUseQueryParams` objects, with no parameter.

Then, at usage, you can do something like:

```ts
return useQuery({
    queryKey: getInstancesQueryKey(projectId, volume?.region),
    queryFn: () => getInstancesByRegion(projectId, volume.region),
    enabled: !!volume,
    select,
    ...(options.forceReload && getForceReloadUseQueryParams()),
  });



```

packages/manager/apps/pci-block-storage/src/pages/retype/Retype.page.tsx
```diff
@@ -37,15 +37,21 @@ const RetypePage = () => {
   const {
     data: instances,
     isPending: isInstancesPending,
-  } = useAttachedInstances(projectId, volumeId);
+    isFetching: isInstancesFetching,
+  } = useAttachedInstances(projectId, volumeId, true);
```
fredericvilcot: Should be invoked with an object as third parameter

aTurmo: olala, not on point today ><

fredericvilcot: BTW it throws an error with your actual code ;)

lolottetheclash: as yesterday, you should have some errors in your IDE, strange no?

packages/manager/apps/pci-block-storage/src/api/hooks/helpers.ts
aTurmo: done

packages/manager/apps/pci-block-storage/src/pages/retype/Retype.page.tsx
aTurmo: fixed

aTurmo: As we say the problem is between the screen and the chair. Don't blame the best IDE ^^

PR #20971, opened 2025-12-10
feat(object-storage): add service type parameter on rclone route

lionel95200x:
ref: #DATATR-2712

## Description

- Add service type parameter on rclone route

Ticket Reference: #DATATR-2712


packages/manager/apps/pci-object-storage/src/data/api/user/user.api.ts
```diff
@@ -37,17 +37,19 @@ export const getUserPolicy = async ({ projectId, userId }: UserProps) =>
...
   projectId,
   userId,
   region,
+  service,
 }: UserPropsWithRegion) =>
   apiClient.v6.get<user.Rclone>(
     `/cloud/project/${projectId}/user/${userId}/rclone?region=${encodeURIComponent(
       region,
-    )}`,
+    )}&service=${encodeURIComponent(service)}`,
```
AkdM: Can't we use this instead here? (not tested)

```suggestion
const region = encodeURIComponent(region);
const service = encodeURIComponent(service);
apiClient.v6.get<user.Rclone>(
`/cloud/project/${projectId}/user/${userId}/rclone`, {
  params: {
    region,
    service,
  }
}
```

PR #20972, opened 2025-12-10
fix(billing): fix difference between interval of commitment option

MaximeBajeux:
## Description

In the current commitment page, some vps with Plesk licence have a price 12 times upper than they should. The thing is the upfront12 option has an interval of 12 that differ from the service 1. So for this reason we check if the option's interval is equal to the service's interval

Ticket Reference: #INC0174363


sachinrameshn:
Managed here : https://github.com/ovh/manager/pull/20988 hence closing this

PR #20975, opened 2025-12-10
fix(okms): reorder versions in secret version selector

mathieumousnier:
## Description

Ticket Reference: #MANAGER-20499


packages/manager/apps/okms/src/modules/secret-manager/pages/drawers/secret-value-drawer/VersionSelector.component.tsx
```diff
@@ -105,3 +90,46 @@ export const VersionSelector = ({
...
+  setSelectedVersion,
+  versions,
+}: VersionSelectorSelectProps) => {
+  // Default to the first version if no default version is provided
+  let defaultVersion: SecretVersion | undefined = versions[0];
+
+  // If a default version is provided, use it
+  if (defaultVersionId) {
+    defaultVersion = versions.find((version) => version.id.toString() === defaultVersionId);
+  }
```
tibs245, line 112: May be you can externalise this logic in a utils : `getVersions(versions, defaultVersionId)`

```ts
const getVersions = (versions, defaultVersionId) => {
     if (defaultVersionId) {
    return versions.find((version) => version.id.toString() === defaultVersionId);
}

return versions[0];
}
```

And you can add TU on this logic

PR #20978, opened 2025-12-10
feat(backup-agent): UX improvements

pauldkn:
## Description

Ticket Reference: #BKP-700

### DIFF
--- a/packages/manager/modules/backup-agent/src/pages/vaults/listing/Listing.page.tsx
+++ b/packages/manager/modules/backup-agent/src/pages/vaults/listing/Listing.page.tsx
@@ -1,4 +1,6 @@
-import React, { Suspense } from 'react';
+import { Suspense } from 'react';
+
+import { Outlet } from 'react-router-dom';
 
 import { useQueryClient } from '@tanstack/react-query';
 import { useTranslation } from 'react-i18next';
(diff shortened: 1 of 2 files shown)


packages/manager/modules/backup-agent/src/pages/vaults/dashboard/general-information/_components/subscription-tile/SubscriptionTile.component.tsx
```diff
@@ -21,13 +21,24 @@ export function SubscriptionTile({ vaultId }: SubscriptionTileProps) {
     BACKUP_AGENT_NAMESPACES.VAULT_DASHBOARD,
   ]);
   const { isLoading } = useBackupVaultDetails({ vaultId });
+  const tooltipId = 'consumption-tooltip';
```
tibs245, line 24: The react hook useId is more safe

```suggestion
  const tooltipId = useId();
```

vovh (approved):

packages/manager/modules/backup-agent/src/pages/vaults/dashboard/general-information/_components/subscription-tile/SubscriptionTile.component.tsx
pauldkn: I was thinking it might be easier in order to add some tests around this tooltip with a predictable id, but we can change it if you prefer to use `useId`

PR #21025, opened 2025-12-12
feat(backup-agent): unmock module with backup services endpoints

tibs245:
## Description

Ticket Reference:  #BKP-704


vovh (approved):

packages/manager/modules/backup-agent/src/data/api/agents/agents.requests.ts
pauldkn, line 1: The create agent request is missing, I don't know if it's better to add it here, in the target PR, or in another PR

packages/manager/modules/backup-agent/src/data/api/vaults/vault.requests.ts
pauldkn, line 1: It seems that the request routes are not correct
On the API console they're listed as `/backupServices/tenant/{backupServicesId}/vault`, for almost every endpoint

PR #21048, opened 2025-12-13
refactor(observability): rename module to metrics to customer

yjaaouane:
## Description

Ticket Reference: #MAOBS-148


packages/manager/modules/metrics-to-customer/src/routes/routes.tsx
```diff
@@ -31,13 +29,9 @@ function lazyRouteConfig<
...
 
 /**
- * @deprecated This export is deprecated. Please use getObservabilityRoute() instead.
+ * @deprecated This export is deprecated. Please use getMetricsRoute() instead.
  */
-export const observabilityRoutes: RouteObject[] = [
-  {
-    path: 'overview',
-    ...lazyRouteConfig(importMetricsOverviewPage),
-  },
```
MarcMonchablon, line 40: Not a blocker, but I guess that it's not a problem that a route was removed since it's deprecated ?

yjaaouane: This route "overview" was added only for testing multiple routes on module, and I forgot to remove it. 
It’s not part of the spec.

PR #21082, opened 2025-12-15
fix(pci-block-storage): block some actions while retyping

aTurmo:
ref: #TAPC-5492


packages/manager/apps/pci-block-storage/src/components/list/Actions.component.spec.tsx
```diff
@@ -7,90 +7,145 @@ import { TVolume } from '@/api/hooks/useVolume';
...
+    expect(getByTestId('actionComponent-edit-button')).toHaveAttribute(
       'href',
-      './delete/1',
+      './1221/edit',
     );
 
     expect(getByTestId('actionComponent-attach-button')).toHaveAttribute(
       'href',
-      './attach/1',
+      './attach/1221',
```
lolottetheclash, line 120: strange urls

aTurmo: will be changed in the future

Taelar (approved):

PR #21105, opened 2025-12-16
feat(pci-kubernetes): initiate new creation page with breadcrumb and title 

Taelar:
## Description

Ticket Reference: #TAPC-5437

### DIFF
--- a/packages/manager/apps/pci-kubernetes/src/components/nav/Breadcrumb.component.tsx
+++ b/packages/manager/apps/pci-kubernetes/src/components/nav/Breadcrumb.component.tsx
@@ -9,7 +9,9 @@ import {
   Breadcrumb as ODSBreadcrumb,
 } from '@ovhcloud/ods-react';
 
-import { computeBreadcrumbUrl } from '@/helpers/nav/url';
+import { useProject } from '@ovh-ux/manager-pci-common';
+
+import { computeBreadcrumbUrl, getProjectUrl } from '@/helpers/nav/url';
 
 type TBreadcrumbEntry = {
   label: string;
@@ -23,15 +25,17 @@ type TBreadcrumbProps = {
 export const Breadcrumb: FC<TBreadcrumbProps> = ({ entries }) => {
   const { t } = useTranslation(['common']);
 
-  const [home, ...crumbs] = useMemo(
-    () => computeBreadcrumbUrl<Partial<TBreadcrumbEntry>>([{}, ...entries], window.location.hash),
+  const { data: project } = useProject();
+  const projectUrl = useMemo(() => getProjectUrl(), []);
+  const crumbs = useMemo(
+    () => computeBreadcrumbUrl<Partial<TBreadcrumbEntry>>(entries, window.location.hash),
     [entries],
   );
 
   return (
     <ODSBreadcrumb aria-label={t('common_breadcrumb')}>
       <BreadcrumbItem>
-        <BreadcrumbLink href={home?.href}>
+        <BreadcrumbLink href={projectUrl} aria-label={project?.description} target="_self">
           <Icon name="home" />
         </BreadcrumbLink>
       </BreadcrumbItem>
--- a/packages/manager/apps/pci-kubernetes/src/helpers/nav/url.spec.ts
+++ b/packages/manager/apps/pci-kubernetes/src/helpers/nav/url.spec.ts
@@ -1,4 +1,6 @@
-import { computeBreadcrumbUrl } from './url';
+import { MockInstance, afterEach, describe, expect, it } from 'vitest';
+
+import { computeBreadcrumbUrl, getProjectUrl } from './url';
 
 describe('computeBreadcrumbUrl', () => {
   it('computes hrefs going back from current path', () => {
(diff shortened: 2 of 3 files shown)


packages/manager/apps/pci-kubernetes/src/components/nav/Breadcrumb.component.tsx
```diff
@@ -0,0 +1,47 @@
...
+
+import { computeBreadcrumbUrl } from '@/helpers/nav/url';
+
+type TBreadcrumbEntry = {
+  label: string;
+  ariaLabel?: string;
+};
+
+type TBreadcrumbProps = {
+  items: Array<TBreadcrumbEntry>;
```
aTurmo:
```suggestion
  entries: Array<TBreadcrumbEntry>;
```

packages/manager/apps/pci-kubernetes/src/helpers/nav/url.ts
```diff
@@ -0,0 +1,16 @@
+import { resolvePath } from 'react-router-dom';
+
+export const computeBreadcrumbUrl = <Entry extends object>(
+  items: Array<Entry>,
+  currentPath: string,
+) => {
+  return items.map((item, index) => {
+    const backIndex = items.length - index - 1;
```
aTurmo:
```suggestion
) => {
  const maxIndex = items.length - 1;
  return items.map((item, index) => {
    const backIndex = maxIndex - index;
```

packages/manager/apps/pci-kubernetes/src/helpers/nav/url.spec.ts
```diff
@@ -0,0 +1,29 @@
...
+
+describe('computeBreadcrumbUrl', () => {
+  it('computes hrefs going back from current path', () => {
+    const items = [{ label: 'a' }, { label: 'b' }, { label: 'c' }, { label: 'd' }];
+    const result = computeBreadcrumbUrl(items, '/a/b/c/d');
+
+    expect(result[0]?.href).toEqual('/a');
+    expect(result[1]?.href).toEqual('/a/b');
+    expect(result[2]?.href).toEqual('/a/b/c');
+    expect(result[3]?.href).toEqual('/a/b/c/d');
```
aTurmo: Remark: Not of fan of the input data, because it make it seems like there could be a usage of the label since there are exactly the same as the path (a is /a, b is /b etc.). To make it clearer it better to have different data for different things (ie here change you label to not equal anything else)

aTurmo: same with the other test

ppprevost (approved):

packages/manager/apps/pci-kubernetes/src/helpers/nav/url.spec.ts
```diff
@@ -0,0 +1,29 @@
+import { computeBreadcrumbUrl } from './url';
+
+describe('computeBreadcrumbUrl', () => {
+  it('computes hrefs going back from current path', () => {
```
fredericvilcot, line 6: We could imagine using `it.each` with an array of data to prevent having too many assertions

packages/manager/apps/pci-kubernetes/src/helpers/nav/url.ts
```diff
@@ -0,0 +1,17 @@
+import { resolvePath } from 'react-router-dom';
+
+export const computeBreadcrumbUrl = <Entry extends object>(
```
fredericvilcot, line 3: I'm not sure to feel the real interest having a Generic Type here? 

`computeBreadcrumbUrl` could directly take an Array of `TBreadcrumbEntry` and returns an Array of `TBreadcrumbEntryWithHref`, no?

packages/manager/apps/pci-kubernetes/src/components/nav/Breadcrumb.component.tsx
```diff
@@ -0,0 +1,47 @@
...
+type TBreadcrumbProps = {
+  entries: Array<TBreadcrumbEntry>;
+};
+
+export const Breadcrumb: FC<TBreadcrumbProps> = ({ entries }) => {
+  const { t } = useTranslation(['common']);
+
+  const [home, ...crumbs] = useMemo(
+    () => computeBreadcrumbUrl<Partial<TBreadcrumbEntry>>([{}, ...entries], window.location.hash),
+    [entries],
```
fredericvilcot, line 32: So crumbs are not reevaluated when hash changes, is it wanted?

packages/manager/apps/pci-kubernetes/src/helpers/nav/url.spec.ts
Taelar: I'm not sure to see how it.each would make it simpler, do you have an example ? 
Also I tend to prefer basic assertions as it is easier to debug when test fail

packages/manager/apps/pci-kubernetes/src/helpers/nav/url.ts
Taelar: The idea was to formally separate the `computeBreadcrumbUrl` function from the component, so the function could be used somewhere else

packages/manager/apps/pci-kubernetes/src/components/nav/Breadcrumb.component.tsx
Taelar: It is, my idea was to make entries the source of truth :slightly_smiling_face:

packages/manager/apps/pci-kubernetes/src/helpers/nav/url.spec.ts
fredericvilcot: IMO when you have multiple assertions to test the same thing, you can enhance using a data-table representation and prevent from having this : 

```ts
    expect(result[0]?.href).toEqual('/a');
    expect(result[1]?.href).toEqual('/a/b');
    expect(result[2]?.href).toEqual('/a/b/c');
    expect(result[3]?.href).toEqual('/a/b/c/d');
```
What will happen if your function handles more edge cases to test? You'll have an entire list of assertions doing the same things :) 

A test is not so far from real code implementation, in this case you'll use a loop :)

So instead you can keep being DRY by easily manipulating your data and test inside a loop : https://vitest.dev/api/#test-each

But it's up to you 😌

packages/manager/apps/pci-kubernetes/src/helpers/nav/url.ts
fredericvilcot: Yes I see the point, but will this function being used out of breadcrumb logic? ;)

packages/manager/apps/pci-kubernetes/src/components/nav/Breadcrumb.component.tsx
fredericvilcot: However, since we still read from `window.location.hash`, the dependency is implicit and React won’t re-evaluate on navigation.

If `computeBreadcrumbUrl` relies on hash, it should be a dependency, or logic should be moved to the function itself no?

packages/manager/apps/pci-kubernetes/src/helpers/nav/url.ts
Taelar: Probably not, but it could be used in another breadcrumb related case. Also we could think it the other way around : Is it useful to strictly bind them through types ? :sweat_smile:

packages/manager/apps/pci-kubernetes/src/components/nav/Breadcrumb.component.tsx
Taelar: Yeah we could consider it a limitation. However if it were a dependency, it could recompute the crumbs with no entries change. Meaning we could end up with a breadcrumb link associated to the wrong label, which is another kind of limitation but is much more risky imo. 
The perfect solution would be to create a dictionnary (or a labelised routing object) which allows us to bind a route fragment to a translation key, in which case we would not even need the parents to give us entries, and the Breadcrumb component would be totally autonomous. But that would be a pretty big refactor 

And about putting window.location.hash directly into the function, I'd say why not, but it would make the function smarter that it needs to be imo

packages/manager/apps/pci-kubernetes/src/helpers/nav/url.spec.ts
Taelar: I do see the use of .each when testing multiple calls of the same function with changing args, but here I'm calling my function once and asserting the result array contains the expected output, so I can't really picture what you're talking about :sweat_smile:

PR #21106, opened 2025-12-16
fix(pci-block-storage): fix reload attached instances for retyping

aTurmo:
ref: #TAPC-5456

### DIFF
--- a/packages/manager/apps/pci-block-storage/src/api/select/volume.ts
+++ b/packages/manager/apps/pci-block-storage/src/api/select/volume.ts
@@ -23,6 +23,7 @@ import {
   BlockStorageListColumn,
   isBlockStorageListColumn,
 } from '@/hooks/useDatagridColumn';
+import { capitalizeFirstLetter } from '@/utils';
 
 const volumeComparator = <T>(extractor: (_: TVolume) => T) => (
   t1: TVolume,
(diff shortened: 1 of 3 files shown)


packages/manager/apps/pci-block-storage/src/api/hooks/useInstance.ts
```diff
@@ -41,18 +41,31 @@ export const useAttachedInstances = (
   volumeId: string,
   options: TApiHookOptions = {},
 ) => {
-  const { data: volume } = useVolume(projectId, volumeId);
+  const {
+    data: volume,
+    isPending: isVolumePending,
+    isFetching: isVolumeFetching,
+  } = useVolume(projectId, volumeId, undefined, options);
```
lolottetheclash: Here I ave no idea about what is `undefined` , you could change a little bit `useVolume` so that we can do :
`useVolume({ projectId, volumeId, options });`
what do you think?

lolottetheclash (approved):
Just something to discuss, not blocking ;)

packages/manager/apps/pci-block-storage/src/api/hooks/useInstance.ts
aTurmo: As discussed I changed it to include the 'capacity' in the options

PR #21113, opened 2025-12-16
fix(pci-project): load willpayment relative path

seven-amid:
## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-project/src/utils/__tests__/environment.spec.ts
+++ b/packages/manager/apps/pci-project/src/utils/__tests__/environment.spec.ts
@@ -1,0 +1,41 @@
+import { describe, expect, it } from 'vitest';
+
+import { isLabeuEnvironment, isProdEnvironment } from '../environment';
+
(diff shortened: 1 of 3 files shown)


packages/manager/apps/pci-project/src/utils/__tests__/environment.spec.ts
```diff
@@ -0,0 +1,34 @@
+import { describe, expect, it } from 'vitest';
+
+import { isLabeuEnvironment, isProdEnvironment } from '../environment';
+
+describe('environment utils', () => {
+  describe('isProdEnvironment', () => {
+    it('returns true when hostname contains one of the prod host suffixes', () => {
```
aTurmo: Suggestion: you could use it.each to avoid repeating yourself in the expect

seven-amid: @aTurmo done :)

PR #21125, opened 2025-12-16
feat(billing): add filter renew in services

achrafHafedh:
ref: #MANAGER-20467

## Description

Ticket Reference: #MANAGER-20467

### DIFF
--- a/packages/manager/modules/billing/src/autoRenew/actions/update/translations/Messages_de_DE.json
+++ b/packages/manager/modules/billing/src/autoRenew/actions/update/translations/Messages_de_DE.json
@@ -1,7 +1,7 @@
 {
   "autorenew_service_update_modal_title": "Verlängerung konfigurieren",
   "autorenew_service_update_modal_description": "Definieren Sie ein Verlängerungsintervall für Ihren Dienst: {{ serviceName }} ",
-  "autorenew_service_update_modal_period_manual": "Keine Frequenz (manuelle Verlängerung)",
+  "autorenew_service_update_modal_period_manual": "Manuelle Erneuerung",
   "autorenew_service_update_modal_period_every_month": "Jeden Monat",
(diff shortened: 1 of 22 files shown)


packages/manager/modules/billing/src/autoRenew/autorenew.routing.js
```diff
@@ -40,6 +40,10 @@ export default /* @ngInject */ ($stateProvider) => {
         value: null,
         squash: true,
       },
+      renew: {
+        value: null,
+        squash: true,
+      },
```
JacquesLarique: This is not necessary as we saw that the "renew" filter is added to the "filter" query parameter

packages/manager/modules/billing/src/autoRenew/autorenew.service.js
```diff
@@ -137,6 +139,18 @@ export default class {
...
+      (translatedStatus, status) => ({
+        ...translatedStatus,
+        [status]: this.$translate.instant(
+          `billing_autorenew_service_renew_${status}`,
+        ),
+      }),
+      {},
+    );
+  }
+
```
JacquesLarique, line 153: Maybe we should create a dedicated service for this kind of functions (creating filters list) as the `BillingAutoRenew` is containing a lot of various functions

packages/manager/modules/billing/src/autoRenew/autorenew.routing.js
achrafHafedh: fixed !

packages/manager/modules/billing/src/autoRenew/autorenew.service.js
achrafHafedh: As discussed, we can handle it in a refactoring ticket

JacquesLarique (approved):

PR #21161, opened 2025-12-17
fix(pci-project): fix project renaming bug

seven-amid:
ref: #MANAGER-20608

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-project/src/pages/creation/hooks/useConfigForm.tsx
+++ b/packages/manager/apps/pci-project/src/pages/creation/hooks/useConfigForm.tsx
@@ -40,24 +40,22 @@ export const useConfigForm = (ovhSubsidiary: OvhSubsidiary, cartId: string | und
   const { data: hdsItem } = useIsHdsChecked(cartId);
 
   useEffect(() => {
-    if (cartConfigurationDescription && form.description !== cartConfigurationDescription.value) {
-      setTimeout(() => {
-        setForm((prev) => ({
-          ...prev,
-          description: cartConfigurationDescription.value || '',
-        }));
-      }, 0);
+    if (cartConfigurationDescription) {
+      // eslint-disable-next-line react-hooks/set-state-in-effect
+      setForm((prev) => ({
+        ...prev,
+        description: cartConfigurationDescription.value || '',
+      }));
     }
-  }, [cartConfigurationDescription, setForm, form.description]);
+  }, [cartConfigurationDescription, setForm]);
 
   useEffect(() => {
     if (hdsItem) {
-      setTimeout(() => {
-        setForm((prev) => ({
-          ...prev,
-          isHdsChecked: !!hdsItem,
-        }));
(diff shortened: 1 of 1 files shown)


FabienHenon (approved):

packages/manager/apps/pci-project/src/pages/creation/hooks/useConfigForm.tsx
```diff
@@ -40,15 +40,15 @@ export const useConfigForm = (ovhSubsidiary: OvhSubsidiary, cartId: string | und
   const { data: hdsItem } = useIsHdsChecked(cartId);
 
   useEffect(() => {
-    if (cartConfigurationDescription && form.description !== cartConfigurationDescription.value) {
-      setTimeout(() => {
+    setTimeout(() => {
```
fredericvilcot: Why do you need here to simulate an asynchronous operation on the form? 

Generally, these kinds of tricks should be avoided because it makes code unpredictable and unmaintainable, as it produces asynchronous side effects.

seven-amid: Added setTimeout to fix the static-kit linter error: 'Calling setState synchronously within an effect can trigger cascading renders.' Open to suggestions if there's a better way to handle this synchronization.

fredericvilcot: @seven-amid 
I think here there is a mismatch somewhere and adding an async setter comes to hide the pain.
I see no issue here, so maybe, according to the Docs you can either :
- remove the unnecessary `settimeout`
- remove the unnecessary useEffect and use render to compute data
- move from `useEffect` to `useLayoutEffect` 

But please, do not add timers everywhere that are a pain to debug and maintain.

https://react.dev/reference/eslint-plugin-react-hooks/lints/set-state-in-effect

seven-amid: @fredericvilcot, I've removed the setTimeout calls and added eslint-disable comments. While not ideal, I don't see a simpler way to synchronize the React Query data with local form state without a more significant architectural change.

fredericvilcot: @seven-amid 
IMO you don't need any effect here 
You should compute data in a simple useMemo.

There is no need for local state and effect

seven-amid: @fredericvilcot Even with useLayoutEffect, the linter is still blocking it.

fredericvilcot: Regarding the current architecture, we should have used React-hook-form library + zod to handle form, to prevent multiple local states with some inconsistencies (effects + setState) 

To be well tested , thanks ☺️

PR #21180, opened 2025-12-18
feat(hub): apply eslint strategy of static analysis kit adaption

sachinrameshn:
## Description
**Static Analysis Kit Adoption**

- Eslint strategy


.eslintignore
anooparveti, line 1: **question:** Why removing communications app from eslintignore? Is it intended?

MehdiZaghouani (approved):

.eslintignore
sachinrameshn: A entry for communication app already exist in same file, so I just updated it with hub app to remove duplication.

anooparveti (dismissed):

PR #21200, opened 2025-12-18
fix(backup-agent): unmocking fixes wave2

pauldkn:
## Description

Ticket Reference: #BKP-717


packages/manager/modules/backup-agent/src/data/hooks/tenants/useVspcTenants.ts
```diff
@@ -14,19 +10,23 @@ import { getVSPCTenantDetails, getVSPCTenants } from '../../api/tenants/tenants.
...
     queryFn: () => getVSPCTenants({ backupServicesId }),
     select: (res) => res.data,
     enabled: !!backupServicesId,
   });
+
+  return {
+    ...query,
+    isLoadingBackupServicesId: isPending,
+    isFullyLoaded: !query.isPending && !isPending,
+  };
```
mathieumousnier: I think you could return `useQuery` and be safe with isPending.
The query is not enabled if `backupServicesId` is undefined.

1. while `useBackupServicesId` loads, `backupServicesId=undefined`, `enabled=false`, and `query.isPending=true`
2. `backupServicesId='string'`, `enabled=true`, the query loads and `query.isPending=true`
3. then the query succeeds, and `isPending=false`

In theory this should work, it's worth a try
But, since you did something different you might have seen something I don't.

In that case, I would advise to stay closer to something more usual, and return `isPending: query.isPending || isPending`.
This would avoid to do something like `isLoading={!isFullyLoaded}` which is a small thing, but breaks the most common conventions and feels a bit hard to understand.

packages/manager/modules/backup-agent/src/pages/services/dashboard/agent/download/AgentDownload.page.tsx
```diff
@@ -68,8 +71,12 @@ export default function DownloadAgentPage() {
...
           <OdsText>{t(`${BACKUP_AGENT_NAMESPACES.AGENT}:download_agent_with_executable`)}</OdsText>
           <div className="flex justify-center align-center">
-            <a href={downloadLink} download>
-              <OdsButton label={t(`${NAMESPACES.ACTIONS}:download`)} isLoading={isLoading} />
+            <a href={isDownloadEnabled ? downloadLink : undefined} download>
+              <OdsButton
+                label={t(`${NAMESPACES.ACTIONS}:download`)}
+                isLoading={isLoading}
+                isDisabled={!isDownloadEnabled}
+              />
```
mathieumousnier, line 79: This is not related to this PR, but this is evil_2000 level.
This breaks all accessibility rules.
Download is an action, not navigation, so a simple button is the perfect fit.

packages/manager/modules/backup-agent/src/pages/services/listing/Listing.page.tsx
```diff
@@ -20,7 +20,7 @@ export default function ListingPage() {
             columns={columns}
             items={data || []}
             totalItems={data?.length || 0}
-            isLoading={isLoading}
+            isLoading={!isFullyLoaded}
```
mathieumousnier: This looks strange, I made a detailed comment in the useVSPCTenants hook

packages/manager/modules/backup-agent/src/data/hooks/consumption/useServiceConsumption.ts
```diff
@@ -0,0 +1,20 @@
...
+
+  return useQuery({
+    queryKey: ['services', 'consumption', resourceName],
+    queryFn: async () => {
+      const res = await queryClient.fetchQuery(serviceQueryOptions);
+      const serviceId = res.data.serviceId;
+      return getServiceConsumption(serviceId);
+    },
+    enabled: !!resourceName,
+  });
```
mathieumousnier, line 18: This `enabled: !!resourceName,` option should be added only if needed.
Here the `resourceName` param is typed as `string`, so `resourceName=undefined` should never happen.

Either:
- type resourceName: string | undefined and use the enable option 
- type resourceName: string and don't use it

packages/manager/modules/backup-agent/src/pages/vaults/dashboard/general-information/_components/subscription-tile/SubscriptionTile.component.tsx
```diff
@@ -14,13 +14,15 @@ type SubscriptionTileProps = {
   vaultId: string;
 };
 
+export const CONSUMPTION_MAX_VALUE_IN_TO = 100;
```
mathieumousnier:
```suggestion
export const CONSUMPTION_MAX_VALUE_IN_TB = 100;
```

packages/manager/modules/backup-agent/src/data/hooks/tenants/useVspcTenants.ts
pauldkn: You're right, I've simplified it by returning the query directly and using only `isPending`.
I keep updated the `BackupServiceAwareQuery` type for now in case we need to distinguish the two loading phase, or display a small loader for the `backupServicesId` query.

Note: I tried but can't override the return of `isPending` because it would break the type of `UseQueryResult`, as `isPending` is not really a boolean, but a literal type computed for the query state (either `true` | `false`)

packages/manager/modules/backup-agent/src/pages/services/dashboard/agent/download/AgentDownload.page.tsx
pauldkn: Indeed, we'll update this in another pull request

packages/manager/modules/backup-agent/src/data/hooks/consumption/useServiceConsumption.ts
pauldkn: Right, although an empty string would be evaluated as falsy, with the use of `useRequiredParams` it should not happen. Updated 👍

PR #21214, opened 2025-12-19
feat(pci-kubernetes): add empty summary cart

Taelar:
Note : The Cart component was copy/pasted from pci-instances creation revamp branch, it will be removed from pci-kubernetes when the component becomes available in pci-ui

## Description

Ticket Reference: #TAPC-5326

### DIFF
--- a/packages/manager/apps/pci-kubernetes/src/components/cart/Cart.component.tsx
+++ b/packages/manager/apps/pci-kubernetes/src/components/cart/Cart.component.tsx
@@ -1,0 +1,71 @@
+import { JSX, useMemo } from 'react';
+
+import { useTranslation } from 'react-i18next';
+
+import { Button, Icon, Text } from '@ovhcloud/ods-react';
+
+import {
+  CartActions,
+  CartContainer,
+  CartContent,
+  CartItem,
+  CartItemDetails,
+  CartItemHeader,
+  CartTotalPrice,
+} from './components';
+
+export type TCartProps = {
+  items: TCartItem[];
+};
+
+export type TCartItem = {
+  id: string;
+  title: string;
+  name?: string;
+  details: TCartItemDetail[];
+  expanded: boolean;
+};
+
+export type TCartItemDetail = {
+  name: string;
+  description?: JSX.Element;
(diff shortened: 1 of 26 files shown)


packages/manager/apps/pci-kubernetes/src/components/cart/Cart.component.tsx
```diff
@@ -0,0 +1,72 @@
+import { JSX, useMemo } from 'react';
+
+import { useTranslation } from 'react-i18next';
+
+import { Button, Icon, Text } from '@ovhcloud/ods-react';
+
+import {
+  Cart as BaseCart,
```
aTurmo: Question: Then why not having the Cart from /cart/components called 'BaseCart' or 'CartLayout' ? Or change the name of this one

It will be very confusing if I want to use the 'Cart' component when there are two in /cart and /cart.component.

aTurmo (commented):
Suggestion: You should add some tests

packages/manager/apps/pci-kubernetes/src/components/cart/Cart.component.tsx
Taelar: I'm gonna dislike my own answer here :sweat: 
Actually the Cart component is a copy/paste from pci-instances while we are waiting for the pci-ui lib so I don't feel like taking time to do non-functionnal changes is really worth 
Although I agree the naming could be improved here

aTurmo: :'(

packages/manager/apps/pci-kubernetes/src/components/cart/Cart.component.tsx
```diff
@@ -0,0 +1,71 @@
...
+
+import {
+  CartActions,
+  CartContainer,
+  CartItem,
+  CartItemDetails,
+  CartItemHeader,
+  CartTotalPrice,
+} from './components';
+import { CartContent } from './components/CartContent.component';
```
aTurmo: I think it could be grouped with the rest no ?

PR #21219, opened 2025-12-19
feat(billing): add badge for indicating renewal mode

sachinrameshn:
ref: #MANAGER-20437

## Description

As an OVHcloud customer, I would like to see the information in the 'Renewal frequency' column for each service on the 'My offers and services' page by indicating Automatic and Manual badges for each renewal modes


packages/manager/modules/billing/src/autoRenew/services/services.html
```diff
@@ -181,8 +181,14 @@
             data-type="options"
             data-filterable
         >
-            <span data-ng-bind-html="::$ctrl.getRenewPeriodTranslation($row)">
-            </span>
+            <div class="d-flex flex-column">
+                <service-renewal-badge service="$row"></service-renewal-badge>
+                <small
+                    class="pl-2"
```
JacquesLarique, line 188: I think you can use "align-self-center" and "justify-content-center" on the encapsulating div in order to center the badge and the text

PR #21228, opened 2025-12-19
fix(backup-agent): unmocking fixes wave3

pauldkn:
## Description

Ticket Reference: #BKP-704


packages/manager/modules/backup-agent/src/data/hooks/tenants/useVspcTenantDetails.ts
```diff
@@ -24,15 +24,15 @@ export const useBackupVSPCTenantDetails = ({
...
+  const query = useQuery({
     queryFn: () => getVSPCTenantDetails(backupServicesId, tenantId),
     queryKey: BACKUP_VSPC_TENANT_DETAILS_QUERY_KEY(tenantId),
     enabled: !!backupServicesId && !!tenantId,
-    select: (data): Resource<WithRegion<VSPCTenant>> =>
-      mapTenantResourceToTenantResourceWithRegion(data),
     ...options,
   });
+
+  return { ...query, isLoadingBackupServicesId: isPending };
```
mathieumousnier, line 33: This looks avoidable, like in the previous PR.

packages/manager/modules/backup-agent/src/data/hooks/agents/getAgents.ts
```diff
@@ -17,15 +17,15 @@ export const useBackupAgentList = (
     tenantId,
     pageSize,
   }: {
-    tenantId?: string;
-    pageSize: number;
-  } = { pageSize: 9999 },
+    tenantId: string;
+    pageSize?: number;
+  } = { tenantId: '', pageSize: 9999 },
```
mathieumousnier: Image

packages/manager/modules/backup-agent/src/data/hooks/tenants/useVspcTenantDetails.ts
pauldkn: Actually it's not the same as the previous PR
TLDR: it uses basic `isPending` and expose `isLoadingBackupServicesId` if needed to display first query loading.

(details here: https://github.com/ovh/manager/pull/21200#discussion_r2634713173)

mathieumousnier: I don't really get why you would need both loaders, and I see you use them doing `isPending || isLoadingBackupServicesId`, which is useless because doing `isPending` would give exactly the same result.
But there might be something I don't understand.

Nevertheless, since you removed the select and there is no "with-region" involved, you might want to simplify a little bit the type of the query like this:

```ts
export const useBackupVSPCTenantDetails = ({
  tenantId,
  ...options
}: {
  tenantId: string;
} & Partial<
  Omit<
    UseQueryOptions<Resource<VSPCTenant>>,
    'queryKey' | 'queryFn'
  >
>): BackupServiceAwareQuery<Resource<VSPCTenant>>
```

pauldkn: Right, hook's type fixed.

Regarding the loaders, it’s not mandatory, but I think exposing `isLoadingBackupServicesId` has several benefits:

- it lets the UX separate loading states (loader for global service ID / skeletons for main query)
- it makes the dependency on `backupServicesId` explicit in the hook’s type
- it keeps hooks uniform (still shaped like a `UseQueryResult`, no surprises)

In the same spirit, it could also be useful to expose its error, because a failure to fetch the service ID is a different problem than a failure to fetch the tenant, vault, or any other request

packages/manager/modules/backup-agent/src/data/hooks/tenants/useVspcTenantDetails.ts
```diff
@@ -17,22 +20,15 @@ export const BACKUP_VSPC_TENANT_DETAILS_QUERY_KEY = (vspcTenantID: string) => [
...
-  tenantId: string;
-} & Partial<
-  Omit<
-    DefinedInitialDataOptions<Resource<VSPCTenant>, unknown, Resource<WithRegion<VSPCTenant>>>,
-    'queryKey' | 'queryFn'
-  >
->) => {
-  const { backupServicesId } = useBackupServicesId();
+}: UseVSPCTenantDetailsParams): BackupServiceAwareQuery<Resource<VSPCTenant>, unknown> => {
+  const { backupServicesId, isPending } = useBackupServicesId();
```
sefiani-ovh, line 24: I think this hook would be simpler if it took backupServicesId as a parameter instead of calling useBackupServicesId internally. That would make the dependency explicit and the hook easier to understand and test.

If the hook is used in many places and you want to avoid calling both hooks every time, it’d be cleaner to create a small wrapper hook (or a context) that composes them, rather than hiding the dependency inside this hook.

packages/manager/modules/backup-agent/src/pages/services/dashboard/agent/AgentsListing.page.tsx
```diff
@@ -12,27 +12,28 @@ import { Datagrid } from '@ovh-ux/manager-react-components';
...
+  const { tenantId } = useRequiredParams('tenantId');
   const { t } = useTranslation([
     BACKUP_AGENT_NAMESPACES.SERVICE_LISTING,
     BACKUP_AGENT_NAMESPACES.AGENT,
     NAMESPACES.ACTIONS,
   ]);
   const navigate = useNavigate();
   const columns = useAgentsListingColumnsHooks();
-  const { flattenData, isLoading } = useBackupAgentList();
+  const { flattenData, isPending } = useBackupAgentList({ tenantId });
```
sefiani-ovh, line 29: useBackupAgentList(param1,param2, ..., {... object options of use query} )
This is just a suggestion: I think it’s fine to keep functional parameters (param1, param2, …) as positional and pass options of useQuery as object

packages/manager/modules/backup-agent/src/pages/vaults/dashboard/general-information/GeneralInformation.page.tsx
```diff
@@ -1,21 +1,29 @@
...
 
   return (
     <section className="flex flex-col sm:flex-row gap-8">
       <GeneralInformationVaultTile vaultId={vaultId!} />
       <SubscriptionTile vaultId={vaultId!} />
-      <BillingInformationsTileStandard resourceName={vaultResource?.currentState?.resourceName} />
+      <BillingInformationsTileStandard
+        resourceName={vaultResource?.currentState?.resourceName}
+        onResiliateLinkClick={() => {
+          navigate(`${subRoutes.delete}`);
```
sefiani-ovh: navigate(subRoutes.delete)

sefiani-ovh (approved):

packages/manager/modules/backup-agent/src/data/hooks/tenants/useVspcTenantDetails.ts
pauldkn: Actually I find it easier to handle like this, because in the component you use just one hook instead of calling the two every time, and handle both loading together

The context can be a solution too, but it's less consistent in handling it or mocking it in tests files

packages/manager/modules/backup-agent/src/pages/services/dashboard/agent/AgentsListing.page.tsx
pauldkn: The change here is only the use of `tenantId` as a parameter and not the way it's defined, but yes it would also be fine 👍

PR #21235, opened 2025-12-20
feat: add warning when manual editing dns zone breaking terraform state

saimonn:
## Description

When using terraform to manage DNS zones, it's can be very (very) annoying to have our terraform state out-of-sync with the current status of the zone. It happens if we modify the zone in text mode.

This PR warns terraform (and other IaC tools) users for this, so they have a chance to abort and use terraform instead.

## Additional Information

This is more or less linked to the OVH terraform provider issue 275: https://github.com/ovh/terraform-provider-ovh/issues/275#issuecomment-1506798086

A warning has been added to the provider documentation (https://github.com/ovh/terraform-provider-ovh/pull/408/), but in my opinion having the warning directly on the OVH web console itself would prevent such issues more efficiently.


saimonn:
sorry for the noise due to copilot. closing this PR in favor of a clean one.

PR #21276, opened 2025-12-23
feat(zimbra): enhance ux in account section

amine-daoussi:
ref: #PRDCOL-301

## Description
In zimbra > email accounts :
- Rename buttons.
- Change some buttons positions.

image

### DIFF
--- a/packages/manager/apps/zimbra/public/translations/common/Messages_de_DE.json
+++ b/packages/manager/apps/zimbra/public/translations/common/Messages_de_DE.json
@@ -48,7 +48,7 @@
   "define_members_mailing_list": "Mitglieder definieren",
   "configure_delegation_mailing_list": "Delegierung konfigurieren",
   "email_account": "E-Mail-Account",
-  "add_email_account": "Einen E-Mail-Account erstellen",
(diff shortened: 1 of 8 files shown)


packages/manager/apps/zimbra/src/pages/dashboard/emailAccounts/DatagridTopBar.component.tsx
```diff
@@ -168,6 +169,9 @@ export const DatagridTopbar: React.FC<DatagridTopbarProps> = ({
           icon={ODS_ICON_NAME.trash}
         />
       )}
+      <div className="ml-auto flex gap-6">
```
atefzed: This style is unnecessary.

amine-daoussi: it is, because in the scope, we have to put this button in the right close to the search bar.

atefzed: You only need ml-auto without the new parent div

ghyenne (approved):

PR #21285, opened 2025-12-23
feat(pci-instances): add monthly windows license & update cart

lolottetheclash:
ref: #TAPC-5543

**Hourly selected**
Capture d’écran 2025-12-23 à 16 01 33

**Monthly selected**
Capture d’écran 2025-12-23 à 16 01 19

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-instances/src/pages/instances/create/view-models/cartViewModel.ts
+++ b/packages/manager/apps/pci-instances/src/pages/instances/create/view-models/cartViewModel.ts
@@ -2,11 +2,8 @@ import { Deps } from '@/deps/deps';
 import { Reader } from '@/types/utils.type';
 import { getRegionNameKey } from './localizationsViewModel';
 import { getRegionalizedFlavorOsTypePriceId } from '@/utils';
-import {
-  TFlavorPrices,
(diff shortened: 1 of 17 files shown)


packages/manager/apps/pci-instances/src/pages/instances/create/view-models/cartViewModel.ts
```diff
@@ -151,18 +122,15 @@ export const selectFlavorDetails: Reader<Deps, TSelectFlavorData> = (deps) => {
...
-        );
+    const price = getFlavorPrice({
+      flavorName: foundFlavor.name,
+      regionId: foundRegionalizedFlavor.regionId,
+      osType,
+      flavorPricesById: data.entities.flavorPrices.byId,
+      billingType,
+    });
+
+    if (!price) return null;
```
yaallco: Question:

Do we want to return null if price is 0 ? What happen if a product is free for a promotion or something ?

lolottetheclash: @ZyaCon Thank you, I always get caught out on checking numbers!

PR #21291, opened 2025-12-23
feat(pci-kubernetes): add name section

Taelar:
ref: #TAPC-5326

## Description

Ticket Reference: #...


packages/manager/apps/pci-kubernetes/src/pages/new/components/CreateClusterForm.component.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+describe('CreateClusterForm name management', () => {
+  let user: UserEvent;
+  let renderedDom: RenderResult;
+  let nameField: Element;
+
+  beforeEach(async () => {
+    user = userEvent.setup();
+    renderedDom = render(<CreateClusterForm />);
+
+    const field = renderedDom.container.querySelector('input[name="name"]');
```
lolottetheclash: Is it not easier to get the field by something like `screen.getByRole`or `screen.getByLabelText`or something like this?

packages/manager/apps/pci-kubernetes/src/components/create/NameSection.component.tsx
```diff
@@ -0,0 +1,44 @@
...
+        {t('kubernetes_add_name')}
+      </Text>
+      <FormField>
+        <FormFieldLabel>{t('kubernetes_add_name_input_label')}</FormFieldLabel>
+        <Input {...register(fieldName)} className="sm:w-1/2" invalid={!!fieldError} />
+        <FormFieldHelper
+          className={clsx('mt-3 text-[--ods-color-text]', {
+            'text-[--ods-theme-critical-color]': !!fieldError,
+          })}
+          aria-live={fieldError ? 'polite' : undefined}
```
ppprevost: {...(fieldError ? { 'aria-live': 'polite' } : {})}

packages/manager/apps/pci-kubernetes/src/pages/new/components/CreateClusterForm.component.tsx
```diff
@@ -1,20 +1,36 @@
...
+      <div className="flex flex-col gap-6 md:flex-row">
+        <section className="flex-[2]">
+          <Text preset="heading-2" className="mb-6">
+            {t('kubernetes_add')}
+          </Text>
+          <NameSection<TCreateClusterSchema>
+            fieldName="name"
+            fieldHelper={t('kubernetes_add_cluster_name_input_pattern_validation_error')}
+          />
+          <Divider spacing="64" />
```
ppprevost: 64 > it must be a const

packages/manager/apps/pci-kubernetes/src/pages/new/components/CreateClusterForm.component.spec.tsx
Taelar: Indeed it's better with a getByRole, thanks for the advice !

packages/manager/apps/pci-kubernetes/src/pages/new_v2/components/CreateClusterForm.component.spec.tsx
```diff
@@ -0,0 +1,46 @@
...
+
+    await user.clear(field);
+    nameField = field;
+  });
+
+  it('keeps synchronized name input and cart subtitle', async () => {
+    await user.type(nameField, validClusterName);
+
+    const cartHeaderSubtitle = renderedDom.getByTestId('cart-header-subtitle');
+    expect(cartHeaderSubtitle).toHaveTextContent(validClusterName);
```
Tsiorifamonjena, line 37: You can still expect with getByText(validClusterName) to be visible no ? Then you can omit the data-testId

packages/manager/apps/pci-kubernetes/src/pages/new_v2/hooks/useCreateClusterForm/useCreateClusterForm.ts
Tsiorifamonjena, line 1: The .hook file extension is unusual 🤭

packages/manager/apps/pci-kubernetes/src/pages/new_v2/components/CreateClusterForm.component.spec.tsx
Taelar: Theorically yes but I find it unsafe for the purpose of this test, as I'm rendering a whole form this text could be displayed somewhere else (for example in the input it comes from) and so the test would pass for the wrong reason

packages/manager/apps/pci-kubernetes/src/pages/new_v2/hooks/useCreateClusterForm/useCreateClusterForm.ts
Taelar: Whoops, wrong project :face_with_peeking_eye:

packages/manager/apps/pci-kubernetes/src/components/create/NameSection.component.tsx
```diff
@@ -0,0 +1,42 @@
...
+      <Text preset="heading-3" className="mb-6">
+        {t('kubernetes_add_name')}
+      </Text>
+      <FormField invalid={!!fieldError}>
+        <FormFieldLabel>{t('kubernetes_add_name_input_label')}</FormFieldLabel>
+        <Input {...register(fieldName)} className="sm:w-1/2" invalid={!!fieldError} />
+        <FormFieldHelper className="flex flex-col">
+          {helperEntries.map((entry, index) => (
+            <Text preset="caption" key={index}>
+              {entry}
```
ppprevost, line 46: great

ppprevost (approved):

packages/manager/apps/pci-kubernetes/src/pages/new_v2/components/CreateClusterForm.component.spec.tsx
Tsiorifamonjena: I'm not sure that you could target input in this way but I know also that ODS is rendered differently so yes I know how difficult is to target element. It's ok for me :)

Tsiorifamonjena (approved):

packages/manager/apps/pci-kubernetes/src/pages/new_v2/components/CreateClusterForm.component.tsx
```diff
@@ -1,20 +1,33 @@
...
-      <section className="flex-[2]">
-        <Text preset="heading-2">{t('kubernetes_add')}</Text>
-      </section>
-      <aside className="flex-1">
-        <CreationCart />
-      </aside>
-    </div>
+    <FormProvider {...form}>
+      <div className="flex flex-col gap-6 md:flex-row">
+        <section className="flex-[2]">
```
ppprevost, line 20: and md:basis-2/3 instead

packages/manager/apps/pci-kubernetes/src/pages/new_v2/components/CreateClusterForm.component.tsx
```diff
@@ -1,20 +1,33 @@
...
+          <Text preset="heading-2" className="mb-6">
+            {t('kubernetes_add')}
+          </Text>
+          <ClusterNameSection />
+          <Divider spacing={dividerSpacing} />
+        </section>
+        <aside className="flex-1">
```
ppprevost, line 27: md:basis-1/3 ?

packages/manager/apps/pci-kubernetes/src/components/cart/Cart.component.tsx
```diff
@@ -16,6 +16,7 @@ import {
 
 export type TCartProps = {
   items: TCartItem[];
+  confirmDisabled: boolean;
```
aTurmo: Question: confirmDisabled ? I don't get the name of this props. it suggest an action since confirm is used as an imperative verb. Since it's used to disable a button why not just "isActionDisabled" or something like that ?

packages/manager/apps/pci-kubernetes/src/components/create/NameSection.component.tsx
```diff
@@ -0,0 +1,42 @@
...
+import { useTranslation } from 'react-i18next';
+
+import { FormField, FormFieldHelper, FormFieldLabel, Input, Text } from '@ovhcloud/ods-react';
+
+type TNameSectionProps<FormSchema extends object> = {
+  fieldName: Path<FormSchema>;
+  helperEntries: Array<string>;
+};
+
+export const NameSection = <FormSchema extends object>({
```
ppprevost: from object to FieldValues

```
import {  FieldValues, Path, useFormContext } from 'react-hook-form';
FormSchema extends FieldValues
```

packages/manager/apps/pci-kubernetes/src/pages/new_v2/components/CreateClusterForm.component.tsx
Taelar: The recommended way is actually to use the flex property rather than just flex-basis (especially cause it also includes a flex-grow property). 
Plus I think we should not change our codes out of fear of custom values when Tailwind intentionally lets us decide which value we should use (https://tailwindcss.com/docs/flex) :sweat_smile:

PR #21310, opened 2025-12-24
feat(dedicated): Show “Status of advanced GAME protection” (3-state)

b-fritz-ovh:
ref: #MANAGER-20588

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/modules/bm-server-components/src/network-tile/controller.js
+++ b/packages/manager/modules/bm-server-components/src/network-tile/controller.js
@@ -3,7 +3,7 @@ import ipaddr from 'ipaddr.js';
 import { GAME_DDOS_STATUS } from './constants';
 
 const parseIpGameUrl = (ipBlock, ip) =>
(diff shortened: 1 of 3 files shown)


packages/manager/modules/bm-server-components/src/network-tile/styles.scss
```diff
@@ -0,0 +1,3 @@
+.d-webkit-box {
+  display: -webkit-box;
+}
```
qpavy, line 3: Why don't use flexbox ?

packages/manager/modules/bm-server-components/src/network-tile/controller.js
```diff
@@ -1,4 +1,5 @@
 import isFunction from 'lodash/isFunction';
+import { filterIpv4List } from './utils';
```
qpavy: You can use ipaddr lib here I think

PR #21330, opened 2025-12-26
feat(dedicated): game ddos modal disclamer

b-fritz-ovh:
ref: #MANAGER-20590

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/modules/bm-server-components/src/network-tile/utils.js
+++ b/packages/manager/modules/bm-server-components/src/network-tile/utils.js
@@ -1,4 +1,0 @@
-const ipv4Regex = /^((25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)\.){3}(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)$/;
-
-export const filterIpv4List = (ips) =>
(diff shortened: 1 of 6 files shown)


packages/manager/modules/bm-server-components/src/network-tile/styles.scss
```diff
@@ -0,0 +1,3 @@
+.d-webkit-box {
+  display: -webkit-box;
```
qpavy, line 2: Why don't use flexbox ?

packages/manager/modules/bm-server-components/src/network-tile/utils.js
```diff
@@ -0,0 +1,4 @@
+const ipv4Regex = /^((25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)\.){3}(25[0-5]|2[0-4]\d|1\d{2}|[1-9]?\d)$/;
+
+export const filterIpv4List = (ips) =>
+  ips.filter((ip = '') => ipv4Regex.test(ip.split('/')[0]));
```
qpavy: We already have ipaddr.js library imported inside dedicated. Maybe we can use it here to filter on IpV4

mhelhali-soufien (approved):

PR #21336, opened 2025-12-26
fix(observability): fix issue on name and descirption limit

yjaaouane:
fixes: #MAOBS-158

## Description

Ticket Reference: #MAOBS-158

### DIFF
--- a/packages/manager/apps/observability/src/utils/schemas/description.schema.ts
+++ b/packages/manager/apps/observability/src/utils/schemas/description.schema.ts
@@ -6,5 +6,4 @@ export const createDescriptionSchema = (t: (key: string, options?: { value?: num
   z
     .string()
     .trim()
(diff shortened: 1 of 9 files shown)


packages/manager/apps/observability/src/utils/schemas/description.schema.ts
```diff
@@ -6,5 +6,4 @@ export const createDescriptionSchema = (t: (key: string, options?: { value?: num
   z
     .string()
     .trim()
-    .max(80, t(`${NAMESPACES.FORM}:error_max_chars`, { value: 80 }))
-    .optional();
+    .min(1, t(`${NAMESPACES.FORM}:required_field`));
```
MarcMonchablon, line 9: No 'max' limit for description ?

MarcMonchablon: Edit: I've checked the associated Jira, so no limit apparently.

PR #21339, opened 2025-12-29
test(ovh-shell): add mock for fetchConfiguration in shell test

anooparveti:
## Description

This Change fixes the flaky test in shell by mocking the fetchConfiguration call.

Ticket Reference: #MANAGER-20425    

## Additional Information

**Test case goal:** When user try to load a µ-app directly without Container (supposed to be loaded only in container), shell should redirect to container and then load the µ-app inside container.

**Steps involved:**
	1.	Get application registry with `/applications` BFF
	2.	Check if µ-app can/cannot be loaded outside of container.
	3.	If µ-app is not supposed be loaded outside container, replace the window.location.href to load the µ-app inside Container (Required URL is obtained from step 1)
	4.	Make a call to `/configuration` BFF and then initialise Environment using the fetchConfiguration utility provided by `@ovh-ux/manager-config` package.

In the above process, the flaky behaviour is occurred at step 4 but we can validate the required test scenario at step 3. So I am mocking the `fetchConfiguration` util from the package to avoid the trigger of flaky code.

### DIFF
--- a/packages/components/ovh-shell/src/client/shell-client.spec.ts
+++ b/packages/components/ovh-shell/src/client/shell-client.spec.ts
@@ -13,6 +13,17 @@ vi.stubGlobal('window', {
   location: { href: '' },
 });
 
+vi.mock('@ovh-ux/manager-config', async (importOriginal) => {
+  const original: any = await importOriginal();
+  return {
+    ...original,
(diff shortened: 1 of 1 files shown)


helabenkhalfallah (approved):
See together, approved !

PR #21367, opened 2025-12-30
Feat/pci kubernetes change condition 3az tapc 5311

ppprevost:
## Description

Ticket Reference: #TAPC-5311, #TAPC-5435

### DIFF
--- a/packages/manager/apps/pci-kubernetes/src/pages/new/New.page.tsx
+++ b/packages/manager/apps/pci-kubernetes/src/pages/new/New.page.tsx
@@ -26,11 +26,11 @@ import {
 import { ShellContext } from '@ovh-ux/manager-react-shell-client';
 
 import { useCreateKubernetesCluster } from '@/api/hooks/useKubernetes';
-import { isMonoDeploymentZone, isMultiDeploymentZones } from '@/helpers';
+import { isStandardPlan } from '@/helpers';
 import use3AZPlanAvailable from '@/hooks/use3azPlanAvaible';
 import useHas3AZRegions from '@/hooks/useHas3AZRegions';
 import { PAGE_PREFIX } from '@/tracking.constants';
-import { DeploymentMode, TClusterPlanEnum } from '@/types';
+import { TClusterPlanEnum } from '@/types';
 
 import {
   TClusterCreationForm,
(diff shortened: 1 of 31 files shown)


packages/manager/apps/pci-kubernetes/src/pages/new/New.page.tsx
```diff
@@ -42,13 +42,10 @@ import stepsConfig from './steps/stepsConfig';
...
 
-  const regionType = form.region.type;
-
-  // TODO:  When 3AZ will be available, update condition to make form.plan mandatory to create a cluster & remove "free" fallback value in createCluster call
   return (
     !!form.network &&
-    ((isMultiDeploymentZones(regionType) && !!form.network.privateNetwork) ||
-      isMonoDeploymentZone(regionType)) &&
+    form.plan !== null &&
+    ((isStandardPlan(form.plan) && !!form.network.privateNetwork) || !isStandardPlan(form.plan)) &&
```
Taelar: It seems we're not using form.region.type in the condition anymore, is the `if (!form.region?.type) return false;` still relevant ?
Also maybe we could do 'return asap' for `!!form.network` et `form/plan != null` to shorten the final condition

packages/manager/apps/pci-kubernetes/src/pages/new/steps/NetworkClusterStep.component.tsx
```diff
@@ -31,15 +31,17 @@ export type TNetworkFormState = {
...
   region: string;
   type: DeploymentMode;
+  plan: TClusterPlanEnum;
   onChange: (networkForm: TNetworkFormState) => void;
 };
 
-export const isValidGateway3AZ = (type: DeploymentMode, gateways: TGateway[]) =>
-  isMultiDeploymentZones(type) && Array.isArray(gateways) && gateways.length;
+export const isValidGateway = (gateways: TGateway[]) =>
+  Boolean(Array.isArray(gateways) && gateways.length);
```
Taelar: I don't think components shoud export function, if this function is used elsewhere, maybe it belongs to ./utils ?

Taelar (approved):
Only non-blocking remarks, up to you to apply them or not :slightly_smiling_face: 
Well done for all the type fixing :raccoon:  :+1:

packages/manager/apps/pci-kubernetes/src/pages/new/steps/NetworkClusterStep.component.tsx
ppprevost: you're right, that is a bad legacy behavior

packages/manager/apps/pci-kubernetes/src/pages/new/New.page.tsx
ppprevost: right didn't see that :(

packages/manager/apps/pci-kubernetes/src/pages/new/steps/NetworkClusterStep.component.tsx
```diff
@@ -113,7 +119,7 @@ export default function NetworkClusterStep({
                 key={form.privateNetwork?.id}
                 region={region}
                 className="mt-2"
-                projectId={projectId}
+                projectId={projectId ?? ''}
```
yaallco: What happen if projectId is null ? The component will render with an empty string as id, is it a possible case ?

yaallco:
Well done ! I just commented non-blocking suggestion too.
And the task id is missing from your description

PR #21371, opened 2025-12-30
Fix/display list before tooltip tapc 5515

ZyaCon:
## Description

Ticket Reference: #TAPC-5515    

## Additional Information
image

### DIFF
--- a/packages/manager/apps/pci-instances/src/pages/instances/create/components/flavor/select/FlavorSelect.component.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/create/components/flavor/select/FlavorSelect.component.tsx
@@ -117,7 +117,7 @@ const FlavorSelect = ({ option, items }: TFlavorSelectOption) => {
               )}
             />
             <SelectContent
-              className="[&>div>span:first-child]:w-full"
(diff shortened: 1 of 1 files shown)


packages/manager/apps/pci-instances/src/pages/instances/create/components/flavor/select/FlavorSelect.component.tsx
```diff
@@ -117,7 +117,7 @@ const FlavorSelect = ({ option, items }: TFlavorSelectOption) => {
               )}
             />
             <SelectContent
-              className="[&>div>span:first-child]:w-full"
+              className="z-[1000] [&>div>span:first-child]:w-full"
```
Taelar, line 120: Did you ask ODS maintainers if the z-index layers hierarchy was made this way on purpose ? 
They may have decided that list should go behind tooltips (which makes more sense to me), so we'd be playing against ODS here

yaallco: Thanks for feedback, I did not think about it. I'll ask ODS maintainers :)

yaallco:
Ticket is canceled, no modification needed

PR #21394, opened 2025-12-31
Feat/update modal display tapc 4635

ZyaCon:
## Description

Ticket Reference: #TAPC-4635

### DIFF
--- a/packages/manager/apps/pci-instances/src/pages/instances/create/view-models/flavorsViewModel.ts
+++ b/packages/manager/apps/pci-instances/src/pages/instances/create/view-models/flavorsViewModel.ts
@@ -397,7 +397,9 @@ export const selectAvailableFlavorMicroRegions: Reader<
           regionalizedFlavorId,
         );
 
-        return regionalizedFlavor?.hasStock ? [regionalizedFlavor] : [];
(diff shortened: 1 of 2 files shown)


packages/manager/apps/pci-instances/src/pages/instances/create/view-models/flavorsViewModel.ts
```diff
@@ -397,7 +397,9 @@ export const selectAvailableFlavorMicroRegions: Reader<
           regionalizedFlavorId,
         );
 
-        return regionalizedFlavor?.hasStock ? [regionalizedFlavor] : [];
+        return regionalizedFlavor?.hasStock && regionalizedFlavor.quota
```
fredericvilcot, line 400: Missing optional chaining, no? 


```suggestion
        return regionalizedFlavor?.hasStock && regionalizedFlavor?.quota
```

yaallco: No need ! Undefined check is done first by regionalizedFlavor?.hasStock. If its true, the right condition after "&&" will know that regionalizedFlavor is not undefined

PR #21404, opened 2026-01-02
fix(zimbra): display organizations label in redirections page and man…

amine-daoussi:
ref: #PRDCOL-323

## Description
In Zimbra > redirections page :
- fix the display of organization label in the data grid.
- manage the filter by organization label.

image

### DIFF
--- a/packages/manager/apps/zimbra/src/data/hooks/redirection/useRedirections.ts
+++ b/packages/manager/apps/zimbra/src/data/hooks/redirection/useRedirections.ts
@@ -25,6 +25,7 @@ type UseRedirectionsParams = Omit<
   shouldFetchAll?: boolean;
 };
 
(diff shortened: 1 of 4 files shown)


packages/manager/apps/zimbra/src/data/hooks/redirection/useRedirections.ts
```diff
@@ -22,18 +22,48 @@ type UseRedirectionsParams = Omit<
   redirection?: string;
   redirectionTargetId?: string;
   organizationLabel?: string;
+  organizationId?: string;
   shouldFetchAll?: boolean;
 };
 
+export const useFetchAllPages = (
```
atefzed: I don't think we need this custom hook, we can keep it inside the component.

PR #21447, opened 2026-01-06
fix(exchange): check domain zones

atefzed:
ref: #PRDCOL-287

## Description

Ticket Reference: #PRDCOL-287

### DIFF
--- a/packages/manager/modules/exchange/src/domain/domain.service.js
+++ b/packages/manager/modules/exchange/src/domain/domain.service.js
@@ -40,6 +40,10 @@ export default class ExchangeDomains {
     );
   }
 
+  getDomainZones() {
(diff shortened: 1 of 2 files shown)


packages/manager/modules/exchange/src/domain/domain.service.js
```diff
@@ -40,6 +40,12 @@ export default class ExchangeDomains {
     );
   }
 
+  getDomainZones() {
+    return this.services.OvhHttp.get('/domain/zone', {
+      rootPath: 'apiv6',
+    });
+  }
```
ghyenne, line 45:
```suggestion
  getDomainZones() {
    return this.services.$http.get('/domain/zone').then(({data}) => data);
  }
```

amine-daoussi (approved):

PR #21486, opened 2026-01-07
feat(backup-agent): add vault consumption

tibs245:
ref: #BKP-505

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/modules/backup-agent/src/pages/billing/listing/_components/BillingPriceCell.components.tsx
+++ b/packages/manager/modules/backup-agent/src/pages/billing/listing/_components/BillingPriceCell.components.tsx
@@ -1,0 +1,26 @@
+import React from 'react';
+
+import { useQuery } from '@tanstack/react-query';
+
+import { OdsSkeleton } from '@ovhcloud/ods-components/react';
+
+import { DataGridTextCell } from '@ovh-ux/manager-react-components';
+
+import { useGetServiceConsumptionOptions } from '@/data/hooks/consumption/useServiceConsumption';
+import { VAULT_PLAN_CODE } from '@/module.constants';
+
+export type BillingPriceCellProps = { vaultId: string };
+
+export const BillingPriceCell = ({ vaultId }: BillingPriceCellProps) => {
+  const { data: priceText, isPending } = useQuery({
+    ...useGetServiceConsumptionOptions()(vaultId),
+    select: (consumptions) =>
+      consumptions?.find((consumption) => consumption.planCode === VAULT_PLAN_CODE)?.price.text,
+  });
+
+  if (isPending) {
+    return <OdsSkeleton className="w-10" />;
+  }
+
+  return <DataGridTextCell>{priceText ?? '-'}</DataGridTextCell>;
+};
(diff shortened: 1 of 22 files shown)


packages/manager/modules/backup-agent/src/data/api/services/consumption.ts
```diff
@@ -3,7 +3,7 @@ import { v6 } from '@ovh-ux/manager-core-api';
 import { ServiceConsumption } from '@/types/Consumption.type';
 import { getServiceConsumptionRoute } from '@/utils/apiRoutes';
 
-export const getServiceConsumption = async (serviceId: number) => {
+export const getServiceConsumption = async (serviceId: string | number) => {
```
vovh, line 6: i'm a bit curious, in which use case we can have either a string or a number as a serviceId ? Is it because of legacy code ?

tibs245: Yes it's to avoid have a breaking change. I not see it's a problem to have two type work so I extended with my functionnality

packages/manager/modules/backup-agent/public/translations/vaults/dashboard/Messages_fr_FR.json
```diff
@@ -17,5 +17,7 @@
   "general-information": "Informations générales",
   "help": "Aide",
   "type_billing": "Type de facturation",
+  "vault_price": "Soit {{vaultPrice}} HT/mois",
+  "without_engagement": "sans engagement",
```
pauldkn: I think both of these translations are more generic than just "vault"

packages/manager/modules/backup-agent/src/pages/billing/listing/_components/BillingPriceCell.components.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+import { useGetServiceConsumptionOptions } from '@/data/hooks/consumption/useServiceConsumption';
+import { VAULT_PLAN_CODE } from '@/module.constants';
+
+export type BillingPriceCellProps = { vaultId: string };
+
+export const BillingPriceCell = ({ vaultId }: BillingPriceCellProps) => {
+  const { data: consumptionData, isPending } = useQuery({
+    ...useGetServiceConsumptionOptions()(vaultId),
+  });
```
pauldkn: I find this pattern quite unusual, it adds a bit of complexity when we read it

(comparing to something like `const { data } = useServiceConsumption(serviceId)`)

packages/manager/modules/backup-agent/public/translations/vaults/dashboard/Messages_fr_FR.json
```diff
@@ -17,5 +17,6 @@
   "general-information": "Informations générales",
   "help": "Aide",
   "type_billing": "Type de facturation",
+  "vault_price": "Soit {{vaultPrice}} HT/mois",
```
pauldkn, line 20: Could be more generic with something like `price_equivalence_excluding_tax`

(might even be in common translations)

tibs245: We have tool to detect duplication. I think this work should be just after. But yes I understand your point

PR #21494, opened 2026-01-07
fix(object-storage): tag implementation and unit test

lionel95200x:
ref: #DATATR-2658

## Description

- Rework tag implementation and add unit test

Ticket Reference: #DATATR-2658


packages/manager/apps/pci-object-storage/src/pages/object-storage/storage/s3Id/replication/_components/form/useReplicationForm.hook.ts
```diff
@@ -137,7 +130,7 @@ export const useReplicationForm = ({
       deleteMarkerReplication:
         storages.ReplicationRuleDeleteMarkerReplicationStatusEnum.disabled,
       isReplicationApplicationLimited: false,
-      tags: {},
+      tags: [] as Tag[],
```
abullet33: Can we avoid to use "as"? Can't it be tags: []

abullet33 (approved):

PR #21500, opened 2026-01-07
fix(okms): no sorting on all datagrids

mathieumousnier:
## Description

Ticket Reference: #EK-2515


packages/manager/apps/okms/src/modules/key-management-service/data/api/utils.ts
```diff
@@ -1,17 +1,8 @@
 import { OKMS } from '@key-management-service/types/okms.type';
-import { OkmsServiceKey } from '@key-management-service/types/okmsServiceKey.type';
 
 export const defaultCompareFunction = (key: keyof OKMS) => (a: OKMS, b: OKMS) => {
   const aValue = a[key] || '';
   const bValue = b[key] || '';
 
   return (aValue as string).localeCompare(bValue as string);
 };
```
rjamet-ovh: this function is not used so you could delete it as well.

mathieumousnier: i removed the whole file

sefiani-ovh (approved):

PR #21506, opened 2026-01-07
feat(pci-instances): disable image type option if unavailable

lolottetheclash:
ref: #TAPC-5444

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-instances/src/pages/instances/create/view-models/imagesViewModel.ts
+++ b/packages/manager/apps/pci-instances/src/pages/instances/create/view-models/imagesViewModel.ts
@@ -1,34 +1,121 @@
+/* eslint-disable max-lines */
 /* eslint-disable max-lines-per-function */
 import { Deps } from '@/deps/deps';
 import {
+  TImageTypeName,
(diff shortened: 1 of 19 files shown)


packages/manager/apps/pci-instances/src/pages/instances/create/view-models/imagesViewModel.ts
```diff
@@ -138,6 +243,33 @@ const hasImageOsTypeStockAvailable = (
...
+      case 'baremetal-linux':
+      case 'bsd':
+      case 'linux':
+        return 'linux';
+      case 'windows':
+        return 'windows';
+    }
+  });
+
+const flavorAcceptsSelectedImageType = (
```
Tsiorifamonjena: suggestion : maybe we can rename this function to make the intention more clear ? like `isFlavorAcceptsSelectedImageType` or something like that

Tsiorifamonjena (commented):
Great work 💪🏽👏🏽

I juste have a little suggestion

ZyaCon (approved):

Issue #21416, opened 2026-01-05
[mrc1/mrc2/muk]: Survey button component


aottr:
### Have you already checked if a similar item is present on muk?

- [x] Yes, I have already checked the existing components/hooks/utils.

### What do you expect from this request?

Component

### Description

Requested is a Button-component (similar to guide, but just a button) which redirects the user to a survey depending on the manager language and app.

Some of the target applications are still using MRCv1 so the change will have to be in all 3 libraries. (MRCv1 and MRCv2 for now, others following)

### Where do you expect to use this?

Having a button component in coherent style + translations

### Do you have mock-up?

yes, internal

### When do you expect this to be delivered?

12-01-2026

### Additional Information

The links have a structure, but we can not guarantee that this structure is always the same. My proposal is to accept a `Record<string,string>` as prop for the links where the key is the manager language and the value the link to the survey. If no value is present for the currently selected language, we don't display the button.

Attribute `target="_blank"` is important for the link since it will always be external.

aottr:
Adjusted scope:
- add a preset enum that contains "survey"
- add an optional preset property to (Manager-) Button to conditionally change label / link of button for surveys
- add translations with button texts in the component package
- add an internal hook for generic link generation for surveys (depending on App-Prefix, Manager language)
- add a slot to the BaseLayout right to the Guide-Button

PR #21536, opened 2026-01-09
feat(muk): extend button for feedback survey

aottr:
gh-21416

ref: #MANAGER-20595

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager-ui-kit/src/commons/utils/TranslationHelper.ts
+++ b/packages/manager-ui-kit/src/commons/utils/TranslationHelper.ts
@@ -151,3 +151,20 @@ const createImportResolver =
  * producing i18next {@link ResourceLanguage} objects.
  */
 export const resolveTranslationModule = createImportResolver<ResourceLanguage>();
+
+/**
+ * Extracts the language code from a locale string.
+ * Converts locale format (e.g., 'fr_FR') to language code (e.g., 'fr').
+ *
+ * @param {string} locale - The locale string in format 'll_RR' or 'll-RR'.
+ * @returns {string} - The language code (e.g., 'fr', 'en', 'de').
+ *
+ * @example
+ * extractLanguageCode('fr_FR') // returns 'fr'
+ * extractLanguageCode('en_GB') // returns 'en'
+ * extractLanguageCode('de-DE') // returns 'de'
+ * extractLanguageCode('invalid') // returns 'fr'
+ */
+export const extractLanguageCode = (locale: string): string => {
+  return locale.split(/[-_]/)[0]?.toLowerCase() ?? 'fr';
+};
--- a/packages/manager-ui-kit/src/components/link/Link.component.tsx
+++ b/packages/manager-ui-kit/src/components/link/Link.component.tsx
@@ -12,10 +12,13 @@ import {
 
 import { NAMESPACES } from '@ovh-ux/manager-common-translations';
 
+import { extractLanguageCode } from '@/commons/utils/TranslationHelper';
+import { LinkProps, LinkType } from '@/components/link/Link.props';
 import { useAuthorizationIam } from '@/hooks/iam/useOvhIam';
 
-import { LinkProps } from './Link.props';
+import { generateSurveyLink } from './Link.util';
 import { LinkIcons } from './LinkIcons.component';
+import './translations';
 
 export const Link = <T extends ElementType = 'a'>({
   children,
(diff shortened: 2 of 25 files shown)


aboungnaseng-ovhcloud (dismissed):

packages/manager-ui-kit/src/commons/utils/TranslationHelper.ts
```diff
@@ -151,3 +151,20 @@ const createImportResolver =
...
+ *
+ * @example
+ * extractLanguageCode('fr_FR') // returns 'fr'
+ * extractLanguageCode('en_GB') // returns 'en'
+ * extractLanguageCode('de-DE') // returns 'de'
+ * extractLanguageCode('invalid') // returns 'fr'
+ */
+export const extractLanguageCode = (locale: string): string => {
+  return locale.split(/[-_]/)[0]?.toLowerCase() ?? 'fr';
+};
```
anooparveti, line 170: **question:** Are we sure this is the right way? What if French user uses `en_GB` language in his Manager? Is Survey link based on subs or just a display language on survey forms?

packages/manager-ui-kit/src/components/link/Link.component.tsx
```diff
@@ -24,11 +27,29 @@ export const Link = <T extends ElementType = 'a'>({
...
 }: LinkProps<T>) => {
-  const { t } = useTranslation(NAMESPACES.IAM);
+  const { t, i18n } = useTranslation([NAMESPACES.IAM, 'link']);
   const { isAuthorized } = useAuthorizationIam(iamActions || [], urn || '', !disableIamCheck);
 
+  if (type === LinkType.survey && surveyApplicationKey) {
+    return (
+      <OdsLink
+        {...odsLinkProps}
+        /* eslint-disable-next-line react-hooks/rules-of-hooks */
```
anooparveti: **question:** Why disabling the linting rule?

packages/manager-ui-kit/src/components/link/LinkIcons.component.tsx
```diff
@@ -29,6 +29,13 @@ export function LinkIcons({ type, children }: LinkIconsProps) {
           <Icon name="arrow-right" />
         </>
       );
+    case LinkType.survey:
+      return (
+        <>
+          {children}
+          <Icon name="emoticon-smile" />
```
anooparveti:
```suggestion
          <Icon name="emoticon-smile" aria-hidden="true" />
```

packages/manager-ui-kit/src/components/link/Link.util.ts
anooparveti, line 1: **nitpick:** I dont see any usage of react state, effect, memo and so on. This hook can be a normal util function added in Link.util.ts file under `src/components/link` folder. And base URL of survey can be added in Link.constants.ts.

packages/manager-ui-kit/src/commons/utils/TranslationHelper.ts
aottr: If the user uses `en_GB` in the manager, it will return the language code `en` and only fall-back to `fr` if the language we get from `i18n` can not be found. (which shouldn't be the case under normal usage)

packages/manager-ui-kit/src/components/link/Link.component.tsx
aottr: This linting warning is because we use a "hook" conditionally. The linting is disabled because:
a) It's not really a react-hook but rather a helper function in hook-style
b) It does not make sense to call this function un-conditionally since then we would have to generate a survey link without existing App-key / modify the "hook" to accept undefined app-key which should be wasted compute.

packages/manager-ui-kit/src/components/link/Link.util.ts
aottr: If we don't want to use the hook anywhere else, sure. It's private anyway. So far this is just mirrored with MRC1/MRC2

packages/manager-ui-kit/src/components/link/Link.util.ts
```diff
@@ -0,0 +1,22 @@
+import { LINK_SURVEY_BASE_URL } from './Link.constants';
+import { GenerateSurveyLinkProps } from './Link.props';
+
+/**
+ * Generate a survey link for a given application key and language
+ * If the language is not provided or is 'fr', the link will be generated without the language suffix
+ * @example
+ * const surveyLink = useSurveyLink({ applicationKey: 'COM', languageCode: 'de' });
```
anooparveti:
```suggestion
 * const surveyLink = generateSurveyLink({ applicationKey: 'COM', languageCode: 'de' });
```

packages/manager-ui-kit/src/components/link/__tests__/Link.util.spec.ts
```diff
@@ -0,0 +1,47 @@
+import { describe, expect, it } from 'vitest';
+
+import { LINK_SURVEY_BASE_URL } from '../Link.constants';
+import { generateSurveyLink } from '../Link.util';
+
+describe('useSurveyLink', () => {
```
anooparveti:
```suggestion
describe('generateSurveyLink', () => {
```

anooparveti (approved):

chipp972 (approved):

PR #21538, opened 2026-01-09
Fix/instances sshkeys reactivity

fredericvilcot:
## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-instances/src/pages/instances/create/view-models/sshKeysViewModel.ts
+++ b/packages/manager/apps/pci-instances/src/pages/instances/create/view-models/sshKeysViewModel.ts
@@ -1,27 +1,6 @@
-import { Deps } from '@/deps/deps';
 import { TSshKey } from '@/domain/entities/configuration';
-import { Reader } from '@/types/utils.type';
 
-export type TSshKeyData = {
-  label: string;
-  value: string;
-};
+export type TSshKeyData = { label: string; value: string };
 
-type TSshKeysSelectData = (
-  projectId: string,
-  microRegion: string,
-) => TSshKeyData[];
-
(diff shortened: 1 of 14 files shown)


packages/manager/apps/pci-instances/src/pages/instances/create/view-models/sshKeysViewModel.ts
```diff
@@ -1,27 +1,10 @@
...
-  const { configurationPort } = deps;
-
-  const sshKeys = configurationPort.selectSshKeys(projectId, microRegion) ?? [];
-
-  return mapSshKeysData(sshKeys);
-};
+/**
+ * Pure transformation function to use as `select` option in useQuery.
+ * Transforms SSH keys from domain entities to display data.
+ */
```
lolottetheclash: I think we can get rid of this comment as it's already clear

packages/manager/apps/pci-instances/src/types/selectableQuery.type.ts
```diff
@@ -0,0 +1,17 @@
+/**
+ * Options type for hooks that require a select function.
+ * Enforces hexagonal architecture by making data transformation mandatory.
+ *
```
lolottetheclash: I would just update the file name with `querySelectOptions`, I think that selectableQuery is not very clear

lolottetheclash (approved):
just details ;)

packages/manager/apps/pci-instances/src/types/selectableQuery.type.ts
Tsiorifamonjena: yes, how about the comments ? I remove it also like your suggestion above.

PR #21565, opened 2026-01-12
feat(pci-instances): replace ssh keys call by aggregated call

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-5545

### DIFF
--- a/packages/manager/apps/pci-instances/src/adapters/tanstack/configuration/right/dto.type.ts
+++ b/packages/manager/apps/pci-instances/src/adapters/tanstack/configuration/right/dto.type.ts
@@ -1,7 +1,4 @@
 export type TSshKeyDTO = {
-  id: string;
-  fingerPrint: string;
   name: string;
-  publicKey: string;
-  region: string;
+  regions: string[];
 };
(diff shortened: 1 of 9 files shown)


packages/manager/apps/pci-instances/src/adapters/tanstack/configuration/right/dto.type.ts
```diff
@@ -5,3 +5,8 @@ export type TSshKeyDTO = {
   publicKey: string;
   region: string;
 };
+
+export type TSshAggregatedDTO = {
```
lolottetheclash: I think that the `aggregated` notion belongs to the backend, maybe just `TSshKeyDTO` is enough, as you want

packages/manager/apps/pci-instances/src/adapters/tanstack/configuration/right/mapper.ts
```diff
@@ -9,5 +9,16 @@ export const mapSshKeyDtoToSshKeyEntity = (
...
+  }));
+
+export const mapSshKeyAggregatedDtoToSshKeyEntity = (
+  sshKeysAggregatedDTO: TSshAggregatedDTO[],
+): TSshKey[] =>
+  sshKeysAggregatedDTO.map(({ name, regions }) => ({
+    id: name,
+    name,
+    regions,
+    fingerPrint: '',
```
lolottetheclash: do we still need `fingerPrint` & `publicKey`?

ZyaCon (approved):

packages/manager/apps/pci-instances/src/adapters/tanstack/configuration/right/mapper.ts
Tsiorifamonjena: not anymore

PR #21568, opened 2026-01-12
fix(database): rework matching subregion to get region Id

pierre-pelloquin:
## Description
-  rework matching subregion to get region Id

Ticket Reference: #DATATR-2358

### DIFF
--- a/packages/manager/apps/pci-databases-analytics/src/pages/services/create/_components/OrderFunnel.component.tsx
+++ b/packages/manager/apps/pci-databases-analytics/src/pages/services/create/_components/OrderFunnel.component.tsx
@@ -115,15 +115,8 @@ const OrderFunnel = ({
         })),
       };
       if (data.network.type === database.NetworkTypeEnum.private) {
-        // Get the subnet Object to get the right subRegion
-        const subnetObject = model.lists.subnets.find(
-          (s) => s.id === data.network.subnetId,
-        );
         // endpoint does not expect the network id, but the linked openstackId instead
-        const networkOpenstackId = model.result.network.network.regions.find(
-          (r) => r.region === subnetObject?.ipPools[0]?.region,
-        )?.openstackId;
-        serviceInfos.networkId = networkOpenstackId;
+        serviceInfos.networkId = model.result.network.networkOpenstackId;
         serviceInfos.subnetId = data.network.subnetId;
       }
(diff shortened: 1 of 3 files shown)


packages/manager/apps/pci-databases-analytics/src/pages/services/create/_components/OrderFunnel.component.tsx
```diff
@@ -115,9 +115,13 @@ const OrderFunnel = ({
...
       };
       if (data.network.type === database.NetworkTypeEnum.private) {
+        // Get the subnet Object to get the right subRegion
+        const subnetObject = model.lists.subnets.find(
+          (s) => s.id === data.network.subnetId,
+        );
         // endpoint does not expect the network id, but the linked openstackId instead
         const networkOpenstackId = model.result.network.network.regions.find(
-          (r) => r.region.includes(data.region),
+          (r) => r.region === subnetObject.ipPools[0].region,
```
lionel95200x: Are we sure that we always have an item ?
```suggestion
          (r) => r.region === subnetObject.ipPools[0]?.region,
```

pierre-pelloquin: I suppose, the Select for the subnet don't check if there is one (src/components/order/cluster-options/NetworkOptions.components.tsx, line 142)
image

jperchoc: I’m not sure this will work reliably.

If I remember correctly, `data.region` corresponds to the region selected for the database, which can be one of the following values coming from the `database/capabilities/regions` endpoint:

`"SGP", "EU-WEST-PAR", "EU-SOUTH-MIL", "SBG", "DE", "WAW", "UK", "BHS", "GRA"`

However, the regions used in `ipPools` follow a different format, for example:

```
{
  "region": "GRA11"
}
```


This mismatch seems to be the root of the issue.

From what I understand, the current logic selects the first `ipPool` whose region matches the service region. This can lead to incorrect behavior. For example:

- The service is configured in the `GRA` region
- The user has two subnets:
  - one in `GRA5`
  - one in `GRA11`

In this case, we will always pick the first matching subnet, regardless of which one the user actually selected.

jperchoc (approved):

PR #21580, opened 2026-01-12
feat(muk): muk available without react context

aboungnaseng-ovhcloud:
ref: #MANAGER-20684

## Description

- muk available without react context

Ticket Reference: #MANAGER-20684

### DIFF
--- a/packages/manager-ui-kit/src/components/error-boundary/ErrorBoundary.component.tsx
+++ b/packages/manager-ui-kit/src/components/error-boundary/ErrorBoundary.component.tsx
@@ -22,12 +22,14 @@ export const ErrorBoundary = ({
   isRouteShellSync = false,
 }: ErrorBoundaryProps) => {
   const error = useRouteError() as ResponseAPIError;
-  const shell = useContext(ShellContext)?.shell;
+
+  const context = useContext(ShellContext);
+  const { shell } = context;
   const navigateToHomePage = () => {
-    shell?.navigation.navigateTo(redirectionApp, '', {});
+    shell?.navigation?.navigateTo(redirectionApp, '', {});
   };
   const errorObject =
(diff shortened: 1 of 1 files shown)


helabenkhalfallah (dismissed):

packages/manager-ui-kit/src/components/error-boundary/ErrorBoundary.component.tsx
```diff
@@ -37,12 +39,12 @@ export const ErrorBoundary = ({
       : {};
 
   const reloadPage = () => {
-    shell?.navigation.reload();
+    shell?.navigation?.reload();
```
anooparveti: **nitpick:** If `reload` is not possible from shell, we should maybe do a window..location.reload() to reload the page.
Also we should document the behaviour regarding error-boundary if it's used outside of the Container/Shell.

packages/manager/core/shell-client/src/hooks/useOvhTracking.ts
```diff
@@ -149,10 +149,25 @@ export const usePageTracking = (): TrackingPageParams | undefined => {
...
+  const context = React.useContext(ShellContext);
+  const { shell, tracking, environment } = context;
+  const isEmptyObject =
+    environment &&
+    typeof environment === 'object' &&
+    Object.keys(environment).length === 0;
+
+  if (isEmptyObject) {
+    return DefaultTrackingContext;
+  }
```
anooparveti, line 169: **question:** Isn't it possible to avoid setting an empty options like this?

packages/manager-ui-kit/src/components/error-boundary/ErrorBoundary.component.tsx
aboungnaseng-ovhcloud: 👍

PR #21582, opened 2026-01-12
fix(iam): disable previous date in permanent tokens

MarcMonchablon:
ref: #MAIAM-70

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/identity-access-management/src/pages/permanentTokens/components/ExpiryDateInput.component.tsx
+++ b/packages/manager/apps/identity-access-management/src/pages/permanentTokens/components/ExpiryDateInput.component.tsx
@@ -38,6 +38,7 @@ export default function ExpiryDateInput({
   const [minutes, setMinutes] = useState<number>(1);
   const [hours, setHours] = useState<number>(1);
   const [days, setDays] = useState<number>(1);
+  const currentDate = new Date();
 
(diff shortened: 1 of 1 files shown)


packages/manager/apps/identity-access-management/src/pages/permanentTokens/components/ExpiryDateInput.component.tsx
```diff
@@ -38,6 +38,7 @@ export default function ExpiryDateInput({
   const [minutes, setMinutes] = useState<number>(1);
   const [hours, setHours] = useState<number>(1);
   const [days, setDays] = useState<number>(1);
+  const currentDate = useMemo(() => new Date(), []);
```
darsene: Why use memoization here,  it won't behave as expected

MarcMonchablon: I was afraid that having a new value for each render passed as prop for the DatePicker would be wasteful.

yjaaouane: @MarcMonchablon  this can affect users who keep the Manager open overnight 🙂(23h59 !!)

PR #21583, opened 2026-01-12
fix(iam): fix token expiry date not removable

MarcMonchablon:
ref: #MAIAM-69

## Description

Ticket Reference: #...


packages/manager/apps/identity-access-management/src/pages/permanentTokens/edit/PermanentTokensEdit.page.tsx
```diff
@@ -117,7 +117,10 @@ export default function PermanentTokensAdd() {
       if (expiryModel.mode === 'duration' && expiryModel.expiresIn !== null) {
         payload.expiresIn = expiryModel.expiresIn;
       }
+    } else {
+      payload.expiresAt = null;
     }
```
darsene, line 122: niptick: should be move into a util dedicated function

PR #21610, opened 2026-01-13
feat(pci.cold-archive): remove unused angular code for cold archive

abullet33:
ref: #DATATR-2867

## Description

Ticket Reference: #...


package.json
```diff
@@ -5,7 +5,222 @@
   "description": "OVH Control Panel also known as Manager",
   "license": "BSD-3-Clause",
   "workspaces": {
-    "packages": []
+    "packages": [
```
lolottetheclash: are you sure about these changes? 
as in manager docs we have:
```
🚫 Do not modify or commit changes to workspaces.packages.
Any local edits will be overwritten or may break the hybrid dependency graph resolution.
```

abullet33: Ouspi yep totally, nice catch. 
We remove that

lolottetheclash (dismissed):

anooparveti (dismissed):

PR #21620, opened 2026-01-14
Feat/cart error message tapc 5595

ZyaCon:
## Description

Ticket Reference: #tapc-5595    
image

### DIFF
--- a/packages/manager/apps/pci-instances/src/pages/instances/create/hooks/useCartItems.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/create/hooks/useCartItems.tsx
@@ -5,6 +5,7 @@ import { FlavorDetails } from '../components/cart/FlavorDetails.component';
 import { useCatalogPrice } from '@ovh-ux/muk';
 import { useInstanceCreation } from './useInstanceCreation';
 import CartOptionDetailItem from '../components/cart/CartOptionDetailItem.component';
+import Banner from '@/components/banner/Banner.component';
 
 export type TCartItem = {
   id: string;
@@ -39,6 +40,7 @@ export const useCartItems = (): TCartItems => {
     isCreationEnabled,
     isCreatingInstance,
     handleCreateInstance,
+    errorMessage,
   } = useInstanceCreation();
 
   const {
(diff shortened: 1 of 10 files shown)


packages/manager/apps/pci-instances/src/pages/instances/create/hooks/useInstanceCreation.ts
```diff
@@ -168,9 +169,11 @@ export const useInstanceCreation = (): TInstanceCreation => {
       ? error.response?.data.message
       : error.message;
     // TODO: update with new error specs to come
```
lolottetheclash: you can remove my previous comment here

packages/manager/apps/pci-instances/src/pages/instances/create/hooks/useInstanceCreation.ts
```diff
@@ -168,9 +169,11 @@ export const useInstanceCreation = (): TInstanceCreation => {
...
-    console.error(errorMessage);
+    setErrorMessage(errorMessage ?? null);
   };
 
+  const [errorMessage, setErrorMessage] = useState<string | null>(null);
```
lolottetheclash: I think you can remove this state and get `error` which is returned by useCreateInstance mutation

you can also get rid of `handleError` which was temporary

lolottetheclash (changes requested):

packages/manager/apps/pci-instances/src/pages/instances/create/hooks/useCartItems.tsx
```diff
@@ -189,6 +196,17 @@ export const useCartItems = (): TCartItems => {
 
   const ActionButtons = (
     <>
+      {errorMessage && (
```
Tsiorifamonjena, line 208: How about adjust the generic `/components/banner/Banner.component.tsx` component to match your need ?

yaallco: Thanks for you comment, i updated my code to use Banner component :)

packages/manager/apps/pci-instances/src/pages/instances/create/hooks/useInstanceCreation.ts
```diff
@@ -218,10 +212,17 @@ export const useInstanceCreation = (): TInstanceCreation => {
     backupConfigurationPrices,
   };
 
+  const errorMessage = error
```
aTurmo, line 215: Question: Isn't this (extracting the error message) done in multiple place? If yes you could add a service method somewhere to mutualize things

yaallco: It may be possible to extract a service somewhere but for this task I think it wouldn't be the right place. For useCreateInstance we extract the error only here

PR #21649, opened 2026-01-15
feat(dedicated-server): edit view

mhelhali-soufien:
ref: #MANAGER-20561

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/dedicated-servers/src/components/manageView/manageViewButton.tsx
+++ b/packages/manager/apps/dedicated-servers/src/components/manageView/manageViewButton.tsx
@@ -31,7 +31,7 @@ import {
   STANDARD_VIEW_ID,
 } from './manageView.constants';
 import { ViewType } from './types';
-import ManageViewDrawer from './manageViewDrawer';
+import ManageViewDrawer from './manageViewDrawer-v2';
 import { useGetViewsPreferences } from '@/hooks/manage-views/useGetViewPreferences';
 
 export const ManageViewButton = () => {
@@ -40,6 +40,7 @@ export const ManageViewButton = () => {
   const [isOpoverOpen, setIsPopoverOpen] = useState<boolean>(false);
   const [isDrawerOpen, setIsDrawerOpen] = useState<boolean>(false);
   const [currentView, setCurrentView] = useState<ViewType>(null);
+  const [selectedView, selectView] = useState<ViewType>(null);
 
   const { preferences, error, isLoading } = useGetViewsPreferences({
     key: PREFERENCES_KEY,
@@ -98,7 +99,16 @@ export const ManageViewButton = () => {
               {t('current_view')}
             </Text>
 
-            <Button role="menuitem" variant={BUTTON_VARIANT.ghost}>
+            <Button
+              onClick={() => {
+                selectView(currentView);
+                setIsPopoverOpen(false);
+                setIsDrawerOpen(true);
+              }}
+              role="menuitem"
+              variant={BUTTON_VARIANT.ghost}
+              disabled={currentView?.id === views[0]?.id}
+            >
               <Icon name={ICON_NAME.pen} aria-hidden={true} />
               <span>{t('configure_current_view')}</span>
             </Button>
@@ -107,6 +117,7 @@ export const ManageViewButton = () => {
               role="menuitem"
               variant={BUTTON_VARIANT.ghost}
               color={BUTTON_COLOR.critical}
+              disabled={currentView?.id === views[0]?.id}
             >
               <Icon name={ICON_NAME.trash} aria-hidden={true} />
               <span>{t('delete_current_view')}</span>
@@ -144,6 +155,7 @@ export const ManageViewButton = () => {
               role="menuitem"
               variant={BUTTON_VARIANT.ghost}
               onClick={() => {
+                selectView(null);
                 setIsPopoverOpen(false);
                 setIsDrawerOpen(true);
               }}
@@ -168,7 +180,7 @@ export const ManageViewButton = () => {
         </PopoverContent>
       </Popover>
       <ManageViewDrawer
-        view={null}
+        view={selectedView}
         views={views}
         isOpen={isDrawerOpen}
         handleCancel={closeDrawer}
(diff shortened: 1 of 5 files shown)


packages/manager/apps/dedicated-servers/src/components/manageView/manageViewDrawer-v2.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  useEffect(() => {
+    setEditingView({
+      name: view?.name || t('new_view'),
+      id: view?.id || `view-${views.length}`,
+      default: view?.default,
+    });
+  }, [isOpen, view, views]);
+
+  const handleNameChange = (value: string) => {
```
b-fritz-ovh, line 49: I don't know. useMemo can to not re-instanciate the functions can be beneficial?

if yes, use it lines 56 & 63 too?

packages/manager/apps/dedicated-servers/src/components/manageView/manageViewDrawer-v2.tsx
```diff
@@ -0,0 +1,135 @@
...
+
+  const handelSetDefault = () => {
+    setEditingView({
+      ...editingView,
+      default: !editingView.default,
+    });
+  };
+
+  const saveViewChanges = () => {
+    const currentDefault = views.find((_view) => _view.default);
```
b-fritz-ovh, line 64: `const currentDefault = views.find(({default}) => default);` ?

packages/manager/apps/dedicated-servers/src/components/manageView/manageViewDrawer-v2.tsx
```diff
@@ -0,0 +1,135 @@
...
+      className={`fixed top-0 right-0 h-full w-96 z-50
+        transition-transform duration-300 ease-in-out
+        ${isOpen ? 'translate-x-0' : 'translate-x-full'}`}
+    >
+      {isOpen && (
+        <aside className="h-full w-full bg-white shadow-lg relative flex flex-col">
+          <div
+            onClick={handleCancel}
+            className="absolute top-9 -left-12 h-12 w-12 z-50 bg-white rounded-l-md shadow-md flex items-center justify-center border-0 outline-none focus:outline-none"
+            aria-label="Fermer le panel"
```
b-fritz-ovh: no string hard coded?

packages/manager/apps/dedicated-servers/src/components/manageView/manageViewDrawer-v2.tsx
```diff
@@ -0,0 +1,135 @@
...
+  views: ViewType[];
+  view?: ViewType;
+  isOpen: boolean;
+  handleDismiss: () => void;
+  handleConfirm: () => void;
+  handleCancel: () => void;
+};
+export const ManageViewDrawer = ({
+  views,
+  view,
```
b-fritz-ovh, line 28: `view = {},` and remove all the optional chainnings?

qpavy: This component should manage the edit but also view creation so I think we should keep like that, it is easier to manage new view

packages/manager/apps/dedicated-servers/src/components/manageView/manageViewDrawer-v2.tsx
qpavy: As we call this function only on click, not sure it will change something to use useCallback. (useMemo is not a good choice anyway here)

packages/manager/apps/dedicated-servers/src/components/manageView/manageViewDrawer-v2.tsx
qpavy: Yes should be translated

packages/manager/apps/dedicated-servers/src/components/manageView/manageViewButton.tsx
```diff
@@ -31,7 +31,7 @@ import {
   STANDARD_VIEW_ID,
 } from './manageView.constants';
 import { ViewType } from './types';
-import ManageViewDrawer from './manageViewDrawer';
+import ManageViewDrawer from './manageViewDrawer-v2';
```
aderghamov, line 34: If manageViewDrawer is no longer used, than remove it and rename manageViewDrawer-v2 into manageViewDrawer

packages/manager/apps/dedicated-servers/src/components/manageView/manageViewDrawer-v2.tsx
aderghamov: The purpose of using `useCallback` is not to regenerate the implementation of the callback at each render. (provided that you have a proper list of dependency to base your implementation on)

packages/manager/apps/dedicated-servers/src/components/manageView/manageViewDrawer-v2.tsx
```diff
@@ -0,0 +1,135 @@
...
+  const handelSetDefault = () => {
+    setEditingView({
+      ...editingView,
+      default: !editingView.default,
+    });
+  };
+
+  const saveViewChanges = () => {
+    const currentDefault = views.find((_view) => _view.default);
+    if (editingView.default) currentDefault.default = null;
```
aderghamov: Provide a default value for `currentDefault` if `views.find(({default}) => default)`  returns undefined.

packages/manager/apps/dedicated-servers/src/components/manageView/manageViewButton.tsx
qpavy: I will do it in my next PR. We keep both for now to avoid unecessary conflicts until I merge all the devs and remove v1

packages/manager/apps/dedicated-servers/src/components/manageView/manageViewDrawer-v2.tsx
qpavy: not possible to use default as the term is use by javascript

PR #21698, opened 2026-01-16
feat(iam): add timepicker to permanent token creation form

MarcMonchablon:
ref: #MAIAM-68

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/identity-access-management/src/utils/expiryDateUtils.spec.ts
+++ b/packages/manager/apps/identity-access-management/src/utils/expiryDateUtils.spec.ts
@@ -1,0 +1,33 @@
+import { vi, describe, expect, it } from 'vitest';
+
+import { extractTimeFromDate, updateDateTime } from './expiryDateUtils';
+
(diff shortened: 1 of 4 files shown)


packages/manager/apps/identity-access-management/src/utils/expiryDateUtils.spec.ts
```diff
@@ -0,0 +1,37 @@
...
+  it('extract the hours and minutes of the given date', () => {
+    const date = new Date('Fri Jan 16 2026 18:24:31 GMT+0100');
+    expect(extractTimeFromDate(date)).toMatch(/\d\d:\d\d/);
+  });
+});
+
+describe('updateDateTime', () => {
+  beforeEach(() => {
+    vi.unstubAllEnvs();
+  });
```
darsene: still needed?

MarcMonchablon: Good catch, no, It's a mistake, removed it :)

PR #21702, opened 2026-01-17
feat(*):  Nx and future tools safe install case of OS

helabenkhalfallah:
## Description

Some CLI tools (e.g. `nx`) are not supported or stable on all developer machines.  
This PR introduces a **dynamic tools injection strategy** to install such tools **only when the current OS is supported**, and optionally **only in CI**, without creating `yarn.lock` noise.

### ✅ What’s implemented

- Added a **tools catalog** defining:
  - the tool version
  - supported OS for developers
  - supported OS for CI

Example `tools-catalog.json`:

```json
{
  "nx": {
    "version": "22.3.3",
    "supportedOs": ["win32", "darwin"],
    "ciSupportedOs": ["linux"]
  }
}
```

The OS check relies on `os.platform()` (Node native identifiers) to avoid any normalization logic.

### ✅ Injection mechanism

During Yarn `preinstall`, supported tools are injected into the root `package.json` as:

```json
{
  "devDependencies": {
    "nx": "22.3.3"
  }
}
```

Rules:
- **CI environment** → use `ciSupportedOs`
- **Local environment** → use `supportedOs`
- If OS is not supported → tool injection is skipped

### ✅ Lockfile stability

`yarn.lock` is committed with `nx` already present, ensuring:
- deterministic resolutions
- **no lockfile diff** during dynamic injection
- no developer `yarn.lock` changes to commit

> Note: if a package is only in `yarn.lock` but not referenced by any `package.json`, Yarn does not install it — so keeping it in the lockfile is harmless.

### ✅ Runner fallback

If the user explicitly requests `--runner nx` but it is not allowed on the current OS/env, the CLI automatically falls back to:

```text
turbo
```

Tests case: https://github.com/ovh/manager/pull/21702#issuecomment-3764206673

---

Ticket Reference: **#MANAGER-20410**

### DIFF
--- a/packages/manager-tools/manager-pm/bin/manager-pm.js
+++ b/packages/manager-tools/manager-pm/bin/manager-pm.js
@@ -33,11 +33,13 @@ import {
   testModule,
 } from '../src/kernel/helpers/tasks-helper.js';
 import { startApp } from '../src/kernel/pnpm/pnpm-start-app.js';
+import { loadToolsCatalog } from '../src/kernel/utils/catalog-utils.js';
 import { logger, setLoggerMode } from '../src/kernel/utils/log-manager.js';
 import {
   attachCleanupSignals,
   handleProcessAbortSignals,
 } from '../src/kernel/utils/process-utils.js';
+import { isToolAllowed } from '../src/kernel/utils/tools-utils.js';
 import {
   clearRootWorkspaces,
   updateRootWorkspacesFromCatalogs,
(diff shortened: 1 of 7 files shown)


helabenkhalfallah:
## If the current OS is not in the known supported list

Change from

```
{
  "nx": {
    "version": "22.3.3",
    "supportedOs": [ "win32", "darwin"],
    "ciSupportedOs": [ "linux"]
  }
}
```

To:
```
{
  "nx": {
    "version": "22.3.3",
    "supportedOs": [ "win32",],
    "ciSupportedOs": [ "linux"]
  }
}
```

Fallback to turbo by default:

image

image

image

## Case of supported OS

image

No lock change after the injection: 
image

We can successfully use the tool if the OS is supported:

image

image

## Case of CI

image

image

image

tibs245 (approved):
Alway a pleasure to read 👌

PR #21714, opened 2026-01-19
feat(pci-kubernetes): product availability endpoint integration

Taelar:
## Description

Ticket Reference: #TAPC-5548

### DIFF
--- a/packages/manager/apps/pci-kubernetes/public/translations/common/Messages_de_DE.json
+++ b/packages/manager/apps/pci-kubernetes/public/translations/common/Messages_de_DE.json
@@ -96,5 +96,5 @@
   "common_continent_label_ASIA": "Asien",
   "common_continent_label_EU": "Europe",
   "common_continent_label_NA": "Nordamerika",
-  "common_continent_label_US": "Vereinigte Staaten von Amerika"
+  "common_continent_label_US": "USA"
 }
--- a/packages/manager/apps/pci-kubernetes/public/translations/common/Messages_en_GB.json
+++ b/packages/manager/apps/pci-kubernetes/public/translations/common/Messages_en_GB.json
@@ -96,5 +96,5 @@
   "common_continent_label_ASIA": "Asia",
   "common_continent_label_EU": "Europe",
   "common_continent_label_NA": "North America",
-  "common_continent_label_US": "United States of America"
+  "common_continent_label_US": "USA"
 }
--- a/packages/manager/apps/pci-kubernetes/public/translations/common/Messages_fr_CA.json
+++ b/packages/manager/apps/pci-kubernetes/public/translations/common/Messages_fr_CA.json
@@ -96,5 +96,5 @@
   "common_continent_label_ASIA": "Asie",
   "common_continent_label_EU": "Europe",
   "common_continent_label_NA": "Amérique du Nord",
-  "common_continent_label_US": "États-Unis d'Amérique"
+  "common_continent_label_US": "États-Unis"
 }
--- a/packages/manager/apps/pci-kubernetes/public/translations/common/Messages_it_IT.json
+++ b/packages/manager/apps/pci-kubernetes/public/translations/common/Messages_it_IT.json
@@ -96,5 +96,5 @@
   "common_continent_label_ASIA": "Asia",
   "common_continent_label_EU": "Europa",
   "common_continent_label_NA": "Nord America",
-  "common_continent_label_US": "Stati Uniti d'America"
+  "common_continent_label_US": "USA "
 }


packages/manager/apps/pci-kubernetes/src/adapters/api/productAvailability/dto.types.ts
```diff
@@ -0,0 +1,28 @@
+export type TDeploymentModeDTO = 'region-3-az' | 'region' | 'localzone';
+
+export type TContinentCodeDTO = 'AF' | 'ASIA' | 'EU' | 'NA' | 'US';
+
+export type TPlanCodeDTO =
```
aTurmo, line 5: could be enum ?

packages/manager/apps/pci-kubernetes/src/adapters/api/productAvailability/mapper.spec.ts
```diff
@@ -0,0 +1,299 @@
...
+      expect(result.entities.microRegions.allIds).toEqual(['GRA9']);
+
+      const macroRegion = result.entities.macroRegions.byId.get('GRA');
+      expect(macroRegion).toBeDefined();
+      expect(macroRegion?.name).toBe('GRA');
+      expect(macroRegion?.countryCode).toBe('fr');
+      expect(macroRegion?.continentCode).toBe('EU');
+      expect(macroRegion?.enabled).toBe(true);
+      expect(macroRegion?.microRegionIds).toEqual(['GRA9']);
+      expect(macroRegion?.plans).toEqual(['mks.free.hour.consumption']);
```
aTurmo: Suggestion: combine all those expect with a `toEqual(fullObject)` or `toMatchObject(partialObject)`
Same with the following expects

packages/manager/apps/pci-kubernetes/src/adapters/api/productAvailability/__tests__/mapper.spec.ts
aTurmo, line 1: specs not yet in a __tests__ folder ?

packages/manager/apps/pci-kubernetes/src/domain/entities/productAvailability.ts
Taelar, line 1: Nota bene : Delete this file

packages/manager/apps/pci-kubernetes/src/adapters/tanstack/queryKey.ts
aTurmo, line 1: Small issue: two file named 'queryKey.ts' with only one folder of difference is not the best I think even if I get why you did that.

packages/manager/apps/pci-kubernetes/src/adapters/api/productAvailability/dto.types.ts
Taelar: I think it could but since I'm describing a DTO, I think a union type is a better option, do you think an enum would suit this typing more ?

packages/manager/apps/pci-kubernetes/src/adapters/api/productAvailability/__tests__/mapper.spec.ts
Taelar: I'm fine with that, but I felt like we were not using tests folders until then, are we supposed to, now ? :sweat_smile:

packages/manager/apps/pci-kubernetes/src/adapters/api/productAvailability/dto.types.ts
yaallco: Question about that, when do you use unions or enum ?

Taelar: For my own case, I almost never use enums, I prefer creating 'as const' arrays and inferring the typing from this. Though it's a personal preference

aTurmo: as you wish

packages/manager/apps/pci-kubernetes/src/adapters/api/productAvailability/__tests__/mapper.spec.ts
```diff
@@ -0,0 +1,303 @@
...
+
+      const result = mapProductAvailabilityToEntity(mockDto);
+
+      expect(result.entities.macroRegions.allIds).toEqual(['GRA']);
+      expect(result.entities.microRegions.allIds).toEqual(['GRA9', 'GRA11']);
+
+      const macroRegion = result.entities.macroRegions.byId.get('GRA');
+      expect(macroRegion).toMatchObject({
+        microRegionIds: ['GRA9', 'GRA11'],
+        enabled: true, // Should be true because at least one micro-region is enabled
```
aTurmo: Tests should make that clear instead of having a comment

aTurmo: And I think you already have tests testing this. So you can remove the comment. You could event remove the `enabled: true`since you specifically test for tit elsewhere

Taelar: You're right about the comment, i removed it and change the title test to make it clearer
About the second point, this is the only test where I test the case in which one regionDTO is disabled and the other is enabled to ensure the resulting macroRegion is enabled, so I think it is still relevant :slightly_smiling_face:

PR #21736, opened 2026-01-19
feat(pci-instances): add disk storage details for cpu and gpu

ZyaCon:
ref: #TAPC-5387

## Description

Ticket Reference: #...    

## Additional Information
image


packages/manager/apps/pci-instances/src/pages/instances/create/components/cart/FlavorDetails.component.tsx
```diff
@@ -40,6 +41,9 @@ export const FlavorDetails: FC<TFlavorDetails> = ({ quantity, flavor }) => {
         <Text className="font-semibold text-[--ods-color-heading]">
           {flavor.vCore} {t('pci_instance_creation_cart_flavor_vCore')}
         </Text>
+        <Text className="font-semibold text-[--ods-color-heading]">
```
Taelar: I'm late to the party as the file already has this kind of entries, but shouldn't this text be a preset="heading-6" instead of using custom style ?

packages/manager/apps/pci-instances/src/pages/instances/create/hooks/useInstanceCreation.ts
```diff
@@ -118,6 +118,7 @@ export const useInstanceCreation = (): TInstanceCreation => {
       ),
     [distributionImageOsType, flavorId, projectId, billingType],
   );
+  console.log('🚀 ~ useInstanceCreation ~ flavorDetails:', flavorDetails);
```
Taelar: :see_no_evil:

packages/manager/apps/pci-instances/src/pages/instances/create/hooks/useInstanceCreation.ts
```diff
@@ -211,6 +212,7 @@ export const useInstanceCreation = (): TInstanceCreation => {
       distributionImageVersion.distributionImageVersionName,
     backupConfigurationPrices,
   };
+  console.log('🚀 ~ useInstanceCreation ~ instanceData:', instanceData);
```
Taelar: :see_no_evil:

packages/manager/apps/pci-instances/src/utils/formatDisks.tsx
```diff
@@ -0,0 +1,57 @@
+import { ReactNode } from 'react';
+import { Text, TEXT_PRESET } from '@ovhcloud/ods-react';
+import { TDisk } from '@/domain/entities/instancesCatalog';
+
+const formatCapacity = (value: number, unit: string): string => {
+  if (unit === 'GB' && value >= 1000) {
+    const tbValue = (value / 1000).toFixed(2);
+    return `${parseFloat(tbValue)} To`;
```
Taelar: I believe "To" is the french unit, but you will need to translate it depending on the current locale

packages/manager/apps/pci-instances/src/pages/instances/create/hooks/useInstanceCreation.ts
yaallco: Oh, forgot to push the clean version , thanks

packages/manager/apps/pci-instances/src/utils/formatDisks.tsx
yaallco: Thats right, I forgot about this, thanks for your comment !

packages/manager/apps/pci-instances/src/pages/instances/create/components/cart/FlavorDetails.component.tsx
yaallco: Why Heading 6 ?
I don't know why the custom was used here

Taelar: Since you're using semibold and heading color, it looks like you're copying the style of heading but with a small size, which sounds like heading-6 to me :slightly_smiling_face: 
But at the same time it's not meaningful title, so maybe my comment is unrelevant :thinking:

packages/manager/apps/pci-instances/src/utils/formatDisks.tsx
Tsiorifamonjena, line 1: You can add unit test for these functions

packages/manager/apps/pci-instances/src/utils/formatDisks.tsx
```diff
@@ -0,0 +1,60 @@
+import { ReactNode } from 'react';
+import { Text, TEXT_PRESET } from '@ovhcloud/ods-react';
+import { TDisk } from '@/domain/entities/instancesCatalog';
+import { t } from 'i18next';
+
+const formatCapacity = (value: number, unit: string): string => {
```
Tsiorifamonjena: When unit does not equal to `GB` this function doesn't preserve unit. Is that what we want ? (eg. When the unit is not GB or value < 1000, the functions returns 1000 instead of 10000 MB)

packages/manager/apps/pci-instances/src/pages/instances/create/components/cart/FlavorDetails.component.tsx
yaallco: I don't know what's the best, since its not a title 🤷 
I think I will keep it like this for now

packages/manager/apps/pci-instances/src/utils/formatDisks.tsx
```diff
@@ -0,0 +1,60 @@
...
+    return (
+      <Text preset={TEXT_PRESET.span} className="font-semibold">
+        -
+      </Text>
+    );
+  }
+
+  return (
+    <div className="flex flex-col">
+      {disks.map((disk, index) => {
```
Tsiorifamonjena: This logic seems the same as `formatDisksAsString` except the join. Maybe you can extract 
DRY ;)

packages/manager/apps/pci-instances/public/translations/creation/Messages_fr_FR.json
```diff
@@ -44,6 +44,7 @@
   "pci_instance_creation_table_header_vcore_unit": "(Go)",
   "pci_instance_creation_table_header_storage": "Stockage",
   "pci_instance_creation_table_header_storage_unit": "(Go)",
+  "pci_instance_creation_table_storage_unit_tb": "To",
```
aTurmo, line 47: Missing translations ?

packages/manager/apps/pci-instances/src/adapters/tanstack/instancesCatalog/right/mapper.ts
```diff
@@ -202,11 +202,15 @@ const mapFlavorSpecifications = (specs: TSpecificationsDTO) => {
   const base = {
     cpu: mapSpecificationsDetails(specs.cpu),
     ram: mapSpecificationsDetails(specs.ram),
-    storage: mapSpecificationsDetails(specs.storage),
     bandwidth: {
       public: mapSpecificationsDetails(specs.bandwidth.public),
       private: mapSpecificationsDetails(specs.bandwidth.private),
     },
+    disks: specs.disks.map((disk) => ({
```
aTurmo: Suggestion : extract to a mapper like other fields

packages/manager/apps/pci-instances/src/pages/instances/create/view-models/cartViewModel.ts
```diff
@@ -47,7 +47,7 @@ export type TSelectFlavorDetails = {
   name: string;
   memory: number;
   vCore: number;
-  storage: number;
+  disks: TDisk[];
```
aTurmo: Issue [blocking]: view model data cannot be the entoty one. You have to remap them even if it's one-to-one. But in your case you could do your 'formatDisk' here instead of doing it everywhere. That's what the view-model/mapper are for

packages/manager/apps/pci-instances/src/pages/instances/create/view-models/flavorsViewModel.ts
```diff
@@ -343,7 +344,7 @@ export type TBaseFlavorDataForTable = {
   name: string;
   memory: number;
   vCore: number;
-  storage: number;
+  disks: TDisk[];
```
aTurmo: Issue [blocking]: same thing here with view model data using directly entity model

packages/manager/apps/pci-instances/src/pages/instances/create/view-models/cartViewModel.ts
yaallco: You're right, thanks

packages/manager/apps/pci-instances/src/utils/formatDisks.tsx
yaallco: I think yes, when value is GB we don't want to print the Unit, and it won't be MB as unit, if it does, we'll make the change ?

packages/manager/apps/pci-instances/src/pages/instances/create/components/cart/FlavorDetails.component.tsx
```diff
@@ -40,6 +40,16 @@ export const FlavorDetails: FC<TFlavorDetails> = ({ quantity, flavor }) => {
         <Text className="font-semibold text-[--ods-color-heading]">
           {flavor.vCore} {t('pci_instance_creation_cart_flavor_vCore')}
         </Text>
+        <div className="flex flex-col">
+          {flavor.disks.map((disk, index) => (
+            <Text
+              key={index}
```
aTurmo: Issue: It's strongly recommended to never use the index as a key. Instead use something that is specific to your disk data, or a combinasion of them if needed (like '${disk.toto}_${disk.toto}')

aTurmo: https://react.dev/learn/rendering-lists#rules-of-keys

packages/manager/apps/pci-instances/src/pages/instances/create/components/flavor/FlavorRowsBuilder.tsx
```diff
@@ -37,7 +37,15 @@ export const FlavorRowsBuilder = (
         name: renderName(flavor),
         memory: <Text preset={TEXT_PRESET.span}>{flavor.memory}</Text>,
         vCore: <Text preset={TEXT_PRESET.span}>{flavor.vCore}</Text>,
-        storage: <Text preset={TEXT_PRESET.span}>{flavor.storage}</Text>,
+        disks: (
+          <div className="flex flex-col">
+            {flavor.disks.map((disk, index) => (
+              <Text key={index} preset={TEXT_PRESET.span}>
```
aTurmo: same thing here

packages/manager/apps/pci-instances/src/pages/instances/create/components/flavor/GpuFlavorRowsBuilder.tsx
```diff
@@ -41,7 +41,15 @@ export function GpuFlavorRowsBuilder(
         vRamTotal: <Text preset={TEXT_PRESET.span}>{flavor.vRamTotal}</Text>,
         memory: <Text preset={TEXT_PRESET.span}>{flavor.memory}</Text>,
         vCore: <Text preset={TEXT_PRESET.span}>{flavor.vCore}</Text>,
-        storage: <Text preset={TEXT_PRESET.span}>{flavor.storage}</Text>,
+        disks: (
+          <div className="flex flex-col">
+            {flavor.disks.map((disk, index) => (
+              <Text key={index} preset={TEXT_PRESET.span}>
```
aTurmo: idem

packages/manager/apps/pci-instances/src/pages/instances/create/components/cart/FlavorDetails.component.tsx
aTurmo: You could also add an id to your view model data since you do this in multiple places. this id has to be unique for each disk and can be constructed from disk data.

Tsiorifamonjena (approved):

PR #21738, opened 2026-01-19
feat(pci-instances): add instances backups images

lolottetheclash:
ref: #TAPC-5264

Capture d’écran 2026-01-28 à 14 57 28

**Backups List with cart & license:**
Capture d’écran 2026-01-28 à 14 59 48

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-instances/src/adapters/tanstack/backups/right/mapper.ts
+++ b/packages/manager/apps/pci-instances/src/adapters/tanstack/backups/right/mapper.ts
@@ -2,9 +2,9 @@ import { TBackup } from '@/domain/entities/backup';
 import { TBackupDTO } from './dto.type';
 
 export const mapBackupDtoToBackupEntity = (
-  BackupsDTO: TBackupDTO[],
+  backupsDTO: TBackupDTO[],
 ): TBackup[] =>
-  BackupsDTO.map(
+  backupsDTO.map(
     ({ id, name, type, minDisk, minRam, region, size, creationDate }) => ({
       id,
       name,
--- a/packages/manager/apps/pci-instances/src/pages/instances/create/components/DistributionImage.component.tsx
+++ b/packages/manager/apps/pci-instances/src/pages/instances/create/components/DistributionImage.component.tsx
@@ -11,8 +11,8 @@ import { deps } from '@/deps/deps';
 import {
   selectImages,
   emptyResult,
-  selectBackupsEligibilityContext,
   selectEligibleBackups,
+  selectFlavorBackupConstraints,
 } from '../view-models/imagesViewModel';
 import { TInstanceCreationForm } from '../CreateInstance.schema';
 import { useBackups } from '@/data/hooks/backups/useBackups';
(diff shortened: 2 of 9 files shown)


packages/manager/apps/pci-instances/src/adapters/tanstack/backups/right/mapper.ts
```diff
@@ -0,0 +1,18 @@
+import { TBackup } from '@/domain/entities/backup';
+import { TBackupDTO } from './dto.type';
+
+export const mapBackupDtoToBackupEntity = (
+  BackupsDTO: TBackupDTO[],
+): TBackup[] =>
+  BackupsDTO.map(
```
yaallco: I think you should name this variable without the uppercase backupsDTO.map

packages/manager/apps/pci-instances/src/data/api/backups.ts
```diff
@@ -0,0 +1,29 @@
+import { TBackupDTO } from '@/adapters/tanstack/backups/right/dto.type';
```
yaallco, line 1: Question about this import : should the API depend on adapters ?

aTurmo: Since we don't have a 100% hexa on the right side (due to the heavy usage of tanstack), if you don't map here you'll have to map in the useBackups hook, that will then depends on the left and the right (dto => entity, for the left, and then entity => data for the right via selector).

So even if it's not 100% ideal that the api depends on the right adapters, hexagonal archi doesn't really enforce anything happening outside the domain. And with the previous context I think it's the better way to do it.

packages/manager/apps/pci-instances/src/pages/instances/create/components/DistributionImage.component.tsx
```diff
@@ -8,53 +8,79 @@ import DistributionVersionList from './distributionImage/DistributionVersionList
...
 
-  const { images: imageVariants, versions } = useMemo(
-    () =>
-      /**
-       * TODO: remove this cast when image backups implementation will be done:
-       * we have 4 image type options in select: linux, apps, windows, backups
-       * but we will have 2 different selectors: 1 for the selected images type (linux, apps, windows)
-       * and another one for backups images.
-       *  */
+  const backupsEligibilityContextSelect = useMemo(
```
aTurmo: Question?: is this intermediate value necessary ? Why not pass directly to the useInstancesCatalogWithSelect ?

If you really really want to keep it, rename it to "Selector" or something like that to make it less confusing (i think it was confusing ^^)

packages/manager/apps/pci-instances/src/pages/instances/create/components/DistributionImage.component.tsx
```diff
@@ -8,53 +8,79 @@ import DistributionVersionList from './distributionImage/DistributionVersionList
...
-      distributionImageType,
-      flavorId,
-      microRegion,
+  const {
+    data: backupsEligibilityContext = null,
+  } = useInstancesCatalogWithSelect({
+    select: backupsEligibilityContextSelect,
+  });
+
+  const backupsSelect = useMemo(
```
aTurmo: Same remark here

packages/manager/apps/pci-instances/src/data/api/backups.ts
lolottetheclash: as we are calling the api, we already know about the DTO that we will receive, but maybe we could have put the api call inside the right adapter ;)

packages/manager/apps/pci-instances/src/pages/instances/create/components/DistributionImage.component.tsx
lolottetheclash: If I don't put it in the useMemo, selectBackupsEligibilityContext will be computed a lot of times, that's why it didn't go straight to the select.

packages/manager/apps/pci-instances/src/pages/instances/create/components/DistributionImage.component.tsx
lolottetheclash: I renamed to make all clearer ;)

Tsiorifamonjena (approved):
well done 👏🏽

PR #21747, opened 2026-01-20
fix(muk): add min height in the datagrid

aboungnaseng-ovhcloud:
ref: #MANAGER-20783

## Description

- add min height in the datagrid

Ticket Reference: #MANAGER-20783


packages/manager-ui-kit/src/components/datagrid/table/table-body/TableBody.component.tsx
```diff
@@ -68,6 +68,7 @@ export const TableBody = <T,>({
     return <EmptyRow />;
   }
 
+  console.info('totalHeight', totalHeight);
```
nfoufeovh:
```suggestion
```

aboungnaseng-ovhcloud: 👍

PR #21773, opened 2026-01-20
fix(web-domains): fix free hosting only available on EU

kb-ovh:
## Description

Ticket Reference: #DCE-152


packages/manager/apps/web-domains/src/domain/components/AssociatedServicesCards/Hosting.tsx
```diff
@@ -101,7 +103,7 @@ export default function Hosting({ serviceName }: HostingProps) {
     });
 
   const actionMenuItems = [
-    ...(!hasFreeHosting
+    ...(region === Region.EU && !hasFreeHosting
```
nfoufeovh, line 106:
```suggestion
    ...(region !== Region.CA && !hasFreeHosting
```

PR #21779, opened 2026-01-20
feat(pci-kubernetes): update 3az conditions

Taelar:
ref: #TAPC-5549

## Description

Ticket Reference: #...


ZyaCon (approved):

packages/manager/apps/pci-kubernetes/src/pages/detail/nodepools/new/components/SizeStep.component.spec.tsx
```diff
@@ -54,6 +54,13 @@ vi.mock('@/pages/new/steps/node-pool/PublicConnectivity.component', () => ({
   default: () => <div data-testid="public-connectivity">PublicConnectivity</div>,
 }));
 
+vi.mock('@/hooks/useFeatureAvailability', () => ({
```
ppprevost: [question](https://pullpo.io/cc?l=question) 
I’m not sure to understand the naming of this file.
I don’t really understand the name. Does this mean it’s mocking all the hooks?

Also, could we set the test setup with the data enabled by default (true), and only set it to false when needed? What do you think?

Taelar: The file naming just means this file is meant to receive every FF hooks so they can be centralized , and it's not supposed to mock every hooks but indeed I forgot the vi.importActual part :sweat_smile: It's done in the setupTests now

PR #21790, opened 2026-01-21
feat(pci-kubernetes): implement kuberegions as entities and enrich filters

Taelar:
## Description

Ticket Reference: #TAPC-5549

### DIFF
--- a/packages/manager/apps/pci-kubernetes/src/pages/new_v2/components/ClusterLocationSection.component.tsx
+++ b/packages/manager/apps/pci-kubernetes/src/pages/new_v2/components/ClusterLocationSection.component.tsx
@@ -8,8 +8,6 @@ import { Icon, Link, Message, MessageBody, MessageIcon, Text } from '@ovhcloud/o
 
 import { ShellContext } from '@ovh-ux/manager-react-shell-client';
 
-import { useAvailabilityRegions } from '@/api/hooks/useAvailabilityRegions';
-import { useKubeRegions } from '@/api/hooks/useKubeRegions';
 import { HelpDrawer } from '@/components/helpDrawer/HelpDrawer.component';
 import { HelpDrawerDivider } from '@/components/helpDrawer/HelpDrawerDivider.component';
 import { PLAN_DOC_LINKS } from '@/constants';
@@ -20,7 +18,7 @@ import { selectAvailablePlanOptions } from '../view-models/plans.viewmodel';
 import {
   filterMacroRegions,
   mapMacroRegionForCards,
-  selectAvailableRegions,
+  useCombinedRegions,
 } from '../view-models/regions.viewmodel';
 import { ContinentSelect } from './location/ContinentSelect.component';
 import { DeploymentModeSelect } from './location/DeploymentModeSelect.component';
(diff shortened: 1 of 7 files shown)


packages/manager/apps/pci-kubernetes/src/pages/new_v2/components/ClusterLocationSection.component.tsx
```diff
@@ -43,32 +52,51 @@ export const ClusterLocationSection: FC<TClusterLocationSectionProps> = ({ is3az
...
-          planField,
-          deploymentModeField,
-        )(selectMacroRegions(regions)),
-      ),
+    select: selectAvailableRegions(deploymentModeField, kubeRegions),
   });
 
+  const continentOptions = useMemo(() => selectAvailableContinentOptions(regions), [regions]);
+
+  const planOptions = useMemo(() => selectAvailablePlanOptions(regions), [regions]);
```
aTurmo: Suggestion: since the calculations. run on the same dependencies, it's better to merge the two useMemo if the calculation are relatively light, by for exemple returning a tuple or an object (recommended by chatGPT and composer)

packages/manager/apps/pci-kubernetes/src/pages/new_v2/components/location/ContinentSelect.component.tsx
```diff
@@ -13,19 +13,23 @@ import {
...
+};
+
+export const ContinentSelect = ({ options }: TContinentSelectProps) => {
   const { t } = useTranslation(['common', 'add']);
 
   const { control } = useFormContext<TCreateClusterSchema>();
 
   const continentOptions = useMemo(() => {
-    return selectContinentOptions().map(({ continentCode, labelKey }) => ({
+    return options.map(({ continentCode, labelKey }) => ({
```
aTurmo: optimisation: you can remove the return and the '{}'

packages/manager/apps/pci-kubernetes/src/pages/new_v2/components/ClusterLocationSection.component.tsx
```diff
@@ -43,32 +52,51 @@ export const ClusterLocationSection: FC<TClusterLocationSectionProps> = ({ is3az
...
-        filterMacroRegions(
-          continentField,
-          planField,
-          deploymentModeField,
-        )(selectMacroRegions(regions)),
-      ),
+    select: selectAvailableRegions(deploymentModeField, kubeRegions),
   });
 
+  const continentOptions = useMemo(() => selectAvailableContinentOptions(regions), [regions]);
```
aTurmo: Thinking (non-blocking): It really feels like there are two many hooks use to get the data. I wonder if we should instead only have one data selector max by ui-component.

Here the main problem I have is that you call useKubeRegions and useAvailabilityRegions, so two differents hooks. It's not just calling two times the same hooks with differents selectors.
It implies that the component have to know how the data is organised on the right side (since two hooks implies two endpoints).
So if you ever get a /catalog endpoint with all data needed like we have for instance and volume, you will have to change ui-component when it should not ever be changed due to this in an hexagonal architecture.

We should think about this as a team. But a potential solution is that the components should use one hook for data, and this hook would call the differents hooks with selector needed to have the needed data.
That way if the api calls change, the ui-component would not change and only the hooks called. Not perfect but I think it's better.

packages/manager/apps/pci-kubernetes/src/pages/new_v2/view-models/location.viewmodel.ts
```diff
@@ -63,25 +45,51 @@ const mapPlanCodeToDeploymentMode = (planCode: TPlanCode): TDeploymentMode => {
...
+
+    return regions
+      .map((region) => ({
+        ...region,
+        microRegionIds: region.microRegionIds.filter((microRegionId) =>
+          kubeRegions.includes(microRegionId),
+        ),
+      }))
+      .filter((region) => region.microRegionIds.length > 0);
+  };
```
aTurmo: if i'm not mistaken those filters act on domain and return domain data. So there should be on domain not view model.

packages/manager/apps/pci-kubernetes/src/pages/new_v2/components/ClusterLocationSection.component.tsx
Taelar: You're not wrong :thinking: 
I think the one data-hook solution would work, I just wonder how beefy this king of centralized hook could be in cases where we have 3+ data hook wrapped, there could be a lot of input parameters and a lot of output data, making it less easy to read. But I'm not sure we have a lot of these cases

packages/manager/apps/pci-kubernetes/src/pages/new_v2/view-models/location.viewmodel.ts
Taelar: Nice catch

PR #21797, opened 2026-01-21
feat(dedicated-servers): column visibility in views and switch views

qpavy:
## Description

- Add context to manage views
- Use new drawer and remove old one
- Add column visibility configuration inside view management
- Add switch views feature
- Add view initialization with default view

Ticket Reference: #MANAGER-20621

### DIFF
--- a/packages/manager/apps/dedicated-servers/src/components/manageView/style.module.scss
+++ b/packages/manager/apps/dedicated-servers/src/components/manageView/style.module.scss
@@ -8,10 +8,6 @@
 
   &__item {
     width: 100%;
-    // display: flex;
-    // flex-flow: row;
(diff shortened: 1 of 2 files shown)


packages/manager/apps/dedicated-servers/src/components/manageView/style.module.scss
```diff
@@ -0,0 +1,32 @@
...
+  padding: 0;
+  list-style: none;
+
+  &__item {
+    width: 100%;
+    // display: flex;
+    // flex-flow: row;
+    // align-items: center;
+    // flex-grow: 1;
+
```
aderghamov: remove this if it not used anymore

PR #21798, opened 2026-01-21
Feat/banner discovery mode tapc 5562

ZyaCon:
## Description

Ticket Reference: #TAPC-5562   

## Additional Information
image

### DIFF
--- a/packages/manager/apps/pci-instances/src/data/utils/project.utils.ts
+++ b/packages/manager/apps/pci-instances/src/data/utils/project.utils.ts
@@ -1,0 +1,10 @@
+import { TProject } from '@ovh-ux/manager-pci-common';
+
+export enum PlanCode {
+  DISCOVERY = 'project.discovery',
+  STANDARD = 'project.2018',
+}
+
+export const isDiscoveryProject = (project: TProject | undefined): boolean => {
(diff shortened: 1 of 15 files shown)


packages/manager/apps/pci-instances/src/data/utils/project.utils.ts
Tsiorifamonjena, line 1: I don't think it should be in domain. First of all you are dependant on an external lib (@ovh-ux/manager-pci-common), and you access it directly in a UI layer below

Tsiorifamonjena (changes requested):

packages/manager/apps/pci-instances/src/data/utils/project.utils.ts
yaallco: Yes that's right, I moved it to utils folder, thanks

packages/manager/apps/pci-instances/src/pages/instances/create/hooks/useCartItems.tsx
```diff
@@ -203,16 +213,62 @@ export const useCartItems = (): TCartItems => {
     },
   ];
 
+  const renderBanner = (
```
aTurmo: maybe this should be in a separate file.

PR #21807, opened 2026-01-21
fix(dedicated.dbaas): fix iam activation from legacy roles

darsene:
fixes: #MABOS-222

## Description

Ticket Reference: #...


packages/manager/modules/dbaas-logs/src/logs/detail/roles/enable-iam/logs-iam-enable.controller.js
```diff
@@ -44,7 +44,7 @@ export default class LogsIamEnableModalCtrl {
     this.CucCloudMessage.flushChildMessage();
     this.saving = this.CucControllerHelper.request.getHashLoader({
       loaderFunction: () =>
-        this.LogsHomeService.enableIam(this.serviceName, this.service)
+        this.LogsHomeService.enableIam(this.service)
```
MarcMonchablon, line 47: Shouldn't it be `this.LogsHomeService.enableIam({ serviceName: this.service })` ? 
Looking at `logs-home.service.js` implementation of `enableIam`, which expects an object.

PR #21821, opened 2026-01-22
feat(muk): add displayValue for FilterList Compo

mhelhali-soufien:
ref: #MANAGER-18416

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager-ui-kit/src/components/filters/Filter.props.ts
+++ b/packages/manager-ui-kit/src/components/filters/Filter.props.ts
@@ -1,7 +1,7 @@
 import type { Filter } from '@ovh-ux/manager-core-api';
 import { FilterComparator } from '@ovh-ux/manager-core-api';
 
-export type FilterWithLabel = Filter & { label: string };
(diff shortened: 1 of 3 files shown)


packages/manager-ui-kit/src/components/filters/filter-list/FilterList.component.tsx
```diff
@@ -22,7 +22,7 @@ export function FilterList({ filters, onRemoveFilter }: Readonly<FilterListProps
           onClick={() => onRemoveFilter(filter)}
           data-testid="filter-list_tag_item"
         >
-          {`${filter.label && `${filter.label} ${tComp(filter.comparator)} ${formatFilter(filter, locale)}`}`}
+          {`${filter.label && `${filter.label} ${tComp(filter.comparator)} ${filter.displayValue || formatFilter(filter, locale)}`}`}
```
anooparveti: **nitpick:** Since you're anyway passing `filter` to formatFilter, you can return the displayValue as base case if it exists.

PR #21823, opened 2026-01-22
Feat/pci instances add instance public connectivity

Tsiorifamonjena:
## Description

Ticket Reference: #TAPC-5319

## Additional Information

image


packages/manager/apps/pci-instances/src/pages/instances/create/components/network/PublicIpConfiguration.component.tsx
```diff
@@ -0,0 +1,266 @@
...
+
+  const handleSelectFloatingIpAssignment = ({
+    value,
+  }: RadioValueChangeDetail) => {
+    const newValue = value as 'createNew' | 'reuseExisting' | null;
+
+    setValue('floatingIpAssignment', newValue);
+
+    if (newValue === 'createNew') setValue('existingFloatingIp', null);
+    else if (newValue === 'reuseExisting')
```
aTurmo, line 97: Suggestion : switch case ?

packages/manager/apps/pci-instances/src/pages/instances/create/components/network/PublicIpConfiguration.component.tsx
aTurmo, line 1: you should add tests for this file

packages/manager/apps/pci-instances/src/pages/instances/create/components/network/PublicIpConfiguration.component.tsx
Tsiorifamonjena: yes but with switch case there will be many more lines of code, won't there ? I think for this specific case it's simply a matter of preference. We can let it as is I think

packages/manager/apps/pci-instances/src/pages/instances/create/components/network/PublicIpConfiguration.component.tsx
Tsiorifamonjena: Initially this was already covered by `AddPublicNetworkConfiguration.component.spec.tsx` but yes for the coverage test rate I split it then and move some tests

PR #21846, opened 2026-01-23
feat(metrics-module): add localization and formatting for dashboard widgets and charts

yjaaouane:
## Description

Ticket Reference: #MAOBS-199


packages/manager/modules/metrics-to-customer/public/translations/dashboard-texts/Messages_fr_FR.json
darsene, line 1: The content is in English, should be in French

PR #21876, opened 2026-01-25
fix(ldp): align storage units

pdepaepe:
## Description

This will fix LDP translations regarding storage units: any of them must be expressed in KiB, MiB, GiB, ...

Ticket Reference: #OB-10802

## Additional Information

- Translations ticket reference: #CMT-34804


pdepaepe:
> You need to fix conflicts. Moreover, you can squash the changes into one commit only

Hello, Thank you for your review. Should be ok now.

PR #21881, opened 2026-01-26
feat(web-domains): add information message about renew mode

louisbensiovh:
ref: #DCE-147


packages/manager/apps/web-domains/src/alldoms/pages/service/serviceDetail/serviceDetail.tsx
```diff
@@ -108,6 +109,20 @@ export default function ServiceDetail() {
...
+          color={MESSAGE_COLOR.information}
+          dismissible={false}
+          className="w-full mb-6"
+        >
+          <MessageIcon name={ICON_NAME.circleInfo} />
+          <MessageBody>
+            <Trans
+              t={t}
+              i18nKey="allDom_detail_page_renew_information"
+              components={{ br: <br /> }}
```
nfoufeovh: Why don't we use 2 separate translation keys rather than a `<br/>` in a Trans ?

PR #21884, opened 2026-01-26
refactor(metrics-module): move all types and use import aliases

yjaaouane:
ref: #MAOBS-206

## Description

Ticket Reference: #...


packages/manager/modules/metrics-to-customer/src/types/widget/ChartWidgetWithData.type.ts
```diff
@@ -1,4 +1,4 @@
-import { ChartWidget } from './ChartWidget.type';
+import { ChartWidget } from "@/types/widget/ChartWidget.type";
```
darsene: niptick
```suggestion
import { ChartWidget } from '@/types/widget/ChartWidget.type';
```

PR #21895, opened 2026-01-26
feat(pci-file-storage): add form buttons and rules for share name

aTurmo:
ref: #TAPC-5427


packages/manager/apps/pci-file-storage/src/pages/create/components/form/CreateShareForm.component.tsx
```diff
@@ -1,8 +1,12 @@
-import { useWatch } from 'react-hook-form';
-import { FormProvider } from 'react-hook-form';
+/* eslint-disable @typescript-eslint/no-misused-promises */
```
Tsiorifamonjena: question: why disabling this typescript rule ?

aTurmo: because of this:
`<form onSubmit={formMethods.handleSubmit(onSubmit)} className="flex w-2/3 flex-col gap-6">`
It doesn't like the use of formMethods.handleSubmit(onSubmit), and if you add the void to make the linter pass, the onSubmit doesn't work

aTurmo: image

Tsiorifamonjena: You can use something like this I think:

```tsx
const handleShare = (event: React.FormEvent) => {
  void formMethods.handleSubmit(onSubmit)(event);
}
```

and then usage :

```tsx
<form onSubmit={handleShare} className="flex w-2/3 flex-col gap-6">
```

But it's up to you. It's just a detail

PR #21914, opened 2026-01-27
fix(web-domains): improve nichandle check for CA environment

kb-ovh:
## Description

Ticket Reference: #INC0185391


packages/manager/apps/web-domains/src/common/hooks/nichandle/useGetConnectedNichandle.tsx
```diff
@@ -0,0 +1,6 @@
+import { useNichandleInformation } from './useNichandleInformation';
+
+export const useGetConnectedNichandle = () => {
```
nfoufeovh:
```suggestion
export const useGetConnectedNichandleId = () => {
```

PR #21928, opened 2026-01-27
feat(pci.os): object pagination

jperchoc:
ref: #DATATR-1234

## Description

Ticket Reference: #...


packages/manager/apps/pci-object-storage/src/data/hooks/s3-storage/useS3ObjectsBrowser.hook.tsx
```diff
@@ -0,0 +1,152 @@
...
+
+  return {
+    type: 'file',
+    ...o,
+    name: getFileName(o.key),
+  };
+}
+
+export interface UseS3ObjectsBrowserParams {
+  projectId: string | undefined;
```
lionel95200x: I think this value shoud not be undefined at this step of app
Already checked in the parent layout 

```suggestion
  projectId: string;
  region: string;
  storageName: string;
```

packages/manager/apps/pci-object-storage/src/data/hooks/s3-storage/useS3ObjectsBrowser.hook.tsx
```diff
@@ -0,0 +1,152 @@
+import { useMemo } from 'react';
+import { useInfiniteQuery } from '@tanstack/react-query';
+import { cloud } from '@datatr-ux/ovhcloud-types';
+import { apiClient } from '@/data/api/api.client';
+import { OBJECTS_PAGE } from '@/pages/object-storage/storage/s3Id/objects/objects.constants';
+
+type StorageObject = cloud.StorageObject & { isCommonPrefix: boolean };
+
+export type S3BrowserItem =
+  | { type: 'parent' }
```
lionel95200x: Why there is a `|` in first ?

packages/manager/apps/pci-object-storage/src/data/hooks/s3-storage/useS3ObjectsBrowser.hook.tsx
```diff
@@ -0,0 +1,152 @@
...
+ *
+ * @param key - The S3 object key (e.g., "photos/2024/")
+ * @returns The folder name (e.g., "2024")
+ *
+ * @example
+ * getFolderName("photos/") // returns "photos"
+ * getFolderName("photos/2024/") // returns "2024"
+ * getFolderName("my folder/") // returns "my folder"
+ */
+function getFolderName(key: string): string {
```
lionel95200x: A unit test could be fine for this use case check :)

packages/manager/apps/pci-object-storage/src/data/hooks/s3-storage/useS3ObjectsBrowser.hook.tsx
```diff
@@ -0,0 +1,152 @@
...
+export function useS3ObjectsBrowser({
+  projectId,
+  region,
+  storageName,
+  prefix,
+  withVersions,
+  limit = OBJECTS_PAGE.DEFAULT_LIMIT,
+  showParent = true,
+}: UseS3ObjectsBrowserParams) {
+  const isEnabled = Boolean(projectId && region && storageName);
```
lionel95200x: Same here i think we could remove this check already in a parent
It's not the responsability of this function to check that

packages/manager/apps/pci-object-storage/src/pages/object-storage/storage/s3Id/objects/objects.utils.ts
```diff
@@ -0,0 +1,39 @@
...
+ * Search query takes precedence over navigation prefix
+ * @param searchQuery - The current search query
+ * @param navigationPrefix - The current navigation prefix
+ * @returns The effective prefix to use
+ */
+export function getEffectivePrefix(
+  searchQuery: string,
+  navigationPrefix: string,
+): string {
+  return searchQuery || navigationPrefix;
```
lionel95200x, line 38: **question** for me we should add the prefix for the actual navigation  ? How to search in a folder with the actual implementation ?

packages/manager/apps/pci-object-storage/src/data/hooks/s3-storage/useS3ObjectsBrowser.hook.tsx
```diff
@@ -0,0 +1,152 @@
...
+      's3-browser',
+      projectId,
+      region,
+      storageName,
+      prefix,
+      withVersions,
+      limit,
+    ],
+
+    initialPageParam: undefined as string | undefined,
```
abullet33: Can we avoid using as?

packages/manager/apps/pci-object-storage/src/pages/object-storage/storage/s3Id/objects/objects.utils.ts
jperchoc: Search works only by providing the full prefix because of current api design. We could inject the current prefix in the search, but that would prevent user from searching in the whole bucket except when viewing the root content.
Maybe we could improve this in the future by letting the user search either in the current folder (where we inject the prefix) or in the whole bucket

abullet33 (approved):

PR #21938, opened 2026-01-28
feat(vrack): adding the public routing bandwidth tiles

aderghamov:
ref: #MANAGER-20377

## Description

Ticket Reference: #MANAGER-20377


packages/manager/apps/network-vrack/src/style.scss
```diff
@@ -1,8 +1,12 @@
-@tailwind utilities;
-
+/* eslint-disable css/no-invalid-at-rules */
+/* eslint-disable css/no-invalid-properties */
 @import '@ovhcloud/ods-themes/default';
 @import '@ovh-ux/muk/dist/style.css';
```
chipp972: muk should be last as indicated in the storybook : https://ovh.github.io/manager/storybook-static/index.html?path=/docs/manager-ui-kit-what-s-new-common-issues-correct-tailwind-issues--technical-information

packages/manager/apps/network-vrack/package.json
```diff
@@ -40,17 +40,18 @@
     "react-hook-form": "7.51.2",
     "react-i18next": "14.1.0",
     "react-router-dom": "6.16.0",
-    "tailwindcss": "^4.1.17",
+    "tailwindcss": "^3.4.4",
     "zod": "4.1.12"
   },
   "devDependencies": {
+    "@ovh-ux/manager-network-common": "^0.8.0",
+    "@ovh-ux/manager-react-components": "^2.43.1",
```
chipp972: We need this ?

chipp972: If you use muk you shouldn't use manager-react-components I think

aderghamov: no longer.
I introduced these dependencies when I was struggling for running the tests.
It seems that I did not even need to include them.
It is now removed.

packages/manager/apps/network-vrack/src/style.scss
aderghamov: Fixed.
I had to change the order of the line because of the linter

PR #21940, opened 2026-01-28
feat(database): add clickhouse to order funnel

pierre-pelloquin:
ref: #DATATR-2816

## Description

Ticket Reference: #...


packages/manager/apps/pci-databases-analytics/src/pages/services/[serviceId]/users/_components/AddEditUser.component.tsx
```diff
@@ -184,45 +184,49 @@ const AddEditUserModal = ({
...
+                          ))}
+                        </SelectContent>
+                      </Select>
+                    )}
+                    <div className="my-2">
+                      <p className="font-semibold">
+                        {t('resetUserConnectionCode')}
+                      </p>
+                      <Clipboard
+                        value={`${selectedEndpoint?.uri.replace(
```
Comment (lionel95200x) [inline packages/manager/apps/pci-databases-analytics/src/pages/services/[serviceId]/users/_components/AddEditUser.component.tsx:219]: Could you encoreURI only once ?

Comment (pierre-pelloquin) [reply packages/manager/apps/pci-databases-analytics/src/pages/services/[serviceId]/users/_components/AddEditUser.component.tsx:219]: the `encodeURIComponent` function would encode also the `:` delimiter between the username and password, I think it would not work

PR #21963, opened 2026-01-29
feat(vrack): adding breadcrumb and tab links to the legacy app

aderghamov:
ref: #MANAGER-20728

## Description

Ticket Reference: #MANAGER-20728

### DIFF
--- a/packages/manager/apps/network-vrack/src/components/breadcrumb/Breadcrumb.tsx
+++ b/packages/manager/apps/network-vrack/src/components/breadcrumb/Breadcrumb.tsx
@@ -1,0 +1,42 @@
+import { useNavigate, useParams } from 'react-router-dom';
+
+import { useTranslation } from 'react-i18next';
+
+import { BreadcrumbItem, BreadcrumbLink, Breadcrumb as OdsBreadcrumb } from '@ovhcloud/ods-react';
+
+import { useNavigateToLegacyApp } from '@/hooks/useNavigateToLegacyApp';
+import { urls } from '@/routes/Routes.constants';
+import { TRANSLATION_NAMESPACES } from '@/utils/constants';
+
+type BreadcrumbSegment = {
+  label: string;
+  onClick: () => void;
+};
+
+export const Breadcrumb = () => {
+  const { t } = useTranslation([TRANSLATION_NAMESPACES.dashboard]);
+  const { serviceName } = useParams();
+  const navigate = useNavigate();
+  const navigateToLegacy = useNavigateToLegacyApp();
+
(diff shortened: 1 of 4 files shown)


packages/manager/apps/network-vrack/src/components/breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,45 @@
...
+      label: t('dashboard_vrackBreadcrumbLabel'),
+      onClick: () => void navigateToLegacy(),
+    },
+  ];
+  if (serviceName) {
+    segments.push({
+      label: serviceName,
+      onClick: () => navigate(urls.dashboard.replace(':serviceName', serviceName)),
+    });
+  }
```
chipp972:
```suggestion
  const segments: { label: string; href?: string; onClick?: () => void }[] = [
    {
      label: t('dashboard_vrackBreadcrumbLabel'),
      onClick: () => void navigateToLegacy(),
    },
    serviceName && {
      label: serviceName,
      onClick: () => navigate(urls.dashboard.replace(':serviceName', serviceName)),
    }
  ].filter(Boolean);
```

packages/manager/apps/network-vrack/src/components/breadcrumb/Breadcrumb.tsx
```diff
@@ -0,0 +1,45 @@
...
+
+  return (
+    <OdsBreadcrumb className="mb-4">
+      {segments
+        .filter((item): item is { label: string; href?: string; onClick?: () => void } =>
+          Boolean(item),
+        )
+        .map((item) => (
+          <BreadcrumbItem key={item.label}>
+            <BreadcrumbLink href={item.href} onClick={() => item.onClick?.()}>
```
chipp972:
```suggestion
            <BreadcrumbLink href={item.href} onClick={item.onClick}>
```

packages/manager/apps/network-vrack/src/components/breadcrumb/Breadcrumb.tsx
aderghamov: this won't work because `navigateToLegacy` returns a promise. It needs to be written this way to pass the lint check.

Steffy29 (approved):

packages/manager/apps/network-vrack/src/components/breadcrumb/Breadcrumb.tsx
chipp972: Added the `void` back but you can disable this rule actually
It's a bit annoying and just add noise in the code IMO

PR #21981, opened 2026-01-30
feat(public-cloud): add listing page

aTurmo:
ref: #TAPC-5482

### DIFF
--- a/packages/manager/apps/pci-file-storage/src/adapters/shares/left/shareStatusMapper.ts
+++ b/packages/manager/apps/pci-file-storage/src/adapters/shares/left/shareStatusMapper.ts
@@ -1,0 +1,50 @@
+/**
+ * Maps API/Openstack share status to Manager display (label key + ODS badge color).
+ * Spec: green → success, yellow → warning, red → critical.
+ * Manager status: Pending (warning), Active (success), Error (critical), NA (neutral).
+ */
+
+export type TShareStatusBadgeColor = 'success' | 'warning' | 'critical' | 'neutral';
+
+export type TShareStatusDisplay = {
+  labelKey: string;
+  badgeColor: TShareStatusBadgeColor;
+};
+
+const STATUS_TO_DISPLAY: Record<string, TShareStatusDisplay> = {
+  available: { labelKey: 'list:status.active', badgeColor: 'success' },
+  creating: { labelKey: 'list:status.pending', badgeColor: 'warning' },
+  updating: { labelKey: 'list:status.pending', badgeColor: 'warning' },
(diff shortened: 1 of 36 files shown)


packages/manager/apps/pci-file-storage/src/adapters/shares/right/mapper.ts
```diff
@@ -0,0 +1,47 @@
...
+    'replication_change',
+    'reverting',
+    'reverting_error',
+    'shrinking',
+    'shrinking_error',
+    'shrinking_possible_data_loss_error',
+    'unmanage_error',
+    'unmanage_starting',
+    'unmanaged',
+  ].includes(s);
```
Taelar, line 34: I think you should use this table as a const so you can extract the `TShareStatusDisplay` type from it. 
This would make your type guard much more reliable as the status list would be only defined once

packages/manager/apps/pci-file-storage/src/adapters/shares/left/shareStatusMapper.ts
```diff
@@ -0,0 +1,50 @@
...
+ */
+
+export type TShareStatusBadgeColor = 'success' | 'warning' | 'critical' | 'neutral';
+
+export type TShareStatusDisplay = {
+  labelKey: string;
+  badgeColor: TShareStatusBadgeColor;
+};
+
+const STATUS_TO_DISPLAY: Record<string, TShareStatusDisplay> = {
```
Taelar, line 14: Could this const be typed as `Record<TShareStatus, TShareStatusDisplay>` ?

packages/manager/apps/pci-file-storage/src/adapters/shares/right/mapper.ts
aTurmo: i'll do it in the next pr (coming really soon)

packages/manager/apps/pci-file-storage/src/adapters/shares/left/shareStatusMapper.ts
aTurmo: i'll do it in the next pr (coming really soon)

aTurmo: Fixed in https://github.com/ovh/manager/pull/22085

packages/manager/apps/pci-file-storage/src/adapters/shares/right/mapper.ts
aTurmo: Fixed in https://github.com/ovh/manager/pull/22085

PR #22025, opened 2026-02-02
fix(pci-instances): checked network dhcp by default

Tsiorifamonjena:
This PR contain these fix : 
- checked network dhcp by default
- avoid returning null for network capabilities

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/pci-instances/src/adapters/tanstack/configuration/right/dto.type.ts
+++ b/packages/manager/apps/pci-instances/src/adapters/tanstack/configuration/right/dto.type.ts
@@ -40,7 +40,7 @@ export type TSubnetDTO = {
   id: string;
   cidr: string;
   gatewayIp: string | null;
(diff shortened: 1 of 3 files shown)


packages/manager/apps/pci-instances/src/adapters/tanstack/configuration/right/dto.type.ts
```diff
@@ -40,7 +40,7 @@ export type TSubnetDTO = {
   id: string;
   cidr: string;
   gatewayIp: string | null;
-  capabilities: TCapabilityDTO[];
```
aTurmo, line 43: Capabilities can be null ?

Tsiorifamonjena: obviously yes. I don't know if this was really intentional on back side

PR #22047, opened 2026-02-03
feat(container): add communication announcement modal

aottr:
ref: #MANAGER-20268

## Description

Ticket Reference: #...

### DIFF
--- a/packages/manager/apps/container/src/public/translations/communication-announce-modal/Messages_de_DE.json
+++ b/packages/manager/apps/container/src/public/translations/communication-announce-modal/Messages_de_DE.json
@@ -1,0 +1,6 @@
+{
+  "communication_announce_modal_title": "Entdecken Sie eine neue Funktion in Ihrem Kundenbereich",
+  "communication_announce_modal_info": "Richten Sie den Versand von Mitteilungen und die Verwaltung der Kontakte ein.",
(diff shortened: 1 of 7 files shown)


packages/manager/apps/container/src/components/communication-announce-modal/CommunicationAnnounceModal.component.tsx
```diff
@@ -0,0 +1,125 @@
...
+    }
+    if (shouldDisplayModal === false) {
+      uxPlugin.notifyModalActionDone(CommunicationAnnounceModal.name);
+    }
+  }, [shouldDisplayModal]);
+
+  if (!showModal) return null;
+
+  return (
+    // eslint-disable-next-line tailwindcss/no-arbitrary-value
```
sachinrameshn, line 75: Why are we disabling the rule here ?

sachinrameshn: UXMAN-manager-build-v4 is failing

aottr: I migrated the modal from ui-kit to tailwindcss and here we need arbitrary values to keep the old behavior. arbitrary values are marked as warning without disabling the rule.

sachinrameshn (approved):

PR #22085, opened 2026-02-04
feat(file-storage): add create share and action column

aTurmo:
ref: #TAPC-5482

### DIFF
--- a/packages/manager/apps/pci-file-storage/public/translations/list/Messages_de_DE.json
+++ b/packages/manager/apps/pci-file-storage/public/translations/list/Messages_de_DE.json
@@ -6,6 +6,12 @@
     "protocol": "Protokoll",
     "allocated_capacity": "Zugewiesene Kapazität",
     "allocated_capacity_value": "{{capacity}} GiB",
-    "status": "Status"
+    "status": "Status",
+    "actions": "Aktionen"
+  },
+  "actionButton": "Ein Share erstellen",
+  "actions": {
+    "manage": "Verwaltung der ",
+    "delete": "Löschen"
   }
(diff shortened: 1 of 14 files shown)


Tsiorifamonjena (approved):

packages/manager/apps/pci-file-storage/src/pages/list/components/menu/ActionMenuItem.component.tsx
```diff
@@ -0,0 +1,34 @@
...
+
+import clsx from 'clsx';
+import { useTranslation } from 'react-i18next';
+
+import { Icon, Link } from '@ovhcloud/ods-react';
+
+import { TShareAction } from '@/pages/list/components/menu/action.type';
+
+const linkClassname =
+  'w-full box-border p-4 bg-none hover:bg-none hover:bg-[--ods-color-primary-100] focus-visible:bg-[--ods-color-primary-100] focus-visible:rounded-sm focus-visible:outline-none';
```
Taelar: non-blocking: I don't feel like this const is needed and I tend to think it's better not to use tailwind classes in const as most tools become unable to lint it or understand it's tailwind related. So maybe these classes could go directly on in the classname attribute ?

aTurmo: copy pasted form instance ^^ but i'll look into it

aTurmo: fixed : const removed and class simplyfied

PR #22160, opened 2026-02-09
feat(vcfaas): generic way to handle async ressource operations for vrack

sefiani-ovh:
ref: #MANAGER-20708

# Handle background operations notifications of OVH api

## Context

OVH APIs return resources with `currentState`, `targetSpec`, `resourceStatus` (`UPDATING`, `READY`, `ERROR`, etc.), and `currentTask`.

## Goal

Display persistent notifications (between user sessions) for async operations:

- Show notification while `resourceStatus` is in progress (`UPDATING`, `CREATING`, `DELETING`) with reference to the resource and which field (first level only) is being updated between `currentState` and `targetSpec`
- Show final notification when resource transitions to `READY` or `ERROR` with reference to the resource and updated field

## Hooks

**`useListUpdatedOperationsStatus`** — Monitors resources and detects status transitions. Returns `updatingResources`, `completedResources`, and `erroredResources` (each containing resource id and changed fields - first level only).

**`useListUpdatedOperationBanner`** — Displays notification banners based on the above hook's output.

## Usage

```typescript
const { updatingResources, completedResources, erroredResources } = 
  useListUpdatedOperationsStatus(vrackSegments, { ignoreFields: ['type'] });

useListUpdatedOperationBanner({
  updatingResources,
  completedResources,
  erroredResources,
  getUpdatingMessage: (updating) => {
    const segment = vrackSegments?.find((s) => s.id === updating.id);
    if (!segment) return null;
    if (updating.changedField === 'vlanId') {
      return `VLAN ID update in progress: ${segment.currentState.vlanId} → ${segment.targetSpec.vlanId}`;
    }
    return `Operation in progress on vRack segment ${segment.targetSpec.vlanId}...`;
  },
  getCompletedMessage: (completed) => {
    const segment = vrackSegments?.find((s) => s.id === completed.id);
    if (!segment) return null;
    return `Operation completed on vRack segment ${segment.targetSpec.vlanId}.`;
  },
  getErroredMessage: (errored) => {
    const segment = vrackSegments?.find((s) => s.id === errored.id);
    if (!segment) return null;
    return `An error occurred on vRack segment ${segment.targetSpec.vlanId}.`;
  },
});
```

## Limitations

Relies on diffing `currentState` vs `targetSpec` to identify changed fields. These objects don't always accurately reflect the actual resource state due to API inconsistencies (raised with Backend team).

### DIFF
--- a/packages/manager/apps/hpc-vmware-public-vcf-aas/src/hooks/updatedOperationBanner/useListUpdatedOperationBanner.ts
+++ b/packages/manager/apps/hpc-vmware-public-vcf-aas/src/hooks/updatedOperationBanner/useListUpdatedOperationBanner.ts
@@ -1,0 +1,134 @@
+import { useEffect, useRef, useMemo } from 'react';
+import { useMessageContext, MessageColor } from '@/context/Message.context';
+import {
+  UpdatingInfo,
+  ChangedFieldInfo,
+} from '@/hooks/updatedOperationStatus/updatedOperationStatus.types';
+
+type UseListUpdatedOperationBannerParams = {
+  updatingResources: UpdatingInfo[];
+  completedResources: UpdatingInfo[];
+  erroredResources: UpdatingInfo[];
+  getUpdatingMessage: (updating: ChangedFieldInfo) => string | null;
+  getCompletedMessage?: (completed: ChangedFieldInfo) => string | null;
+  getErroredMessage?: (errored: ChangedFieldInfo) => string | null;
+  updatingMessageType?: MessageColor;
+  completedMessageType?: MessageColor;
+  erroredMessageType?: MessageColor;
+};
+
+const flattenChangedFields = (items: UpdatingInfo[]): ChangedFieldInfo[] => {
+  const map = new Map<string, ChangedFieldInfo>();
+
+  items.forEach(({ id, changedFields }) => {
+    if (changedFields.length === 0) {
+      const key = `${id}:null`;
+      if (!map.has(key)) map.set(key, { id, changedField: null });
+    } else {
+      changedFields.forEach((field) => {
+        const key = `${id}:${field}`;
+        if (!map.has(key)) map.set(key, { id, changedField: field });
+      });
+    }
+  });
+
+  return Array.from(map.values());
+};
+
+export const useListUpdatedOperationBanner = ({
+  updatingResources,
+  completedResources,
+  erroredResources,
+  getUpdatingMessage,
+  getCompletedMessage,
+  getErroredMessage,
+  updatingMessageType = 'information',
+  completedMessageType = 'success',
+  erroredMessageType = 'danger',
+}: UseListUpdatedOperationBannerParams): void => {
+  const { addMessage, clearMessage } = useMessageContext();
+  const updatingMsgUidsRef = useRef<Map<string, number>>(new Map());
+
+  const hasCompletedResources = completedResources.length > 0;
+  const hasErroredResources = erroredResources.length > 0;
+
+  const updating = useMemo<ChangedFieldInfo[]>(
+    () => flattenChangedFields(updatingResources),
+    [updatingResources],
+  );
+
+  const completed = useMemo<ChangedFieldInfo[]>(
+    () => flattenChangedFields(completedResources),
+    [completedResources],
+  );
+
+  const errored = useMemo<ChangedFieldInfo[]>(
+    () => flattenChangedFields(erroredResources),
+    [erroredResources],
+  );
+
+  useEffect(() => {
+    const currentKeys = new Set(
+      updating.map((item) => `${item.id}:${item.changedField}`),
+    );
+
+    const shownMessages = new Set<string>();
+    updating.forEach((item) => {
+      const key = `${item.id}:${item.changedField}`;
+      if (!updatingMsgUidsRef.current.has(key)) {
+        const message = getUpdatingMessage(item);
+        if (message && !shownMessages.has(message)) {
+          shownMessages.add(message);
+          const msgId = addMessage({
+            type: updatingMessageType,
+            content: message,
+            isDismissible: false,
+          });
+          updatingMsgUidsRef.current.set(key, msgId);
+        }
+      }
+    });
+
+    // Clear banners for items that finished updating
+    updatingMsgUidsRef.current.forEach((msgId, key) => {
+      if (!currentKeys.has(key)) {
+        clearMessage(msgId);
+        updatingMsgUidsRef.current.delete(key);
+      }
+    });
+  }, [updating]);
+
+  useEffect(() => {
+    if (hasCompletedResources && getCompletedMessage) {
+      const shownMessages = new Set<string>();
+      completed.forEach((item) => {
+        const message = getCompletedMessage(item);
+        if (message && !shownMessages.has(message)) {
+          shownMessages.add(message);
+          addMessage({
+            type: completedMessageType,
+            content: message,
+            isDismissible: true,
+          });
+        }
+      });
+    }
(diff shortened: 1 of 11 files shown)


packages/manager/apps/hpc-vmware-public-vcf-aas/src/hooks/updatedOperationStatus/useUpdatedOperationStatus.ts
```diff
@@ -0,0 +1,131 @@
...
+  options: UpdateStatusOptions = {},
+): UpdatingInfo | null => {
+  const { ignoreFields = [] } = options;
+
+  return useMemo(() => {
+    if (!isResourceUpdating(item)) return null;
+
+    const changedFields = getChangedFields(
+      item.currentState as Record<string, unknown>,
+      item.targetSpec as Record<string, unknown>,
```
mathieumousnier, line 35: You can remove ` as Record<string, unknown>,`.
Type seems ok, so you don't need it, and removing them will avoid hidding type errors if the underlying type change in the future

packages/manager/apps/hpc-vmware-public-vcf-aas/src/hooks/updatedOperationStatus/useUpdatedOperationStatus.ts
```diff
@@ -0,0 +1,131 @@
...
+    if (!items) {
+      return { updatingResources, completedResources, erroredResources };
+    }
+
+    items.forEach((item) => {
+      const prev = prevStatusMapRef.current.get(item.id);
+      const currentStatus = item.resourceStatus;
+      const changedFields = getChangedFields(
+        item.currentState as Record<string, unknown>,
+        item.targetSpec as Record<string, unknown>,
```
mathieumousnier, line 71: You can remove  `as Record<string, unknown>,`

packages/manager/apps/hpc-vmware-public-vcf-aas/src/hooks/updatedOperationStatus/useUpdatedOperationStatus.ts
```diff
@@ -0,0 +1,131 @@
...
+  isResourceUpdating,
+  getChangedFields,
+} from './updatedOperationStatus.utils';
+
+type UpdateStatusOptions = {
+  ignoreFields?: string[];
+};
+
+type PrevStatusInfo = {
+  status: string;
```
mathieumousnier, line 18: Most of the time, a status field can be narrowed to a union type like this 'updating' | 'success' | 'error' (this is a general example, i have no idea which statuses you should use here

packages/manager/apps/hpc-vmware-public-vcf-aas/src/hooks/updatedOperationStatus/useUpdatedOperationStatus.ts
```diff
@@ -0,0 +1,131 @@
...
+      const changedFields = getChangedFields(
+        item.currentState as Record<string, unknown>,
+        item.targetSpec as Record<string, unknown>,
+        ignoreFields,
+      );
+
+      // Currently updating
+      if (
+        UPDATING_STATUSES.includes(
+          currentStatus as typeof UPDATING_STATUSES[number],
```
mathieumousnier, line 78: You need to find a way to remove this `as typeof UPDATING_STATUSES[number]`.
This is saying: "the types are not maching, so i force them to be what the code wants"
This thing hides a design issue, and might hide bugs

packages/manager/apps/hpc-vmware-public-vcf-aas/src/hooks/updatedOperationStatus/useUpdatedOperationStatus.ts
```diff
@@ -0,0 +1,131 @@
...
+type UpdateStatusOptions = {
+  ignoreFields?: string[];
+};
+
+type PrevStatusInfo = {
+  status: string;
+  changedFields: string[];
+};
+
+export const useUpdatedOperationStatus = <
```
mathieumousnier, line 22: This hook does not seem to be used anywhere. In that case it would be better to remove it

packages/manager/apps/hpc-vmware-public-vcf-aas/src/hooks/updatedOperationBanner/useListUpdatedOperationBanner.ts
```diff
@@ -0,0 +1,134 @@
...
+        const key = `${id}:${field}`;
+        if (!map.has(key)) map.set(key, { id, changedField: field });
+      });
+    }
+  });
+
+  return Array.from(map.values());
+};
+
+export const useListUpdatedOperationBanner = ({
```
mathieumousnier, line 38: The details in this code are very complicated to understand so I'll give a "feeling" feedback more than a precise one.

My feeling: This is very complicated for displaying a few messages.
I mean, I imagine having to debug this because there is a bug in a few months and it looks a bit worrying.

On thing for example: the way the "changedFields" needs to be transformed several times.
1. There is the original shape you built in the first step: `UpdatingInfo`
2. Then you use this already quiet complicated function called flattenChangedFields to get now `ChangedFieldInfo`
3. Then you build keys for comparison likes this: `${item.id}:${item.changedField}`

I am not sure, but maybe we could remove step 2 ?

packages/manager/apps/hpc-vmware-public-vcf-aas/src/hooks/updatedOperationBanner/useListUpdatedOperationBanner.ts
```diff
@@ -0,0 +1,134 @@
...
+    () => flattenChangedFields(completedResources),
+    [completedResources],
+  );
+
+  const errored = useMemo<ChangedFieldInfo[]>(
+    () => flattenChangedFields(erroredResources),
+    [erroredResources],
+  );
+
+  useEffect(() => {
```
mathieumousnier, line 70: `useEffect` has a readability issue because it encloses some code without any name like a function, so one thing I do is to always add a comment on top of a useEffect, this improves readability by just saying "This code do this"

```ts
// Find updating fields
useEffect(() => {
..

// Find completed fields
useEffect(() => {
..
```

packages/manager/apps/hpc-vmware-public-vcf-aas/src/hooks/updatedOperationStatus/updatedOperationStatus.utils.ts
```diff
@@ -0,0 +1,48 @@
...
+      !isEqual(currentState[key], targetSpec[key])
+    ) {
+      changedFields.push(key);
+    }
+  });
+
+  return changedFields;
+};
+
+export const findUpdatingItems = <
```
mathieumousnier, line 41: Is that function used anywhere ?

packages/manager/apps/hpc-vmware-public-vcf-aas/src/hooks/updatedOperationBanner/useListUpdatedOperationBanner.ts
sefiani-ovh: i agree with you , i will do some more refacto to simplify the code and better naming

packages/manager/apps/hpc-vmware-public-vcf-aas/src/hooks/updatedOperationStatus/useUpdatedOperationStatus.ts
sefiani-ovh: i will update the pr (so this hook normaly will be used  in a page where we do a get to one ressource

PR #22652, opened 2026-03-24
fix(*): lazy load tracking lib

MaximeBajeux:
## Description

Ticket Reference: #...


anooparveti (approved):

anooparveti (changes requested):
Check if account creation app needs to be updated accordingly as the app is loading piano-analytics as well.

PR #23076, opened 2026-07-29
fix(public-vcf): add polling on edit edge gateway name

pauldkn:
## Description

Ticket Reference: #MANAGER-22097


igorms-pro (approved):
LGM

PR #23128, opened 2026-08-13
Fix: instance backup cron

ErikssonJoakim:
"A backup is taken every day between 22:00 and 06:00 (UTC), and the rotation maintains a log of the 7 latest entries." The current cron is for every rotation day

## Description

Ticket Reference: #TAPC-7035


lolottetheclash (approved):
👌

PR #23136, opened 2026-08-18
feat(backup-licenses): track the delivery between the checkout and the service

tibs245:
## Description

Between a paid checkout and the first resource Agora creates, a Backup Licenses account is
indistinguishable from a brand-new one: `GET /backupServices/tenant` is empty, the onboarding guard
sends the customer back to the onboarding screen, and its CTA invites them to buy — again — a
product they have already paid for. In the meantime the funnel had already redirected them to a
service page for a service that does not exist yet.

This PR gives the account three states with one reachable screen each:

| State | How it is established | Screen |
|---|---|---|
| No subscription | No Backup Licenses VSPC tenant **and** no local order trace | Onboarding, order funnel reachable from its CTA |
| Order in flight | A local order trace exists, **or** a Backup Licenses VSPC tenant exists whose `resourceStatus` is not `READY` | The order funnel, frozen, topped by a delivery tracker |
| Subscription delivered | At least one Backup Licenses VSPC tenant is `READY` | The service pages, as today |

**The trace is written on the OK response of the engaging `POST /order/cart/{cartId}/checkout` and
on nothing else.** A failed checkout writes nothing, keeps the form editable with its values, and
shows the refusal — quoting an actionable 4xx as the API worded it, keeping the generic wording for
a 5xx. A trace written optimistically would lock a customer out of a purchase that never happened.

While tracking, the funnel stays readable but inert (fields disabled, step buttons and back link
gone) under a tracker card: indeterminate progress bar, order number when the checkout returned one,
a warning past 30 min that setup is longer than expected (polling continues), and — when every
tenant of the product is in `ERROR` — the failure plus a control to order again. A `READY` tenant
erases the trace and hands over to "Linked servers".

### Notable implementation points

- **State, not a boolean.** `useHasActiveBackupLicensesSubscription` is replaced by
  `useBackupLicensesSubscriptionStatus` (`NONE` / `PENDING` / `READY` / `ERROR`), resolved from the
  two reads the service pages already use — **no new endpoint**.
- **zustand + `persist`** for the local trace, the montage the `backup-agent` module already uses
  for `useShowNoAgentEnabledMessage`. A store rather than direct storage access because localStorage
  emits no event in the tab that writes it: the routing guard would otherwise read the value from
  before the checkout. Rehydration filters malformed, outdated and expired payloads (24 h) and
  erases them from disk, so a delivery that never lands cannot strand a customer on a progress
  screen. `localStorage` and not `sessionStorage`: a paid order must survive a closed tab.
- **Polling** is React Query's `refetchInterval` (10 s while `PENDING`), never a hand-rolled
  interval. Distinct from BKP-1220's polling, which covers the server rows of a *live* service.
- **`ods-progress-bar` has no indeterminate mode** (only `value`/`max`), so the bar is swept in a
  loop rather than showing an invented percentage — Agora publishes no real progress for this
  delivery. It stops, full, under `prefers-reduced-motion`.
- `useOrderCartPreparation` is disabled while tracking: preparing a cart behind a progress screen
  would create carts nobody will ever order.

Ticket Reference: #BKP-1161 — no dedicated story: the gap was found while finishing 0.2 (BKP-1208)
and is tracked under the epic.

## Additional Information

**Checks run:** `tsc --noEmit` clean · `lint:modern` 0 error · module suite green (98 files,
755 tests), including 25 new tests covering the trace lifecycle, the status resolution, the frozen
funnel and the fact that a failed checkout records nothing.

**Not done, and worth a reviewer's attention:**

- **Never run against a real delivery.** The `PENDING → READY` transition has only been exercised
  through MSW. `GET /checkout` still refuses every `license-type` value on Agora's side (see
  BKP-1208), so a real first order cannot complete today.
- **The eight locales carry the seven new keys**, matching what the last translation commits on this
  module do. `fr_FR` and `en_GB` are hand-written (British spelling, "licence" as the `order`
  namespace already spells it); `de`, `es`, `it`, `pl`, `pt` and `fr_CA` are mine and deserve the
  translation team's pass.
- **Two thresholds are display guards I chose, not product SLAs**: 24 h of trace validity and the
  30 min "longer than expected" warning. Neither cancels or alters an order. Confirmation pending
  from PO + Agora on the product's real delivery times.
- **Three deliberate calls** a reviewer may want to challenge: the order funnel becomes unreachable
  for an account that already holds a `READY` tenant (first-order funnel only, per BKP-1208); a
  delivery is considered lost only when *every* Backup Licenses tenant is in `ERROR`; and the
  tracker keeps polling past the 30 min warning rather than giving up.

Spec: `specs/backup-licenses/order-subscription` R8 and `qa.md` Family 5 in the `ovh-team-specs`
repo.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


igorms-pro (approved):
LGTM

maelovh (approved):

maelovh (approved):

PR #23155, opened 2026-08-24
fix(databases): warn about version upgrade downtime

Toumash:
## Description

Add a warning to the Managed Databases version-upgrade dialog explaining that the service can be unavailable for up to five minutes and recommending a maintenance window.

The repository contribution policy permits contributors to add French translations only, so the new user-facing string is added to `Messages_fr_FR.json`.

## Testing

- `pnpm exec vitest run "src/pages/services/[serviceId]/settings/update/updateVersion/UpdateVersion.spec.tsx"`
- `git diff --check`


Toumash:
sorry, wrong target branch caused a lot of teams to be pinged, moving to https://github.com/ovh/manager/pull/23156

PR #23156, opened 2026-08-24
fix(databases): warn about version upgrade downtime

Toumash:
## Description

Add a warning to the Managed Databases version-upgrade dialog explaining that a service interruption should be expected and recommending that the operation be planned during a maintenance window.

The repository contribution policy permits external contributors to add French translations only, so the new user-facing string is added to Messages_fr_FR.json.

## Commit structure

The documentation link is intentionally isolated in the second commit because I am not certain whether the linked documentation is still current or required. If the documentation link is not desired, that commit can simply be dropped and the first commit containing only the downtime warning can be merged on its own.

## Testing

- Targeted UpdateVersion Vitest test
- git diff --check


Toumash:
can we merge it?

PR #23176, opened 2026-08-28
feat(hub): change support widget for Digital Agent API (FR only) + links for fr

AlexisAfonso:
## Description

Change for FR customer only widget and link from support to Digital Agent directly on manager v7 

Ticket Reference:  #MANAGER-22282


MaximeBajeux (approved):

sachinrameshn (approved):

sachinrameshn:
As far as I know comment message ref should start in small letter. So can you keep it kike that 
Ex: ref: #MANAGER-ticket

PR #23178, opened 2026-08-29
feat(account): account creation for FR customers only - remove Administration without siret skip option

AlexisAfonso:
## Description

- Administration and Corporate legal form cannot create account without Siret
- Only Non profit can create an account without Siret

Ticket Reference:: #MANAGER-22249


sachinrameshn (dismissed):

sachinrameshn:
As far as I know comment message ref should start in small letter. So can you keep it kike that 
Ex: ref: #MANAGER-ticket

aottr (dismissed):

aottr (approved):

